PR #475, opened 2014-12-31
fix two bugs

rousya:
2814237 create-diff-object: fix a potential overflow for rela type
f1683eb kpatch-build: fix ERROR: two parent matches for x.o

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -133,7 +133,7 @@ struct rela {
 	struct list_head list;
 	GElf_Rela rela;
 	struct symbol *sym;
-	unsigned char type;
+	unsigned int type;
 	int addend;
 	int offset;
 	char *string;


kpatch-build/kpatch-build
```diff
@@ -165,7 +165,7 @@ find_parent_obj() {
 }
 
 find_kobj() {
-	arg=$1
+	arg=$(echo $1 |sed 's,^\S*\.\./,,g')
```
jpoimboe: Hm, I think I'm missing something.  How can the argument to find_kobj ever be something like `arch/x86/kvm/../../../virt/kvm/kvm_main.o`?  The caller of find_kobj is:
```
FILES="$(find * -type f)"
cd "$TEMPDIR"
mkdir output
declare -A objnames
CHANGED=0
ERROR=0
for i in $FILES; do
	mkdir -p "output/$(dirname $i)"
	cd "$OBJDIR"
	find_kobj $i
```
It gets the filename from the "find" command.  So I would expect the argument to find_kobj to be `virt/kvm/kvm_main.o` already.

rousya: Oh, it's my fault. I checked changed_objs and found 'arch/x86/kvm/../../../virt/kvm/kvm_main.o'.
Please ingore this patch.
Sorry.
Need I re-open a pull-request?

jpoimboe: No need to open a new pull request.  You can just rebase this patch and git push --force to the same branch.

rousya: Sorry for late. I've rebased and pushed again. Thanks.

jpoimboe:
@rousya Looks good to me, can you please rebase to the latest head?

rousya:
Hi @jpoimboe,
I've rebase it to the latest master.

jpoimboe:
:+1:

Issue #476, opened 2015-01-21
add livepatch compatibility


jpoimboe:
Ideally kpatch-build would make a runtime decision: if the kernel has CONFIG_LIVEPATCH enabled, use livepatch, otherwise use the kpatch core module (kpatch.ko).

This might also include changing the kpatch core module interface to more closely resemble the livepatch interface.

jpoimboe:
Fixed by #477.

PR #477, opened 2015-01-22
add livepatch API support

spartacus06:
Adds a new patch module scaffold for use when building against a kernel
with CONFIG_LIVE_PATCHING=y.

Fixes issue #476

Signed-off-by: Seth Jennings <sjenning@redhat.com>

### DIFF
--- a/kmod/patch/livepatch-patch-hook.c
+++ b/kmod/patch/livepatch-patch-hook.c
@@ -29,7 +29,25 @@
 
 #include "kpatch-patch.h"
 
-struct klp_patch *patch;
+/*
+ * There are quite a few similar structures at play in this file:
+ * - livepatch.h structs prefixed with klp_*
+ * - kpatch-patch.h structs prefixed with kpatch_patch_*
+ * - local scaffolding structs prefixed with patch_*
+ *
+ * The naming of the struct variables follows this convention:
+ * - livepatch struct being with "l" (e.g. lfunc)
+ * - kpatch_patch structs being with "k" (e.g. kfunc)
+ * - local scaffolding structs have no prefix (e.g. func)
+ *
+ *  The program reads in kpatch_patch structures, arranges them into the
+ *  scaffold structures, then creates a livepatch structure suitable for
+ *  registration with the livepatch kernel API.  The scaffold structs only
+ *  exist to allow the construction of the klp_patch struct.  Once that is
+ *  done, the scaffold structs are no longer needed.
+ */
+
+struct klp_patch *lpatch;
 
 static LIST_HEAD(patch_objects);
 static int patch_objects_nr;
@@ -99,7 +117,7 @@ static int patch_add_func_to_object(struct kpatch_patch_func *kfunc)
 	return 0;
 }
 
-static int patch_add_reloc_to_object(struct kpatch_patch_dynrela *dynrela)
+static int patch_add_reloc_to_object(struct kpatch_patch_dynrela *kdynrela)
 {
 	struct patch_reloc *reloc;
 	struct patch_object *object;
@@ -108,9 +126,9 @@ static int patch_add_reloc_to_object(struct kpatch_patch_dynrela *dynrela)
 	if (!reloc)
 		return -ENOMEM;
 	INIT_LIST_HEAD(&reloc->list);
-	reloc->kdynrela = dynrela;
+	reloc->kdynrela = kdynrela;
 
-	object = patch_find_object_by_name(dynrela->objname);
+	object = patch_find_object_by_name(kdynrela->objname);
 	if (!object) {
 		kfree(reloc);
 		return -ENOMEM;
@@ -164,7 +182,7 @@ extern struct kpatch_patch_dynrela __kpatch_dynrelas[], __kpatch_dynrelas_end[];
 static int __init patch_init(void)
 {
 	struct kpatch_patch_func *kfunc;
-	struct kpatch_patch_dynrela *dynrela;
+	struct kpatch_patch_dynrela *kdynrela;
 	struct klp_object *lobjects, *lobject;
 	struct klp_func *lfuncs, *lfunc;
 	struct klp_reloc *lrelocs, *lreloc;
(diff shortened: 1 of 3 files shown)


jpoimboe:
YES!

jpoimboe:
This is awesome (see animated gif ;-)

The scaffolding is ok for now.  Hopefully we can eventually figure out a way to build the klp_patch struct at build time instead of run time, so that the patch template can be nice and lean.  We might need to add more intelligence to the linking step somehow.

I'll add some code comments inline.

kmod/patch/livepatch-patch-hook.c
```diff
@@ -0,0 +1,324 @@
...
+#include <linux/kernel.h>
+#include <linux/module.h>
+#include <linux/list.h>
+#include <linux/slab.h>
+
+#include <linux/livepatch.h>
+
+#include "kpatch-patch.h"
+
+struct klp_patch *patch;
```
jpoimboe: Because of the scaffolding, there are three different structure "languages" being used: kpatch-build, scaffolding, and klp.  It kind of threw me for a loop when trying to read this code.  It would be easier to follow if the same variable, struct, and function naming conventions were used consistently throughout the entire file.  Something like:

- for kpatch-build: kpatch/kobject/kreloc/kfunc
- for temporary scaffolding: tpatch/tobject/treloc/tfunc (or spatch/sobject/sreloc/sfunc)
- for klp: lpatch/lobject/lreloc/tfunc

And also maybe a comment describing the fact that there are three languages and the different naming conventions.

kmod/patch/livepatch-patch-hook.c
```diff
@@ -0,0 +1,324 @@
...
+
+	func = kzalloc(sizeof(*func), GFP_KERNEL);
+	if (!func)
+		return -ENOMEM;
+	INIT_LIST_HEAD(&func->list);
+	func->kfunc = kfunc;
+
+	object = patch_find_object_by_name(kfunc->objname);
+	if (!object)
+		return -ENOMEM;
```
jpoimboe: need to free func here

kmod/patch/livepatch-patch-hook.c
```diff
@@ -0,0 +1,324 @@
...
+
+	reloc = kzalloc(sizeof(*reloc), GFP_KERNEL);
+	if (!reloc)
+		return -ENOMEM;
+	INIT_LIST_HEAD(&reloc->list);
+	reloc->kdynrela = dynrela;
+
+	object = patch_find_object_by_name(dynrela->objname);
+	if (!object)
+		return -ENOMEM;
```
jpoimboe: need to free reloc here

kmod/patch/livepatch-patch-hook.c
```diff
@@ -0,0 +1,324 @@
...
+	}
+}
+
+static void patch_free_livepatch(struct klp_patch *patch)
+{
+	struct klp_object *object;
+
+	if (patch) {
+		object = patch->objs;
+		while (object && object->funcs) {
```
jpoimboe: I think a for loop would be clearer here

kmod/patch/livepatch-patch-hook.c
```diff
@@ -0,0 +1,324 @@
...
+	/* past this point, only possible return code is -ENOMEM */
+	ret = -ENOMEM;
+
+	/* allocate and fill livepatch structures */
+	patch = kzalloc(sizeof(*patch), GFP_KERNEL);
+	if (!patch)
+		goto out;
+
+	lobjects = kzalloc(sizeof(*lobjects) * (patch_objects_nr+1),
+	                      GFP_KERNEL);
```
jpoimboe: indentation is off

kmod/patch/livepatch-patch-hook.c
```diff
@@ -0,0 +1,324 @@
...
+		list_for_each_entry(func, &object->funcs, list) {
+			lfunc = &lfuncs[j];
+			lfunc->old_name = func->kfunc->name;
+			lfunc->new_func = (void *)func->kfunc->new_addr;
+			lfunc->old_addr = func->kfunc->old_addr;
+			j++;
+		}
+
+		lrelocs = kzalloc(sizeof(struct klp_reloc) *
+		                    (object->relocs_nr+1), GFP_KERNEL);
```
jpoimboe: indentation is off

kmod/patch/livepatch-patch-hook.c
```diff
@@ -0,0 +1,324 @@
...
+		if (!lrelocs)
+			goto out;
+		lobject->relocs = lrelocs;
+		j = 0;
+		list_for_each_entry(reloc, &object->relocs, list) {
+			lreloc = &lrelocs[j];
+			lreloc->loc = reloc ->kdynrela->dest;
```
jpoimboe: extra space in "reloc ->"

kmod/patch/livepatch-patch-hook.c
```diff
@@ -0,0 +1,324 @@
...
+	return 0;
+out:
+	patch_free_livepatch(patch);
+	patch_free_scaffold();
+	return ret;
+}
+
+static void __exit patch_exit(void)
+{
+	WARN_ON(klp_disable_patch(patch));
```
jpoimboe: We shouldn't disable here.  That will be done by the user (which in the future will result in a module_put on this module, allowing this exit routine to be called).

kmod/patch/livepatch-patch-hook.c
```diff
@@ -0,0 +1,324 @@
...
+			object++;
+		}
+		if (patch->objs)
+			kfree(patch->objs);
+		kfree(patch);
+	}
+}
+
+/* debug */
+#if 0
```
jpoimboe: is this debug code still needed?

jpoimboe:
end comments

spartacus06:
Regarding making the klp_patch structure at build time, I think that would be quite an undertaking.  You'd need a separate program to read in the all the kpatch sections from all of the changed objects, create the section that contains the klp_patch structure, and exclude the (now temporary) kpatch section from the final linking.  I'm pretty confident that it would be more code than this scaffold.

jpoimboe:
I agree it would be more code than this, but I don't think it would be too bad.  It could come along after ld has done its work.  That way it would only have to read output.o instead of all the objects.

It would have the benefit of vastly simplifying the patch template code, and build time complexity is much better than run time complexity IMO.

jpoimboe:
Or it might even be possible to implement it in the linker script... that would be ideal.

jpoimboe:
Looks good.  One more comment: The global `patch` should be `lpatch`, and doesn't need to be global.

I still need to do some testing with it.

jpoimboe:
We also need to update the kpatch script to work with both (can use a separate GH issue to track that)

spartacus06:
new "add naming" commit renames patch to lpatch.  it does need to be global though for unregister in patch_exit()

PR #488, opened 2015-04-29
Replace initramfs code with systemd service

spartacus06:
Remove the dracut support for adding the kpatch modules to the initramfs.  This creates a sizable delay in installation time and doesn't offer any added protection over just applying the patches at boot time using a systemd service.  Additional, we are seeing more platforms, namely atomic and netboot environments, where changing the initramfs can't be done.


Makefile.inc
```diff
@@ -11,6 +11,7 @@ MODULESDIR = $(DESTDIR)$(PREFIX)/$(LIBDIR)/kpatch
 LIBEXECDIR = $(DESTDIR)$(PREFIX)/libexec/kpatch
 DATADIR    = $(DESTDIR)$(PREFIX)/share/kpatch
 MANDIR     = $(DESTDIR)$(PREFIX)/share/man/man1
+SYSTEMDDIR = $(DESTDIR)$(PREFIX)/lib/systemd/system
 DRACUTDIR  = $(DESTDIR)/usr/$(LIBDIR)/dracut/modules.d/99kpatch
```
jpoimboe: This DRACUTDIR line can be removed

contrib/kpatch.spec
```diff
@@ -70,7 +68,7 @@ make %{_smp_mflags}
 %install
 rm -rf %{buildroot}
 
-make install DESTDIR=%{buildroot}
+PREFIX=/usr make install DESTDIR=%{buildroot}
```
jpoimboe: Probably should be `PREFIX=/%{_usr}`.  Also, putting it next to DESTDIR (or putting DESTDIR next to PREFIX) would make it a little more readable.

contrib/kpatch.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description="Apply kpatch kernel patches"
+
+[Service]
+Type=oneshot
```
jpoimboe: Should it also have `RemainAfterExit=yes` so that it shows "active" and nobody tries to restart the service?

jpoimboe:
End comments (though I still haven't tested it yet)

contrib/kpatch.service
jpoimboe: Sorry for waffling on this.  Been thinking about this a little more.

I think RemainAfterExit and ExecStop aren't ideal either, because they result in unpatching the kernel when rebooting, which is unnecessary and potentially harmful.

How about getting rid of RemainAfterExit and ExecStop, and instead putting RefuseManualStart=true in the Unit section?  That way it'll only ever be started once (at boot), and should never be started again.  It probably doesn't really matter that the service state won't be "active" anyway.

jpoimboe:
```
2015-04-29 01:57:00 PM < jpoimboe> sjennings: found an issue in my testing.  by default, kpatch is installed to /usr/local/sbin/kpatch, but the service file looks for it at /usr/sbin/kpatch
2015-04-29 01:57:10 PM < sjennings> ahh yes
2015-04-29 01:57:13 PM < sjennings> :-/
2015-04-29 01:57:24 PM < jpoimboe> may have to do some kind of sed hack when installing the service file?
2015-04-29 01:57:41 PM < sjennings> or just trust $PATH and run "kpatch" :D
2015-04-29 01:57:46 PM < jpoimboe> i.e. replace /usr/sbin/kpatch with $PREFIX/kpatch or something
2015-04-29 01:58:16 PM < jpoimboe> I actually tried just "kpatch", and systemd refused to do it without an absolute path :-)
2015-04-29 01:58:24 PM < sjennings> booo
2015-04-29 01:58:28 PM < jpoimboe> :-)
```

jpoimboe:
One more minor fix -- kpatch.service needs the executable bits removed:
```
Jun 03 09:02:04 treble systemd[1]: Configuration file /usr/local/lib/systemd/system/kpatch.service is marked executable. Please remove executable permission bits. Proceeding anyway.
```

spartacus06:
sigh. fixed :)

PR #489, opened 2015-04-29
readme: update dependencies for f21

jpoimboe:
Change the supported Fedora version to F21 and add a new dependency.

For some reason, numactl-devel is needed by "rpmbuild -bp kernel", but
isn't detected by "yum-builddep kernel".

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Installation
 
 ###Prerequisites
 
-####Fedora 20
+####Fedora 21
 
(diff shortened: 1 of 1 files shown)


spartacus06:
Yeah :-/  We've hit things like that before where there are arch specific build dependencies that yum-builddep <package>, it doesn't include, but if you do yum-builddep <kernelspecfile>, it will.

PR #490, opened 2015-05-31
Fix kpatch-build failure with srpm on centos

kragniz:
kpatch-build was failing on centos7 with

    mv: cannot stat '/home/vagrant/rpmbuild/BUILD/kernel-*/linux-3.10.0-229.el7.x86_64': No such file or directory

in the error log. This was due to the actual directory being named
linux-3.10.0-229.el7.centos.x86_64. This patch avoids this failure by
adding a wildcard before the arch.

Signed-off-by: Louis Taylor <louis@kragniz.eu>

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -365,7 +365,7 @@ else
 
 		clean_cache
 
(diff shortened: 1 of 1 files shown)


jpoimboe:
Thanks @kragniz!

:+1:

PR #492, opened 2015-06-25
Add new r_info field to kpatch_patch_dynrela and rela structs

flaming-toast:
Save the `r_info` field in create-diff-object to pass into `klp_write_module_reloc()`.

### DIFF
--- a/kmod/patch/kpatch-patch.h
+++ b/kmod/patch/kpatch-patch.h
@@ -35,6 +35,7 @@ struct kpatch_patch_dynrela {
 	unsigned long dest;
 	unsigned long src;
 	unsigned long type;
+	unsigned long r_info;
 	char *name;
 	char *objname;
 	int external;
--- a/kmod/patch/livepatch-patch-hook.c
+++ b/kmod/patch/livepatch-patch-hook.c
@@ -252,6 +252,7 @@ static int __init patch_init(void)
 			lreloc->loc = reloc->kdynrela->dest;
 			lreloc->val = reloc->kdynrela->src;
 			lreloc->type = reloc->kdynrela->type;
+			lreloc->r_info = reloc->kdynrela->r_info;
 			lreloc->name = reloc->kdynrela->name;
 			lreloc->addend = reloc->kdynrela->addend;
 			lreloc->external = reloc->kdynrela->external;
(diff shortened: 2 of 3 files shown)


jpoimboe:
Thanks!  Looks good to me.

Need to keep this PR on hold for now, because it depends on a future upstream `struct klp_reloc` change.

flaming-toast:
Sounds good, upstream patch soon to follow :-)

spartacus06:
I haven't been following the discussion closely, but it seems to me that all of the data contained in r_info is already store, in a form, in the fields we have.  r_info contains the type, which is has a fields, and the symbol index, which is indirectly stored in the associated symbol.  The symbols are reindexed by create-diff-object, so the original symbol index doesn't matter.

flaming-toast:
@spartacus06 Thanks for the feedback! We eventually need to use the symbol index in `klp_write_module_reloc` (for s390x), as PLT and GOT offsets (used in certain s390 relocation types) are accessed via module->arch.syminfo + symbol index. I didn't see a way to access symbol index information once we're already in `klp_write_object_relocations()`, since we don't save this information in the `klp_reloc` structs. I decided saving `r_info` might be useful since we can then extract both the relocation type and sym index from it. Do you happen see another way to access symbol indices?

jpoimboe:
Since type is encoded in `r_info`, I guess we can remove `type` from `klp_reloc`?

@spartacus06, good point about the original symbol index being irrelevant.  The new index is generated in `kpatch_reindex_elements()` and encoded into the new `r_info` in `kpatch_rebuild_rela_section_data()`.

flaming-toast:
Hmmm... Actually now considering the reindexing, I think the old `r_info`'s are being saved in the `.kpatch.dynrelas` section in `kpatch_create_dynamic_rela_sections()` instead of the new ones generated in `kpatch_rebuild_rela_section_data()`, since the former is called before the latter. Pretty sure we want to save the new one :-) I'll have to rethink this more..

spartacus06:
Wow, my English was horrible in my comment.  It's like I've been doing the same 4 demos for 6 hours at a conference and extremely sleep deprived or something :P

jpoimboe:
I think we decided that a different approach is needed.  Closing...

PR #499, opened 2015-09-25
bugfix: correlate the rela sections of bundled static variables

zhouchengming1:
If a static variable is a pointer, it has rela section.

Example:
	static int *p = &a;
changed to:
	static int *p = &b;
so its rela section has changed.

Then this change of data should be found and report error.
But if we don't correlate its rela section, we won't
find this change.

Signed-off-by: Zhou ChengMing <zhouchengming1@outlook.com>


jpoimboe:
Thanks @zhouchengming1.  The patch looks good to me.  Do you happen to have an example patch which produces this bug?

zhouchengming1:
Sorry, I don't have an example patch like this. I just find the bug when watch the code.

PR #500, opened 2015-09-30
examples: Added an example of a problematic patch with an explanation

euspectre:
examples/tcp_cubic-better-follow-cubic-curve-original.patch is the
original patch, combined from two mainline commits (see the description
in the patch). It cannot be used with Kpatch as it is because the
change is in the initialization of a global structure.

examples/tcp_cubic-better-follow-cubic-curve-converted.patch is a
modification of the patch that Kpatch can process. Still, this
modification has its issues, see the description there.


jpoimboe:
@euspectre Thanks for starting this!  and congrats on opening our 500th PR :)

:+1:

euspectre:
:-) Yes, let us start from here.

PR #502, opened 2015-10-12
kpatch-build: fix shadow_get function

libin2015:
The shadow_get function does't consider the case that
'shadow is inpace', and after the shadow->data be set to the data,
it will not be the pointer. This patch fix it.


jpoimboe:
:+1:

Issue #494, opened 2015-08-03
NULL pointer dereference in patch modules with multiple objects


flaming-toast:
A patch module will crash due to null pointer dereference in cases where it contains patches to multiple objects (example: a patch module built from patches changing vmlinux and ext4)

This happens in `patch_find_object_by_name()` in `kmod/patch/livepatch-patch-hook.c`. While we're processing each function in `__kpatch_funcs`,  if we've already created a vmlinux `patch_object` and we come across a function from another object (say, ext4) then we'll end up dereferencing a null pointer when we try to `strcmp` `object->name` and the passed in `name` argument. Here are the offending lines from `patch_find_object_by_name()`: 
````
static struct patch_object *patch_find_object_by_name(const char *name) {
[ snip ]
list_for_each_entry(object, &patch_objects, list) {
    if ((!strcmp(name, "vmlinux") && !object->name) ||
          !strcmp(object->name, name)) // Crash here
           return object;
}
```
So order here was important to trigger the bug. Say no `patch_object`s have been allocated yet. We process a vmlinux function first and a vmlinux `patch_object` is alloc'd and added to `&patch_objects`. Then we come across an ext4 function next and when we enter `list_for_each_entry`, `name` obviously doesn't match vmlinux so that case fails, but `object->name` will be NULL because `object->name` is *not* set for the vmlinux `patch_object` (see `patch_alloc_new_object`), and `strcmp` crashes. Only module `patch_object`s get `name` fields. Had we processed the kpatch funcs the other order around, this bug wouldn't have triggered since the ext4 object would have been allocated first and that *does* have a `name` field. 

I think the best fix is to just have the vmlinux patch object to have its `object->name` field filled in, that way we don't even risk having a null pointer to begin with. 

Stacktrace:
````
BUG: unable to handle kernel NULL pointer dereference at (null)
IP: [<ffffffff812c81e8>] strcmp+0x18/0x40
[snip]
Call Trace:
[<ffffffffa032f0af>] patch_init+0xaf/0x1000 [kpatch_multiple_objnames]
[<ffffffff81002128>] do_one_initcall+0xb8/0x200
[<ffffffff8119f1d2>] ? __vunmap+0xa2/0x100
[<ffffffff811bc916>] ? kfree+0x176/0x180
[<ffffffff8157536b>] do_init_module+0x60/0x1cb
[snip...]
````

Sample patch to reproduce bug:
````
diff --git a/mm/vmalloc.c b/mm/vmalloc.c 
[...]

PR #505, opened 2015-10-14
livepatch-patch-hook: check for object->name before calling strcmp

flaming-toast:
Fixes issue #494. A null pointer dereference can result with patch modules for multiple objects since the "vmlinux" patch object's "name" field is null. strcmp therefore crashes trying to compare object->name if the current object is vmlinux and the supplied "name" argument is not. 
Check that object->name is not null before invoking strcmp.


jpoimboe:
:+1:

PR #506, opened 2015-10-16
kpatch-build: support adding new files in patch

libin2015:
Geting the changed objects from the patched dir, in order to support
adding new files in patch.

The reason that not using 'cat $TEMPDIR/changed_objs' to get the
changed objects is to avoid following case:
arch/x86/kvm/../../../virt/kvm/kvm_main.o
that find_parent_obj will fail with "two parent matches for *.o".

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -463,7 +463,7 @@ do
 done
 
 echo "Extracting new and modified ELF sections"
-cd "$TEMPDIR/orig"
+cd "$TEMPDIR/patched"
 FILES="$(find * -type f)"
 cd "$TEMPDIR"
 mkdir output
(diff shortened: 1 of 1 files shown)


kpatch-build/kpatch-build
```diff
@@ -482,6 +482,7 @@ for i in $FILES; do
 	cd $TEMPDIR
 	debugopt=
 	[[ $DEBUG -eq 1 ]] && debugopt=-d
+	[[ ! -e "orig/$i" ]] && cp -f "patched/$i" "output/$i" && continue
```
jpoimboe: In this case we still need to set objnames[$KOBJFILE]=1

jpoimboe:
> The reason that not using 'cat $TEMPDIR/changed_objs' to get the
> changed objects is to avoid following case:
> arch/x86/kvm/../../../virt/kvm/kvm_main.o
> that find_parent_obj will fail with "two parent matches for *.o".

Can you clarify this comment?  As far as I can tell, the patch didn't make any changes related to the use of changed_objs.

kpatch-build/kpatch-build
libin2015: Yes

libin2015:
Maybe i'm missing something? I will test this case soon.

jpoimboe:
Continued with #510.

libin2015:
> Can you clarify this comment? As far as I can tell, the patch didn't make any changes related to the use of changed_objs.

This comment is only to explain the reason that not using the changed_objs but using the pathed dir.

For the following case,

PR #507, opened 2015-10-16
support patching weak function

libin2015:


jpoimboe:
Instead of creating a new lookup_weak_global_symbol() function, how about just adding the WEAK check to lookup_global_symbol()?

libin2015:
Yes, if the two patch is exactly correct, and can be merged together,  we can just adding the WEAK check to lookup_global_symbol.  Thanks.

jpoimboe:
Yes, otherwise it looks right to me, so if you can rebase them into a single commit with the change to lookup_global_symbol, I'll give it a :+1: .  Thanks for the fix!

libin2015:
OK, I will do it and create a new pull request. Thanks.

jpoimboe:
Continued with #508.  (next time, please rebase and force push the existing branch instead of opening a new pull request)

PR #510, opened 2015-10-16
kpatch-build: support adding new files in patch

libin2015:
Geting the changed objects from the patched dir, in order to support
adding new files in patch.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -463,7 +463,7 @@ do
 done
 
 echo "Extracting new and modified ELF sections"
-cd "$TEMPDIR/orig"
+cd "$TEMPDIR/patched"
 FILES="$(find * -type f)"
 cd "$TEMPDIR"
 mkdir output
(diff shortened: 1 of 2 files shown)


jpoimboe:
@libin2015 FYI, next time there's no need to open a new pull request.  You can just update the old pull request by rebasing and force pushing the existing branch.

jpoimboe:
:+1:

kpatch-build/kpatch-build
```diff
@@ -482,21 +482,26 @@ for i in $FILES; do
...
+		# create-diff-object returns 3 if no functional change is found
+		[[ $rc -eq 0 ]] || [[ $rc -eq 3 ]] || ERROR=$(expr $ERROR "+" 1)
+		if [[ $rc -eq 0 ]]; then
+			[[ -n $ERROR_IF_DIFF ]] && die $ERROR_IF_DIFF
+			CHANGED=1
+			objnames[$KOBJFILE]=1
+		fi
+	else
+		cp -f "patched/$i" "output/$i"
 		objnames[$KOBJFILE]=1
```
spartacus06: do we need to set CHANGED=1 here?  it seems that all that changes was the new file, it will bail out a few lines down with "no functional changes found".

jpoimboe: I think the "no functional changes found" error would still be appropriate.  If they're not changing any functions, why use kpatch?

spartacus06: yeah, after i thought about it for a little while, i tired to think of a situation where someone would do this i.e. call new functions in a new object without changing an existing object.

only situation i could come up with is registering in a new function in a notification chain with a kpatch load hook or something.

i'm fine with it returning an error though.

jpoimboe: > only situation i could come up with is registering in a new function in a notification chain with a kpatch load hook or something.

Good point.  I could go either way... the current patch is fine.

libin2015:
>@libin2015 FYI, next time there's no need to open a new pull request. You can just update the old pull request by rebasing and force pushing the existing branch
But if donging in this way, the previous comments based on the previous patch will be viewed odd?

jpoimboe:
> But if donging in this way, the previous comments based on the previous patch will be viewed odd?

In the past we have noticed patch comments that disappear after rebasing the branch.  But that problem only happened when adding a comment to a "Commit" rather than the "Files Changed" tab.  I don't know if GitHub still has that quirk.  But our policy to avoid this problem is to always add review comments to the "Files Changed" tab rather than the "Commits" tab.

Issue #478, opened 2015-01-22
create klp_patch struct at build time


jpoimboe:
Create the klp_patch struct at build time, rather than run time, as discussed in #477.  This will greatly simplify the patch module template code.

arges:
I've started hacking on this. How does this sound:

1) Modify create-diff-object.c with an '-l' flag for livepatch. This flag will output .klp.funcs, .klp.relocs that match klp_* structures (much like kpatch-patch.h).

2) Modify kpatch-build to invoke create-diff-object with -l when klp patching is detected.

3) Create a program that links all patch objects and inserts constructed klp_objects. Each klp_object references its own .klp.<objectname>.relocs and .klp.<objectname>.funcs sections.

4) The livepatch hook script constructs the klp_patch and simply inserts the linked klp_objects.

jpoimboe:
@arges Sounds great.  Though I didn't quite follow this:

> 3) Create a program that links all patch objects and inserts constructed klp_objects. Each klp_object references its own .klp..relocs and .klp..funcs sections.

Why create a program?  Doesn't kpatch-build do the object linking already?  (I must admit I haven't looked at the code in quite a while, so I could be missing something obvious...)

arges:
A few challenges I'm trying to figure out:

1)
The create-diff-object program outputs funcs/relocs for one klp_object. However creating a section like .klp.funcs / .klp.relocs would require an additional field (like objname) to correlate them an object later on (like how .kpatch.funcs already does this). However doing this would mean the livepatch hook will still need to dynamically create the structures.

I'm trying to figure out how to do this with just a linker and templating such that the livepatch hook code externs the variables from the linked object and registers the structures. One option would be to create the klp_object(s) in create-diff-object which already have pre-populated indicies for funcs/relocs into .klp.funcs/.klp.relocs. I'd have to ensure that there are zero'ed elements in between each.

2)
Including the livepatch.h header in create-diff-objects from linux/livepatch.h would be ideal. However, this is challenging because of the list_head conflicts etc. Currently I have a version of this header and put it into kmod/patch/klp_patch.h, such that the 'internal' fields are just char _pad[X]. My worry here is these structures getting out of sync with the kernel (or other CONFIG discrepancies.)

3) 
Separating out kpatch/klp code. Right now I can write lots of if statements and pass flags and what not. Is there a better way to handle this? I could also #ifdef code if it makes things easier.

I'll continue to hack, let me know any advise here.
Thanks,

[...]

PR #511, opened 2015-10-16
Flags/variables to help split livepatch and kpatch

arges:
These patches will allow for developing features that are specific to livepatch to be exposed in create-diff-object as well as in the kpatch-build script. This is a base for #478, which will use these flags to make the livepatch hook much simpler.

### DIFF
--- a/kmod/patch/Makefile
+++ b/kmod/patch/Makefile
@@ -1,17 +1,18 @@
 KPATCH_NAME ?= patch
 KPATCH_BUILD ?= /lib/modules/$(shell uname -r)/build
 KPATCH_MAKE = $(MAKE) -C $(KPATCH_BUILD) M=$(PWD)
+KPATCH_TYPE ?= kpatch
 
-obj-m += kpatch-$(KPATCH_NAME).o
+obj-m += $(KPATCH_TYPE)-$(KPATCH_NAME).o
 
-kpatch-$(KPATCH_NAME)-objs += patch-hook.o kpatch.lds output.o
+$(KPATCH_TYPE)-$(KPATCH_NAME)-objs += patch-hook.o $(KPATCH_TYPE).lds output.o
 
-all: kpatch-$(KPATCH_NAME).ko
+all: $(KPATCH_TYPE)-$(KPATCH_NAME).ko
 
-kpatch-$(KPATCH_NAME).ko:
-	$(KPATCH_MAKE) kpatch-$(KPATCH_NAME).ko
+$(KPATCH_TYPE)-$(KPATCH_NAME).ko:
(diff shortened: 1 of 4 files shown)


kmod/patch/Makefile
```diff
@@ -1,17 +1,18 @@
...
-all: kpatch-$(KPATCH_NAME).ko
+all: $(KPATCH_TYPE)-$(KPATCH_NAME).ko
 
-kpatch-$(KPATCH_NAME).ko:
-	$(KPATCH_MAKE) kpatch-$(KPATCH_NAME).ko
+$(KPATCH_TYPE)-$(KPATCH_NAME).ko:
+	$(KPATCH_MAKE) $(KPATCH_TYPE)-$(KPATCH_NAME).ko
 
-patch-hook.o: patch-hook.c kpatch-patch-hook.c livepatch-patch-hook.c
+patch-hook.o: patch-hook.c
```
jpoimboe: I think we still need a $(KPATCH_TYPE)-patch-hook.c dependency here?

kpatch-build/kpatch-build
```diff
@@ -462,6 +463,12 @@ do
 	cp -f "$OBJDIR/$i" "$TEMPDIR/patched/$i" || die
 done
 
+# Detect patching type
+if grep -q "CONFIG_LIVEPATCH=y" "$CONFIGFILE"; then
+  PATCHTYPE="klp"
```
jpoimboe: I'd prefer to s/klp/livepatch/ and s/KLP/LIVEPATCH/, since it's not obvious to all what klp stands for.

Also, use tabs for indentation to be consistent with the rest of the file.

kmod/patch/Makefile
arges: This built just fine for me since patch-hook.c includes the file.

jpoimboe: Yes, but the dependency needs to be specified so that make knows to rebuild patch-hook.o if one of the source files changes.

arges:
Ok updated to address comments.

kpatch-build/kpatch-build
```diff
@@ -462,6 +463,12 @@ do
 	cp -f "$OBJDIR/$i" "$TEMPDIR/patched/$i" || die
 done
 
+# Detect patching type
+if grep -q "CONFIG_LIVEPATCH=y" "$CONFIGFILE"; then
+  PATCHTYPE="livepatch"
```
jpoimboe: Please use a tabs instead of spaces to be consistent with the rest of the file.

arges:
Ok v3 pushed with indentation fixes.

PR #512, opened 2015-10-19
kpatch-build: a couple of symvers-related fixes

euspectre:
The first commit ensures kpatch-build looks for Module.symvers for the core module in the directory corresponding to the target kernel rather than for the currently running one.

The second one handles the case when Kpatch tools are distributed in a package and the core module is installed to /lib/modules/<kernel_version>/extra/kpatch/


jpoimboe:
:+1:

PR #513, opened 2015-10-19
kpatch-build: do not clobber ~/rpmbuild directory

euspectre:
If a source RPM is used to obtain the kernel sources, kpatch-build
executes rpmdev-setuptree to prepare ~/rpmbuild directory tree, installs
the source RPM there. Then it calls 'rpmbuild -bp' to prepare the
kernel source tree.

All this, however, may clobber the existing contents of ~/rpmbuild,
which is very inconvenient if one uses rpmbuild to build other packages.

To avoid that, I could not find a better way than to specify a fake home
directory (~/.kpatch/tempsrc) for that portion of kpatch-build. It seems,
neither rpmdev-setuptree nor rpm have appropriate options for that.

I put the affected commands into a subshell so that the changes in $HOME
could not propagate to other parts of kpatch-build.


jpoimboe:
:+1:

PR #515, opened 2015-10-21
kpatch-build: do not use -ffunction-sections for a patch module itself

euspectre:
-ffunction-sections and -fdata-sections are needed when building the
original and the patched kernels.

It is not necessary, however, to use these options when building a
patch module itself, its functions and data are OK in the sections they
are.

Let us remove these options from KCGLAGS after the kernels have been
built.


jpoimboe:
:+1:

PR #516, opened 2015-10-21
Integration test cleanups

jpoimboe:
Some integration test cleanups:

- Move the tests to a f22 directory so that we can have distro-specific test suites.
- Create an f22 vagrant box for a consistent test environment.
- Rebase some of the patches so they apply again.

There are still a few outstanding test failures which point to real bugs.  I'll open separate issues for those.


jpoimboe:
vagrant is causing me problems, and isn't quite flexible enough.  May do something like an ansible playbook instead.  Closing for now.

PR #520, opened 2015-10-22
kpatch-build: revisit checking for fentry calls

euspectre:
If it possible in some cases that a kernel function has no relocations
at all (and no __fentry__ call, of course). create-diff-object should
take it into account and output an error instead of segfaulting.

I observed such crashes when experimenting with a modified version of
the patch "tcp_cubic: better follow cubic curve after idle period" in
CentOS 7 x64.

create-diff-object crashed with SIGSEGV when processing
__tcp_fast_path_on() function:

create-diff-object.c:
2602           insn[4] = 0;
2603
> 2604         rela = list_first_entry(&sym->sec->rela->relas, struct rela,
2605                                            list);
2606           if (rela->type != R_X86_64_NONE ||

sym->sec->rela was NULL in that case because __tcp_fast_path_on() has no
relocations. This function was declared inline but was uninlined by GCC.
Such functions have tracing disabled and, hence, no fentry call.

By the way, the previous check (if (insn[0] != 0xf)) passes because the
function does start with 0xf:

0000000000000000 <__tcp_fast_path_on>:
   0:   0f b7 87 18 05 00 00    movzwl 0x518(%rdi),%eax
   7:   81 ce 00 00 10 00       or     $0x100000,%esi
   d:   55                      push   %rbp
   e:   48 89 e5                mov    %rsp,%rbp
  11:   5d                      pop    %rbp
  12:   c1 e0 1a                shl    $0x1a,%eax
  15:   09 c6                   or     %eax,%esi
  17:   0f ce                   bswap  %esi
  19:   89 b7 1c 05 00 00       mov    %esi,0x51c(%rdi)
  1f:   c3                      retq

I moved checking of relocations before the operations on the first
instruction and also added a check that there are actually relocations
for the function.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -127,6 +127,7 @@ struct symbol {
 		int include; /* used in the patched elf */
 		int strip; /* used in the output elf */
 	};
+	int has_fentry_call;
 };
 
 struct rela {
@@ -472,6 +473,24 @@ void kpatch_create_symbol_list(struct kpatch_elf *kelf)
 
 }
 
+/* Check which functions have fentry calls; save this info for later use. */
+static void kpatch_find_fentry_calls(struct kpatch_elf *kelf)
+{
+	struct symbol *sym;
+	struct rela *rela;
+	list_for_each_entry(sym, &kelf->symbols, list) {
+		if (sym->type != STT_FUNC || !sym->sec->rela)
+			continue;
+
+		rela = list_first_entry(&sym->sec->rela->relas, struct rela,
+					list);
+		if (rela->type != R_X86_64_NONE ||
+		    strcmp(rela->sym->name, "__fentry__"))
+			continue;
+
+		sym->has_fentry_call = 1;
+	}
+}
 
 struct kpatch_elf *kpatch_elf_open(const char *name)
 {
@@ -510,6 +529,7 @@ struct kpatch_elf *kpatch_elf_open(const char *name)
 		kpatch_create_rela_list(kelf, sec);
 	}
 
+	kpatch_find_fentry_calls(kelf);
 	return kelf;
 }
 
@@ -1341,6 +1361,7 @@ next:
 void kpatch_verify_patchability(struct kpatch_elf *kelf)
 {
 	struct section *sec;
+	struct symbol *sym;
 	int errs = 0;
 
 	list_for_each_entry(sec, &kelf->sections, list) {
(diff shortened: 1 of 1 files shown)


jpoimboe:
@euspectre Thanks a lot for reporting the issue and submitting the fix!

Some comments:
- I think this check is still not bulletproof.  It's possible for a function to have relocations and not have a fentry call.  It would fall through both of these checks if its first instruction started with 0xf.  I think we should either check for the fentry relocation at offset 0x1 of the function, or check for "0f 00 00 00 00" at the beginning of the function.
- The check would probably be better suited in kpatch_verify_patchability() instead of kpatch_create_mcount_sections().

euspectre:
You are right. I will update the patch, hopefully tomorrow.

jpoimboe:
FYI, no need to close the PR, you can reuse this PR by rebasing the same branch and doing a force push.

euspectre:
I see.

By the way, another question came to mind: why only check if the *patched* functions have fentry call? Perhaps, create-diff-object should check the *original* functions instead? I mean, the original functions that are to be replaced with the patched ones.

I suppose, kpatch_verify_patchability should check if the *original* functions have fentry calls and report an error if some of these does not.

If I understand it correctly, kpatch_create_mcount_sections() prepares placeholders for Ftrace in each new and patched function so that Ftrace could be used for these after the patch has been applied. So kpatch_create_mcount_sections() should check if fentry calls are there. But I think, if a new function does not have fentry call, kpatch_create_mcount_sections could safely skip that function. This is the case with __tcp_fast_path_on() by the way.

euspectre:
And it seems, the relocations for the original functions are no longer available at the time kpatch_verify_patchability() is called. Perhaps, kpatch_elf_teardown/kpatch_elf_free(kelf_base) are involved.

If so, it might be needed to save the first relocation before that. Or, may be, check each function earlier, see if it has fentry call and save the result in struct symbol for later use.

jpoimboe:
> By the way, another question came to mind: why only check if the patched functions have fentry call? Perhaps, create-diff-object should check the original functions instead? I mean, the original functions that are to be replaced with the patched ones.
> 
> I suppose, kpatch_verify_patchability should check if the original functions have fentry calls and report an error if some of these do not.
> 
> If I understand it correctly, kpatch_create_mcount_sections() prepares placeholders for Ftrace in each new and patched function so that Ftrace could be used for these after the patch has been applied. So kpatch_create_mcount_sections() should check if fentry calls are there. But I think, if a new function does not have fentry call, kpatch_create_mcount_sections could safely skip that function. This is the case with __tcp_fast_path_on() by the way.

Ah, right.  We confused two separate issues: fentry for old functions and fentry for new functions.

Yes, the fentry check for the new functions should go in kpatch_create_mcount_sections(), and it shouldn't be a fatal error.

Ideally we would also have a fentry check for old functions.  However, even without that check it will still fail to load the patch at runtime if the old function is missing the fentry call.  (But failing at build time would be much better.)

> And it seems, the relocations for the original functions are no longer available at the time kpatch_verify_patchability() is called. Perhaps, kpatch_elf_teardown/kpatch_elf_free(kelf_base) are involved.
> 
> If so, it might be needed to save the first relocation before that. Or, may be, check each function earlier, see if it has fentry call and save the result in struct symbol for later use.

Yeah, I think either one of those would work.

euspectre:
I have prepared a patch that implements what we discussed above. As you suggested, I reused this pull request and the branch.

euspectre:
Let us close this pull request to avoid confusion. I will open a new one when the patch is ready.

PR #525, opened 2015-10-26
kpatch-build: revisit checking for fentry calls

euspectre:
// Github did not allow me to reopen PR #520 after I force-pushed the changes, so I opened another pull request for these.

create-diff-object now checks if the original functions have fentry calls.
If an original function to be affected by the patch does not have the
fentry call, it cannot be patched. Error is reported in that case.

kpatch_create_mcount_sections() now also takes into account if a changed
or a new function has fentry call. If it does, mcount record is
generated for it as before. If a changed or a new function has no fentry
call, it is not an error in this case.

All this fixes the following issues.

1. If an original function has no fentry call (e.g. a "notrace" function)
but the patched function has it, the original function can not be
patched, but it would only be detected when applying the patch.

2. kpatch_create_mcount_sections() crashed if a patched function had no
relocation at all.

I observed such crashes when experimenting with a modified version of
the patch "tcp_cubic: better follow cubic curve after idle period" in
CentOS 7 x64.

Besides that, for a function with the first instruction starting with
0x0f, it would be incorrectly detemined that the function had fentry call.
The first bytes of the function would be overwritten in that case.

3. create-diff-object output an error if a new (an added) function had
no fentry call. This restriction is not necessary.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -127,6 +127,7 @@ struct symbol {
 		int include; /* used in the patched elf */
 		int strip; /* used in the output elf */
 	};
+	int has_fentry_call;
 };
 
 struct rela {
@@ -472,6 +473,24 @@ void kpatch_create_symbol_list(struct kpatch_elf *kelf)
 
 }
 
+/* Check which functions have fentry calls; save this info for later use. */
+static void kpatch_find_fentry_calls(struct kpatch_elf *kelf)
+{
+	struct symbol *sym;
+	struct rela *rela;
+	list_for_each_entry(sym, &kelf->symbols, list) {
+		if (sym->type != STT_FUNC || !sym->sec->rela)
+			continue;
+
+		rela = list_first_entry(&sym->sec->rela->relas, struct rela,
+					list);
+		if (rela->type != R_X86_64_NONE ||
+		    strcmp(rela->sym->name, "__fentry__"))
+			continue;
+
+		sym->has_fentry_call = 1;
+	}
+}
 
 struct kpatch_elf *kpatch_elf_open(const char *name)
 {
@@ -510,6 +529,7 @@ struct kpatch_elf *kpatch_elf_open(const char *name)
 		kpatch_create_rela_list(kelf, sec);
 	}
 
+	kpatch_find_fentry_calls(kelf);
 	return kelf;
 }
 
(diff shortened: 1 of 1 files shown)


kpatch-build/create-diff-object.c
```diff
@@ -1377,6 +1398,19 @@ void kpatch_verify_patchability(struct kpatch_elf *kelf)
...
+
+	list_for_each_entry(sym, &kelf->symbols, list) {
+		if (sym->type != STT_FUNC || sym->status != CHANGED)
+			continue;
+		if (!sym->has_fentry_call) {
+			log_normal("function %s has no fentry call, unable to patch\n",
+				   sym->name);
+			errs++;
+		}
+	}
```
jpoimboe: I think this won't work, because kpatch_verify_patchability() is called with the patched object, not the original (base) object.  The base object has the original function which will be patched.

Instead of having this check here, I think a similar check will need to be made after kpatch_compare_correlated_elements() and before kpatch_elf_teardown().  For any patched function symbols with status CHANGED, their twin should have sym->has_fentry_call set.

kpatch-build/create-diff-object.c
```diff
@@ -2605,7 +2647,8 @@ void kpatch_create_mcount_sections(struct kpatch_elf *kelf)
 					list);
 		if (rela->type != R_X86_64_NONE ||
 		    strcmp(rela->sym->name, "__fentry__"))
-			ERROR("function '%s' has no fentry call; unable to patch", sym->name);
+			ERROR("%s: the first relocation is not for a fentry call",
+			      sym->name);
```
jpoimboe: I think this check can be removed now.  Or is it still possible to hit this error?

kpatch-build/create-diff-object.c
euspectre: Ah, right.

kpatch-build/create-diff-object.c
euspectre: I cannot say if it is possible to hit the error there. I kept it a kind of a sanity check, like the one for 'index' below:
  if (index != nr)
		ERROR("size mismatch in funcs sections");

If new symbol structures cannot appear in the list somehow after kpatch_find_fentry_calls() is called, then, I think, the check is not necessary.

jpoimboe: Ok.  I'd say let's remove it.

euspectre: Yes, I am checking the updated patch now, will update this pull request if everything is OK.

euspectre:
Here is the v2 of the patch. I removed that excessive error check and moved the checking for fentry calls after kpatch_compare_correlated_elements(). The twins for the patched symbols should be valid then.

jpoimboe:
:+1:

PR #526, opened 2015-10-27
kpatch-build: consolidate build-artifacts and place into CACHEDIR

arges:
It would be very useful to be able to direct build artifacts into specific directories and files. This pull request adds these features to kpatch-build.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -37,16 +37,16 @@
 # - Runs kpatch tools to create and link the patch kernel module
 
 BASE="$PWD"
-LOGFILE="/tmp/kpatch-build-$(date +%s).log"
 SCRIPTDIR="$(readlink -f $(dirname $(type -p $0)))"
 ARCHVERSION="$(uname -r)"
 CPUS="$(getconf _NPROCESSORS_ONLN)"
-CACHEDIR="$HOME/.kpatch"
+CACHEDIR="${CACHEDIR:-$HOME/.kpatch}"
 SRCDIR="$CACHEDIR/src"
 OBJDIR="$CACHEDIR/obj"
 TMPSRCDIR="$CACHEDIR/tempsrc"
 VERSIONFILE="$CACHEDIR/version"
-TEMPDIR=
+TEMPDIR="$CACHEDIR/tmp"
+LOGFILE="$TEMPDIR/build.log"
 APPLIEDPATCHFILE="kpatch.patch"
 DEBUG=0
 SKIPGCCCHECK=0
@@ -274,7 +274,10 @@ while [[ $# -gt 0 ]]; do
 	shift
 done
 
-TEMPDIR="$(mktemp -d /tmp/kpatch-build-XXXXXX)" || die "mktemp failed"
+# ensure cachedir and tempdir are setup properly and cleaned
+mkdir -p "$TEMPDIR" || die "Couldn't create $TEMPDIR"
+rm -rf "$TEMPDIR"/*
+
 trap cleanup EXIT INT TERM
 
 KVER=${ARCHVERSION%%-*}


spartacus06:
:+1:

jpoimboe:
@arges even with a custom TEMPDIR, the directory will be removed on exit in cleanup().  Is that what you want?

arges:
I could make the --tempdir argument the name of the directory to be created. This way a user doesn't accidentally clobber an important directory when not using the -d flag as well.

-               [[ ! -d $TEMPDIR ]] && die "$TEMPDIR does not exist"
+               [[ -d $TEMPDIR ]] && die "$TEMPDIR already exists!"
+               mkdir $TEMPDIR || die "Couldn't create $TEMPDIR"

If this is ok with you, I'll update the branch.

jpoimboe:
Looks good.  :+1:

jpoimboe:
@arges, what instead of /tmp/kpatch-build-XXXXXX, we were to make the TEMPDIR always be ~/.kpatch/tmp ?  Then I think we wouldn't need --tempdir anymore?

I'm considering that because of some out-of-space issues I've been seeing on /tmp when building a big patch which results in rebuilding and comparing the entire tree.

jpoimboe:
Similarly, we might want to put the log file in ~/.kpatch/log ?  I'm not sure if there's any real value in giving them unique names so that they can be saved indefinitely.

spartacus06:
@jpoimboe i like that idea.  then the user could symlink ~/.kpatch/tmp to /tmp or where ever. or mount it as its own tmpfs.

jpoimboe:
@spartacus06 Yeah, to support that use case we could make a permanent ~/.kpatch/tmp directory which has a ~/.kpatch/tmp/kpatch-tmp directory inside of it.  The kpatch-tmp directory could be wiped out and recreated for each kpatch-build run.  That way, even if the "-d" option is used, we never leak more than a single tmp directory.  Thoughts?

spartacus06:
@jpoimboe yeah, we can either create a subdirectory in the tmpdir, remove the contents of the tmpdir on cleanup instead of the tmpdir itself.

arges:
Having this in ~/.kpatch/tmp would make scripting easier to know where the build log and artifacts are stored. So this would work for me. My suggestion would be to make the "~/.kpatch/" variable changeable via an environment variable or flag.

jpoimboe:
> @jpoimboe yeah, we can either create a subdirectory in the tmpdir or remove the contents of the tmpdir on cleanup instead of the tmpdir itself.

True, either works for me.

> Having this in ~/.kpatch/tmp would make scripting easier to know where the build log and artifacts are stored. So this would work for me. My suggestion would be to make the "~/.kpatch/" variable changeable via an environment variable or flag.

Making ~/.kpatch changeable is fine with me (as long as you plan on using that option).

arges:
Yes, plan on using that option. I can hack on this and resubmit the pull request.

arges:
Ok pushed latest patch which has the above comments. Let me know if any adjustments need to be made.

jpoimboe:
Looking good.  How about we simplify things even further by getting rid of TEMPDIR (~/.kpatch/tmp/kpatch-tmp) and just putting everything in TMPDIR (~/.kpatch/tmp) as @spartacus06 suggested?  And then just `mkdir -p $TMPDIR` and `rm -rf $TMPDIR/*` on startup?

arges:
Ok another version pushed.

kpatch-build/kpatch-build
```diff
@@ -274,7 +274,13 @@ while [[ $# -gt 0 ]]; do
 	shift
 done
 
-TEMPDIR="$(mktemp -d /tmp/kpatch-build-XXXXXX)" || die "mktemp failed"
+# ensure cachedir and tempdir are setup properly
+mkdir -p "$CACHEDIR" || die "Couldn't create $CACHEDIR"
+mkdir -p "$TEMPDIR" && rm -rf $TEMPDIR/* || true
+
+# remove old LOGFILE
+rm -f $LOGFILE
```
jpoimboe: A few nits:
- it should die if `mkdir -p $TEMPDIR` fails
- no need to `mkdir -p $CACHEDIR` since `mkdir -p $TEMPDIR` will do that
- similarly, no need to `rm -f $LOGFILE` since `rm -rf $TEMPDIR/*` will do that

So I think this can be reduced to:
```
mkdir -p "$TEMPDIR" || die "Couldn't create $TEMPDIR"
rm -rf "$TEMPDIR"/*
```

arges:
Ok nit-fixes added. Hopefully this works!

jpoimboe:
:+1:

Thanks @arges !

PR #537, opened 2015-10-30
test: integration test cleanup

jpoimboe:
- fix kpatch tool path in multiple.test
- remove "kpatch replace" tests (replace is deprecated)


test/integration/f22/Makefile
```diff
@@ -36,4 +36,4 @@ remote_sync: remote_setup
 	ssh $(SSH_USER)@$(SSH_HOST) "cd kpatch-test/kpatch && make"
 
 remote_slow: remote_sync
-	ssh $(SSH_USER)@$(SSH_HOST) "cd kpatch-test/kpatch/test/integration/f22 && make slow"
+	ssh $(SSH_USER)@$(SSH_HOST) "cd kpatch-test/kpatch/test/integration/f22 && make local"
```
spartacus06: The target is "remote_slow" so why not "make slow"?  I get that the local target requires the slow target and so the end result is the same.  Just wondering what the reason for the change was.

jpoimboe: Because of
```
all:
 	$(error please specify local or remote)
```
So my intention is that the main user visible targets are "local" and "remote".  Where "local" defaults to "slow" and "remote" defaults to "remote_slow".  The other targets like "slow", "quick", and "cached" are hidden for more advanced users.

jpoimboe: But um, yeah I see your point.  The target is "remote_slow" so why not "make slow"?  You're probably right :-)

jpoimboe:
Rebased based on the comment.

PR #539, opened 2015-10-30
Minor comment changes

arges:


spartacus06:
I think you probably meant for your PR to only contain the last commit "kpatch-build: edit script description".  Looks like you may have rebased or something.

arges:
This should be better now.

spartacus06:
:+1:

PR #541, opened 2015-11-03
kpatch-build: fix searching for the sizes of special structures

euspectre:
readelf -wi may output trailing spaces in the lines with section names ('alt_instr', etc.). 

The regexps should take this into account, otherwise kpatch-build may fail with error:
    "can't find special struct size"


jpoimboe:
:+1:

PR #543, opened 2015-11-03
kpatch-build: clean up rpmbuild tmp directory handling

jpoimboe:
Setting HOME in a subshell is too hacky.  Instead just pass the rpmbuild
directory to the rpm and rpmbuild commands.


jpoimboe:
Added a commit to address the review comment.

PR #546, opened 2015-11-12
kpatch-build: allow to specify the target kernel version

euspectre:
If the source tree of the kernel is used rather than its SRPM, the
version of the currently running kernel will be assumed as the version
of the target kernel (ARCHVERSION=$(uname -r)). This can be
inconvenient if the user actually prepares the binary patch for a
different kernel: the paths to some files used in the build contain
$ARCHVERSION, etc.

This patch adds -k|--kernel-version option to kpatch-build, so that it
becomes possible to specify the target kernel version explicitly.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -206,6 +206,7 @@ usage() {
 	echo "		-r, --sourcerpm	   Specify kernel source RPM" >&2
 	echo "		-s, --sourcedir	   Specify kernel source directory" >&2
 	echo "		-c, --config	   Specify kernel config file" >&2
+	echo "		-k, --kernel-version   Specify target kernel version" >&2
 	echo "		-v, --vmlinux	   Specify original vmlinux" >&2
 	echo "		-t, --target	   Specify custom kernel build targets" >&2
 	echo "		-d, --debug	   Keep scratch files in /tmp" >&2
@@ -213,7 +214,7 @@ usage() {
 	echo "		                   (not recommended)" >&2
 }
 
-options=$(getopt -o hr:s:c:v:t:d -l "help,sourcerpm:,sourcedir:,config:,vmlinux:,target:,debug,skip-gcc-check" -- "$@") || die "getopt failed"
+options=$(getopt -o hr:s:c:k:v:t:d -l "help,sourcerpm:,sourcedir:,config:,kernel-version:,vmlinux:,target:,debug,skip-gcc-check" -- "$@") || die "getopt failed"
 
 eval set -- "$options"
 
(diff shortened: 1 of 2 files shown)


jpoimboe:
Can you clarify why you need ARCHVERSION for the `--sourcedir` option?  Is it because it can't find the Module.symvers file?

euspectre:
It finds a wrong Module.symvers (the one for the currently running kernel) in my case, because ARCHVERSION remains $(uname -r) without this patch.

euspectre:
Oops, I lost the arch in ARCHVERSION somehow. I have restored it now and updated the patch.

jpoimboe:
Can we instead get the kernel version from an existing source, like `make kernelversion` in the kernel tree, or from the vmlinux binary?

euspectre:
May be, haven't checked this yet. 

There is another problem however. 

If I download the SRPM with CentOS kernel, say 3.10.0-229.7.2.el7, do rpmbuild -bp like kpatch-build does, the kernel source tree is prepared. Then I pass the path to the prepared source tree to kpatch-build. 

In this case, "3.10.0" is used as the kernel version and the vermagic of the module, not "3.10.0-229.7.2.el7" as it should be. As a result, 'kpatch install' will fail to install the built patch module, for example.

So, I suppose, `make kernelversion` alone won't help.

The kernel config files do not always specify the full version as well (my first attempt was to extract it from there).

As for vmlinux with debug info that kpatch-build needs anyway - I don't know yet.

jpoimboe:
I guess that info could be figured out from parsing the Makefile
```
VERSION = 3
PATCHLEVEL = 10
SUBLEVEL = 0
EXTRAVERSION =
NAME = Unicycling Gorilla
RHEL_MAJOR = 7
RHEL_MINOR = 1
RHEL_RELEASE = 229
```
But I guess it would need distro-specific code to figure out the full version, which is less than ideal.

vmlinux seems to know:
```
# strings /lib/debug/lib/modules/3.10.0-229.el7.x86_64/vmlinux |grep "Linux version"
Linux version 3.10.0-229.el7.x86_64 (mockbuild@x86-035.build.eng.bos.redhat.com) (gcc version 4.8.3 20140911 (Red Hat 4.8.3-7) (GCC) ) #1 SMP Thu Jan 29 18:37:38 EST 2015
```

Not sure if there's a less hacky way to do that.

euspectre:
Oh, nice. So we could extract it from vmlinux then. 
If this way is not distro-specific, I think, adding an option to kpatch-build is not necessary.

jpoimboe:
Yeah, I think vmlinux is the way to go.

euspectre:
Hmm. As far as I can see in kpatch-build, if vmlinux is not specified explicitly, ARCHVERSION is needed to locate it in /usr/lib/debug/<...>. In such cases, obviously, vmlinux cannot be used to get the target kernel version.

jpoimboe:
In the case of `--sourcedir`, it doesn't look for vmlinux in /usr/lib/debug.  Instead it looks in the kernel source tree (unless `--vmlinux` is used).

euspectre:
Indeed. OK, closing.

PR #547, opened 2015-11-12
modification of correlating static local variables

zhouchengming1:
This patch has the same restrictions as before:
    - the static local variables can't be removed;
    - the order in which they're referenced remains the same.


zhouchengming1:
Only the second commit is useful, please disregard the other two commits.
I don't know how github created three commits. : (

jpoimboe:
@zhouchengming1 You probably need to rebase your commit onto the master branch.

jpoimboe:
@zhouchengming1 Thanks for the patch.  I like it, but I want to add some additional protections to ensure that each static local variable is used in the same functions between the old and new object.  Soon I'll submit a new pull request which has your patch and a few others which extend it.

PR #548, opened 2015-11-13
kpatch-build: fix gcc_version_check

libin2015:
gcc version string format may be 'gcc (xxx xxx) x.x.x [xxx]'
fix gcc_version_check to adapt to it.


jpoimboe:
Seems to work on Fedora and Ubuntu.

:+1:

PR #550, opened 2015-11-13
kpatch-build: fix find_parent_obj

libin2015:
When find kobj, it should use 'cat changed_objs' to get the changed
objects, in order to process the following object format:
a/b/c/../../object.o. If using patched dir to get changed object,
the object will be a/object.o, but it is a/b/c/../../object.o in
*.cmd file.
This patch also fix the find_parent_obj that change the format
'a/b/c/../../object.o' to 'a/object.o' in deep find, otherwise
it will fail with "two parent matches for *.o".

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -142,18 +142,21 @@ gcc_version_check() {
 
 find_parent_obj() {
 	dir=$(dirname $1)
+	absdir=$(readlink -f $dir)
+	pwddir=$(readlink -f .)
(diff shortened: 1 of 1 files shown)


jpoimboe:
@libin2015 Do you have a specific example for how to reproduce these errors?

libin2015:
I don't find a suitable example in kernel code for x86, and i provide a simple module to reproduce this error.

[kpatch-test-module](https://github.com/libin2015/kpatch/tree/kpatch-test-module)

jpoimboe:
:+1:

PR #551, opened 2015-11-13
kpatch-build: reduce dependency on bash version >4.0

libin2015:
Before this patch, kpatch_build dependends on bash version >4.0
that support declare -A. This patch remove this dependency by
replacing dict(declare -A) with array.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -506,7 +506,7 @@ cd "$TEMPDIR/patched"
 FILES="$(find * -type f)"
 cd "$TEMPDIR"
 mkdir output
-declare -A objnames
+declare -a objnames
 CHANGED=0
 ERROR=0
 for i in $FILES; do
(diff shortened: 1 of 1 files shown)


kpatch-build/kpatch-build
```diff
@@ -554,7 +554,10 @@ if [[ $CHANGED -eq 0 ]]; then
 fi
 
 echo -n "Patched objects:"
-for i in "${!objnames[@]}"; do echo -n " $(basename $i)"; done
+for i in $(echo "${objnames[@]}" | tr ' ' '\n' | sort -u | tr '\n' ' ')
```
jpoimboe: Why reverse the sort order?

libin2015: In order to remove the duplicate in array.

jpoimboe: Oh, right, sorry.  I don't even remember what I was thinking.  Is it possible for there to be duplicates?

spartacus06: Also "sort -u" only works in the expected way if the input is already sorted.  so might need to "sort | sort -u" for the expected result.

libin2015: @jpoimboe If different objects in one module have been changed, there will be duplicates.

libin2015: @spartacus06 "sort -u" can sort and retain only unique elements, only with -c it will check for strict ordering. Right?

spartacus06: @libin2015  ah you are right.  i was thinking about the uniq command.  sort -u keeps the last occurrence if there are duplicates (for the record).

spartacus06:
:+1:

Issue #553, opened 2015-11-13
Error: data section .data..read_mostly selected for inclusion


rosslagerwall:
The following patch fails to compile against 3.10.0-229.14.1.el7.x86_64:
```
diff --git a/arch/x86/kernel/acpi/boot.c b/arch/x86/kernel/acpi/boot.c
index 45a51de..3b6852f 100644
--- a/arch/x86/kernel/acpi/boot.c
+++ b/arch/x86/kernel/acpi/boot.c
@@ -108,9 +108,7 @@ static unsigned int gsi_to_irq(unsigned int gsi)
 	unsigned int i;
 
 	for (i = 0; i < NR_IRQS_LEGACY; i++) {
-		if (isa_irq_to_gsi[i] == gsi) {
-			return i;
-		}
+		printk("%u\n", isa_irq_to_gsi[i]);
 	}
 
 	/* Provide an identity mapping of gsi == irq
```

The issue here is that the section reference to .data..read_mostly does not get replaced with the correct symbol because of the check in https://github.com/dynup/kpatch/blob/master/kpatch-build/create-diff-object.c#L1372 and the way that GCC compares the ptr with the address at the end of the array (one past the index).

```
  [90] .data..read_mostly PROGBITS        0000000000000000 0033c0 000040 00  WA  0   0 32

Relocation section '.rela.text.gsi_to_irq' at offset 0x28738 contains 6 entries:
    Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
0000000000000001  000000b900000000 R_X86_64_NONE          0000000000000000 __fentry__ - 4
0000000000000012  000000310000000b R_X86_64_32S           0000000000000000 .data..read_mostly + 0
0000000000000027  000000120000000b R_X86_64_32S           0000000000000000 .rodata.str1.1 + 0
0000000000000030  000000d800000002 R_X86_64_PC32          0000000000000000 printk - 4
0000000000000037  000000310000000b R_X86_64_32S           0000000000000000 .data..read_mostly + 40
000000000000003f  000000df00000002 R_X86_64_PC32          0000000000000000 gsi_top - 4
```

I'm struggling to think of a decent way of fixing this. You could "bundle" the sections manually....

#define __read_mostly(name) __attribute__((__section__(".data.read_mostly." # name))) name
static u32 __read_mostly(isa_irq_to_gsi)[NR_IRQS_LEGACY] = {

jpoimboe:
Thanks, this is an interesting problem.

[...]

PR #554, opened 2015-11-13
create-diff-object: handle reference to end of section

jpoimboe:
Deal with a special case where gcc needs a pointer to the address at the end of
a data section.

This is usually used with a compare instruction to determine when to end a
loop.  The code doesn't actually dereference the pointer so this is "normal"
and we just replace the section reference with a reference to the last symbol
in the section.

Note that this only catches the issue when it happens at the end of a section.
It can also happen in the middle of a section.  In that case, the wrong symbol
will be associated with the reference.  But that's ok because:

1) This situation only occurs when gcc is trying to get the address of the
   symbol, not the contents of its data; and

2) Because kpatch doesn't allow data sections to change, &(var1+sizeof(var1))
   will always be the same as &var2.

Fixes: #553

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1369,8 +1369,45 @@ void kpatch_replace_sections_syms(struct kpatch_elf *kelf)
 				start = sym->sym.st_value;
 				end = sym->sym.st_value + sym->sym.st_size;
 
-				if (rela->addend + add_off < start ||
-				    rela->addend + add_off >= end)
+				if (!is_text_section(sym->sec) &&
+				    rela->type == R_X86_64_32S &&
+				    rela->addend == sym->sec->sh.sh_size &&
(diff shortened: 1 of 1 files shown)


kpatch-build/create-diff-object.c
```diff
@@ -1369,8 +1369,45 @@ void kpatch_replace_sections_syms(struct kpatch_elf *kelf)
...
+					 * 1) This situation only occurs when
+					 *    gcc is trying to get the address
+					 *    of the symbol, not the contents
+					 *    of its data; and
+					 *
+					 * 2) Because kpatch doesn't allow data
+					 *    sections to change,
+					 *    &(var1+sizeof(var1)) will always
+					 *    be the same as &var2.
+					 */
```
spartacus06: Is this an empty if block?

jpoimboe: Yes, unless you count the large comment :-)

Issue #545, opened 2015-11-11
Questions about the current solution of correlating static local variables


zhouchengming1:
1. Why undo the previous correlation in the iteration of base->symbols?
   If               xxx.123  --->   xxx.123 (previous correlation by coincidence)
                    xxx.256  --->   xxx.256 (previous correlation by coincidence) 
   But real     xxx.123  --->  xxx.256  
   In this case, the code doesn't work. Because when find patched_sym for xxx.123, 
   the xxx.256 in patched_object hasn't been de-correlated.
   So I think should undo all previous correlation before the new correlation.
2. Is the iteration of base-sections wrong?
   old-object         |         new-object
   func1                |         func1
   xxx.123             |         xxx.123        (inline)
   func2                |          func2
   xxx.256             |          xxx.256
   xxx.123              |         xxx.123        (inline)
   
   When find patched_sym for xxx.123, first find xxx.123 in func1 of new-object,
   But then find xxx.256 in func2 of new-object.
   So I think should not iterate the base-sections, when find one, just go out to next symbol.

jpoimboe:
> Why undo the previous correlation in the iteration of base->symbols?
> If xxx.123 ---> xxx.123 (previous correlation by coincidence)
> xxx.256 ---> xxx.256 (previous correlation by coincidence)
> But real xxx.123 ---> xxx.256
>
> In this case, the code doesn't work. Because when find patched_sym for xxx.123,
> the xxx.256 in patched_object hasn't been de-correlated.
> So I think should undo all previous correlation before the new correlation.

Ah, right.  Instead of uncorrelating and correlating in the same loop, they should be done in two separate loops, so that *all* of the uncorrelations are done before starting the correlations.

> Is the iteration of base-sections wrong?
> old-object | new-object
> func1 | func1
> xxx.123 | xxx.123 (inline)
> func2 | func2
> xxx.256 | xxx.256
> xxx.123 | xxx.123 (inline)
> 
> When find patched_sym for xxx.123, first find xxx.123 in func1 of new-object,
> But then find xxx.256 in func2 of new-object.
> So I think should not iterate the base-sections, when find one, just go out to next symbol.

Ah, right, this is another problem.  I need to think about it a little bit.  Can you clarify your suggested solution?

This static local correlation has been really hard to get right :-/

zhouchengming1:
@jpoimboe , there is a solution that works fine in the same restrictions.
And I just have implemented it simply, also did some test on it.
[...]

PR #555, opened 2015-11-13
create-diff-object: more static local variable rework

jpoimboe:
Refine the static local variable handling again.  This builds on a
previous patch by Zhou Chengming.

This fixes the following bugs reported by Zhou:

1.          xxx.123 ---> xxx.123 (previous correlation by coincidence)
            xxx.256 ---> xxx.256 (previous correlation by coincidence)
   But real xxx.123 ---> xxx.256

   In this case, the code doesn't work. Because when find patched_sym for
   xxx.123, the xxx.256 in patched_object hasn't been de-correlated.

2. old-object | new-object
        func1 | func1
      xxx.123 | xxx.123 (inline)
        func2 | func2
      xxx.256 | xxx.256
      xxx.123 | xxx.123 (inline)

   When find patched_sym for xxx.123, first find xxx.123 in func1 of new-object,
   But then find xxx.256 in func2 of new-object.
   So I think should not iterate the base-sections, when find one, just go out to next symbol.

Both of these problems can be fixed by splitting the code up into
multiple passes:

  1. uncorrelate all static locals
  2. correlate all static locals
  3. ensure each static local is referenced by all the same sections in
     both objects
  4. print warning on any new static locals

Fixes: #545

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1086,193 +1086,206 @@ static struct symbol *kpatch_find_static_twin(struct section *sec,
 	return NULL;
 }
 
+static int kpatch_is_normal_static_local(struct symbol *sym)
+{
+	if (sym->type != STT_OBJECT || sym->bind != STB_LOCAL)
+		return 0;
+
+	if (!strchr(sym->name, '.'))
+		return 0;
+
+	if (is_special_static(sym))
+		return 0;
+
+	return 1;
+}
+
 /*
  * gcc renames static local variables by appending a period and a number.  For
  * example, __foo could be renamed to __foo.31452.  Unfortunately this number
  * can arbitrarily change.  Correlate them by comparing which functions
  * reference them, and rename the patched symbols to match the base symbol
  * names.
+ *
+ * Some surprising facts about static local variable symbols:
+ *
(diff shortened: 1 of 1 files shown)


kpatch-build/create-diff-object.c
```diff
@@ -1086,193 +1086,206 @@ static struct symbol *kpatch_find_static_twin(struct section *sec,
...
-			    is_debug_section(sec))
+		list_for_each_entry(rela, &sec->relas, list) {
+
+			sym = rela->sym;
+			if (!kpatch_is_normal_static_local(sym))
 				continue;
 
-			list_for_each_entry(rela, &sec->relas, list) {
+			if (sym->twin)
+				continue;
```
spartacus06: i'm not seeing how this case can be hit.  we set all the normal_static_local sym->twin = NULL in the previous code block.  there doesn't seem do be a way for sym->twin to get set out of order.  am i missing something?

jpoimboe: A static local can be referenced multiple times.  The first time it's referenced, it goes through an iteration of this loop and gets its twin set.  If it's referenced again, it hits this check on a future iteration.

Incidentally that's why the correlations have to be in a separate loop from the uncorrelations.

PR #560, opened 2015-11-17
Get kernel version from vmlinux if the kernel source tree is used

euspectre:
If a kernel SRPM is used to get the kernel sources, the target kernel
version is determined from the name of the SRPM.

One cannot obtain the target kernel version this way if the source tree
is used instead of an SRPM, so let us extract that information from
vmlinux.


jpoimboe:
:+1:

PR #561, opened 2015-11-17
kpatch-build: fix gcc_version_check: both "GNU" and "GCC" are possible

euspectre:
This fix is an addition to 9fedd0d283 "kpatch-build: fix
gcc_version_check".

On some systems, the GCC version stored in vmlinux may have the
following format:
  (GNU) 4.8.3 20140911 (Red Hat 4.8.3-9)
while GCC returns
  (GCC) 4.8.3 20140911 (Red Hat 4.8.3-9)

As a result, binary patches cannot be built, although the compiler is
the same.

gcc_version_check() now takes this into account.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -119,8 +119,8 @@ find_dirs() {
 
 gcc_version_check() {
 	# ensure gcc version matches that used to build the kernel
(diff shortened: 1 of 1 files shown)


jpoimboe:
:+1:

jpoimboe:
@spartacus06 Merging this one now because it fixes a regression (kpatch-build always reports a gcc mismatch).

PR #563, opened 2015-11-18
revert gcc check changes

### LINKED ISSUE #562: ERROR: gcc >= 4.8 required
Seeing this error on RHEL 7.

jpoimboe:
This reverts #561 and #548, as gcc version checking is still broken.

@libin2015 , can you come up with an approach that works on all supported distros (Fedora, RHEL/CentOS 7, Ubuntu, Debian)?

Fixes #562.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -119,9 +119,9 @@ find_dirs() {
 
 gcc_version_check() {
 	# ensure gcc version matches that used to build the kernel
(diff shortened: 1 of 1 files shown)


libin2015:
> @libin2015 , can you come up with an approach that works on all supported distros (Fedora, RHEL/CentOS 7, Ubuntu, Debian)?

Ensure gcc version > 4.8 is to ensure gcc support -mfentry feature, right?

Issue #246, opened 2014-06-16
gcc version checks


goldyfruit:
Hi,

I try to use kpatch on Debian GNU/Linux Wheezy with a 3.14.5 kernel. The kernel compilation works but when kpatch-build try to create the patch, it fail !

```
+ /usr/local/libexec/kpatch/create-diff-object orig/fs/proc/meminfo.o patched/fs/proc/meminfo.o /usr/lib/debug/lib/modules/3.14.5-kpatch/vmlinux output/fs/proc/meminfo.o
+ tee -a /tmp/kpatch-build-1402941756.log
meminfo.o: meminfo.o: changed section .rela.text not selected for inclusion
meminfo.o: meminfo.o: changed section .rodata.str1.8 not selected for inclusion
meminfo.o: 2 unsupported section change(s)
/usr/local/libexec/kpatch/create-diff-object: unreconcilable difference
+ [[ 2 -eq 0 ]]
+ die
+ [[ -z '' ]]
+ echo 'ERROR: kpatch build failed. Check /tmp/kpatch-build-1402941756.log for more details.'
ERROR: kpatch build failed. Check /tmp/kpatch-build-1402941756.log for more details.
+ exit 1
+ cleanup
+ [[ 1 -eq 0 ]]
+ [[ -e /media/.kpatch/src/applied-patch ]]
```

The command line used : kpatch-build -s /usr/src/linux-source-3.14.5-kpatch -c /boot/config-3.14.5-kpatch -d meminfo-string.patch

I found some discussions about _.rela._ issues but everytime it's fixed by a commit. I use the last kpatch version.

Thanks for your help.

xdel:
Just for search simplicity` sake, the necessary commit is https://github.com/gcc-mirror/gcc/commit/e83660d17a3ed9fb20339118d07452f5d95b27b2

PR #565, opened 2015-11-19
kpatch-build: fix gcc_version_check

libin2015:
gcc version string format may be 'gcc (xxx xxx) x.x.x [xxx]'
fix gcc_version_check to adapt to it.

Signed-off-by: Li Bin huawei.libin@huawei.com
Signed-off-by: Evgenii Shatokhin eshatokhin@odin.com

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -119,9 +119,9 @@ find_dirs() {
 
 gcc_version_check() {
 	# ensure gcc version matches that used to build the kernel
-	local gccver=$(gcc --version |head -n1 |cut -d' ' -f3-)
-	local kgccver=$(readelf -p .comment $VMLINUX |grep GCC: | tr -s ' ' | cut -d ' ' -f6-)
-	if [[ $gccver != $kgccver ]]; then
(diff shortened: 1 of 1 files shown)


jpoimboe:
What's the `xxx` value in `(xxx xxx)`?  Is it a number that should be compared, or can it be stripped out?

Just trying to figure out if there's a cleaner way we can do this.

libin2015:
> What's the xxx value in (xxx xxx)? Is it a number that should be compared, or can it be stripped out?

Such as on suse system
$ gcc --version
(SUSE Linux) 4.3.4 [gcc-4_3-branch revision 152973]

sm00th:
> What's the xxx value in (xxx xxx)? Is it a number that should be compared, or can it be stripped out?
On debian it contains a package version:
gcc (Debian 5.2.1-23) 5.2.1 20151028

jpoimboe:
I think there are just too many version string conventions to be able to parse the string sanely.  Probably a better approach would be to compile a dummy .c file and ensure that `readelf -p .comment <file> |grep GCC` matches for both the dummy .o and the vmlinux.

As for the gcc >= 4.8 check, mfentry isn't the right check.  Instead it should check if `-ffunction-sections -pg` produces a warning.  See #246 for more info.

PR #569, opened 2015-11-26
readme: add support for Fedora 23

terrywang:
Add support for Fedora 23, `dnf` finally replaced `yum` as the default package manager.

Tested and confirmed to work on Fedora 23 x86_64 (Server Edition).

> BTW: should we remove Fedora 21 or leave it?

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,29 @@ Installation
 
 ###Prerequisites
 
+####Fedora 23
+
+*NOTE: You'll need about 15GB of free disk space for the kpatch-build cache in
+`~/.kpatch` and for ccache.*
+
+Install the dependencies for compiling kpatch:
+
+```bash
+sudo dnf install gcc kernel-devel elfutils elfutils-devel
+```
+
+Install the dependencies for the "kpatch-build" command:
+
+```bash
+sudo dnf install rpmdevtools pesign yum-utils openssl wget numactl-devel
+sudo dnf builddep kernel
+sudo dnf debuginfo-install kernel
+
+# optional, but highly recommended
+sudo dnf install ccache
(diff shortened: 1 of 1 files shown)


jpoimboe:
> BTW: should we remove Fedora 21 or leave it?

I'd say let's remove it.  Would you mind adding a commit to do so?

terrywang:
@jpoimboe Done, we are up-to-date now ;-D

jpoimboe:
Thanks!
:+1:

README.md
```diff
@@ -265,7 +265,7 @@ Quick start
 
 > NOTE: While kpatch is designed to work with any recent Linux
 kernel on any distribution, the `kpatch-build` command has **ONLY** been tested
-and confirmed to work on Fedora 20, RHEL 7, Oracle Linux 7, CentOS 7 and Ubuntu 14.04.
+and confirmed to work on Fedora {20,21,23}, RHEL 7, Oracle Linux 7, CentOS 7 and Ubuntu 14.04.
```
spartacus06: Do we want to say "Fedora 20 and later"? 20,21,23 gives the impression that 22 was not supported... which is weird.

terrywang: Good point, It can be misleading. I personally haven't tested `kpatch` on Fedora 22 (technically it should, just don't want to take it for granted) that's why I didn't put it in the list. Let me know if any of you have tested on Fedora 22.

If so, I'll add 1 more commit to the PR ;-D

spartacus06: @terrywang I have tested it and it works

terrywang: @spartacus06 Thanks! Updated again.

PR #571, opened 2015-12-07
Use Module.symvers to get rid of "external" dynrelas

flaming-toast:
Use Module.symvers to remove external dynrelas. Instead of using an external flag, the  `kpatch_patch_dynrela` struct has a new field `sym_objname` that will eventually be passed to `klp_find_object_symbol()` (in livepatch) to find the symbol. `sym_objname` is simply the object that the symbol (referenced by the dynrela) belongs to. For what used to be the "external" case, create-diff-object will now call `find_exported_symbol_objname()` which will parse Module.symvers and look for the module/object an exported symbol belongs to. If it is not there, then `sym_objname` will be set to the patch module's name. 

Still a WIP as I haven't checked to see what modifications are still required in the core kpatch module. It works with livepatch, but requires a new upstream patch to remove the external flag check and replace it with a single `klp_find_object_symbol` call (should be sent out soon). So far I've tested it with https://github.com/dynup/kpatch/blob/master/test/integration/f22/module-call-external.patch with the modified livepatch, along with the meminfo and a patch with multiple objs for good measure.

A couple of potential issues:
- create-diff-object now takes 6 arguments, which is a bit yucky. (the new ones being, the path to the Module.symvers file, and the second being the name of the to-be-built patch module)
- The Module.symvers parsing code is based off of `scripts/mod/modpost.c`, with only the parsing functions taken and a few modifications to make them fit create-diff-object's needs. Didn't want to rewrite code that was already there. But I'm not sure how licensing works in this case (i.e. should I retain the original copywrites from modpost.c and append a "Modified by .." statement?)

### DIFF
--- a/kmod/patch/kpatch-patch.h
+++ b/kmod/patch/kpatch-patch.h
@@ -37,7 +37,7 @@ struct kpatch_patch_dynrela {
 	unsigned long type;
 	char *name;
 	char *objname;
-	int external;
+	char *sym_objname;
 	int addend;
 };
 
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -69,6 +69,8 @@
 		printf(format, ##__VA_ARGS__); \
 })
 
+extern char *find_exported_symbol_objname(const char *fname, char *symname);
+
 char *childobj;
 
 enum loglevel {
@@ -2592,7 +2594,8 @@ static int kpatch_is_core_module_symbol(char *name)
 
 void kpatch_create_dynamic_rela_sections(struct kpatch_elf *kelf,
                                          struct lookup_table *table, char *hint,
-                                         char *objname)
+                                         char *objname, char *mod_symvers,
+                                         char *pmod)
 {
 	int nr, index, objname_offset;
 	struct section *sec, *dynsec, *relasec;
@@ -2600,7 +2603,8 @@ void kpatch_create_dynamic_rela_sections(struct kpatch_elf *kelf,
 	struct symbol *strsym;
 	struct lookup_result result;
 	struct kpatch_patch_dynrela *dynrelas;
-	int vmlinux, external;
+	int vmlinux;
+	char *sym_objname, *tmp_objname;
 
 	vmlinux = !strcmp(objname, "vmlinux");
 
@@ -2648,7 +2652,12 @@ void kpatch_create_dynamic_rela_sections(struct kpatch_elf *kelf,
 			if (kpatch_is_core_module_symbol(rela->sym->name))
 				continue;
 
-			external = 0;
+			/*
+			 * Default: sym_objname is the component being patched
+			 * (vmlinux or module). If it's an "external" symbol
+			 * sym_objname will get reassigned appropriately
+			 */
+			sym_objname = objname;
 
 			if (rela->sym->bind == STB_LOCAL) {
 				/* An unchanged local symbol */
@@ -2695,13 +2704,22 @@ void kpatch_create_dynamic_rela_sections(struct kpatch_elf *kelf,
 				 * patched.
 				 */
 				if (lookup_global_symbol(table, rela->sym->name,
-							 &result))
+							 &result)) {
 					/*
-					 * Not there, assume it's either an
-					 * exported symbol or provided by
-					 * another .o in the patch module.
+					 * Not there, see if the symbol is
+					 * exported, and set sym_objname to the
+					 * object the exported symbol belongs to
 					 */
-					external = 1;
+					tmp_objname = find_exported_symbol_objname(mod_symvers,
+										   rela->sym->name);
+					if (tmp_objname) {
+						 /* Strip the path, only want the module name */
+						sym_objname = basename(tmp_objname);
+					} else {
+						/* Not exported, must be in another .o in patch module */
+						sym_objname = pmod;
+					}
+				}
 			}
 			log_debug("lookup for %s @ 0x%016lx len %lu\n",
 			          rela->sym->name, result.value, result.size);
@@ -2714,7 +2732,6 @@ void kpatch_create_dynamic_rela_sections(struct kpatch_elf *kelf,
 				dynrelas[index].src = 0;
 			dynrelas[index].addend = rela->addend;
 			dynrelas[index].type = rela->type;
-			dynrelas[index].external = external;
 
 			/* add rela to fill in dest field */
 			ALLOC_LINK(dynrela, &relasec->relas);
@@ -2740,6 +2757,15 @@ void kpatch_create_dynamic_rela_sections(struct kpatch_elf *kelf,
 			dynrela->addend = objname_offset;
 			dynrela->offset = index * sizeof(*dynrelas) +
 				offsetof(struct kpatch_patch_dynrela, objname);
+
+			/* add rela to fill in sym_objname field */
+			ALLOC_LINK(dynrela, &relasec->relas);
+			dynrela->sym = strsym;
+			dynrela->type = R_X86_64_64;
+			dynrela->addend = offset_of_string(&kelf->strings,
+							   sym_objname);
+			dynrela->offset = index * sizeof(*dynrelas) +
+				offsetof(struct kpatch_patch_dynrela, sym_objname);
 
 			rela->sym->strip = 1;
 			list_del(&rela->list);
@@ -3041,11 +3067,11 @@ void kpatch_write_output_elf(struct kpatch_elf *kelf, Elf *elf, char *outfile)
 }
 
 struct arguments {
-	char *args[4];
+	char *args[5];
 	int debug;
 };
 
-static char args_doc[] = "original.o patched.o kernel-object output.o";
+static char args_doc[] = "original.o patched.o kernel-object output.o Module.symvers patch-module-name";
 
 static struct argp_option options[] = {
 	{"debug", 'd', 0, 0, "Show debug output" },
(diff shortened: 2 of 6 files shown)


jpoimboe:
> Still a WIP as I haven't checked to see what modifications are still required in the core kpatch module.

I'd say let's leave the 'external' field in the kpatch_dynrela struct so that  kpatch-patch-hook.c and kpatch.ko can still use it.  No need to add any new features to kpatch.ko at this point because its days are numbered.

>     create-diff-object now takes 6 arguments, which is a bit yucky. (the new ones being, the path to the Module.symvers file, and the second being the name of the to-be-built patch module)

That's fine with me, we can always refactor later if it turns out anybody actually cares about the create-diff-object cmdline interface.

>     The Module.symvers parsing code is based off of scripts/mod/modpost.c, with only the parsing functions taken and a few modifications to make them fit create-diff-object's needs. Didn't want to rewrite code that was already there. But I'm not sure how licensing works in this case (i.e. should I retain the original copywrites from modpost.c and append a "Modified by .." statement?)

I would copy the copyright lines from the original file:
```
 * Copyright 2003       Kai Germaschewski
 * Copyright 2002-2004  Rusty Russell, IBM Corporation
 * Copyright 2006-2008  Sam Ravnborg
 * Based in part on module-init-tools/depmod.c,file2alias
```
Then add your own copyright line, then some statement about where it was copied from, then add the full GPL license statement (which you already have).

I have a few more comments which I'll add inline.

kmod/patch/kpatch-patch.h
```diff
@@ -37,7 +37,7 @@ struct kpatch_patch_dynrela {
 	unsigned long type;
 	char *name;
 	char *objname;
-	int external;
+	char *sym_objname;
```
jpoimboe: I'd leave 'external' here, and still set it like before so that kpatch can use it.

Also, we need to try to stay backwards compatible as much as possible, so that newer kpatch-build will still work with old kernels which don't have this livepatch change.  Maybe with some ifdefs in `livepatch-patch-hook.c` like what @arges did in #573.

kpatch-build/create-diff-object.c
```diff
@@ -69,6 +69,8 @@
 		printf(format, ##__VA_ARGS__); \
 })
 
+extern char *find_exported_symbol_objname(const char *fname, char *symname);
```
jpoimboe: This function is kind of in the same realm as the functions in lookup.c, though the lookup functions look at the vmlinux or original module file rather than Module.symvers.  I wonder if it would be better to abstract this functionality in lookup.c?  lookup_open() could take an additional argument (the path to Module.symvers) and then there could be a lookup_exported_symbol() or something.

Or alternatively,

kpatch-build/create-diff-object.c
```diff
@@ -2695,13 +2704,22 @@ void kpatch_create_dynamic_rela_sections(struct kpatch_elf *kelf,
...
+					 * Not there, see if the symbol is
+					 * exported, and set sym_objname to the
+					 * object the exported symbol belongs to
 					 */
-					external = 1;
+					tmp_objname = find_exported_symbol_objname(mod_symvers,
+										   rela->sym->name);
+					if (tmp_objname) {
+						 /* Strip the path, only want the module name */
+						sym_objname = basename(tmp_objname);
```
jpoimboe: The file name needs to be converted to the "module name" by converting dashes ('-') to underlines ('_').

kpatch-build/create-diff-object.c
```diff
@@ -3041,11 +3067,11 @@ void kpatch_write_output_elf(struct kpatch_elf *kelf, Elf *elf, char *outfile)
 }
 
 struct arguments {
-	char *args[4];
+	char *args[5];
```
jpoimboe: Should be args[6] ?

kpatch-build/kpatch-build
```diff
@@ -527,7 +527,10 @@ for i in $FILES; do
 	debugopt=
 	[[ $DEBUG -eq 1 ]] && debugopt=-d
 	if [[ -e "orig/$i" ]]; then
-		"$TOOLSDIR"/create-diff-object $debugopt "orig/$i" "patched/$i" "$KOBJFILE" "output/$i" 2>&1 |tee -a "$LOGFILE"
+		# create-diff-object orig.o patched.o kernel-object Module.symvers patch-mod-name
+		"$TOOLSDIR"/create-diff-object $debugopt "orig/$i" \
+			"patched/$i" "$KOBJFILE" "output/$i" \
+			$OBJDIR/Module.symvers "kpatch_${PATCHNAME//-/_}" 1>&1 |tee -a "$LOGFILE"
```
jpoimboe: Should be `2>&1`

kpatch-build/create-diff-object.c
flaming-toast: Yeah I think that makes more sense, instead of having the out-of-place extern declaration here it could go in lookup.h. Technically the current parse-mod-symvers stuff can stay its own file, but we can move it to lookup.c, but how does licensing work in that case?

kpatch-build/create-diff-object.c
flaming-toast: Interestingly, this didn't segfault when it *should have*. Yay for undefined C behavior!

kpatch-build/create-diff-object.c
jpoimboe: License and copyright are two separate things.  The license is the same (GPLv2) so it's ok to copy the code -- you don't need to do anything there because lookup.c already has the license notice.

For the copyright, the original file's copyrights should be preserved.  So you can add something to the existing lookup.c copyright statement like "For functions A and B:" and then copy the original copyright statements.  You can add your own copyright statement at the end if you modify the functions.

kpatch-build/create-diff-object.c
jpoimboe: Yeah, unfortunately C doesn't have bounds checking and you can't expect a buffer overrun to cause a seg fault.

kpatch-build/create-diff-object.c
flaming-toast: Ah, ok. Then I'll simply move the parse-mod-symvers functions to lookup.c, rename the function to lookup_exported_symbol_objname(), and add in the copyrights. Thanks for the clarifications.

zhouchengming1:
Hi, I have a question.
In the "external" case, the symbol is an exported symbol or provided by another .o in the patch module. So the load_module() will do relocations of these symbols for the patch module. I don't know why use "externel" and strip the original relas that reference these symbols.
I think it's just ok to replace the line "externel = 1" with "continue" ; b/c we don't need to create dynamic
relas for these symbols.

flaming-toast:
@zhouchengming1 You bring up a good point. Yes, the module loader would be able to resolve the exported symbols if we don't strip the original relas that reference those symbols. However this will only work for exported symbols in vmlinux, not exported symbols that come from other modules, since those modules might not be loaded when we load the patch module (and therefore `load_module()` would not be able to find them). So technically, we would need to do a little more work (i.e. check if the module name in Module.symvers is vmlinux or not) to see if we can leave the original rela in-place.

zhouchengming1:
@flaming-toast  I understand your point. But if a module references some symbols exported by other modules, those modules should be loaded first.  Even the module is a kpatch-module.

zhouchengming1:
And the "externel" solution has other problems. If a kpatch-module references a symbol exported by a module that is not in the kpatch-module patched list, and the module is not loaded when we load the kpatch-module, then later when the module loaded, our kpatch-module won't try to find the exported symbol, see the function klp_module_notify.

zhouchengming1:
So I think we should load those modules first. If those modules are not loaded when our kpatch-module loaded, the load_module()  will report an error, just like load a normal module, it's fine.

jpoimboe:
> @flaming-toast I understand your point. But if a module references some symbols exported by other modules, those modules should be loaded first. Even the module is a kpatch-module.

So I think you're arguing that we shouldn't allow the loading of patches to unloaded modules, like we do with klp_module_notify()?

I think both RH and SUSE agree that klp_module_notify() is a nice feature.  If a cumulative patch module has fixes for multiple modules, we don't have to require users to load any unneeded modules before loading the patch module.  There was general agreement on the livepatching mailing list in support of it, but if you disagree, please start a broader discussion on the list.

> And the "externel" solution has other problems. If a kpatch-module references a symbol exported by a module that is not in the kpatch-module patched list, and the module is not loaded when we load the kpatch-module, then later when the module loaded, our kpatch-module won't try to find the exported symbol, see the function klp_module_notify.

I don't think that's much of a problem because it would only happen if the patch module were to add a new module dependency that the patched module didn't already have.  I can't think of a good reason to support that scenario.

flaming-toast:
@zhouchengming1 Yeah, we had a discussion about removing the module notify code and requiring to-be-patched modules to be loaded before the patch module. Just to reiterate what @jpoimboe already said, the general consensus was to keep the feature since we want to enable users to be able to load patch modules that can potentially affect many different modules without forcing the user to unnecessarily pre-load all those modules. The thread to the discussion is here for reference: http://www.spinics.net/lists/live-patching/msg00946.html
In short: we want to be able to load patch modules even if the target module(s) isn't loaded.

> If a kpatch-module references a symbol exported by a module that is not in the kpatch-module patched list, and the module is not loaded when we load the kpatch-module, then later when the module loaded, our kpatch-module won't try to find the exported symbol, see the function klp_module_notify.

So modules.dep and depmod will take care of these types of dependencies. If a module references an exported symbol in another module, a module dependency is created and modprobe takes care to make sure to load modules in the correct order. If livepatch patches module A that references an exported symbol in module B, module B is marked as a dependency of module A. So in order for module A to be loaded at all, module B must be loaded beforehand. So if the klp_module_notify handler runs for module A, we can assume that module B must also be loaded and the symbol will be found.

flaming-toast:
Just leaving a note here. I'm putting this PR on hold, since if the arch-independent patchset gets merged relatively soon, the changes that must be made to create-diff-object and friends to accommodate the patchset will include getting rid of external dynrelas (+ dynrelas completely).

flaming-toast:
I think we can close this PR, since we have Module.symvers parsing support now, and external dynrelas are no longer used in upstream livepatch. Do we also want to get rid of external dynrelas for kpatch.ko? Not sure if it's worth implementing this change for the core module if we plan to deprecate it eventually..

jpoimboe:
I guess it depends on how much work it would be to get rid of 'external' completely.  If it's not that much work, I'd say let's do it because we'll have to support kpatch.ko some for older kernels even after we deprecate it, and it still might be a while before we can deprecate it anyway.

Issue #493, opened 2015-07-09
sysfs issue for symbols in a patch module with the same name


flaming-toast:
This is a small edge case but I thought I'd bring it up. If a patch involves two local functions that happen to have the same name, sysfs will complain upon patch module load saying that it cannot create duplicate filenames. 
Here's the test patch I tried:
```
diff --git a/kernel/kallsyms.c b/kernel/kallsyms.c
index 5c5987f..8e2829f 100644
--- a/kernel/kallsyms.c
+++ b/kernel/kallsyms.c
@@ -511,6 +511,7 @@ static void *s_next(struct seq_file *m, void *p, loff_t *pos)
 
        if (!update_iter(m->private, *pos))
                return NULL;
+       pr_info("s_next");
        return p;
 }

diff --git a/mm/vmalloc.c b/mm/vmalloc.c
index 2faaa29..837b2f4 100644
--- a/mm/vmalloc.c
+++ b/mm/vmalloc.c
@@ -2578,7 +2578,7 @@ static void *s_next(struct seq_file *m, void *p, loff_t *pos)
        next = list_entry(va->list.next, typeof(*va), list);
        if (&next->list != &vmap_area_list)
                return next;
-
+       pr_info("s_next");
        return NULL;
 }
```
Here are the backtraces:
```
[  600.686129] WARNING: CPU: 1 PID: 28425 at fs/sysfs/dir.c:31 sysfs_warn_dup+0x68/0x80()
[  600.686133] sysfs: cannot create duplicate filename '/kernel/livepatch/kpatch_s_next_ambig/vmlinux/s_next'
[ snip ]
[  600.686234] CPU: 1 PID: 28425 Comm: insmod Tainted: G           O    4.0.4-ARCH-gcfa8c8d #1
[ snip ]
[  600.686261] Call Trace:
[  600.686271]  [<ffffffff81575f1b>] dump_stack+0x4c/0x6e
[  600.686279]  [<ffffffff81074dda>] warn_slowpath_common+0x8a/0xc0
[  600.686286]  [<ffffffff81074e65>] warn_slowpath_fmt+0x55/0x70
[  600.686293]  [<ffffffff812502f8>] ? kernfs_path+0x48/0x60
[  600.686298]  [<ffffffff81253b18>] sysfs_warn_dup+0x68/0x80
[  600.686303]  [<ffffffff81253bbd>] sysfs_create_dir_ns+0x8d/0xa0
[...]

PR #573, opened 2015-12-10
livepatch-patch-hook: add support for livepatch sympos structures

arges:
Support for patching objects that have duplicated function names will be
introduced upstream in v4.5. This structure change needs to be handled by
the livepatch patch hook.

In addition, we need to be able to look up the symbol position in order to
populate old_sympos and sympos fields in klp_object and klp_reloc
respectively.

Fixes: #493

Signed-off-by: Chris J Arges <chris.j.arges@canonical.com>

### DIFF
--- a/kmod/patch/kpatch-patch.h
+++ b/kmod/patch/kpatch-patch.h
@@ -27,6 +27,7 @@ struct kpatch_patch_func {
 	unsigned long new_size;
 	unsigned long old_addr;
 	unsigned long old_size;
+	unsigned long sympos;
 	char *name;
 	char *objname;
 };
@@ -35,6 +36,7 @@ struct kpatch_patch_dynrela {
 	unsigned long dest;
 	unsigned long src;
 	unsigned long type;
+	unsigned long sympos;
 	char *name;
 	char *objname;
 	int external;
--- a/kmod/patch/livepatch-patch-hook.c
+++ b/kmod/patch/livepatch-patch-hook.c
@@ -24,6 +24,7 @@
 #include <linux/module.h>
 #include <linux/list.h>
 #include <linux/slab.h>
+#include <linux/version.h>
 
 #include <linux/livepatch.h>
 
@@ -237,7 +238,11 @@ static int __init patch_init(void)
 			lfunc = &lfuncs[j];
 			lfunc->old_name = func->kfunc->name;
 			lfunc->new_func = (void *)func->kfunc->new_addr;
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 5, 0)
+			lfunc->old_sympos = func->kfunc->sympos;
+#else
 			lfunc->old_addr = func->kfunc->old_addr;
+#endif
 			j++;
 		}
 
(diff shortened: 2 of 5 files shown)


jpoimboe:
Doing the sympos lookup at runtime will be problematic because `old_addr` for a module is useless: a module can be loaded at any address (which is one of the reasons we're getting rid of `old_addr`).

Instead I think we'll need to do it in create-diff-object.  One way to do it:

- add sympos field to kpatch_func and kpatch_dynrela structs (but probably still keep the obsoleted kpatch_func.old_addr and kpatch_dynrela.src fields around for use by kpatch-patch-hook.c).
- change lookup_local_symbol() to find the sympos to allow kpatch_create_patches_sections() and kpatch_create_dynamic_rela_sections() to populate the new fields (sympos will always be 0 for global and exported symbols).

arges:
On 12/14/2015 04:23 PM, Josh Poimboeuf wrote:
> Doing the sympos lookup at runtime will be problematic because
> |old_addr| for a module is useless: a module can be loaded at any
> address (which is one of the reasons we're getting rid of |old_addr|).
> 
> Instead I think we'll need to do it in create-diff-object. One way to do it:
> 
>   * add sympos field to kpatch_func and kpatch_dynrela structs (but
>     probably still keep the obsoleted kpatch_func.old_addr and
>     kpatch_dynrela.src fields around for use by kpatch-patch-hook.c).
>   * change lookup_local_symbol() to find the sympos to allow
>     kpatch_create_patches_sections() and
>     kpatch_create_dynamic_rela_sections() to populate the new fields
>     (sympos will always be 0 for global and exported symbols).
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dynup/kpatch/pull/573#issuecomment-164579005>.
> 

Ok, I'm actually working on a way to generate the livepatch structures
after create-diff-object. This 'create-livepatch-object' could calculate
the needed fields and output the necessary structures, while the normal
kpatch generate would remain the same. I could roll this logic into that
patchset.

--chris

flaming-toast:
Quick question. Since we're doing the sympos lookup during patch module compile time rather than during runtime, will symbol positions be altered at all if KASLR is enabled? i.e. Will symbol positions found during compile time ever differ from kallsym positions during runtime?

jpoimboe:
KASLR only changes the base address where all the code is loaded.  It doesn't affect any of the symbol offsets from that base, nor their order in kallsyms.

flaming-toast:
Pinging @arges, are you still currently working on adding livepatch support for the new sympos scheme? Just trying to see where we're at. :-)

arges:
Yes.

I had a comment here about getting issue #478 resolved first (statically
generating klp structures), in that I was planning on making sympos
structures work.

Here was my question:
https://github.com/dynup/kpatch/issues/478#issuecomment-171062242

I can also re-try without the static klp structures if needed.

--chris

arges:
Ok since #478 is closed. I'll pick this back up without the static klp structures bit.

arges:
I was able to get this working with this latest PR. I've done some light testing using the original patch as shown in #493.

$ ls /sys/kernel/livepatch/kpatch_duplicate/vmlinux/
s_next,2  s_next,5

flaming-toast:
Thanks @arges, looks good from my side. :+1: Tested with the `s_next` patch and a patch that modifies multiple `m_next` functions in vmlinux.

PR #575, opened 2016-01-13
Changes to make packaging easier

arges:
Here are a series of commits that will make packaging kpatch in distributions such as Debian and Ubuntu a bit easier.

### DIFF
--- a/kmod/Makefile
+++ b/kmod/Makefile
@@ -1,18 +1,26 @@
 include ../Makefile.inc
 
 all: clean
+ifeq ($(BUILDMOD),yes)
 	$(MAKE) -C core
+endif
 
 install:
+ifeq ($(BUILDMOD),yes)
 	$(INSTALL) -d $(MODULESDIR)/$(shell uname -r)
 	$(INSTALL) -m 644 core/kpatch.ko $(MODULESDIR)/$(shell uname -r)
 	$(INSTALL) -m 644 core/Module.symvers $(MODULESDIR)/$(shell uname -r)
+endif
 	$(INSTALL) -d $(DATADIR)/patch
 	$(INSTALL) -m 644 patch/* $(DATADIR)/patch
 
 uninstall:
+ifeq ($(BUILDMOD),yes)
 	$(RM) -R $(MODULESDIR)
+endif
 	$(RM) -R $(DATADIR)
 
 clean:
+ifeq ($(BUILDMOD),yes)
 	$(MAKE) -C core clean
+endif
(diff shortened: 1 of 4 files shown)


kmod/Makefile
```diff
@@ -1,18 +1,30 @@
...
-	$(MAKE) -C core clean
+all: $(BUILD_DIRS)
+$(BUILD_DIRS):
+	$(MAKE) -C $(@:build-%=%)
+
+install: $(INSTALL_DIRS)
+$(INSTALL_DIRS):
+	$(MAKE) -C $(@:install-%=%) install
+
+ninstall: $(UNINSTALL_DIRS)
```
flaming-toast: You meant `uninstall:` right?

arges: You are correct! Already pushed a fix.

kmod/core/Makefile
```diff
@@ -13,11 +14,18 @@ kpatch.ko: core.c
 
 all: kpatch.ko
 
+install:
+	$(INSTALL) -d $(MODULESDIR)/$(shell uname -r)
+	$(INSTALL) -m 644 core/kpatch.ko $(MODULESDIR)/$(shell uname -r)
+	$(INSTALL) -m 644 core/Module.symvers $(MODULESDIR)/$(shell uname -r)
+
```
flaming-toast: Don't forget to define `INSTALL` somewhere, or include Makefile.inc

kmod/patch/Makefile
```diff
@@ -14,6 +14,13 @@ kpatch-$(KPATCH_NAME).ko:
 patch-hook.o: patch-hook.c kpatch-patch-hook.c livepatch-patch-hook.c
 	$(KPATCH_MAKE) patch-hook.o
 
+install:
+	$(INSTALL) -d $(DATADIR)/patch
+	$(INSTALL) -m 644 patch/* $(DATADIR)/patch
+
```
flaming-toast: Same here, `INSTALL` needs to be defined.

kmod/core/Makefile
flaming-toast: No need to specify `core/` in the path, since we would already be in the `core/` directory when this target is called. So it'd just  be `$(INSTALL) -m 644 kpatch.ko`

kmod/patch/Makefile
flaming-toast: No need to specify `patch/` in the path here, for the same reasons above.

flaming-toast:
@arges Thanks for the PR! And thanks for your patience :-) Other than the comments I already made, everything else looks good on my side.

arges:
I thought about this a bit more, and pushed an earlier approach to just add conditionals to kmod/Makefile. This is a bit more simple than moving makefile logic around and having to worry about including sub makefiles. Thanks for the reviews.

jpoimboe:
:+1:

flaming-toast:
Looks good to me, thanks.

PR #582, opened 2016-02-27
kpatch-build: process the patch name correctly

libin2015:
Process the patch name correctly that only concern the fuffix with
.patch or .diff. Otherwise if the patch name is not end with .patch
or .diff but has it as substring, the fuffix will be removed
unreasonably.


jpoimboe:
:+1:

PR #583, opened 2016-03-15
kpatch-build: fix building of livepatch-based patches

euspectre:
kpatch-build currently requires Module.symvers for the Kpatch core
module unconditionally and fails if it is not found. This does not allow
using kpatch-build to prepare livepatch-based patches.

This patch fixes the problem.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -97,24 +97,30 @@ find_dirs() {
 		# git repo
 		TOOLSDIR="$SCRIPTDIR"
 		DATADIR="$(readlink -f $SCRIPTDIR/../kmod)"
-		SYMVERSFILE="$DATADIR/core/Module.symvers"
-		return
 	elif [[ -e "$SCRIPTDIR/../libexec/kpatch/create-diff-object" ]]; then
 		# installation path
 		TOOLSDIR="$(readlink -f $SCRIPTDIR/../libexec/kpatch)"
 		DATADIR="$(readlink -f $SCRIPTDIR/../share/kpatch)"
+	else
+		return 1
+	fi
+}
+
+find_core_symvers() {
+	SYMVERSFILE=""
+	if [[ -e "$SCRIPTDIR/create-diff-object" ]]; then
+		# git repo
+		SYMVERSFILE="$DATADIR/core/Module.symvers"
+	elif [[ -e "$SCRIPTDIR/../libexec/kpatch/create-diff-object" ]]; then
+		# installation path
 		if [[ -e $SCRIPTDIR/../lib/kpatch/$ARCHVERSION/Module.symvers ]]; then
 			SYMVERSFILE="$(readlink -f $SCRIPTDIR/../lib/kpatch/$ARCHVERSION/Module.symvers)"
 		elif [[ -e /lib/modules/$ARCHVERSION/extra/kpatch/Module.symvers ]]; then
 			SYMVERSFILE="$(readlink -f /lib/modules/$ARCHVERSION/extra/kpatch/Module.symvers)"
-		else
-			warn "unable to find Module.symvers for kpatch core module"
-			return 1
 		fi
-		return
 	fi
 
-	return 1
+	[[ -e "$SYMVERSFILE" ]]
 }
 
 gcc_version_check() {
@@ -340,9 +346,6 @@ fi
 
 find_dirs || die "can't find supporting tools"
 
-[[ -e "$SYMVERSFILE" ]] || die "can't find core module Module.symvers"
-
-
 if [[ $SKIPGCCCHECK -eq 0 ]]; then
 	gcc_version_check || die
 fi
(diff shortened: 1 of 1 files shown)


flaming-toast:
Thanks for the patch @euspectre. This has indeed been a slightly annoying issue for a while. I think it might be more helpful though, to have this implemented as a command line option/switch for kpatch-build. Something like `-coremod` or `-nocoremod`, that would indicate whether we are building with the core kpatch module or building a livepatch module. This way we won't be dependent on the host system which could be building patch modules intended for another kernel/system.

jpoimboe:
Maybe we can look for CONFIG_LIVEPATCH in the .config file?

flaming-toast:
@jpoimboe That works (and is easier) too :-)

euspectre:
Good point. I will revisit the patch to check CONFIG_LIVEPATCH in the target's config instead.

euspectre:
Here is the updated patch.

kpatch-build/kpatch-build
```diff
@@ -97,24 +97,27 @@ find_dirs() {
...
+		SYMVERSFILE="$DATADIR/core/Module.symvers"
+	elif [[ -e "$SCRIPTDIR/../libexec/kpatch/create-diff-object" ]]; then
+		# installation path
 		if [[ -e $SCRIPTDIR/../lib/kpatch/$ARCHVERSION/Module.symvers ]]; then
 			SYMVERSFILE="$(readlink -f $SCRIPTDIR/../lib/kpatch/$ARCHVERSION/Module.symvers)"
 		elif [[ -e /lib/modules/$ARCHVERSION/extra/kpatch/Module.symvers ]]; then
 			SYMVERSFILE="$(readlink -f /lib/modules/$ARCHVERSION/extra/kpatch/Module.symvers)"
 		else
 			warn "unable to find Module.symvers for kpatch core module"
-			return 1
```
jpoimboe: In this case I think we still want to return 1 so that it fails.

kpatch-build/kpatch-build
```diff
@@ -97,24 +97,27 @@ find_dirs() {
...
 		fi
-		return
 	fi
-
-	return 1
```
jpoimboe: Same here.

kpatch-build/kpatch-build
```diff
@@ -338,10 +341,7 @@ elif [[ $DISTRO = ubuntu ]] || [[ $DISTRO = debian ]]; then
 	export PATH=/usr/lib/ccache:$PATH
 fi
 
-find_dirs || die "can't find supporting tools"
-
-[[ -e "$SYMVERSFILE" ]] || die "can't find core module Module.symvers"
-
+find_dirs
```
jpoimboe: || die "can't find supporting tools"

kpatch-build/kpatch-build
```diff
@@ -97,24 +97,27 @@ find_dirs() {
...
 		DATADIR="$(readlink -f $SCRIPTDIR/../kmod)"
-		SYMVERSFILE="$DATADIR/core/Module.symvers"
-		return
 	elif [[ -e "$SCRIPTDIR/../libexec/kpatch/create-diff-object" ]]; then
 		# installation path
 		TOOLSDIR="$(readlink -f $SCRIPTDIR/../libexec/kpatch)"
 		DATADIR="$(readlink -f $SCRIPTDIR/../share/kpatch)"
+	fi
+}
+
```
flaming-toast: Let's have `find_dirs()` return an error if it could not find `create-diff-object` in either location.
Then later when `find_dirs` is invoked, we can check for the error: 
`find_dirs || die "can't find supporting tools"`

kpatch-build/kpatch-build
flaming-toast: If we have `find_core_symvers` return 1 for an error, then I think we should move the `[[ -e "$SYMVERSFILE" ]]` check from line 454 to the very end of `find_core_symvers` and have it return 1 if the check fails. Then we could just do `find_core_symvers || die "can't find core module Module.symvers"`

kpatch-build/kpatch-build
euspectre: Yes, that would be useful.

kpatch-build/kpatch-build
euspectre: Reasonable.

euspectre:
Here is v.3 of the patch.

As you suggested, find_dir() now returns error if create-diff-object is missing. find_core_symvers() now checks if the symvers file is actually present.

As for the returns in find_core_symvers(), I do not think they are necessary. In case of errors, SYMVERSFILE will remain empty and [[ -e "" ]] will return false. Added initialization of SYMVERSFILE to make things clearer.

I also removed one of the warnings about missing symvers file. One should be enough, I suppose.

jpoimboe:
Thanks @euspectre !

:+1:

flaming-toast:
Thanks @euspectre.

euspectre:
Thanks for the reviews! Three heads are much better than one ;-)

Issue #580, opened 2016-02-10
Crashes when patching kvm_arch_vm_ioctl


arges:
I've noticed kernel oopses, invalid opcodes, or corrupted stacks when trying to
produce a livepatch for CVE-2015-7513. In starting to debug this I've pared
down the problem and tested a few things but have not root caused this yet.

Steps to reproduce:
1. Produce a livepatch using [1]. This simply adds a printk to `kvm_arch_vm_ioctl`.
2. I built this using 'kpatch-build' and have tested it using kernel livepatch and kpatch.ko.
3. Insert the KVM module, and insert the Livepatch module. (Order doesn't matter for reproduciblilty.)
4. Start a VM

Here's you'll notice crashes as follows:

```
[  128.546935] BUG: unable to handle kernel paging request at ffffffff87e626c2
[  128.547750] IP: [<ffffffff87e626c2>] 0xffffffff87e626c2
[  128.548283] PGD 1c0f067 PUD 1c10063 PMD 0 
[  128.548808] Oops: 0010 [#1] SMP 
[  128.549215] Modules linked in: vhost_net vhost macvtap macvlan xt_conntrack ipt_REJECT nf_reject_ipv4 kpatch_patch(OE) isofs ebtable_filter ebtables ip6table_filter ip6_tables xt_CHECKSUM iptable_mangle ipt_MASQUERADE nf_nat_masquerade_ipv4 iptable_nat nf_conntrack_ipv4 nf_defrag_ipv4 nf_nat_ipv4 nf_nat nf_conntrack xt_tcpudp bridge stp llc iptable_filter ip_tables x_tables ppdev joydev serio_raw parport_pc parport kvm_intel kvm autofs4 psmouse floppy
[  128.554499] CPU: 0 PID: 1249 Comm: qemu-system-x86 Tainted: G           OE K 4.2.0-16-generic #19-Ubuntu
[  128.555335] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS Ubuntu-1.8.2-1ubuntu1 04/01/2014
[  128.556178] task: ffff88007c0d1900 ti: ffff880036d08000 task.ti: ffff880036d08000
[  128.556847] RIP: 0010:[<ffffffff87e626c2>]  [<ffffffff87e626c2>] 0xffffffff87e626c2
[  128.557683] RSP: 0018:ffff880036d0bcf0  EFLAGS: 00010282
[...]

PR #590, opened 2016-04-11
kmod/core: Skip relocations of already altered instructions

euspectre:
When a patch module is loaded, the kernel facilities like alternatives
and paravirt may alter some of its instructions. This happens before
Kpatch core module is notified and tries to apply dynrelas to it. If an
instruction to apply a dynrela to has already been changed by these
facilities, an incorrect instruction might be written as a result.

The core module now detects such conditions and does not apply dynrela
to the changed instructions.

Suggested by Josh Poimboeuf in the discussion of
https://github.com/dynup/kpatch/issues/580.

Changes in v.2:
* Used pr_notice to give more emphasis to the messages.
* Added an explanation message.

### DIFF
--- a/kmod/core/core.c
+++ b/kmod/core/core.c
@@ -43,6 +43,7 @@
 #include <linux/uaccess.h>
 #include <linux/kallsyms.h>
 #include <linux/version.h>
+#include <linux/string.h>
 #include <asm/stacktrace.h>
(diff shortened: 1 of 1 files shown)


jpoimboe:
:+1:

flaming-toast:
Thanks @euspectre!

PR #592, opened 2016-04-27
kmod/core: fix stacktrace_ops 'address' function prototype for 4.6

### LINKED ISSUE #584: Recent struct stacktrace_ops change breaks kpatch core module
The kpatch core module currently breaks with linux-next. A recent commit upstream (568b329a "perf: generalize perf_callchain") modifies the `struct stacktrace_ops` declaration in arch/x86/include/asm/stacktrace.h. This issue **doesn't** apply to the current 4.5 kernel release, so just noting for now that this might have to be fixed for 4.6.
````
./kpatch/kmod/core/core.c:255:13: error: initialization from incompatible pointer type [-Werror=incompatible-pointer-types]
  .address = kpatch_backtrace_address_verify,
             ^
./kpatch/kmod/core/core.c:255:13: note: (near initialization for ‘kpatch_backtrace_ops.address’)
./kpatch/kmod/core/core.c:275:13: error: initialization from incompatible pointer type [-Werror=incompatible-pointer-types]
  .address = kpatch_print_trace_address,
```

flaming-toast:
Fixes #584.

Since `kpatch_backtrace_address_verify` returns an int now, `args->ret` isn't used in the new version, since returning an error will prompt `print_context_stack_bp` to stop walking the stack. Tested on 4.5.0 and 4.6.0-rc5.

### DIFF
--- a/kmod/core/core.c
+++ b/kmod/core/core.c
@@ -195,6 +195,7 @@ static inline int kpatch_compare_addresses(unsigned long stack_addr,
 	return 0;
 }
 
+#if LINUX_VERSION_CODE < KERNEL_VERSION(4, 6, 0)
 static void kpatch_backtrace_address_verify(void *data, unsigned long address,
 					    int reliable)
 {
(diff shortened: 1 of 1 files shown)


kmod/core/core.c
```diff
@@ -195,6 +195,7 @@ static inline int kpatch_compare_addresses(unsigned long stack_addr,
 	return 0;
 }
 
+#if LINUX_VERSION_CODE <= KERNEL_VERSION(4, 5, 0)
```
jpoimboe: Would this break with a kernel 4.5.1?

Maybe it would better to say `#if LINUX_VERSION_CODE < KERNEL_VERSION(4, 6, 0)`.  Not sure if that would exclude the 4.6-rcX releases, but we don't need to support them anyway.

kmod/core/core.c
```diff
@@ -264,12 +318,23 @@ static int kpatch_print_trace_stack(void *data, char *name)
...
```
jpoimboe: Same here.

kmod/core/core.c
flaming-toast: Ah yeah, you are right.

PR #594, opened 2016-05-01
kpatch-build: Add "CONFIG_DEBUG_KERNEL" kernel config checking

NonerKao:
While the officially supported distributions all have CONFIG_DEBUG_KERNEL enabled, this is not true for some other distributions.  This option is necessary when kpatch-build retrieves the SPECIAL_VARS using readelf command.

Signed-off-by: Quey-Liang Kao <s101062801@m101.nthu.edu.tw>

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -448,6 +448,8 @@ else
 	fi
 fi
 
+# kernel option checking: CONFIG_DEBUG_KERNEL and CONFIG_LIVEPATCH
+grep -q "CONFIG_DEBUG_KERNEL=y" "$OBJDIR/.config" || die "kernel doesn't have 'CONFIG_DEBUG_KERNEL' enabled"
 if grep "CONFIG_LIVEPATCH=y" "$OBJDIR/.config" > /dev/null; then
 	# The kernel supports livepatch.
 	KBUILD_EXTRA_SYMBOLS=""


flaming-toast:
Thanks for catching this. I think we also need to add a check for CONFIG_DEBUG_KERNEL in the kpatch-build script itself, before SPECIAL_VARS is set, since livepatch modules don't depend on the kpatch core module.

kpatch-build/kpatch-build
```diff
@@ -464,19 +464,24 @@ cp "$PATCHFILE" "$APPLIEDPATCHFILE" || die
...
+		       b == 0 && /DW_AT_name.* bug_entry\s*$/ {b = 1; next}
+		       p == 0 && /DW_AT_name.* paravirt_patch_site\s*$/ {p = 1; next}
+		       e == 0 && /DW_AT_name.* exception_table_entry\s*$/ {e = 1; next}
+		       a == 1 {printf("export ALT_STRUCT_SIZE=%d\n", $4); a = 2}
+		       b == 1 {printf("export BUG_STRUCT_SIZE=%d\n", $4); b = 2}
+		       p == 1 {printf("export PARA_STRUCT_SIZE=%d\n", $4); p = 2}
+		       e == 1 {printf("export EX_STRUCT_SIZE=%d\n", $4); e = 2}
+		       a == 2 && b == 2 && p == 2 && e == 2 {exit}')
+else
+	die "no support for debug infos"
```
jpoimboe: This error message should be more specific, like "kernel doesn't have 'CONFIG_DEBUG_KERNEL' enabled".

NonerKao: Sure.

jpoimboe:
I think the first commit isn't needed, since the issue is specific to kpatch-build, not the core module.  Can you remove the first commit, squash the other two commits into a single commit, and `git push --force`?

kpatch-build/kpatch-build
```diff
@@ -464,19 +464,24 @@ cp "$PATCHFILE" "$APPLIEDPATCHFILE" || die
...
-	       BEGIN { a = b = p = e = 0 }
-	       a == 0 && /DW_AT_name.* alt_instr\s*$/ {a = 1; next}
-	       b == 0 && /DW_AT_name.* bug_entry\s*$/ {b = 1; next}
-	       p == 0 && /DW_AT_name.* paravirt_patch_site\s*$/ {p = 1; next}
-	       e == 0 && /DW_AT_name.* exception_table_entry\s*$/ {e = 1; next}
-	       a == 1 {printf("export ALT_STRUCT_SIZE=%d\n", $4); a = 2}
-	       b == 1 {printf("export BUG_STRUCT_SIZE=%d\n", $4); b = 2}
-	       p == 1 {printf("export PARA_STRUCT_SIZE=%d\n", $4); p = 2}
-	       e == 1 {printf("export EX_STRUCT_SIZE=%d\n", $4); e = 2}
-	       a == 2 && b == 2 && p == 2 && e == 2 {exit}')
```
flaming-toast: Small nit...I think we can simplify the entire if/else block with just one line:
```
grep -q "CONFIG_DEBUG_KERNEL=y" "$OBJDIR/.config" || die "kernel doesn't have 'CONFIG_DEBUG_KERNEL' enabled"
```
Ideally placed after (or before, doesn't matter) the CONFIG_LIVEPATCH check above. Then we can leave this block of code above as-is.

NonerKao: I add this checking before that of CONFIG_LIVEPATCH, and add a comment at the beginning of this section.  If some other kernel options are required in the future, one may add them here as a whole.

flaming-toast:
:+1:

PR #595, opened 2016-05-06
kmod: Fix install path

rosslagerwall:
When building against a different kernel from the one that is currently
running, install the kernel module to a directory based on the version
of the target kernel rather than the currently running kernel.


jpoimboe:
:+1:

PR #601, opened 2016-07-10
Introduce a common kpatch-elf "api"

flaming-toast:
I'm separating the 4.7-readiness changes into two parts for easier code review. This is the first part, which moves all kpatch elf related functions into a new source file `kpatch-elf.c`. Any source file that needs to access to these functions (and any related structs) must include `kpatch-elf.h`. The code that regenerates klp rela sections and sets the appropriate markings (`SHN_LIVEPATCH` etc) (which is part 2) will use this kpatch-elf api.

This change should be functionally equivalent to the current `create-diff-object`, the diff is only large because it is a large migration of code to a new source file `kpatch-elf.c` and new header file `kpatch-elf.h`, which `create-diff-object` includes. Each function that was copied over to `kpatch-elf.c` should be exactly the same as the version from `create-diff-object` (in other words, every subtraction from `create-diff-object` should correlate to an addition in `kpatch-elf.{c,h}`). I only moved functions that were not directly related to the diffing logic in `create-diff-object`. Any function related to correlating elfs, including/excluding elements, comparison of sections, etc were left in `create-diff-object`. All the functions that were plucked from `create-diff-object` are nicely listed in the `kpatch-elf.h` header file.

Additional tweaks I made are reflected in the latter 2 commits, and I also had to tweak some of the logging macros. Specifically, I had to tweak the `log_normal` macro, because it referred to a global variable in `create-diff-object` named `childobj`, and I also decided to let `create-diff-object` redefine the `ERROR` macro so it could prepend `childobj` to its error message.

Did some light testing with these changes and was able to build a couple of patch modules without issue (tested with the kpatch.ko core module).

Also, I'm wondering if we should create a separate branch for 4.7-related changes? Since rhel still relies on the current code on `master`. Unfortunately this would create two separate development streams..


jpoimboe:
> Also, I'm wondering if we should create a separate branch for 4.7-related changes? Since rhel still relies on the current code on master. Unfortunately this would create two separate development streams..

Any reason why we can't make the changes backwards compatible by using the `KERNEL_VERSION()` macro like we did with 4.6?

kpatch-build/kpatch-elf.h
```diff
@@ -0,0 +1,186 @@
...
+	struct list_head sections;
+	struct list_head symbols;
+	struct list_head strings;
+	int fd;
+};
+
+struct special_section {
+	char *name;
+	int (*group_size)(struct kpatch_elf *kelf, int offset);
+};
```
jpoimboe: I think all the special, static local, and group_size code is more specific to kpatch than elf and would fit better in create-diff-object.c.

kpatch-build/kpatch-elf.h
```diff
@@ -0,0 +1,186 @@
...
+	if (loglevel <= (level)) \
+		printf(format, ##__VA_ARGS__); \
+})
+
+enum loglevel {
+	DEBUG,
+	NORMAL
+};
+
+extern enum loglevel loglevel;
```
jpoimboe: Maybe the log code could be moved to a small log.h file (though loglevel could still be defined in create-diff-object.c).

kpatch-build/kpatch-elf.h
```diff
@@ -0,0 +1,186 @@
...
+ */
+
+#ifndef _KPATCH_ELF_H_
+#define _KPATCH_ELF_H_
+
+#include <gelf.h>
+#include "list.h"
+
+#define ERROR(format, ...) \
+	error(1, 0, "ERROR: %s: %d: " format, __FUNCTION__, __LINE__, ##__VA_ARGS__)
```
jpoimboe: It would be really nice if kpatch-elf.c still had access to childobj so that ERROR() and other logging macros can still use it.  Maybe it could just be part of the "interface": when using log.h you have to define childobj and loglevel in your .c file.

flaming-toast: Ok sure!

jpoimboe:
OK, sounds good.  I'd rather avoid passing the kernel version to create-diff-object, so I think leaving old fields in place and adding new fields is fine.

flaming-toast:
I moved all the `*_group_size()` functions, `special_sections` array, `is_special_static()`, `kpatch_is_normal_static_local()` back to create-diff-object. I also moved the `inc_printf` macro (that I originally grouped with the rest of the logging macros) back to create-diff-object because that macro needs `recurselevel` defined and it doesn't make sense outside of its original context.

A new header file named `log.h` contains the logging macros, and __requires `childobj` and `loglevel` to be defined__ by whatever includes it.

I must note that moving the logging macros outside of kpatch-elf.h was slightly problematic, because both kpatch-elf.c and create-diff-object.c need the macros, but only create-diff-object.c defines both `loglevel` and `childobj`. It doesn't make sense for kpatch-elf.c to define those variables by itself. So even though kpatch-elf.c requires log.h, it must depend on another source file (namely create-diff-object) to define `loglevel` and `childobj`...

flaming-toast:
>OK, sounds good. I'd rather avoid passing the kernel version to create-diff-object, so I think leaving old fields in place and adding new fields is fine.

Hm, but create-diff-object still needs to decide whether or not to create dynrelas or klp relas, so we need to figure out what version vmlinux is from inside create-diff-object. Either we pass in an additional argument, grab some environment variable (as you do with the group_size() funcs), or dig through the vmlinux's sections for it. Too bad the version magic is not neatly in its own section (like in .comment or something), but rather stuffed in the .rodata section...

jpoimboe:
I'm not crazy about create-diff-object knowing more about the kernel than it already does, because it could be a useful tool in other non-kernel environments.  We need to make it as kernel-agnostic as possible.  So I really don't like the idea of adding more environment variables (yes, I know I created the group_size ones ;-) ) or kernel-specific arguments, or reading the kernel string, as they all make create-diff-object more kernel-centric, moving it in the opposite direction we want it to go.

So I think we should do something like:
- have create-diff-object emit both old-style and new-style relas, and have a linker script or another tool strip the ones we don't need; or
- have create-diff-object emit some intermediary rela metadata which is then converted by another tool to the right format.

I have a feeling the 1st option would be easier (though maybe the 2nd would be something to strive for long-term).  Do you think it's reasonably possible?

Hopefully that all makes sense, my brain is fried from rewriting the x86 kernel unwinder.

flaming-toast:
>I'm not crazy about create-diff-object knowing more about the kernel than it already does, because it could be a useful tool in other non-kernel environments. We need to make it as kernel-agnostic as possible. So I really don't like the idea of adding more environment variables (yes, I know I created the group_size ones ;-) ) or kernel-specific arguments, or reading the kernel string, as they all make create-diff-object more kernel-centric, moving it in the opposite direction we want it to go.

All good points.

So I think solution 1 is the easiest, since we are keeping all the old fields in all the structs anyway. I would have to think more about how solution 2 would work out...Anyway I'll have a go at solution 1 for the 2nd (part 2) pull request.

> Hopefully that all makes sense, my brain is fried from rewriting the x86 kernel unwinder.

:-)

PR #605, opened 2016-07-25
kpatch-build: do not assume that TEMPDIR is unexported on entry to script

bearieq:
The user's environment might have TEMPDIR exported.  If so, then kpatch-build
dies with a bogus "invalid ancestor" error. If you turn those bogus errors into
warnings, then the script goes on to incorrectly put into the generated .ko file
every single function that was compiled in the *original* kernel build, thereby
producing an immense .ko file with more than 64k sections that the linux kernel
cannot load.  This fix makes sure that TEMPDIR is unexported on the build of the
original kernel.  Actually, this fix uses a separate KPATCH_GCC_TEMPDIR variable,
so that if the kernel build is interrupted, the cleanup function in the kpatch-kbuild
script will still have TEMPDIR set correctly.

Signed-off-by: Martin Carroll <martin.carroll@alcatel-lucent.com>

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -495,12 +495,14 @@ done
 echo "Building original kernel"
 ./scripts/setlocalversion --save-scmversion || die
 make mrproper >> "$LOGFILE" 2>&1 || die
+unset KPATCH_GCC_TEMPDIR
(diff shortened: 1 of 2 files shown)


jpoimboe:
Thanks @bearieq !

:+1:

PR #607, opened 2016-08-10
4.7 changes (WIP)

flaming-toast:
This pull request attempts to bridge the two ways we have of building "dynamic relocations" i.e. dynrelas and klp relas.

I had to scrap the first implementation I had which tried to have create-diff-object create intermediary, implementation-neutral `.kpatch.symbols` and `.kpatch.relocations` sections, which we agreed seemed like the most "correct" way to go about it. But unfortunately I later bumped into a unresolvable limitation: After the final object.o is linked, there was no deterministic way to figure out which entry belonged to which target section just from looking at each entry in `.kpatch.relocations`. The target section is needed to fill in the sh_info field for klp rela sections. Storing the section name in `.kpatch.relocations` wouldn't work since names aren't required to be unique, and section indices are jumbled in linking the final object, so we can't use that either.

So a new approach was needed, and I'll try to describe it as succinctly as possible...

Currently, create-diff-object strips relas and symbols that are replaced by dynrelas. Instead, we're going to have create-diff-object not only create `.kpatch.dynrela` sections but also keep these relas and symbols, and additionally rename these symbols with the ".klp.sym." prefix so after the objects are linked they remain identifiable. The relas that refer to these symbols are also identifiable in this way (just check if they refer to a .klp.sym symbol). In addition, create-diff-object must tack on the objname to the name of each section (in the form of ".klp.objname.secname") that refers to klp relas because we will need to know the objname when building .klp.rela sections. These modified section names are only temporary and their original section names are restored in "phase 2".

Thus that concludes "phase 1", the intermediary objects produced by create-diff-object will contain `.kpatch.dynrelas` sections and retain the formerly stripped symbols and relas.

After the intermediary objects are linked, we have then move to phase 2 and we have two options depending on the kernel version we are dealing with. We either (1) strip all klp elements or (2) further refine the klp symbols and relas (actually create the klp rela sections) and strip the dynrela sections. We do option (1) if kernel version < 4.7.0 and option (2) otherwise.

Option (1) - strip all klp elements is pretty straightforward; free all the relas that refer to symbols prefixed with .klp.sym. and free the corresponding symbols. Restore any sections with the intermediary .klp.objname prefix back to their original section name.

Option (2) - further build upon the klp symbols and relas. This involves iterating through each rela section, sorting out the ones that refer to .klp.sym. symbols, and moving those relas to new .klp.rela sections (these are the ones marked `SHF_RELA_LIVEPATCH` and `SHT_RELA`). We also mark those .klp.syms. with `SHN_LIVEPATCH`. We have to wait until _after_ the patch module is linked to apply any special livepatch-specific elf flags because the linker tends to throw these values away. The `.kpatch.dynrelas` section is discarded via kpatch.lds.

Other notable changes, and things worth noting:
- kpatch.lds -> kpatch.lds.S: The kbuild system lets us run .lds.S linker scripts under the preprocessor before generating the corresponding .lds script. This is handy since we can use one linker script to include or discard certain sections (i.e. .kpatch.dynrelas) based on a variable we set in kpatch-build (`KCPPFLAGS="-D__KPATCH_DYNRELAS__"`)
- I only discard `.kpatch.dynrelas` but I think we can also extend that to discarding `.kpatch.hooks.*` and `.kpatch.force` sections for livepatch modules, since these sections aren't used in livepatch.

UPDATE 8/29/16:
Commits are still in a rough state in that they need to be better split up and reorganized; nonetheless the groundwork implementation is there and general comments on this approach are welcome.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1584,6 +1584,8 @@ static void kpatch_regenerate_special_section(struct kpatch_elf *kelf,
 	/* include both rela and base sections */
 	sec->include = 1;
 	sec->base->include = 1;
+	/* include secsym so .kpatch.arch relas can point to section symbols */
+	sec->base->secsym->include = 1;
 
 	/*
 	 * Update text section data buf and size.
@@ -1705,6 +1707,56 @@ static void kpatch_mark_ignored_functions_same(struct kpatch_elf *kelf)
 		if (rela->sym->sec->rela)
 			rela->sym->sec->rela->status = SAME;
 	}
+}
+
+static void kpatch_create_kpatch_arch_section(struct kpatch_elf *kelf, char *objname)
+{
+	struct special_section *special;
+	struct kpatch_arch *entries;
+	struct symbol *strsym;
+	struct section *sec, *karch_sec;
+	struct rela *rela;
+	int nr, index = 0;
+
+	nr = sizeof(special_sections) / sizeof(special_sections[0]);
+	karch_sec = create_section_pair(kelf, ".kpatch.arch", sizeof(*entries), nr);
+	entries = karch_sec->data->d_buf;
+
+	/* lookup strings symbol */
+	strsym = find_symbol_by_name(&kelf->symbols, ".kpatch.strings");
+	if (!strsym)
+		ERROR("can't find .kpatch.strings symbol");
+
+	for (special = special_sections; special->name; special++) {
+		if (strcmp(special->name, ".parainstructions") &&
+		    strcmp(special->name, ".altinstructions"))
+			continue;
+
+		sec = find_section_by_name(&kelf->sections, special->name);
+		if (!sec)
+			continue;
+
+		/* entries[index].sec */
+		ALLOC_LINK(rela, &karch_sec->rela->relas);
+		rela->sym = sec->secsym;
+		rela->type = R_X86_64_64;
+		rela->addend = 0;
+		rela->offset = index * sizeof(*entries) + \
+			       offsetof(struct kpatch_arch, sec);
+
+		/* entries[index].objname */
+		ALLOC_LINK(rela, &karch_sec->rela->relas);
+		rela->sym = strsym;
+		rela->type = R_X86_64_64;
+		rela->addend = offset_of_string(&kelf->strings, objname);
+		rela->offset = index * sizeof(*entries) + \
+			       offsetof(struct kpatch_arch, objname);
+
+		index++;
+	}
+
+	karch_sec->data->d_size = index * sizeof(struct kpatch_arch);
+	karch_sec->sh.sh_size = karch_sec->data->d_size;
 }
 
 static void kpatch_process_special_sections(struct kpatch_elf *kelf)
@@ -2462,6 +2514,7 @@ int main(int argc, char *argv[])
 	kpatch_create_strings_elements(kelf_out);
 	kpatch_create_patches_sections(kelf_out, lookup, hint, objname);
 	kpatch_create_intermediate_sections(kelf_out, lookup, hint, objname, pmod_name);
+	kpatch_create_kpatch_arch_section(kelf_out, objname);
 	kpatch_create_hooks_objname_rela(kelf_out, objname);
 	kpatch_build_strings_section_data(kelf_out);
 
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2246,11 +2246,11 @@ static void kpatch_build_strings_section_data(struct kpatch_elf *kelf)
 }
 
 struct arguments {
-	char *args[4];
+	char *args[5];
 	int debug;
 };
 
-static char args_doc[] = "original.o patched.o kernel-object output.o";
+static char args_doc[] = "original.o patched.o kernel-object output.o Module.symvers";
 
 static struct argp_option options[] = {
 	{"debug", 'd', NULL, 0, "Show debug output" },
@@ -2269,13 +2269,13 @@ static error_t parse_opt (int key, char *arg, struct argp_state *state)
 			arguments->debug = 1;
 			break;
 		case ARGP_KEY_ARG:
-			if (state->arg_num >= 4)
+			if (state->arg_num >= 5)
 				/* Too many arguments. */
 				argp_usage (state);
 			arguments->args[state->arg_num] = arg;
 			break;
 		case ARGP_KEY_END:
-			if (state->arg_num < 4)
+			if (state->arg_num < 5)
 				/* Not enough arguments. */
 				argp_usage (state);
 			break;
@@ -2296,6 +2296,7 @@ int main(int argc, char *argv[])
 	struct section *sec, *symtab;
 	struct symbol *sym;
 	char *hint = NULL, *objname, *pos;
+	char *mod_symvers_path;
 
 	arguments.debug = 0;
 	argp_parse (&argp, argc, argv, 0, NULL, &arguments);
@@ -2305,6 +2306,8 @@ int main(int argc, char *argv[])
 	elf_version(EV_CURRENT);
 
 	childobj = basename(arguments.args[0]);
+
+	mod_symvers_path = arguments.args[4];
 
 	kelf_base = kpatch_elf_open(arguments.args[0]);
 	kelf_patched = kpatch_elf_open(arguments.args[1]);
@@ -2378,7 +2381,7 @@ int main(int argc, char *argv[])
 		ERROR("FILE symbol not found in output. Stripped?\n");
 
 	/* create symbol lookup table */
-	lookup = lookup_open(arguments.args[2]);
+	lookup = lookup_open(arguments.args[2], mod_symvers_path);
 
 	/* extract module name (destructive to arguments.modulefile) */
 	objname = basename(arguments.args[2]);
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2295,7 +2295,7 @@ int main(int argc, char *argv[])
 	struct lookup_table *lookup;
 	struct section *sec, *symtab;
 	struct symbol *sym;
-	char *hint = NULL, *name, *pos;
+	char *hint = NULL, *objname, *pos;
 
 	arguments.debug = 0;
 	argp_parse (&argp, argc, argv, 0, NULL, &arguments);
@@ -2381,15 +2381,15 @@ int main(int argc, char *argv[])
 	lookup = lookup_open(arguments.args[2]);
 
 	/* extract module name (destructive to arguments.modulefile) */
-	name = basename(arguments.args[2]);
-	if (!strncmp(name, "vmlinux-", 8))
-		name = "vmlinux";
+	objname = basename(arguments.args[2]);
+	if (!strncmp(objname, "vmlinux-", 8))
+		objname = "vmlinux";
 	else {
-		pos = strchr(name,'.');
+		pos = strchr(objname,'.');
 		if (pos) {
 			/* kernel module */
 			*pos = '\0';
-			pos = name;
+			pos = objname;
 			while ((pos = strchr(pos, '-')))
 				*pos++ = '_';
 		}
@@ -2397,9 +2397,9 @@ int main(int argc, char *argv[])
 
 	/* create strings, patches, and dynrelas sections */
 	kpatch_create_strings_elements(kelf_out);
-	kpatch_create_patches_sections(kelf_out, lookup, hint, name);
-	kpatch_create_dynamic_rela_sections(kelf_out, lookup, hint, name);
-	kpatch_create_hooks_objname_rela(kelf_out, name);
+	kpatch_create_patches_sections(kelf_out, lookup, hint, objname);
+	kpatch_create_dynamic_rela_sections(kelf_out, lookup, hint, objname);
+	kpatch_create_hooks_objname_rela(kelf_out, objname);
 	kpatch_build_strings_section_data(kelf_out);
 
 	kpatch_create_mcount_sections(kelf_out);
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1311,26 +1311,6 @@ static void kpatch_reorder_symbols(struct kpatch_elf *kelf)
 	list_replace(&symbols, &kelf->symbols);
 }
 
-static void kpatch_reindex_elements(struct kpatch_elf *kelf)
-{
-	struct section *sec;
-	struct symbol *sym;
-	int index;
-
-	index = 1; /* elf write function handles NULL section 0 */
-	list_for_each_entry(sec, &kelf->sections, list)
-		sec->index = index++;
-
-	index = 0;
-	list_for_each_entry(sym, &kelf->symbols, list) {
-		sym->index = index++;
-		if (sym->sec)
-			sym->sym.st_shndx = sym->sec->index;
-		else if (sym->sym.st_shndx != SHN_ABS)
-			sym->sym.st_shndx = SHN_UNDEF;
-	}
-}
-
 static int bug_table_group_size(struct kpatch_elf *kelf, int offset)
 {
 	static int size = 0;
@@ -2263,38 +2243,6 @@ static void kpatch_build_strings_section_data(struct kpatch_elf *kelf)
 		strcpy(strtab, string->name);
 		strtab += strlen(string->name) + 1;
 	}
-}
-
-static void kpatch_rebuild_rela_section_data(struct section *sec)
-{
-	struct rela *rela;
-	int nr = 0, index = 0, size;
-	GElf_Rela *relas;
-
-	list_for_each_entry(rela, &sec->relas, list)
-		nr++;
-
-	size = nr * sizeof(*relas);
-	relas = malloc(size);
-	if (!relas)
-		ERROR("malloc");
-
-	sec->data->d_buf = relas;
-	sec->data->d_size = size;
-	/* d_type remains ELF_T_RELA */
-
-	sec->sh.sh_size = size;
-
-	list_for_each_entry(rela, &sec->relas, list) {
-		relas[index].r_offset = rela->offset;
-		relas[index].r_addend = rela->addend;
-		relas[index].r_info = GELF_R_INFO(rela->sym->index, rela->type);
-		index++;
-	}
-
-	/* sanity check, index should equal nr */
-	if (index != nr)
-		ERROR("size mismatch in rebuilt rela section");
 }
 
 struct arguments {
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -51,6 +51,7 @@
 #include "asm/insn.h"
 #include "kpatch-patch.h"
 #include "kpatch-elf.h"
+#include "kpatch-intermediate.h"
 
 #define DIFF_FATAL(format, ...) \
 ({ \
@@ -1887,17 +1888,20 @@ static int kpatch_is_core_module_symbol(char *name)
 		!strcmp(name, "kpatch_shadow_get"));
 }
 
-static void kpatch_create_dynamic_rela_sections(struct kpatch_elf *kelf,
-						struct lookup_table *table, char *hint,
-						char *objname)
+static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
+						struct lookup_table *table,
+						char *hint, char *objname,
+						char *pmod_name)
 {
-	int nr, index, objname_offset;
-	struct section *sec, *dynsec, *relasec;
-	struct rela *rela, *dynrela, *safe;
-	struct symbol *strsym;
+	int nr, index;
+	struct section *sec, *ksym_sec, *krela_sec;
+	struct rela *rela, *rela2, *safe;
+	struct symbol *strsym, *ksym_sec_sym;
+	struct kpatch_symbol *ksyms;
+	struct kpatch_relocation *krelas;
 	struct lookup_result result;
-	struct kpatch_patch_dynrela *dynrelas;
-	int vmlinux, external, ret;
+	char *sym_objname;
+	int ret, vmlinux, external;
 
 	vmlinux = !strcmp(objname, "vmlinux");
 
@@ -1909,21 +1913,29 @@ static void kpatch_create_dynamic_rela_sections(struct kpatch_elf *kelf,
 		if (!strcmp(sec->name, ".rela.kpatch.funcs"))
 			continue;
 		list_for_each_entry(rela, &sec->relas, list)
-			nr++; /* upper bound on number of dynrelas */
+			nr++; /* upper bound on number of kpatch relas and symbols */
 	}
 
-	/* create text/rela section pair */
-	dynsec = create_section_pair(kelf, ".kpatch.dynrelas", sizeof(*dynrelas), nr);
-	relasec = dynsec->rela;
-	dynrelas = dynsec->data->d_buf;
+	/* create .kpatch.relocations text/rela section pair */
+	krela_sec = create_section_pair(kelf, ".kpatch.relocations", sizeof(*krelas), nr);
+	krelas = krela_sec->data->d_buf;
+
+	/* create .kpatch.symbols text/rela section pair */
+	ksym_sec = create_section_pair(kelf, ".kpatch.symbols", sizeof(*ksyms), nr);
+	ksyms = ksym_sec->data->d_buf;
+
+	/* create .kpatch.symbols section symbol (to set rela->sym later) */
+	ALLOC_LINK(ksym_sec_sym, &kelf->symbols);
+	ksym_sec_sym->sec = ksym_sec;
+	ksym_sec_sym->sym.st_info = GELF_ST_INFO(STB_LOCAL, STT_SECTION);
+	ksym_sec_sym->type = STT_SECTION;
+	ksym_sec_sym->bind = STB_LOCAL;
+	ksym_sec_sym->name = ".kpatch.symbols";
 
 	/* lookup strings symbol */
 	strsym = find_symbol_by_name(&kelf->symbols, ".kpatch.strings");
 	if (!strsym)
 		ERROR("can't find .kpatch.strings symbol");
-
-	/* add objname to strings */
-	objname_offset = offset_of_string(&kelf->strings, objname);
 
 	/* populate sections */
 	index = 0;
@@ -1946,6 +1958,17 @@ static void kpatch_create_dynamic_rela_sections(struct kpatch_elf *kelf,
 				continue;
 
 			external = 0;
+			/*
+			 * sym_objname is the name of the object to which
+			 * rela->sym belongs. We'll need this to build
+			 * ".klp.sym." symbol names later on.
+			 *
+			 * By default sym_objname is the name of the
+			 * component being patched (vmlinux or module).
+			 * If it's an external symbol, sym_objname
+			 * will get reassigned appropriately.
+			 */
+			sym_objname = objname;
 
 			if (rela->sym->bind == STB_LOCAL) {
 				/* An unchanged local symbol */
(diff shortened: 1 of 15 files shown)


jpoimboe:
First off, sorry for taking so long to look at this.  I was distracted with other things and it really takes some time and effort for me to page this stuff back into my brain.  That said, it's only *partially* paged in, so excuse me if I ask some dumb questions :-)

My biggest question is about why `.kpatch.symbols` and `.kpatch.relocations` didn't work.  Would you mind spelling it out with some examples?

Also now that kpatch-build has more moving parts, it would be good to document somewhere the output of each piece, with examples.  (And that would also help me with reviewing it.)

BTW, looking at `lookup_exported_symbol_objname()` reminded me that I had actually implemented something similar a few years ago: https://github.com/jpoimboe/kpatch/commit/8cdca59c8844505bcf40b59b5c5050b03e974fea .  I can't remember if that code works or not, but it at least seems to do a lot less pointer arithmetic, and it seems safer to me at first glance.

flaming-toast:
> First off, sorry for taking so long to look at this. I was distracted with other things and it really takes some time and effort for me to page this stuff back into my brain. That said, it's only partially paged in, so excuse me if I ask some dumb questions :-)

Completely understand, and there are no dumb questions :-) I don't expect this to be merged anytime soon, so we can work on it bit by bit. It's on me to unfurl this giant ball of confusion somehow...so currently I'm working on some documentation/better explanation of how this all works together, and I'll comment on here again when that's done.

> My biggest question is about why .kpatch.symbols and .kpatch.relocations didn't work. Would you mind spelling it out with some examples?

Yeah, this was an unfortunate limitation I ran into. The short explanation is: there is no way to stuff enough information into .kpatch.relocations to figure out which base section and which object each rela originally belonged to when it comes time to create the klp rela sections (which are created after the patch module is built to circumvent the linker), because all sections will get shuffled when the final object (output.o) is linked. Section names are not required to be unique (see `ld --unique`), and section indices are also not reliable since they get shuffled around. The main limitation we need to work with is the fact the linker will only accept one rela section per base section, but klp rela sections are real SHT_RELA sections that apply to base sections (so we potentially get two rela sections per base section).

Q: __Hm, then why not just create the .klp.rela. sections during create-diff-object?__
The problem with that though is that the linker would reject these sections when output.o is linked (only 1:1 mapping between base section->rela section is allowed) and it'd also throw away our SHF_RELA_LIVEPATCH flag (and also the SHN_LIVEPATCH flag).

Q: __Well, why not make them temporarily SHT_PROGBITS instead of SHT_RELA?__
Here we will have the same problem of figuring out which klp rela progbits section applies to which base section after output.o is linked. Given a klp rela progbits section, how can we tell which base section it's supposed to apply to, when section names aren't required to be unique? If they were SHT_RELA sections instead, we'd get automatic tracking of which base section a rela sec belongs to.

Q: __Then how does this PR attempt to solve this problem__? 
To get around this, this patchset just forces create-diff-object not only create dynrelas, but to retain the original relas and symbols in the intermediary objects create-diff-object outputs (these relas/syms were formally stripped). Then depending on the kernel version, kpatch-build will either call a script (`strip-klp-elements`) to strip these relas/syms and keep dynrelas, or call a different script __after__ the patch module is compiled (`rebuild-klp-elements`) to create the klp rela sections from the retained relas/syms from the create-diff-object phase.

Well, that wasn't as short as I had hoped :-( If that is still confusing, I'm working on a better explanation/documentation that I'll post here tomorrow.

> BTW, looking at lookup_exported_symbol_objname() reminded me that I had actually implemented something similar a few years ago: jpoimboe@8cdca59 . I can't remember if that code works or not, but it at least seems to do a lot less pointer arithmetic, and it seems safer to me at first glance.

Ah, I haven't tested your implementation yet, but `lookup_exported_symbol_objname()` was actually pilfered from scripts/mod/modpost.c (save a few modifications to fit kpatch's needs), so I was just reusing the code there.

flaming-toast:
Here is a better commit roadmap/list because I dislike Github's UI...
Each commit is classified as a setup-commit, bugfix, or core change:
1. 6ac1d28 log: include error.h `[bugfix]`
2. a2e7745 list: add list_empty() function `[setup for commit 15]`
3. 6b57ee6 kpatch-elf: add kpatch/livepatch related Elf constants `[setup]`
4. be675d5 kpatch-elf: add kpatch_reindex_elements() and... `[setup for 14 and 15]`
5. b14dadc kpatch-elf: ensure SHN_LIVEPATCH syms don't... `[bugfix]`
6. c05889e kpatch-elf: fix kpatch_find_fentry_calls() sym->sec check `[bugfix]`
7. 3c01de6 kpatch-elf: add 'objname' fields to symbol and section structs `[setup for 11]`
8. e009133 kpatch-elf: add 'pos' field to symbol struct  `[setup]`
9. 0a0ff30 kpatch-elf: add is_klp_sym() helper function `[setup]`
10. 753092f lookup: add lookup_exported_symbol_objname() `[setup for 11]`
11. b703e82 create-diff-object: retain formerly stripped relas and symbols `[core]`
12. b1c4bf3 create-diff-object: create intermediary names for special sections `[core]`
13. cceee33 create-diff-object: rename 'name' variable to 'objname' `[bugfix]`
14. 5114ae8 strip-klp-elements: add new program to remove extraneous klp elements `[core]`
15. f2ca0f5 rebuild-klp-elements: add new program to rebuild klp elements `[core]`
16. 98f6d4d kpatch-build: build dynrelas or klp relas depending on kernel version `[core]`
17. 9111b57 kpatch.lds: add preprocessor checks to define symbols... `[core]`
18. e9ac55d livepatch-patch-hook: ensure compatibility with kernels... `[core]`

flaming-toast:
I put together a document that better shows the overarching changes made (and hopefully makes the pull request more palatable...). I think this effectively illustrates the changes without writing a gigantic wall of text :-) It can be found here: http://eisen.io/work/pr607.pdf

The diagram roughly follows kpatch-build and shows what changes can be expected at each step. The colored text in each blue square (and these blue squares represent objects outputted by `create-diff-object`) shows what new elements are added or changed to the resulting object. The colored elements are new changes and don't exist in our current implementation on master. Hope this helps explaining what all these commits ultimately do together.

jpoimboe:
Thanks, the chart does help a lot.  I think I have a decent understanding of what the PR does.

But I'm still trying to understand why the original idea of .kpatch.symbols and .kpatch.relocations didn't work.  I was hoping we could have something like:

1. create-diff-object creates an intermediate object (with e.g., .kpatch.symbols and .kpatch.relocations).

2. link all intermediate objects together to form a linked object.

3. convert linked object to its final format (with e.g., convert-klp-module, convert-old-klp-module, convert-kpatch-module).

And I think you're saying that's not possible, right?  You said there's no way to put enough information into .kpatch.relocations to figure out which base section each rela belonged to.  Can you elaborate on why that is?

Or, maybe it would help if I clarify why I'm confused with an example.  I built the meminfo_proc_show patch .ko, before and after, and dumped the relevant relocations.

Before:

```
Relocation section [26] '.rela.kpatch.dynrelas' for section [25] '.kpatch.dynrelas' at offset 0x478f8 contains 27 entries:
  000000000000000000  X86_64_64       000000000000000000     +73 meminfo_proc_show
  0x0000000000000020  X86_64_64       000000000000000000     +26 .kpatch.strings
  0x0000000000000028  X86_64_64       000000000000000000      +0 .kpatch.strings
  0x0000000000000038  X86_64_64       000000000000000000     +80 meminfo_proc_show
  0x0000000000000058  X86_64_64       000000000000000000     +38 .kpatch.strings
  0x0000000000000060  X86_64_64       000000000000000000      +0 .kpatch.strings
  0x0000000000000070  X86_64_64       000000000000000000    +101 meminfo_proc_show
  0x0000000000000090  X86_64_64       000000000000000000     +54 .kpatch.strings
  0x0000000000000098  X86_64_64       000000000000000000      +0 .kpatch.strings
  0x00000000000000a8  X86_64_64       000000000000000000    +238 meminfo_proc_show
  0x00000000000000c8  X86_64_64       000000000000000000     +76 .kpatch.strings
  0x00000000000000d0  X86_64_64       000000000000000000      +0 .kpatch.strings
  0x00000000000000e0  X86_64_64       000000000000000000    +284 meminfo_proc_show
  0x0000000000000100  X86_64_64       000000000000000000     +91 .kpatch.strings
  0x0000000000000108  X86_64_64       000000000000000000      +0 .kpatch.strings
  0x0000000000000118  X86_64_64       000000000000000000    +296 meminfo_proc_show
  0x0000000000000138  X86_64_64       000000000000000000    +110 .kpatch.strings
  0x0000000000000140  X86_64_64       000000000000000000      +0 .kpatch.strings
  0x0000000000000150  X86_64_64       000000000000000000    +528 meminfo_proc_show
  0x0000000000000170  X86_64_64       000000000000000000     +54 .kpatch.strings
  0x0000000000000178  X86_64_64       000000000000000000      +0 .kpatch.strings
  0x0000000000000188  X86_64_64       000000000000000000   +1352 meminfo_proc_show
  0x00000000000001a8  X86_64_64       000000000000000000    +126 .kpatch.strings
  0x00000000000001b0  X86_64_64       000000000000000000      +0 .kpatch.strings
  0x00000000000001c0  X86_64_64       000000000000000000   +1360 meminfo_proc_show
  0x00000000000001e0  X86_64_64       000000000000000000    +149 .kpatch.strings
  0x00000000000001e8  X86_64_64       000000000000000000      +0 .kpatch.strings
```

After:
```
Relocation section [52] '.klp.rela.vmlinux..text.meminfo_proc_show' for section [10] '.text.meminfo_proc_show' at offset 0x853d8 contains 9 entries:
  0x0000000000000550  X86_64_PC32     000000000000000000      -4 .klp.sym.vmlinux.arch_report_meminfo,0
  0x0000000000000548  X86_64_PC32     000000000000000000      -4 .klp.sym.vmlinux.hugetlb_report_meminfo,0
  0x0000000000000210  X86_64_PC32     000000000000000000      -4 .klp.sym.vmlinux.total_swapcache_pages,0
  0x0000000000000128  X86_64_PC32     000000000000000000      -4 .klp.sym.vmlinux.vm_commit_limit,0
  0x000000000000011c  X86_64_PC32     000000000000000000      -4 .klp.sym.vmlinux.num_poisoned_pages,0
  0x00000000000000ee  X86_64_PC32     000000000000000000      -4 .klp.sym.vmlinux.totalcma_pages,0
  0x0000000000000065  X86_64_PC32     000000000000000000      -4 .klp.sym.vmlinux.total_swapcache_pages,0
  0x0000000000000050  X86_64_PC32     000000000000000000      +4 .klp.sym.vmlinux.vm_committed_as,0
  0x0000000000000049  X86_64_PC32     000000000000000000      -4 .klp.sym.vmlinux.si_swapinfo,0
```

Specifically, compare the first dynrela entry with its corresponding klp rela entry:
```
  000000000000000000  X86_64_64       000000000000000000     +73 meminfo_proc_show
  0x0000000000000020  X86_64_64       000000000000000000     +26 .kpatch.strings
  0x0000000000000028  X86_64_64       000000000000000000      +0 .kpatch.strings
```
```
  0x0000000000000049  X86_64_PC32     000000000000000000      -4 .klp.sym.vmlinux.si_swapinfo,0
```

Now you can't see the entire dynrela struct being dumped here, but you can see an important part.  Notice the dynrela itself has a rela.  The meminfo_proc_show rela points to the function address where the dynrela will be written: meminfo_proc_show+73 (aka 0x49 which is also shown in the klp rela).

The dynrela's rela will still be valid, even after linking.  So it follows that the information about which section the dynrelas apply to, the base section, should still be available after linking.  In this case, we just have to look for the section which contains the meminfo_proc_show symbol: .text.meminfo_proc_show.

So AFAICT, that is how you figure out which base section each rela belongs to.  Am I missing something?

flaming-toast:
Ah, I think you are right. Now I am not sure how I even reached that conclusion in the first place, as that solution was right under my nose the whole time!! :-/ 

This goes to show that I should have discussed the whole design first, and code second. Well, we can do that now :-) I'll change this PR to implement the .kpatch.symbols and .kpatch.relocations approach (and that was the original idea...). I've scrapped this design and wrote a new one (with pseudocode) [here](http://eisen.io/work/pr607_v2.pdf) to get a feel for how this will be implemented...

Basically, all objects outputted by create-diff-object will have .kpatch.relocations and .kpatch.symbols. These will be arrays of struct kpatch_relocation and struct kpatch_symbol that will contain enough information to either build klp relas or dynrelas. To correlate each kpatch_relocation with its corresponding kpatch_symbol, the .kpatch.relocations section will have relas pointing to kpatch_symbols in .kpatch.symbols. Do you think you could could skim through the linked design/algorithm above to see if there are any potential issues with it?

jpoimboe:
So if I understand correctly, .kpatch.symbols and .kpatch.relocations will replace .kpatch.dynrelas.  Is there a reason we can't instead just keep .kpatch.dynrelas, and have it be the intermediary section (which just happens to also be the final section for kpatch modules)?

flaming-toast:
>So if I understand correctly, .kpatch.symbols and .kpatch.relocations will replace .kpatch.dynrelas. Is there a reason we can't instead just keep .kpatch.dynrelas, and have it be the intermediary section (which just happens to also be the final section for kpatch modules)?

Technically yes, we could jam everything into one section (that being .kpatch.dynrelas), but we'll need to add more fields into the dynrela struct in order to be able to reconstruct the symbols to add back to the symtab in convert-klp-module. IMO this new combined "super" struct (with both symbol and rela information) looks clumsy and potentially confusing. We'll probably have to prefix some fields with `rela_` or `sym_` to remember which field deals with what. Example of what that might look like:
```C
struct kpatch_patch_dynrela { 
        unsigned long rela_dest; 
        unsigned long sym_src; // value of symbol
        unsigned long rela_type; // rela type
        unsigned char sym_type, sym_bind; // symbol type and bind
        unsigned long sympos;
        char *sec_objname; // objname of base section
        char *sym_name; // name of symbol
        char *sym_objname; // objname of symbol
        int rela_addend; 
        int rela_external; 
};
```

So that's why I felt that separating these fields into two different structs `kpatch_relocation` and `kpatch_symbol` was cleaner and more readable. In addition, I wasn't sure if we wanted the intermediary sections to be implementation-agnostic, which was what .kpatch.symbols and .kpatch.relocations were supposed to be.

jpoimboe:
Ok, I think I got it now.  I didn't realize that .kpatch.symbols is needed for reconstructing the klp symbols.

So your design sounds good to me!

flaming-toast:
This new set of changes implements the design described [in this comment](https://github.com/dynup/kpatch/pull/607#issuecomment-249990455). Though it's quite possible we may be deprecating kpatch-build soon, I'll continue to leave this PR open with the code available for anyone interested in building patch modules for kernel versions 4.7+ during that transition period. 
1. c752900 log: include error.h `[bugfix]`
2. f2c996c args.h: move any argument-related funcs/data to args.h `[cleanup]`
3. 83c2673 kpatch-elf: add kpatch/livepatch related Elf constants `[setup]`
4. 0519ca2 kpatch-elf: add kpatch_reindex_elements() and... `[setup]`
5. 497c500 kpatch-elf: ensure SHN_LIVEPATCH syms don't get set to SHN_UNDEF... `[bugfix]`
6. 9db0c0c kpatch-elf: fix kpatch_find_fentry_calls() sym->sec check `[bugfix]`
7. ac96b52 kpatch-elf: find base section by index rather than name `[bugfix]`
8. 3240600 lookup: add lookup_exported_symbol_objname() `[setup]`
9. e61cce6 create-diff-object: rename 'name' variable to 'objname' `[cleanup]`
10. 618c99f create-diff-object: create .kpatch.relocations and .kpatch.symbols... `[core]`
11. 4c1c847 create-klprela-module: add new program that creates patch modules... `[core]`
12. 49c8fd5 create-klprela-module: create .klp.arch. sections `[core]`
13. 521c273 create-dynrela-module: add new program that creates dynrelas `[core]`
14. 26ac5ab kpatch.lds: add preprocessor checks to define symbols/discard... `[core]`
15. 9f7e7d1 livepatch-patch-hook: ensure compatibility with kernels < 4.7 and >= 4.7 `[core]`
16. 532483d kpatch-build: build dynrelas or klp relas depending on kernel...`[core]`

kpatch-build/lookup.c
```diff
@@ -223,6 +236,132 @@ int lookup_is_exported_symbol(struct lookup_table *table, char *name)
...
+/*
+ * lookup_exported_symbol_objname - find the object/module an exported
+ * symbol belongs to.
+ *
+ * Based on read_dump() from scripts/mod/modpost.c with a few modifications.
+ *
+ * Module.symvers line format:
+ * 0x12345678<tab>symbol<tab>module[[<tab>export]<tab>something]
+ */
+char *lookup_exported_symbol_objname(const char *fname, char *symname)
```
jpoimboe: This opens and reads the file every time the function is called.   I think it would be cleaner to read the file once in lookup_open() like my version from a few years ago: https://github.com/jpoimboe/kpatch/commit/8cdca59c8844505bcf40b59b5c5050b03e974fea

kpatch-build/create-klprela-module.c
```diff
@@ -0,0 +1,400 @@
...
+		len = strlen(KLP_ARCH_PREFIX) + strlen(objname) + \
+		      strlen(base->name) + 2;
+		newname = malloc(len);
+		if (!newname)
+			ERROR("malloc");
+		strcpy(newname, KLP_ARCH_PREFIX);
+		strcat(newname, objname);
+		strcat(newname, ".");
+		strcat(newname, base->name);
+		base->name = newname;
```
jpoimboe: Also, for kernels older than 4.10, which don't have the `.klp.arch` stuff, I think we want to skip calling this function altogether so the special sections don't get renamed?

kpatch-build/kpatch-build
```diff
@@ -93,6 +95,26 @@ clean_cache() {
...
+		warn "$1 SIGSEGV"
+		if ls core* &> /dev/null; then
+			cp core* /tmp
+			die "core file at /tmp/$(ls core*)"
+		fi
+		die "no core file found, run 'ulimit -c unlimited' and try to recreate"
+	fi
+}
+
+# $1 <= $2
```
jpoimboe: should be '>=' ?

kpatch-build/kpatch-build
```diff
@@ -93,6 +95,26 @@ clean_cache() {
...
+function version_gte {
+	[  "$1" = "`echo -e "$1\n$2" | sort -rV | head -n1`" ]
```
jpoimboe: clever!

kpatch-build/args.h
```diff
@@ -0,0 +1,41 @@
...
+struct arguments {
+	int debug;
+	char **args;
+};
+
+extern char args_doc[];
+extern struct argp_option options[];
+extern struct argp argp;
+
+error_t parse_opt (int key, char *arg, struct argp_state *state)
```
flaming-toast: Just to be clear, are you referring to the entire args.h file? If so, I'll remove this header and paste the args code back into the files that need it.

jpoimboe: Yeah, I think it would be good to get rid of the whole args.h and just copy/paste where needed.

kpatch-build/create-diff-object.c
```diff
@@ -2296,7 +2319,10 @@ int main(int argc, char *argv[])
 	struct lookup_table *lookup;
 	struct section *sec, *symtab;
 	struct symbol *sym;
-	char *hint = NULL, *name, *pos;
+	char *hint = NULL, *objname, *pos;
+	char *mod_symvers_path, *pmod_name;
+
+	arguments.args = malloc(sizeof(char *) * NUM_ARGS);
```
flaming-toast: I think I had changed it to be dynamically allocated because of the shared args.h header, where `struct arguments` is defined. The `args` field was formerly a static array with set size, but since different files had varying number of arguments, I had changed it to `char **` to accommodate differently sized arrays.

But I think you suggested we get rid of args.h anyway (maybe that was overkill), so I'll revert back to static arrays.

kpatch-build/kpatch-build
```diff
@@ -459,10 +481,15 @@ grep -q "CONFIG_DEBUG_KERNEL=y" "$OBJDIR/.config" || die "kernel doesn't have 'C
 if grep "CONFIG_LIVEPATCH=y" "$OBJDIR/.config" > /dev/null; then
 	# The kernel supports livepatch.
 	KBUILD_EXTRA_SYMBOLS=""
+	if version_gte ${ARCHVERSION//-*/} 4.7.0; then
+		USE_KLPRELAS=true
+		KPATCH_LDFLAGS="--unique=.parainstructions --unique=.altinstructions"
```
flaming-toast: The original rationale for this is the following:

If we have a patch that ends up generating multiple .parainstructions sections for different objects (e.g., kvm and vmlinux), these sections get merged into one big `.parainstructions` section when `output.o` is linked. However, these sections need to be applied per-object, with separate `.klp.arch.$objname.` sections. It seemed easier to pass in `--unique` than to have `create-klp-module` (which is invoked after the patch module .ko is built) iterate through a merged `.parainstructions` section and tease out which parts of it belong to what object.

But looking at this again, I think this does the wrong thing when we deal with multiple .o's from the same object that have their .parainstructions sections included. The upstream livepatch code doesn't handle the case when a single object has multiple .klp.arch. sections. :-/ 

So maybe the correct thing to do is to not use --unique and have `create-klp-module` do the harder thing (which is to iterate through the arch sections and manually separate them per-object...)

kpatch-build/create-klprela-module.c
jpoimboe: Yeah, that sounds good.  Or maybe the opposite, make .klp.arch creation the default and have a --no-klp-arch-sections flag, to make it more futuristic.

kpatch-build/kpatch-build
jpoimboe: Ah, when I made this comment, I think I had forgotten that create-klp-module runs on the .ko rather than the individual .o files.  Is there any reason why it couldn't be renamed to `create-klp-object` and run on the .o file instead?

jpoimboe: Oh right.  I'm sure you've explained this to me several times before :-)

flaming-toast:
Phew. Updated PR with all the comments in the previous review addressed...

TODO:
- Split up commits to be neater/progress more logically.
- Reword some commit messages so as to describe the changes better
- More comments are needed to explain how an intermediate object with .kpatch.symbols, .kpatch.relocations, and .kpatch.arch sections turn into a klp module or a kpatch module

flaming-toast:
v4:
- Change create-klprela-module and create-dynrela-module to create-klp-module and create-kpatch-module instead.
- Remove args.h and just paste in the argument handling code in each program that needs it
- Instead of repeating `#if LINUX_VERSION_CODE < KERNEL_VERSION(4, 7, 0)` all over livepatch-patch-hook.c, define `NEED_KLP_RELOCS` once and use it.
- Instead of using that weird loop+switch statement (see: https://github.com/dynup/kpatch/pull/607#discussion_r87620272) to find relas with a specific offset, abstract that away into a new, simpler function: `find_rela_by_offset()`
- Create `.klp.arch.` sections for only `.parainstructions` and `.altinstructions` sections, and only for kernels >= 4.9. We enable the creation of these sections through a new switch for `create-klp-module`, called `--no-klp-arch-sections`
- There was a discovered issue with the handling of multiple arch sections under one object (See: https://github.com/dynup/kpatch/pull/607#discussion_r90160341, which also explains why the `--unique` flag for ld was used). The new create-klp-module program will now manually merge multiple .parainstructions sections (same for .altinstructions) under the same object.
- Base the Module.symvers parsing code on Josh's patch here:  https://github.com/jpoimboe/kpatch/commit/8cdca59c8844505bcf40b59b5c5050b03e974fea
---
New commit list for v4:
4c0d69a log: include error.h [bugfix]
6300273 kpatch-elf: add livepatch related Elf constants [setup]
46c4f0d4 kpatch-elf: add kpatch_reindex_elements() and kpatch_rebuild_rela_section_data() [setup]
c3579501 kpatch-elf: ensure SHN_LIVEPATCH syms don't get set to SHN_UNDEF when...[bugfix]
12721bd9f kpatch-elf: skip livepatch syms in kpatch_find_fentry_calls() [bugfix]
323c3d480 kpatch-elf: for rela sections, find base section by index rather than name [bugfix]
be8b50f54 create-diff-object: rename 'name' variable to 'objname' [cleanup]
1ac2c43db lookup: parse Module.symvers in lookup_open() [setup]
377a96186 create-diff-object: create .kpatch.relocations and .kpatch.symbols sections [core]
a6508a849 create-diff-object: create intermediate .kpatch.arch section [core]
6641a950e kpatch-elf: add find_rela_by_offset() [setup]
998f5684d7 create-klp-module: add new program that creates patch modules that have klp elements [core]
90d57e5782 create-kpatch-module: add new program that creates kpatch modules [core]
ddcfc8ea1f2 livepatch-patch-hook: ensure compatibility with kernels < 4.7 and >= 4.7 [core]
50544bbdfe kpatch-build: build dynrelas or klp relas depending on kernel version [core]

---
This patchset is huge and dense, so I'll give a high level explanation of all the moving parts here.
The gist of what we are doing is the following.
1. Instead of generating `.kpatch.dynrelas` sections, `create-diff-object` will now output objects with the following intermediate sections:
 `.kpatch.relocations`
  `.kpatch.symbols`
  `.kpatch.arch` (optional, if .parainstructions and .altinstructions exist)
These sections are implementation-agnostic and contain just enough information to build either a kpatch (dynrela) module or a klp (klp rela + sym) module.

2.  From here, the build process will take a route depending on what version of the kernel we are working with.
2a. For >= 4.7 kernels, we take the completely built patch module .ko with the intermediate sections from Step 1 and and create new sections:
 `.kpatch.relocations` → `.klp.rela.$objname.section_name`
 `.kpatch.arch` → `.klp.arch.$objname.section_name`
And with klp symbols following the following format:
 `.kpatch.symbols` → `.klp.sym.$objname.symname,pos`
2b. For < 4.7 kernels, we will operate on the outputted objects from Step 1 and create `.kpatch.dynrelas` sections. (This route is actually much easier than step 2a)
 `.kpatch.relocations` + `.kpatch.symbols` → `.kpatch.dynrelas`

Of course this description is missing a lot of details, but this is basically the bird's eye view of what is going on in this patchset.

jpoimboe:
Also just noticed that create-*-module need to be added to .gitignore.

flaming-toast:
@jpoimboe Ah, I was able to reproduce the segfaults. The null dereference was caused by a last-minute change I had made to kpatch_find_fentry_calls(), called from kpatch_elf_open() (from this commit: https://github.com/dynup/kpatch/commit/12721bd9f150c7ab06e35c7e9618fb4863e6c222). So both programs were calling kpatch_elf_open() and segfaulting. I updated this PR with the following commit: https://github.com/dynup/kpatch/pull/607/commits/2e2983c075d1f24b3a9922700bd77718bd51cc9e, which should hopefully fix the issue. I'm still working on addressing the remaining comments, but you can force pull the changes and try the builds again to see if the segfaults go away.

jpoimboe:
@flaming-toast Yep, verified that the quick integration test works now.

kpatch-build/create-klp-module.c
```diff
@@ -0,0 +1,590 @@
...
+	ALLOC_LINK(ks, &klp_symbols);
+	ks->sym = sym;
+	ks->pos = ksym->pos;
+	ks->oldname = name;
+	ks->objname = objname;
+
+	return sym;
+}
+
+static struct section *kpatch_find_klp_relasec(struct kpatch_elf *kelf,
```
flaming-toast: How about I remove those `kpatch_` prefixes in the create-*.c and add a `kelf_` prefix  to the kpatch-elf functions? I think kelf_ sounds better than elf_, to distinguish from elf.h functions. I can add those two changes to this PR, although they can go into a separate PR as well...

jpoimboe: Yeah, the 'elf_' prefix is the same namespace used by libelf.h, and might confuse things.  Then again, I don't like 'kelf_' either because ideally our elf code should have no knowledge of kpatch. It should really be a layer above libelf to make it easier to interact with it.

FWIW, I have a very similar library in upstream Linux in tools/objtool/elf.{ch}, though it has diverged a bit -- it has some hashes for fast lookups.  Since I'm using it in two projects, and others might find it useful, it might be nice to move that code to another project someday.  No idea what to call it, or prefix the code with, though.

But anyway, such changes should indeed probably go into a separate PR.

flaming-toast: Alright! In any case I'll probably take out the existing kpatch_ prefixes in create-klp-module and create-kpatch-module...and we can save the rest of the renaming for a separate PR.

kpatch-build/create-klp-module.c
```diff
@@ -0,0 +1,590 @@
...
+
+			sec = find_section_by_name(&kelf->sections, secname);
+			if (!sec)
+				sec = kpatch_create_klp_arch_section(kelf, secname);
+
+			old_size = sec->data->d_size;
+			new_size = old_size + entry->sec->data->d_size;
+			sec->data->d_buf = realloc(sec->data->d_buf, new_size);
+			sec->data->d_size = new_size;
+			memcpy(sec->data->d_buf, entry->sec->data->d_buf, new_size);
```
flaming-toast: Gah, you're right! Fixed in v5..

flaming-toast:
v5:
- Removed all the bookkeeping lists from create-klp-module.c (`klp_symbols`, `klp_objects`, `klp_relasecs`). They were extraneous and did not help with code readability. Instead, just search `kelf->symbols` and `kelf->sections` for already-allocated klp symbols and klp rela sections.
- Removed all `kpatch_` prefixes from create-klp-module.c and create-kpatch-module.c. However, I did keep the prefixes in create-diff-object and kpatch-elf, to avoid making this PR bigger than it already is.
- Add create-klp-module and create-kpatch-module in .gitignore
- Move struct definitions for kpatch_symbol, kpatch_relocation, and kpatch_arch to a separate header file (kpatch-intermediate.h) instead of putting them in kpatch-elf.h.
- Added more comments to create-klp-module and create-kpatch-module to better explain the process
- Since the section removal code (formerly kpatch_remove_intermediate_sections() and kpatch_remove_arch_sections()) in create-*module was duplicated 3 times, I moved that code to kpatch-elf.c as a new function, "kpatch_remove_and_free_section()".
- Addressed the remaining minor comments from the last review.

Lightly tested on a 4.8.12 kernel with livepatch disabled and on 4.10-rc2 with livepatch enabled.

To test a patch module with multiple .klp.arch. sections (with both .parainstructions and .altinstructions from different objects), [this](http://paste.fedoraproject.org/535277/9978148/) patch was used.

kpatch-build/kpatch-build
```diff
@@ -610,14 +640,36 @@ if [[ ! -z $UBUNTU_KERNEL ]]; then
...
+KPATCH_BUILD="$SRCDIR" KPATCH_NAME="$PATCHNAME" \
+KBUILD_EXTRA_SYMBOLS="$KBUILD_EXTRA_SYMBOLS" \
+KPATCH_LDFLAGS="$KPATCH_LDFLAGS" \
 	make "O=$OBJDIR" >> "$LOGFILE" 2>&1 || die
 
+if ! $KPATCH_MODULE; then
+	if [[ -z "$KPATCH_LDFLAGS" ]]; then
+		extra_flags="--no-klp-arch-sections"
+	fi
+	cp $TEMPDIR/patch/kpatch-$PATCHNAME.ko $TEMPDIR/patch/kpatch-tmp.ko || die
```
jpoimboe: I think this would break if the patch were named `tmp.patch`.  Maybe just call it tmp.ko?

kpatch-build/kpatch-intermediate.h
```diff
@@ -0,0 +1,48 @@
...
+
+struct kpatch_symbol {
+	unsigned long src;
+	unsigned long pos;
+	unsigned char bind, type;
+	char *name;
+	char *objname; /* object to which this sym belongs */
+};
+
+/* For .kpatch.{symbols,relocations,arch} sections */
```
jpoimboe: This comment should be higher, above kpatch_symbol

flaming-toast: It seems my copy/paste skills are lacking ;-)

kpatch-build/kpatch-build
flaming-toast: Ah yeah, good point! Will fix.

flaming-toast:
Very glad to hear this is finally shaping up (at last!) :-) Addressed the comments in the additional commits.

PR #610, opened 2016-09-29
Add kpatch-build --skip-cleanup option

joe-lawrence:
Add a switch to kpatch-build that provides an opt-out to the cleanup
portion of the script.  This can be handy when debugging $TEMPDIR or
$RPMTOPDIR contents, as well as inspecting the patched source code
itself.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -47,6 +47,7 @@ TEMPDIR="$CACHEDIR/tmp"
 LOGFILE="$CACHEDIR/build.log"
 APPLIEDPATCHFILE="kpatch.patch"
 DEBUG=0
+SKIPCLEANUP=0
 SKIPGCCCHECK=0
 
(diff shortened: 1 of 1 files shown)


kpatch-build/kpatch-build
```diff
@@ -287,7 +293,7 @@ mkdir -p "$TEMPDIR" || die "Couldn't create $TEMPDIR"
 rm -rf "$TEMPDIR"/*
 rm -f "$LOGFILE"
 
-trap cleanup EXIT INT TERM HUP
+[[ $SKIPCLEANUP -eq 0 ]] && trap cleanup EXIT INT TERM HUP
```
flaming-toast: Actually, scratch that comment. It'd be useful to keep $TMPDIR/$RPMTOPDIR under those other conditions.

flaming-toast:
:+1:

PR #616, opened 2016-11-10
create-diff-object: Update fixup offsets in .rela__ex_table

### LINKED ISSUE #615: ex_table relocation offsets are wrong
The following patch (against 4.4) generates an exception table with relocations pointing to an offset in fixup table that is wrong:
```
diff --git a/fs/readdir.c b/fs/readdir.c
index ced6791..7389fb8 100644
--- a/fs/readdir.c
+++ b/fs/readdir.c
@@ -286,6 +286,8 @@ SYSCALL_DEFINE3(getdents64, unsigned int, fd,
 	};
 	int error;
 
+	printk("hello\n");
+
 	if (!access_ok(VERIFY_WRITE, dirent, count))
 		return -EFAULT;
 
 ```

```
  [12] .fixup            PROGBITS        0000000000000000 0008c8 00000a 00  AX  0   0  1
  [13] .rela.fixup       RELA            0000000000000000 030e08 000018 18     61  12  8

Relocation section '.rela.fixup' at offset 0x30e08 contains 1 entries:
0000000000000006  0000006700000002 R_X86_64_PC32          0000000000000000 SyS_getdents64 + b5

Relocation section '.rela__ex_table' at offset 0x31090 contains 2 entries:
0000000000000000  0000006700000002 R_X86_64_PC32          0000000000000000 SyS_getdents64 + b5
0000000000000004  0000000700000002 R_X86_64_PC32          0000000000000000 .fixup + 9b
```

The problem is that the fixup table is rewritten (with several entries removed), but the offsets in the ex_table are not updated.

### PR DESCRIPTION
**Pull request by rosslagerwall**
When pruning entries from the fixup table, update the offsets in
.rela__ex_table otherwise the relas might point to the wrong fixup entry
or even out of the .fixup section.

Fixes #615.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1490,6 +1490,31 @@ int should_keep_rela_group(struct section *sec, int start, int size)
 	return found;
 }
 
+/*
+ * When updating .fixup, the corresponding addends in .ex_table need to be
+ * updated too. Stash the result in rela.r_addend so that the calculation in
+ * fixup_group_size() is not affected.
+ */
+void kpatch_update_ex_table_addend(struct kpatch_elf *kelf,
+				   struct special_section *special,
+				   int src_offset, int dest_offset,
+				   int group_size)
+{
+	struct rela *rela;
+	struct section *sec;
+
+	sec = find_section_by_name(&kelf->sections, ".rela__ex_table");
+	if (!sec)
+		ERROR("missing .rela__ex_table section");
+
+	list_for_each_entry(rela, &sec->relas, list) {
+		if (!strcmp(rela->sym->name, ".fixup") &&
+		    rela->addend >= src_offset &&
+		    rela->addend < src_offset + group_size)
+			rela->rela.r_addend = rela->addend - (src_offset - dest_offset);
+	}
+}
+
 void kpatch_regenerate_special_section(struct kpatch_elf *kelf,
 				       struct special_section *special,
 				       struct section *sec)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by rosslagerwall** [comment]
I think this fixes the issue correctly, but would appreciate a careful review.

**Comment by rosslagerwall** [comment]
It's also kind of a hack. There's probably a better way.

File: kpatch-build/create-diff-object.c [hunk 1]
```diff
@@ -1490,6 +1490,32 @@ int should_keep_rela_group(struct section *sec, int start, int size)
...
+	if (strcmp(special->name, ".fixup"))
+		return;
+
+	sec = find_section_by_name(&kelf->sections, ".rela__ex_table");
+	if (!sec)
+		ERROR("missing .rela__ex_table section");
+
+	list_for_each_entry(rela, &sec->relas, list) {
+		if (!strcmp(rela->sym->name, ".fixup") &&
+		    rela->addend >= src_offset)
```
rosslagerwall: In the usual case it probably wouldn't matter as `rela->rela.r_addend` would just get overwritten later with the correct value.

But what you suggested would certainly be better.

rosslagerwall:
OK, I've updated it.

jpoimboe:
:+1:

kpatch-build/create-diff-object.c
```diff
@@ -1533,6 +1569,11 @@ void kpatch_regenerate_special_section(struct kpatch_elf *kelf,
 				rela->rela.r_offset = rela->offset;
 
 				rela->sym->include = 1;
+
+				kpatch_update_ex_table_addend(kelf, special,
+							      src_offset,
+							      dest_offset,
+							      group_size);
```
rosslagerwall: Fixed.

PR #619, opened 2016-11-10
kpatch-build: remove ~/.kpatch/src/.git dir on Fedora

jpoimboe:
Source RPMs for recent Fedora kernels have a '.git' subdirectory, which
causes '+' to be appended to the module version magic, causing the
module to fail to load:

  kpatch_readdir: version magic '4.8.6-201.fc24.x86_64+ SMP mod_unload ' should be '4.8.6-201.fc24.x86_64 SMP mod_unload '


flaming-toast:
:+1: Thanks, I noticed this too and it was starting to irk me :-) I additionally discovered a similar problem ("+" being appended to the version string) for user-specified source directories...will open a separate PR for that.

PR #620, opened 2016-11-16
Add CONFIG_RANDOMIZE_BASE KASLR support

### LINKED ISSUE #617: KASLR support
kpatch is broken on the latest Fedora 24 kernels which have CONFIG_RANDOMIZE_BASE enabled.

joe-lawrence:
Backport the symbol lookup and checking code from upstream livepatch
code that relies on a symbol position enumeration rather than a fixed
memory address.  Kpatch only verifies symbol matches for symbols found
in vmlinux, so drop the parts that deal with module symbols.

Fixes #617.

### DIFF
--- a/kmod/core/core.c
+++ b/kmod/core/core.c
@@ -75,8 +75,11 @@ struct kpatch_backtrace_args {
 };
 
 struct kpatch_kallsyms_args {
+	const char *objname;
 	const char *name;
 	unsigned long addr;
+	unsigned long count;
+	unsigned long pos;
 };
 
 /* this is a double loop, use goto instead of break */
@@ -582,58 +585,75 @@ static int kpatch_kallsyms_callback(void *data, const char *name,
 					 unsigned long addr)
 {
 	struct kpatch_kallsyms_args *args = data;
+	bool vmlinux = !strcmp(args->objname, "vmlinux");
 
-	if (args->addr == addr && !strcmp(args->name, name))
+	if ((mod && vmlinux) || (!mod && !vmlinux))
+		return 0;
+
+	if (strcmp(args->name, name))
+		return 0;
+
+	if (!vmlinux && strcmp(args->objname, mod->name))
+		return 0;
+
+	args->addr = addr;
+	args->count++;
+
+	/*
+	 * Finish the search when the symbol is found for the desired position
+	 * or the position is not defined for a non-unique symbol.
+	 */
+	if ((args->pos && (args->count == args->pos)) ||
+	    (!args->pos && (args->count > 1))) {
 		return 1;
-
-	return 0;
-}
-
-static int kpatch_verify_symbol_match(const char *name, unsigned long addr)
-{
-	int ret;
-
-	struct kpatch_kallsyms_args args = {
-		.name = name,
-		.addr = addr,
-	};
-
-	ret = kallsyms_on_each_symbol(kpatch_kallsyms_callback, &args);
-	if (!ret) {
-		pr_err("base kernel mismatch for symbol '%s'\n", name);
-		pr_err("expected address was 0x%016lx\n", addr);
-		return -EINVAL;
 	}
 
 	return 0;
 }
 
-static unsigned long kpatch_find_module_symbol(struct module *mod,
-					       const char *name)
+static int kpatch_find_object_symbol(const char *objname, const char *name,
+				     unsigned long sympos, unsigned long *addr)
 {
-	char buf[KSYM_SYMBOL_LEN];
+	struct kpatch_kallsyms_args args = {
+		.objname = objname,
+		.name = name,
+		.addr = 0,
+		.count = 0,
+		.pos = sympos,
+	};
 
-	/* check total string length for overrun */
-	if (strlen(mod->name) + strlen(name) + 1 >= KSYM_SYMBOL_LEN) {
-		pr_err("buffer overrun finding symbol '%s' in module '%s'\n",
-		       name, mod->name);
+	mutex_lock(&module_mutex);
+	kallsyms_on_each_symbol(kpatch_kallsyms_callback, &args);
+	mutex_unlock(&module_mutex);
+
+	/*
+	 * Ensure an address was found. If sympos is 0, ensure symbol is unique;
+	 * otherwise ensure the symbol position count matches sympos.
+	 */
+	if (args.addr == 0)
+		pr_err("symbol '%s' not found in symbol table\n", name);
+	else if (args.count > 1 && sympos == 0) {
+		pr_err("unresolvable ambiguity for symbol '%s' in object '%s'\n",
+		       name, objname);
+	} else if (sympos != args.count && sympos > 0) {
+		pr_err("symbol position %lu for symbol '%s' in object '%s' not found\n",
+		       sympos, name, objname);
+	} else {
+		*addr = args.addr;
 		return 0;
 	}
 
-	/* encode symbol name as "mod->name:name" */
-	strcpy(buf, mod->name);
-	strcat(buf, ":");
-	strcat(buf, name);
-
-	return kallsyms_lookup_name(buf);
+	*addr = 0;
+	return -EINVAL;
 }
 
 /*
  * External symbols are located outside the parent object (where the parent
  * object is either vmlinux or the kmod being patched).
  */
-static unsigned long kpatch_find_external_symbol(struct kpatch_module *kpmod,
-						 const char *name)
+static int kpatch_find_external_symbol(const char *objname, const char *name,
+				       unsigned long sympos, unsigned long *addr)
+
 {
 	const struct kernel_symbol *sym;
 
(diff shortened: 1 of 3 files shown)


kmod/core/core.c
```diff
@@ -583,29 +585,57 @@ static int kpatch_kallsyms_callback(void *data, const char *name,
...
+	if ((args->pos && (args->count == args->pos)) ||
+	    (!args->pos && (args->count > 1))) {
 		return 1;
+	}
 
 	return 0;
 }
 
-static int kpatch_verify_symbol_match(const char *name, unsigned long addr)
+static int kpatch_verify_symbol_match(const char *name,
```
jpoimboe: This should probably be made 100% identical to klp_find_object_symbol(), with its objname parameter, and renamed accordingly.  And the same for making kpatch_kallsyms_callback() similar to klp_find_callback().  Then I think we can get rid of kpatch_find_module_symbol() and symbol resolution will be more nicely klp-esque.

flaming-toast: I agree, we can combine the functionality of `kpatch_verify_symbol_match` with `kpatch_find_module_symbol` into the kpatch-analog version of `klp_find_object_symbol()`. Plus, `kpatch_verify_symbol_match` in the context of kaslr doesn't make sense anymore, since the original function's purpose was to verify if the recorded address (at compile time) in dynrela->src matches the one we find with kallsyms.

In livepatch, `klp_find_object_symbol()` interprets a NULL `objname` argument to mean vmlinux. If the new `kpatch_find_object_symbol()` function is implemented the same way,  we can also get rid of all the `if (vmlinux)` checks and probably keep only the `if (dynrela->external)` checks (though it would be great if we could also get rid of externals eventually).

joe-lawrence:
Pushed a v2 to the 617_KASLR_support branch:

- Renamed ```kpatch_verify_symbol_match``` to ```kpatch_find_object_symbol``` and fully sync'd to livepatch's ```klp_find_object_symbol``` and ```klp_find_callback```
- ```kpatch_link_object``` changed to call ```patch_find_object_symbol``` for both vmlinux and module cases.
- ```kpatch_write_relocations``` similarly modified for both cases (external dynrela's still handled as a special case though).

Tested on CentOS 7 with kernel cmdline patch, as well as fuse module patch (both pre and post module load).

kmod/core/core.c
```diff
@@ -583,29 +586,64 @@ static int kpatch_kallsyms_callback(void *data, const char *name,
...
+		*addr = args.addr;
+		return 0;
 	}
 
-	return 0;
+	*addr = 0;
+	return -EINVAL;
 }
 
 static unsigned long kpatch_find_module_symbol(struct module *mod,
```
flaming-toast: I think we can also get rid of `kpatch_find_module_symbol` and reuse `kpatch_find_object_symbol` in `kpatch_find_external_symbol`.

kmod/core/core.c
```diff
@@ -661,31 +699,26 @@ static int kpatch_write_relocations(struct kpatch_module *kpmod,
...
-	unsigned long src;
+	bool vmlinux = !strcmp(object->name, "vmlinux");
 
 	list_for_each_entry(dynrela, &object->dynrelas, list) {
-		if (!strcmp(object->name, "vmlinux")) {
-			ret = kpatch_verify_symbol_match(dynrela->name,
-							 dynrela->src);
-			if (ret)
-				return ret;
+		if (vmlinux || !dynrela->external) {
```
flaming-toast: If I'm not mistaken, I think we can simplify this even further to
```
if (dynrela->external)
   /* call kpatch_find_external_symbol */
else
   /* call kpatch_find_object_symbol */
```
When `dynrela->external` is set, we assume the symbol is either exported or in the patch module itself (kpmod). For all other cases, it should suffice to use kpatch_find_object_symbol() (for vmlinux, or the object being patched (object->name))

kmod/core/core.c
```diff
@@ -661,31 +699,26 @@ static int kpatch_write_relocations(struct kpatch_module *kpmod,
...
+			ret = kpatch_find_object_symbol(vmlinux ? NULL : object->name,
+							dynrela->name,
+							dynrela->sympos,
+							&dynrela->src);
 		} else {
 			/* module, dynrela->src needs to be discovered */
```
flaming-toast: This comment can be removed, since it doesn't really apply anymore (we are now resolving/discovering dynrela->src in all cases, not just in the case of patching modules)

kmod/core/core.c
joe-lawrence: @flaming-toast : Sure, what's the easiest way to generate one of these external dynrela entries for testing?

kmod/core/core.c
joe-lawrence: @flaming-toast : In v2 I left the conditional ordering the same, but you're right, this arrangement would be make more readable source.

kmod/core/core.c
joe-lawrence: @flaming-toast : Good catch, will remove in next version.\

kmod/core/core.c
flaming-toast: @joe-lawrence You can patch a function in a module that uses an exported symbol (like printk) and that will generate a dynrela with `external` set to 1. You could try using the patch below to test (It should generate dynrelas with `external` set, and to test you can cat any file in cat /sys/fs/btrfs/features/* to test if the kpatch module successfully patched)

```
diff --git a/fs/btrfs/sysfs.c b/fs/btrfs/sysfs.c
index e498a47..2e4c5d6 100644
--- a/fs/btrfs/sysfs.c
+++ b/fs/btrfs/sysfs.c
@@ -96,6 +96,8 @@ static ssize_t btrfs_feature_attr_show(struct kobject *kobj,
        int val = 0;
        struct btrfs_fs_info *fs_info = to_fs_info(kobj);
        struct btrfs_feature_attr *fa = to_btrfs_feature_attr(a);
+
+       pr_info("btrfs_feature_attr_show\n");
        if (fs_info) {
                u64 features = get_features(fs_info, fa->feature_set);
                if (features & fa->feature_bit)
```

joe-lawrence: @flaming-toast : in this case, isn't ```kpatch_find_external_symbol``` going to pick them up as an exported symbol by way of ``find_symbol```?  I need to test the "it's in another .o within the patch module" case.

flaming-toast: @joe-lawrence Ah, that's what you meant. In that case, you can use the `module-call-external` patch located in the `test/integration/f22/` directory. An external dynrela will be generated when create-diff-object tries to find `kpatch_string`, but that function is defined in another object (af_netlink.o) in the patch module

joe-lawrence:
v3, the code-composting continues:
- Modify ```kpatch_find_external_symbol```'s interface** to match ```kpatch_find_object_symbol```, replace its call to ```kpatch_find_module_symbol``` with a call to ```kpatch_find_object_symbol``` for the given module name.
- With no more callers remaining, remove ```kpatch_find_module_symbol```
- Simplify the ```if (dynrela->external)``` conditional in ```kpatch_write_relocations```
- Drop the "module, dynrela->src needs to be discovered" comment from ```kpatch_write_relocations```

** Would it be more accurate to call this "modname" rather than "objname"?

flaming-toast:
> Would it be more accurate to call this "modname" rather than "objname"

It's not intuitive, but usage of "objname" is everywhere in both livepatch and in kpatch code, with a consistent meaning. We should probably keep it that way just to be consistent.

jpoimboe:
> Would it be more accurate to call this "modname" rather than "objname"
> 
> It's not intuitive, but usage of "objname" is everywhere in both livepatch and in kpatch code, with a consistent meaning. We should probably keep it that way just to be consistent.

I think we decided that a "module" implied kernel module, whereas an "object" was a superset of that: it can be a module or the kernel proper.

kmod/core/core.c
```diff
@@ -661,31 +682,22 @@ static int kpatch_write_relocations(struct kpatch_module *kpmod,
...
-			}
-
-			dynrela->src = src;
+		if (dynrela->external)
+			ret = kpatch_find_external_symbol(kpmod->mod->name,
+							  dynrela->name,
+							  dynrela->sympos,
+							  &dynrela->src);
+		else
+			ret = kpatch_find_object_symbol(vmlinux ? NULL : object->name,
```
jpoimboe: One minor nit here.  kpatch and klp have a discrepancy between how they specify objname for vmlinux.  For kpatch, it's "vmlinux".  For klp, it's NULL.  Someday we might want to change kpatch to be consistent with klp.

But until then, to reduce confusion, I think kpatch should be consistent with itself.  So here, and for the other caller of this function, I think object->name should be passed directly.  And then kpatch_find_object_symbol() and kpatch_kallsyms()callback should be prepared to deal with "vmlinux" instead of NULL.

flaming-toast: Ah, I think I was the one who originally implied we should use NULL for vmlinux in kpatch. Sorry for any confusion!

joe-lawrence: @jpoimboe : I think ```kpatch_find_object_symbol``` can just setup ```args.objname``` accordingly (NULL or objname) and then ```kpatch_kallsyms_callback``` should just work without any changes.  Depends how far down the callchain you want to go.

joe-lawrence:
v4, turn the crank:
- Push ```vmlinux object->name``` check all the way down into ```kpatch_kallsyms_callback``` and adjust call chain accordingly.

flaming-toast:
Lightly tested it, looks good from my end. :+1:

jpoimboe:
Actually, I think 'old_addr' and 'src' are still used for targeting kernels with old versions of livepatch.  So I guess we still need them for the livepatch case.  So instead of all the above, I think all that's needed is just to stop setting func 'old_addr' and dynrela 'src' in kpatch-patch-hook.c.

joe-lawrence:
v5, tested kpatch-cmdline-string.ko and kpatch-module-call-external.ko on fedora 24:
- in ```kpatch-patch-hook.c``` stop setting ```p_func->old_addr``` and ```p_dynrela->src``` as they are know dynamically set on patch load.

jpoimboe:
Looks good!

👍

PR #624, opened 2016-12-05
Update integration tests

joe-lawrence:
This patchset updates the integration test patches for Fedora 25 and adds CentOS 7.   A few highlights:

- The ```rebase_patches``` script automates much of the rebasing process.  In my experience, I only had to manually apply a few patches after running this script.
- Given multiple patch directories, I eliminated redundant Makefiles and modified a few pieces to use a common, top-level Makefile.

Note: a few of the patches aren't building for Fedora, for which I will open separate issues.


test/integration/rebase_patches
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+#
+# rebase a set of patches, assumes the kernel has already been downloaded into
+# the kpatch $CACHEDIR.Output patches go into ./${ID}_${VERSION_ID}/
+#
+# Example:
+#
+# ./rebase_patches old_dir/*.patch
```
jpoimboe: I like this.  Can you rename it to `rebase-patches` to be consistent with other kpatch program names?

joe-lawrence:
v2, update patch rebasing script:
- rename to rebase-patches
- un-echo the kernel source directory cleanup
- remove gratuitous space in patch header awk script

jpoimboe:
👍

joe-lawrence:
v3, add the corresponding "directory" long option to kpatch-test's getopt call.

joe-lawrence:
v4, put "directory" long option in the *correct* order :blush:

jpoimboe:
@joe-lawrence do the changes involve changes to the existing commits?  Or new commits?  If it's just new stuff then we can probably go ahead and merge this PR as it is, and you could open another PR for the new stuff.

joe-lawrence:
@jpoimboe - it includes more fixups for the last patch, a clean up patch (symlink the tests instead of separate copies), and a new patch to allow the user to specify a list of patches to test.  I can split the new ones off into a new branch if that's easier for review.

jpoimboe:
No problem, it's fine to keep using this PR.

joe-lawrence:
v5:
- kpatch-test, shift the ```--directory xyz/``` option (not critical, but if the directory happens to match on another getopt option, things could get weird).
- kpatch-test, a few more PATCHDIR fixups (remove ```./``` instances, set test paths correctly, etc)
- Setup a ```common_tests``` directory and symlink ```centos_7``` and ```fedora_25``` instances to these.
- Enhance kpatch-test (and the Makefile) to support testing a set of patches (instead of all *.patch).  This was helpful in verifying Josh's fixes for individual tests.

joe-lawrence:
Additional thought (perhaps for a follow-on PR)... is there interest in a fedora_upstream variant?  I think this would require plumbing down to ```kpatch-build --sourcedir``` to force the tests to use a local kernel tree.  If there is interest, I could hack on that once the dust settles on these changes.

jpoimboe:
Yeah, I think something like fedora_upstream would be useful.  Especially when we have some kind of CI which could alert of us of new issues.

jpoimboe:
Everything looks good to me, except a nit: I'm not sure about the common_tests directory and symlinks.  I guess it makes it slightly easier to modify the tests, since you only have to modify one instead of multiple.  But then I think it makes things more complicated when porting the tests to a new distro, since copying files is less error-prone than symlinks.  I also like having the real test files next to the real patch files.  Anyway I don't feel strongly about it.

joe-lawrence:
Fair enough, how about I take that change out of the PR and if we later feel that redundant test script copies are becoming a nuisance, do the symlinking then.

jpoimboe:
Sounds good.

joe-lawrence:
v6:
- pulled the .test symlink change from the set
- removed redundant "--help" in kpatch-test usage introduced in v5

jpoimboe:
👍

joe-lawrence:
v7:
- Since I added Ubuntu patch build issues #635 and #634, I figured I should add the Ubuntu rebased patches to the lot.

jpoimboe:
In keeping with our convention of using a dash instead of an underline for a separator, can you rename the distro directories accordingly?

joe-lawrence:
I think I got them all...  I retested ```rebase-patches``` and ```make -C test/integration quick``` on each OS to verify that they're all using ```${ID}-${VERSION_ID}/``` form.  A commit msg updated, too.  Thank god for git rebase.

jpoimboe:
👍

PR #631, opened 2016-12-06
kpatch-build: fix tree-wide rebuild

### LINKED ISSUE #626: testing: meminfo-cmdline-rebuild-SLOW.patch fails to build on Fedora 25
As seen when running #624 :

```
Using cache at /home/vagrant/.kpatch/src
Testing patch file
Reading special section data
Building original kernel
Building patched kernel
ERROR: invalid ancestor arch/x86/kernel/ebda.o for arch/x86/kernel/ebda.o. Check /home/vagrant/.kpatch/build.log for more details.
kpatch-build.out (END)
```

**OS:** Fedora 25

**kernel version**: 4.8.6-300.fc25.x86_64
**gcc version**: gcc (GCC) 6.2.1 20160916 (Red Hat 6.2.1-2)

jpoimboe:
For newer kernels, some new objects have been added to the 'head-y'
build target.  These objects aren't directly traceable to vmlinux so
they have to be added manually.

Fixes #626.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -199,6 +199,8 @@ find_kobj() {
 				*/built-in.o|\
 				arch/x86/lib/lib.a|\
 				arch/x86/kernel/head*.o|\
+				arch/x86/kernel/ebda.o|\
+				arch/x86/kernel/platform-quirks.o|\
 				lib/lib.a)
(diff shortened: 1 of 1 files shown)


joe-lawrence:
I thought I tested this the other day, but apparently not... the fix works for me in my F25 setup.
:+1:

PR #632, opened 2016-12-07
kpatch-build: fix tree-wide rebuild on RHEL 7

### LINKED ISSUE #625: testing: meminfo-cmdline-rebuild-SLOW.patch fails to build on CentOS 7
As seen when running #624 :

**OS:** CentOS 7
**kernel version**: 3.10.0-327.36.3.el7.x86_64
**gcc version**: gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-4)

```
Using cache at /home/vagrant/.kpatch/src
Testing patch file
Reading special section data
Building original kernel
Building patched kernel
/usr/local/libexec/kpatch/create-diff-object: ERROR: copy_user_64.o: fixup_group_size: 1423: can't find .fixup rela group at offset 0

cmdline.o: changed function: cmdline_proc_show

ERROR: 1 error(s) encountered. Check /home/vagrant/.kpatch/build.log for more details.
```

jpoimboe:
On RHEL 7 based kernels, copy_user_64.o misuses the .fixup section by
placing a normal function in it.  That confuses create-diff-object.

Work around it by just skipping the file altogether, which is fine to do
because it's an assembly file which should never change anyway.

Fixes #625.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -541,6 +541,12 @@ declare -a objnames
 CHANGED=0
 ERROR=0
 for i in $FILES; do
+	# In RHEL 7 based kernels, copy_user_64.o misuses the .fixup section,
+	# which confuses create-diff-object.  It's fine to skip it, it's an
+	# assembly file anyway.
+	[[ $DISTRO = rhel ]] || [[ $DISTRO = centos ]] || [[ $DISTRO = ol ]] && \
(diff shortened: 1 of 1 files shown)


joe-lawrence:
With this change on CentOS 7: 
Building patch module: kpatch-meminfo-cmdline-rebuild-SLOW.ko
SUCCESS
:+1:

PR #637, opened 2016-12-14
kpatch and patch module builds fail on Ubuntu 16.04 #636

### LINKED ISSUE #636: kpatch and patch module builds fail on Ubuntu 16.04
**OS**: Ubuntu 16.04.1
**kernel version**:  4.4.0-53-generic
**gcc version**: gcc (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609

For kpatch:
```
make[3]: Entering directory '/usr/src/linux-headers-4.4.0-53-generic'
  CC [M]  /home/vagrant/kpatch/kmod/core/core.o
/home/vagrant/kpatch/kmod/core/core.c: In function ‘kpatch_write_relocations’:
/home/vagrant/kpatch/kmod/core/core.c:683:54: error: ‘struct module’ has no member named ‘module_core’
        unsigned long core = (unsigned long)kpmod->mod->module_core;
                                                      ^
/home/vagrant/kpatch/kmod/core/core.c:684:44: error: ‘struct module’ has no member named ‘core_size’
        unsigned long core_size = kpmod->mod->core_size;
                                            ^
/home/vagrant/kpatch/kmod/core/core.c:752:43: error: ‘struct module’ has no member named ‘core_ro_size’
                if (loc < core + kpmod->mod->core_ro_size)
                                           ^
scripts/Makefile.build:258: recipe for target '/home/vagrant/kpatch/kmod/core/core.o' failed
```
The build fails because Ubuntu backported upstream [7523e4dc5057e](https://github.com/torvalds/linux/commit/7523e4dc5057e) "module: use a structure to encapsulate layout".  The kpatch code currently wraps the code in question around ```#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 5, 0)``` blocks, however the distribution's kernel numbering defeats this check.

For patch modules:
```
/home/vagrant/.kpatch/tmp/patch/livepatch-patch-hook.c:261:10: error: ‘struct klp_reloc’ has no member named ‘val’
    lreloc->val = reloc->kdynrela->src;
```
The klp_reloc structure was updated in [064c89df6247](https://github.com/torvalds/linux/commit/064c89df6247) "livepatch: add sympos as disambiguator field to klp_reloc" which was also introduced in upstream v4.5.  Kpatch source's ```#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 5, 0)``` guards fail to protect when building against Ubuntu's 4.4.0-53-generic kernel.

arges:
These patches address Ubuntu specific patches that change livepatch structures in a way that breaks checking for kernel versions strictly on 4.X

FIxes #636

### DIFF
--- a/kmod/core/core.c
+++ b/kmod/core/core.c
@@ -46,7 +46,12 @@
 #include <linux/string.h>
 #include <asm/stacktrace.h>
 #include <asm/cacheflush.h>
+#include <generated/utsrelease.h>
 #include "kpatch.h"
+
+#ifndef UTS_UBUNTU_RELEASE_ABI
+#define UTS_UBUNTU_RELEASE_ABI 0
+#endif
 
 #if !defined(CONFIG_FUNCTION_TRACER) || \
 	!defined(CONFIG_HAVE_FENTRY) || \
@@ -676,7 +681,10 @@ static int kpatch_write_relocations(struct kpatch_module *kpmod,
 	int ret, size, readonly = 0, numpages;
 	struct kpatch_dynrela *dynrela;
 	u64 loc, val;
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 5, 0)
+#if (( LINUX_VERSION_CODE >= KERNEL_VERSION(4, 5, 0) ) || \
+     ( LINUX_VERSION_CODE >= KERNEL_VERSION(4, 4, 0) && \
+      UTS_UBUNTU_RELEASE_ABI >= 7 ) \
+    )
        unsigned long core = (unsigned long)kpmod->mod->core_layout.base;
        unsigned long core_size = kpmod->mod->core_layout.size;
 #else
@@ -746,7 +754,10 @@ static int kpatch_write_relocations(struct kpatch_module *kpmod,
 		}
 
 #ifdef CONFIG_DEBUG_SET_MODULE_RONX
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 5, 0)
+#if (( LINUX_VERSION_CODE >= KERNEL_VERSION(4, 5, 0) ) || \
+     ( LINUX_VERSION_CODE >= KERNEL_VERSION(4, 4, 0) && \
+      UTS_UBUNTU_RELEASE_ABI >= 7 ) \
+    )
                if (loc < core + kpmod->mod->core_layout.ro_size)
 #else
                if (loc < core + kpmod->mod->core_ro_size)
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -602,6 +602,14 @@ echo "Building patch module: kpatch-$PATCHNAME.ko"
 cp "$OBJDIR/.config" "$SRCDIR"
 cd "$SRCDIR"
 make prepare >> "$LOGFILE" 2>&1 || die
+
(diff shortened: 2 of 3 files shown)


kmod/core/core.c
```diff
@@ -48,6 +48,10 @@
 #include <asm/cacheflush.h>
 #include "kpatch.h"
 
+#ifdef DISTRO_UBUNTU
+#include <generated/utsrelease.h>
+#endif
```
jpoimboe: I think DISTRO_UBUNTU isn't needed because generated/utsrelease.h also exists for upstream kernels so it can be included unconditionally.

kmod/core/core.c
```diff
@@ -676,7 +680,10 @@ static int kpatch_write_relocations(struct kpatch_module *kpmod,
 	int ret, size, readonly = 0, numpages;
 	struct kpatch_dynrela *dynrela;
 	u64 loc, val;
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 5, 0)
+#if (( LINUX_VERSION_CODE >= KERNEL_VERSION(4, 5, 0) ) || \
+     ( LINUX_VERSION_CODE >= KERNEL_VERSION(4, 4, 0) && \
+      UTS_UBUNTU_RELEASE_ABI >= 7 ) \
```
jpoimboe: I think this will break for non-Ubuntu kernels between 4.4 and 4.5 because UTS_UBUNTU_RELEASE_ABI isn't defined.  Maybe do something at the top like:
```
#ifndef UTS_UBUNTU_RELEASE_ABI
#define UTS_UBUNTU_RELEASE_ABI 0
#endif
```

arges:
Updated based on @jpoimboe's comments.

joe-lawrence:
:+1:  builds for me on CentOS7+Fedora25+Ubuntu16

Thanks, @arges !

kpatch-build/kpatch-build
```diff
@@ -602,6 +602,13 @@ echo "Building patch module: kpatch-$PATCHNAME.ko"
 cp "$OBJDIR/.config" "$SRCDIR"
 cd "$SRCDIR"
 make prepare >> "$LOGFILE" 2>&1 || die
+
+if [[ $DISTRO == ubuntu ]]; then
+  # UBUNTU: add UTS_UBUNTU_RELEASE_ABI to utsrelease.h after regenerating it
+  UBUNTU_ABI=$(echo "$ARCHVERSION" | awk '{split($1, a, "-"); print a[2]}')
```
jpoimboe: The shelling out can be avoided here by using bash parameter expansion.  I think `UBUNTU_ABI=${ARCHVERSION#*-}` would work.  Also please use tabs instead of spaces.

arges: We'd need to do it twice since we want 'A' from X.Y.Z-A-flavor.
So either the one liner to awk, or:
```
UBUNTU_ABI=${ARCHVERSION#*-}
UBUNTU_ABI=${UBUNTU_ABI%-*}
```

jpoimboe: Ah, I see.  I think doing it in two lines is fine and still would be better than having to spawn processes.

arges: Fixed.

jpoimboe:
Looks good to me, thanks @arges!

:+1:

PR #638, opened 2016-12-14
kmod/core: fix activeness safety checks for kernels >= 4.6

jpoimboe:
If an activeness safety check fails for kernels newer than 4.6, the
error is silently ignored because the newer version of
kpatch_backtrace_address_verify() doesn't set args.ret on error.

It would be an easy fix to just set args->ret on error, but I think a
better approach is just to combine the two versions of the function into
a single function with the use of a little macro trickery.

### DIFF
--- a/kmod/core/core.c
+++ b/kmod/core/core.c
@@ -199,8 +199,15 @@ static inline int kpatch_compare_addresses(unsigned long stack_addr,
 }
 
 #if LINUX_VERSION_CODE < KERNEL_VERSION(4, 6, 0)
-static void kpatch_backtrace_address_verify(void *data, unsigned long address,
-					    int reliable)
+#define BACKTRACE_ADDRESS_VERIFY_RETURN_TYPE void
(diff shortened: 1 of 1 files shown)


kmod/core/core.c
```diff
@@ -199,16 +199,23 @@ static inline int kpatch_compare_addresses(unsigned long stack_addr,
...
 #if LINUX_VERSION_CODE < KERNEL_VERSION(4, 6, 0)
-static void kpatch_backtrace_address_verify(void *data, unsigned long address,
-					    int reliable)
+#define BACKTRACE_ADDRESS_VERIFY_RETURN_TYPE void
+#define BACKTRACE_ADDRESS_VERIFY_RETURN_ARG
+#else
+#define BACKTRACE_ADDRESS_VERIFY_RETURN_TYPE int
+#define BACKTRACE_ADDRESS_VERIFY_RETURN_ARG args->ret
+#endif
+
```
flaming-toast: What an interesting trick :-)

PR #639, opened 2016-12-14
kpatch: retry kpatch load on failure

jpoimboe:
If a kpatch load fails, try again, up to 5 times.

It will usually only fail if the activeness safety check failed, in
which case trying again might work.  Unfortunately, insmod doesn't
return the error reported by the module, so we can't distinguish
activeness safety errors from other types of errors.

### DIFF
--- a/kpatch/kpatch
+++ b/kpatch/kpatch
@@ -165,7 +165,27 @@ load_module () {
 	fi
 
 	echo "loading patch module: $module"
-	insmod "$module"
+	local i=0
+	while true; do
+		out=$(insmod "$module" 2>&1)
+		[[ -z $out ]] && break
+		echo $out 1>&2
+		[[ ! $out =~ "Device or resource busy" ]] &&
+			die "failed to load module $module"
+
+		# "Device or resource busy" means the activeness safety check
+		# failed.  Retry in a few seconds.
+		i=$((i+1))
+		if [[ $i = 5 ]]; then
+			die "failed to load module $module"
+			break
+		else
+			warn "retrying..."
+			sleep 2
+		fi
+	done
+
+	return 0
 }
 
 unload_module () {
--- a/kpatch/kpatch
+++ b/kpatch/kpatch
@@ -136,6 +136,8 @@ verify_module_checksum () {
 }
 
 load_module () {
+	local module="$1"
+
 	if ! core_module_loaded; then
 		if modprobe -q kpatch; then
 			echo "loaded core module"
@@ -146,11 +148,11 @@ load_module () {
 		fi
 	fi
 
-	modname=$(get_module_name $1)
-	moddir=/sys/kernel/kpatch/patches/$modname
+	local modname=$(get_module_name $module)
+	local moddir=/sys/kernel/kpatch/patches/$modname
 	if [[ -d $moddir ]] ; then
 		if [[ $(cat "${moddir}/enabled") -eq 0 ]]; then
-			if verify_module_checksum $1; then # same checksum
+			if verify_module_checksum $module; then # same checksum
 				echo "module already loaded, re-enabling"
 				echo 1 > ${moddir}/enabled || die "failed to re-enable module $modname"
 				return
@@ -162,8 +164,8 @@ load_module () {
 		fi
 	fi
 
-	echo "loading patch module: $1"
-	insmod "$1" "$2"
+	echo "loading patch module: $module"
+	insmod "$module"
 }
 
 unload_module () {


kpatch/kpatch
```diff
@@ -163,7 +163,16 @@ load_module () {
...
 	echo "loading patch module: $1"
-	insmod "$1" "$2"
+	for i in `seq 5`; do
+		if [[ $i != 1 ]]; then
+			echo "retrying..."
+			sleep 2
+		fi
+		insmod "$1" "$2" && break
+	done
+	[[ $i = 5 ]] && die "failed to load module $modname"
```
flaming-toast: If `insmod` happens to succeed on the 5th try (i = 5), then we will `die` here. Maybe set a fail flag within the loop (at i=5, if insmod fails, then set fail=1...) and then check it outside of the for loop?

kpatch/kpatch
```diff
@@ -163,7 +163,16 @@ load_module () {
 	fi
 
 	echo "loading patch module: $1"
-	insmod "$1" "$2"
+	for i in `seq 5`; do
+		if [[ $i != 1 ]]; then
+			echo "retrying..."
+			sleep 2
+		fi
+		insmod "$1" "$2" && break
```
flaming-toast: Do we still need to call `insmod` with `$2`? Looks like nothing in the script sets it anymore. It used to be used to pass the `replace` param, but IIRC the `kpatch replace` option was deprecated in commit 8e8de4718d8

kpatch/kpatch
joe-lawrence: Alternatively move the conditional check at the bottom of the loop.  Any successful ```insmod``` will break the loop and skip the check...  Or do something like: ```insmod "$1" "$2" && loaded=1 && break``` to employ a success flag without having duplicate "5" constants laying about.

flaming-toast: @joe-lawrence Ah, I like your idea better than mine :-) (ditching the conditional and moving the sleep and echo after the insmod)

kpatch/kpatch
jpoimboe: Oops, yeah.  This is some ugly code.  These functions should really be using local variables to deal with the function arguments.

kpatch/kpatch
jpoimboe: Will fix.

jpoimboe:
Updated.

- rewrote loop based on all the comments
- cleaned up load_module() to use local variables
- removed unused 2nd argument to insmod
- added specific checking for insmod error, retry only makes sense for EBUSY.

kpatch/kpatch
```diff
@@ -165,7 +165,27 @@ load_module () {
 	fi
 
 	echo "loading patch module: $module"
-	insmod "$module"
+	local i=0
+	while true; do
+		out=$(insmod "$module" 2>&1)
+		[[ -z $out ]] && break
+		echo $out 1>&2
+		[[ ! $out =~ "Device or resource busy" ]] &&
```
joe-lawrence: I have no i18n experience, but is it assured that insmod output will be in English?  (Too bad it doesn't propagate out the error code as insmod exit status.)

kpatch/kpatch
```diff
@@ -165,7 +165,27 @@ load_module () {
...
+			die "failed to load module $module"
+
+		# "Device or resource busy" means the activeness safety check
+		# failed.  Retry in a few seconds.
+		i=$((i+1))
+		if [[ $i = 5 ]]; then
```
joe-lawrence: I always get tripped up with shell scripts... is ```=``` legit here, or do you need ```-eq``` for numerical test?

kpatch/kpatch
flaming-toast: @joe-lawrence That's an interesting question. I looked at the insmod sources from kmod and old module-init-tools sources (superceded by kmod) and they both have always used `strerror` to print the message, *not* `strerror_l`,  which would print the locale-dependent/translated version of the error message. So I think the message will always be in English.

(For the curious, the translated glibc error messages used by `strerror_l` and others can be found here: https://sourceware.org/git/?p=glibc.git;a=tree;f=po;h=b6fb91b5b5b9730f24a0960bf0b0db766fc54c17;hb=HEAD)

kpatch/kpatch
jpoimboe: `=` is legit because all variables are really strings in bash.  So it's testing for the string "5".  `-eq` would have worked too.

joe-lawrence: Okay, like I said, it just gets confusing, especially with ```-gt``` vs ```>```.  I did test this in a shell and you're right, the string comparison will match.

PR #641, opened 2016-12-16
Fix static code analyzer complaints

joe-lawrence:
These are mostly trivial, but really simple to clean up.  With a clean slate, we can monitor future pull requests that may introduce new complaints  (I have a script in the works).

### DIFF
--- a/kmod/core/core.c
+++ b/kmod/core/core.c
@@ -66,7 +66,7 @@ static DEFINE_HASHTABLE(kpatch_func_hash, KPATCH_HASH_BITS);
 
 static DEFINE_SEMAPHORE(kpatch_mutex);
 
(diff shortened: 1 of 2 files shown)


jpoimboe:
@joe-lawrence nice!  BTW, I'm surprised it didn't complain about all the non-static functions in create-diff-object.c.

:+1:

jpoimboe:
Also surprised it didn't find more bugs!

joe-lawrence:
Ahh, I was only running the checks on the driver directories...  Let me look into checking the entire project.

flaming-toast:
I'll go ahead and merge this now since it has a 👍 :-) If you find more smatch/sparse complaints, feel free to open another pull request.

PR #642, opened 2016-12-19
kpatch-build: make patches for debug kernels too

iredko:
GCC with KASAN instrumentation creates section ".rodata" with some static strings (i.e. some of them go to ".rodata.str1.1" for release build).
This change makes possible to build patch and check if it fixes issue found with KASAN, such as CVE-2016-9555.


jpoimboe:
👍

Issue #629, opened 2016-12-05
testing: create-diff-object segfaults when building replace-section-references.patch on Fedora 25


joe-lawrence:
As seen when running #624 :

**OS:** Fedora 25

**kernel version**: 4.8.6-300.fc25.x86_64
**gcc version**: gcc (GCC) 6.2.1 20160916 (Red Hat 6.2.1-2)

```
Core was generated by `/usr/local/libexec/kpatch/create-diff-object orig/arch/x86/kvm/x86.o patched/ar'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000000000407fd7 in kpatch_rebuild_rela_section_data (sec=0xa980f0) at create-diff-object.c:2280
2280                            relas[index].r_info = GELF_R_INFO(rela->sym->index, rela->type);
Missing separate debuginfos, use: dnf debuginfo-install elfutils-libelf-0.167-2.fc25.x86_64 glibc-2.24-3.fc25.x86_64 zlib-1.2.8-10.fc24.x86_64
(gdb) p index
$1 = 0
(gdb) p *rela
$2 = {list = {next = 0xa98280, prev = 0xa98178}, rela = {r_offset = 0, r_info = 0, r_addend = 0}, sym = 0x0, type = 1, addend = 4, offset = 0, string = 0x0}
                                                                                                  ^^^^^^^^^
```

joe-lawrence:
A little bit more information from another generated core file:
```
  (gdb) p rela
  $1 = (struct rela *) 0x12f8240
  (gdb) p *rela
  $2 = {list = {next = 0x12f8290, prev = 0x12f8188}, rela = {r_offset = 0,
        r_info = 0, r_addend = 0}, sym = 0x0, type = 1, addend = 4,
        offset = 0, string = 0x0}

  (gdb) p sec
  $3 = (struct section *) 0x12f8100
  (gdb) p *sec
  $4 = {list = {next = 0x12f8c40, prev = 0x12f7e80}, twin = 0x0,
        sh = {sh_name = 0, sh_type = 4, sh_flags = 0, sh_addr = 0,
        sh_offset = 0, sh_size = 432, sh_link = 22, sh_info = 27,
        sh_addralign = 8, sh_entsize = 24}, data = 0x12f7ff0,
        name = 0x12f7db0 ".rela.kpatch.dynrelas", index = 28, status = NEW,
        include = 0, ignore = 0, grouped = 0, {{base = 0x12f7e80,
[...]

PR #644, opened 2016-12-19
create-diff-object: print error message instead of seg faulting

jpoimboe:
Related to issue #629, print a useful error message at the time of the
original error condition instead of seg faulting later.


joe-lawrence:
:+1:

PR #645, opened 2016-12-19
kpatch-build: reduce the debug firehose

jpoimboe:
When passing '-d' to kpatch-build, it prints out some useful information
and keeps the related files around in ~/.kpatch/tmp.  However, it also
passes '-d' to create-diff-object, which spits out way too much
information, drowning out all the other useful messages printed by
kpatch-build.

In my experience, the create-diff-object debug info is overkill for
debugging most issues, so disable it.  The flag can still be used when
running create-diff-object manually.


joe-lawrence:
:+1:  Makes sense to me.  When I was debugging one of the integration test patches, the debug fire house output to xterm would lockup my desktop during ```create-diff-object```.

PR #646, opened 2016-12-20
Sparse smatch round 2

joe-lawrence:
I let the smatch and sparse static code analysers loose on the entire product tree and sparse came up with two new complaints.  These are pretty trivial, so my heart won't be broken if we drop these over the code churn.  After these two changes, the output is relatively* clean as far as I can tell.

\* With a few caveats:
1 - sparse gets confused about ```insn/asm/insn.h :: struct insn :: unsigned char x86_64``` ... if I rename all references to the ```x86_64``` member to ```foo_x86_64```, it's happy.  So it must think that "x86_64" is defined as something else?  This is more than false-negative chatter as the parser gives up on .c files which include insn.h as it counts too many errors.

2 - I had to add ```error``` to smatch's ```no_return_funcs``` file, else it would complain about possible NULL dereferences after a call via the ```kpatch-build/log.h :: ERROR``` macro.  This is a documented step for non-kernel projects.

3 - Red Hat's kABI macros still confuse smatch.

That said, I think if we can eliminate caveat 1 (or plow through the false positives) then we can at least run the smatch ```new_bugs.sh``` script to detect the introduction of new complaints.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -143,7 +143,7 @@ static int kpatch_mangled_strcmp(char *s1, char *s2)
 	return 1;
 }
 
-int rela_equal(struct rela *rela1, struct rela *rela2)
+static int rela_equal(struct rela *rela1, struct rela *rela2)
 {
 	if (rela1->type != rela2->type ||
 	    rela1->offset != rela2->offset)
@@ -162,7 +162,7 @@ int rela_equal(struct rela *rela1, struct rela *rela2)
 	return !strcmp(rela1->sym->name, rela2->sym->name);
 }
 
-void kpatch_compare_correlated_rela_section(struct section *sec)
+static void kpatch_compare_correlated_rela_section(struct section *sec)
 {
 	struct rela *rela1, *rela2 = NULL;
 
@@ -179,7 +179,7 @@ void kpatch_compare_correlated_rela_section(struct section *sec)
 	sec->status = SAME;
 }
 
-void kpatch_compare_correlated_nonrela_section(struct section *sec)
+static void kpatch_compare_correlated_nonrela_section(struct section *sec)
 {
 	struct section *sec1 = sec, *sec2 = sec->twin;
 
@@ -190,7 +190,7 @@ void kpatch_compare_correlated_nonrela_section(struct section *sec)
 		sec->status = SAME;
 }
 
-void kpatch_compare_correlated_section(struct section *sec)
+static void kpatch_compare_correlated_section(struct section *sec)
 {
 	struct section *sec1 = sec, *sec2 = sec->twin;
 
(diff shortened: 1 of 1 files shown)


joe-lawrence:
> 1 - sparse gets confused about insn/asm/insn.h :: struct insn :: unsigned char x86_64 ... if I rename all references to the x86_64 member to foo_x86_64, it's happy. So it must think that "x86_64" is defined as something else? This is more than false-negative chatter as the parser gives up on .c files which include insn.h as it counts too many errors.

Added a ```-v``` for debugging verbosity and noticed that sparse's cgcc script calls gcc with ```-Dx86_64=1``` (along with a host of other constants) ... my guess is that's why it gets confused when it sees a variable/definition with the same name.  IMHO, ```x86_64``` seems like a reasonable name that folks would want to use, no underscores or other decoration, so I'm going to ping the sparse-list to see if they have any advice.

joe-lawrence:
> so I'm going to ping the sparse-list to see if they have any advice.

Linus replied and stated that this was bogus... perhaps (much) older gcc versions defined arch macros without preceding underscores like ```x86_64```, but no longer.  I'm going to consider this a non-issue and look forward to a sparse fix for this.

kpatch-build/insn/inat-tables.c
```diff
@@ -2,7 +2,7 @@
 /* Do not change this code. */
 
 /* Table: one byte opcode */
-const insn_attr_t inat_primary_table[INAT_OPCODE_TABLE_SIZE] = {
+static const insn_attr_t inat_primary_table[INAT_OPCODE_TABLE_SIZE] = {
```
jpoimboe: Is it possible to remove this file from your checking?  I'd rather leave it stock because it was copied from the kernel and we should try to keep it in sync without diverging.

joe-lawrence: There are a few ways to invoke the static parsers:  For the earlier (merged) patchset round I set the ```CHECK``` variable that the kernel's Makefile reads.  For this patchset, I set ```CC=~/smatch/cgcc``` to run the analyser over _all_ of the code.

As far as I can see, there is no easy way to pass to smatch/sparse a list of source files to ignore nor any inline annotation to instruct the parsers to ignore a particular source code line.

I suppose our Makefile(s) could be modified in a way to accomplish this, but that seems like overkill.  I think the best compromise would be to drop ```inat-tables.c``` from this PR... smatch comes with a ```new_bugs.sh``` script that can report just the _new_ warnings if that's what we're ultimately interested in.

We could also push the same fixes for ```inat-tables.c``` upstream and then resync.  (I think this file is auto-generated, but perhaps its owner would take these fixups.)

Let me know and I'll push out a v2.

jpoimboe: Dropping inat-tables.c from this PR sounds good and adding new_bugs.sh (now or later, if needed) is fine.

I agree that fixing inat-tables.c upstream would be the best long term approach.  (If you wanted to do that yourself, be aware that there are three copies of that code in the kernel tree which are all kept in sync, so they would need to be updated simultaneously.  I own one of those copies (objtool), my eventual goal is to share the code with perf so that there are only two copies in the tree, but it's way down on my TODO list.)

joe-lawrence:
v2 - dropped inat-tables.c from this PR

jpoimboe:
👍

PR #647, opened 2016-12-20
kpatch: better support for livepatch modules

### LINKED ISSUE #479: kpatch script compatibility with livepatch
Make the kpatch script compatible with both the kpatch core module and livepatch.

This might involve making the kpatch core module sysfs interface identical to the livepatch interface.

joe-lawrence:
Livepatch modules can be supported with minimal changes to the kpatch
script.  Adjust for appropriate sysfs paths, core-patching code (in
kernel for livepatch, kpatch.ko for kpatch), and checksum verification
(only verify the checksum if it exists).

Fixes #479.

### DIFF
--- a/kpatch/kpatch
+++ b/kpatch/kpatch
@@ -27,6 +27,14 @@ INSTALLDIR=/var/lib/kpatch
 SCRIPTDIR="$(readlink -f $(dirname $(type -p $0)))"
 VERSION="0.3.4"
 
+# Livepatch is built into the kernel, if it's not present
+# we must use kpatch core module.
+if [[ -e /sys/kernel/livepatch ]] ; then
+	SYSFS="/sys/kernel/livepatch"
+else
+	SYSFS="/sys/kernel/kpatch/patches"
+fi
+
 usage_cmd() {
 	printf '   %-20s\n      %s\n' "$1" "$2" >&2
 }
@@ -116,8 +124,8 @@ find_core_module() {
 	return 1
 }
 
-core_module_loaded () {
-	grep -q "T kpatch_register" /proc/kallsyms
+core_loaded () {
+	grep -q -e "T klp_register_patch" -e "T kpatch_register" /proc/kallsyms
 }
 
 get_module_name () {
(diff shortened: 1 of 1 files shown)


kpatch/kpatch
```diff
@@ -116,29 +124,31 @@ find_core_module() {
...
-    [[ -z $checksum ]] && return 1
+	checksum=$(readelf -p .kpatch.checksum $1 | grep '\[.*\]' | awk '{print $3}')
 
-    sysfs_checksum=$(cat /sys/kernel/kpatch/patches/${modname}/checksum)
-    [[ $checksum == $sysfs_checksum ]] || return 1
+	# Fail checksum match only if both exist and diverge
+	if [[ ! -z $checksum ]] && [[ -e "$SYSFS/${modname}/checksum" ]] ; then
+		sysfs_checksum=$(cat $SYSFS/${modname}/checksum)
+		[[ $checksum == $sysfs_checksum ]] || return 1
+	fi
```
jpoimboe: Bash functions return the return the result of the last command -- in this case, the "if; fi" construct.  It would probably be safer to explicitly return 0 at the end of the function.

kpatch/kpatch
```diff
@@ -27,6 +27,14 @@ INSTALLDIR=/var/lib/kpatch
 SCRIPTDIR="$(readlink -f $(dirname $(type -p $0)))"
 VERSION="0.3.4"
 
+# Assume livepatch core is built into the kernel, otherwise
+# use kpatch model.
```
jpoimboe: s/model/core module/ ?

joe-lawrence: Reworded in v2

kpatch/kpatch
joe-lawrence: Good catch, fixed in v2

joe-lawrence:
v2 - reworded live/kpatch comment, explicitly return 0 from ```verify_module_checksum()```

jpoimboe:
👍

flaming-toast:
Thanks @joe-lawrence!

PR #649, opened 2016-12-22
kpatch-build: Add UTS_UBUNTU_RELEASE_ABI tag for non-stock kernel

kamalesh-babulal:
commit eb55adc52d0c ("use livepatch 4.5 features in
Ubuntu Xenial kernel") will trigger following build
failure, while building stock kernel on Ubuntu:
make[2]: Entering directory '/root/.kpatch/obj'
In file included from /root/.kpatch/tmp/patch/livepatch-patch-hook.c:28:0,
                 from /root/.kpatch/tmp/patch/patch-hook.c:21:
/root/.kpatch/tmp/patch/livepatch-patch-hook.c: In functionpatch_ini:
/root/linux-4.8.15/include/generated/utsrelease.h:2:32: error: too many decimal points in number
 #define UTS_UBUNTU_RELEASE_ABI 4.8.15
                                ^
/root/.kpatch/tmp/patch/livepatch-patch-hook.c:252:7: note: in expansion of macro UTS_UBUNTU_RELEASE_ABI
       UTS_UBUNTU_RELEASE_ABI >= 7 ) \
       ^
Stock kernel version string might differ from the ubuntu
kernel versioning format. This patch sets UBUNTU_KERNEL
flag, when kpatch module is being build for ubuntu distro
kernel and check for this flag before echoing
UTS_UBUNTU_RELEASE_ABI tag.

Cc: Chris J Arges <christopherarges@gmail.com>

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -419,6 +419,7 @@ else
 			extension="bz2"
 			sublevel="SUBLEVEL = 0"
 			taroptions="xvjf"
+			UBUNTU_KERNEL=1
 
 		elif [[ $DISTRO = debian ]]; then
 
@@ -601,7 +602,7 @@ cp "$OBJDIR/.config" "$SRCDIR"
 cd "$SRCDIR"
 make prepare >> "$LOGFILE" 2>&1 || die
 
-if [[ $DISTRO == ubuntu ]]; then
+if [[ ! -z $UBUNTU_KERNEL ]]; then
 	# UBUNTU: add UTS_UBUNTU_RELEASE_ABI to utsrelease.h after regenerating it
 	UBUNTU_ABI=${ARCHVERSION#*-}
 	UBUNTU_ABI=${UBUNTU_ABI%-*}


kpatch-build/kpatch-build
```diff
@@ -602,10 +602,13 @@ cd "$SRCDIR"
 make prepare >> "$LOGFILE" 2>&1 || die
 
 if [[ $DISTRO == ubuntu ]]; then
-	# UBUNTU: add UTS_UBUNTU_RELEASE_ABI to utsrelease.h after regenerating it
-	UBUNTU_ABI=${ARCHVERSION#*-}
-	UBUNTU_ABI=${UBUNTU_ABI%-*}
-	echo "#define UTS_UBUNTU_RELEASE_ABI "$UBUNTU_ABI"" >> "$SRCDIR"/include/generated/utsrelease.h
+	#UTS_UBUNTU_RELEASE_ABI: is added only, if the kernel is not stock kernel.
+	if [[ ! -z $UBUNTU_KERNEL ]]; then
```
jpoimboe: I think the above `if [[ $DISTRO == ubuntu ]]` can be removed altogether?

jpoimboe:
@arges does this look ok?

arges:
This seems Ok, but I have not tested it.

kamalesh-babulal:
@jpoimboe Thanks for the review. How do I push v2.

jpoimboe:
@kamalesh-babulal You can force push to the same branch.

kamalesh-babulal:
> @jpoimboe Thanks. I have pushed the v2.

kpatch-build/kpatch-build
```diff
@@ -454,6 +454,7 @@ else
 		# upstream SUBLEVEL; they are always at SUBLEVEL 0
 		sed -i "s/^SUBLEVEL.*/${sublevel}/" "$SRCDIR/Makefile" || die
 		echo $ARCHVERSION > "$VERSIONFILE" || die
+		UBUNTU_KERNEL=1
```
jpoimboe: I think this will break debian.  Maybe this variable can be set in the `if [[ $DISTRO = ubuntu ]]` clause at line 415.

kamalesh-babulal:
> @jpoimboe Thanks for the review. Yes, it will break the Debian build. Pushed v3, addressing the review comment.

kpatch-build/kpatch-build
```diff
@@ -601,7 +604,8 @@ cp "$OBJDIR/.config" "$SRCDIR"
 cd "$SRCDIR"
 make prepare >> "$LOGFILE" 2>&1 || die
 
-if [[ $DISTRO == ubuntu ]]; then
+# UTS_UBUNTU_RELEASE_ABI: is added only, if the kernel is not a stock kernel.
```
jpoimboe: I actually think this comment is confusing, because I interpret "stock kernel" to mean "the kernel which comes with Ubuntu by default", whereas I think you really mean "the upstream kernel".  And actually I don't think either of the comments in this patch are necessary, so if you just remove all the comments, I'll give it a 👍 :-)

kamalesh-babulal:
> @jpoimboe Yeah, it should have been non-stock. I have pushed v4, without comments.

jpoimboe:
👍

Issue #613, opened 2016-10-20
PPC64LE support


kamalesh-babulal:
I am currently looking into getting livepatch hook to work on PPC64LE. As suggested on "Getting Involved " section, using this issue to discuss and brainstorm on idea's.

kamalesh-babulal:
Design is to introduce PPC64 specific bits to enhance kpatch, to build livepatch supported modules on PPC64 and guard kpatch specific information with #ifdef's. 

To start with make kmod module to print, about the state of support on powerpc, while kmod
module is loaded.

diff --git a/kmod/core/core.c b/kmod/core/core.c
index 77c7290..5922d55 100644
--- a/kmod/core/core.c
+++ b/kmod/core/core.c
@@ -32,9 +32,11 @@
  * return instruction pointer so that ftrace will return to the new function.
  */
 
+#include <linux/printk.h>
+#include <linux/module.h>
+#ifndef __powerpc__
 #define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
 
-#include <linux/module.h>
 #include <linux/slab.h>
 #include <linux/stop_machine.h>
 #include <linux/ftrace.h>
@@ -1199,6 +1201,14 @@ static void kpatch_exit(void)
        kobject_put(kpatch_patches_kobj);
        kobject_put(kpatch_root_kobj);
 }
+#else
+static int kpatch_init(void)
+{
+       printk(KERN_INFO "kpatch on PPC only supports Livepatching infrastructure");
+       return 0;
+}
+static void kpatch_exit(void) { }
+#endif
 
 module_init(kpatch_init);
 module_exit(kpatch_exit);
 
2. Modify kpatch-build/kpatch-build to address or capture special section required for PPC64LE v2 ELF.
3. Introduce code to read the _mcount location for PPC64LE, which are little tricky.
4. Modify kpatch/kmod/patch/livepatch-patch-hook.c to guard kpatch specific details with #ifdefery.

Can let me your thoughts on it.

jpoimboe:
I think the hardest part of this, by far, would be porting kpatch-build to work with ppc64le.  Before spending much time on that, it would probably be best to wait a few weeks, because we're going to be discussing the future of the livepatch patch creation tooling at Linux Plumber's Conference on Nov 2.

kamalesh-babulal:
[...]

PR #650, opened 2016-12-22
[RFC] ppc64le livepatch hook support

kamalesh-babulal:
This patchset gets the livepatch hook based kernel patching working on ppc64le.
kpatch build script can create modules and successfully patch the kernel for
the test case's posted at https://lkml.org/lkml/2016/3/24/190. Which includes
the in kernel sample test case samples/livepatch/livepatch-sample.c. 

First, four patch(s) are pre-checks and helpers functions. Those attempts to
unify both x86 and ppc64le into common functions and introduce ppc64le
based checks. Patch 'Add PPC64le livepatch support' handles ppc64le specific
ELF details.

Last patch address marking livepatch module as livepatch from
kernel commit 425595a7fc20 ("livepatch: reuse module loader code to write
relocations"). ppc64le requires more enhancements, to cover all of the cases
currently supported.


jpoimboe:
@kamalesh-babulal This is awesome!  I'll be out for most of the rest of the year but will give it a proper review in 2017.  Thanks!

flaming-toast:
Hurray for more arch support! Thanks @kamalesh-babulal for working on this!

joe-lawrence:
@kamalesh-babulal - are there corresponding kernel patches?  I get this build error on both CentOS 7 and Fedora 25 ppc64le: ```/root/kpatch/kmod/core/core.c:47:28: fatal error: asm/stacktrace.h: No such file or directory```

flaming-toast:
@joe-lawrence I haven't looked at the patches too closely yet, but I think this is only for livepatch patch modules. asm/stacktrace.h only exists for some arches and for kpatch, the core module is only supported on x86, as we use the x86-specific stacktrace_ops struct and dump_trace function. (And note that the core module would also no longer work with current upstream, see Josh's x86 unwinder patchset: https://marc.info/?l=linux-kernel&m=147405352424700)

joe-lawrence:
@flaming-toast - Ahh, right.  Will try again with livepatch patch modules.. sorry for the noise.

kamalesh-babulal:
@joe-lawrence kernel support for ppc64le livepatch was merged into v4.7. First, the step towards supporting kpatch on the ppc64le arch is to ensure it works with the livepatch kernel infrastructure (livepatch hook). For the build failure your seeing, I use a workaround mentioned in https://github.com/dynup/kpatch/issues/613.
and  building the kernel module, the command line used it
../kpatch/kpatch-build/kpatch-build -v vmlinux -c ./.config -s . --skip-gcc-check ../proc-cmdline.patch -d 
, then insmod <patched.ko> to test the module.

@jpoimboe @flaming-toast @joe-lawrence Thank you. I will also be out for rest of the year and first half of January 2017. Merry Christmas and happy new year.

kpatch-build/create-diff-object.c
```diff
@@ -2109,6 +2148,9 @@ void kpatch_create_hooks_objname_rela(struct kpatch_elf *kelf, char *objname)
 	}
 }
 
+#ifdef __powerpc__
+void kpatch_create_mcount_sections(struct kpatch_elf *kelf) { }
```
jpoimboe: Isn't this needed for powerpc? Otherwise, I think patched functions won't be traceable by ftrace.

kpatch-build/kpatch-build
```diff
@@ -36,6 +36,7 @@
 
 BASE="$PWD"
 SCRIPTDIR="$(readlink -f $(dirname $(type -p $0)))"
+ARCH=$"$(uname -m)"
```
jpoimboe: Looks like a superfluous '$'?

kpatch-build/kpatch-build
```diff
@@ -140,12 +153,54 @@ gcc_version_check() {
...
 
-	# ensure gcc version is >= 4.8
-	gccver=$(echo $gccver |cut -d'.' -f1,2)
-	if [[ $gccver < 4.8 ]]; then
-		warn "gcc >= 4.8 required"
-		return 1
+	if [[ $ARCH = "ppc64le" ]]; then
+		# On ppc64le livepatch requires -mprofile-kernel support.
+		# Adopt <linux-sources>/arch/powerpc/scripts/gcc-check-mprofile-kernel.sh
+		# for gcc support checks.
```
jpoimboe: Maybe this comment should be moved to the header of the gcc_mprofile_test() function since it describes its implementation.

kamalesh-babulal:
> @jpoimboe Thanks for the reviews. I will work on them and push v2 sooner.

kpatch-build/create-diff-object.c
```diff
@@ -1034,7 +1041,12 @@ void kpatch_verify_patchability(struct kpatch_elf *kelf)
 		/*
 		 * ensure we aren't including .data.* or .bss.*
 		 * (.data.unlikely is ok b/c it only has __warned vars)
+		 * and .toc sections for powerpc.
 		 */
+#ifdef __powerpc__
+		if (!strncmp(sec->name, ".toc", 4) && sec->status == CHANGED)
+			continue;
```
kamalesh-babulal: Sorry, It's fallout from old code. Will remove it.

kpatch-build/create-diff-object.c
```diff
@@ -1864,7 +1879,11 @@ void kpatch_create_patches_sections(struct kpatch_elf *kelf,
 			 */
 			ALLOC_LINK(rela, &relasec->relas);
 			rela->sym = sym;
+#ifdef __powerpc__
+			rela->type = R_PPC64_ADDR64;
+#else
 			rela->type = R_X86_64_64;
+#endif
```
kamalesh-babulal: Looks neat, will use this instead of #ifdefs.

kpatch-build/create-diff-object.c
kamalesh-babulal: - load_module
  - ftrace_module_init
     - ftrace_process_locs
        - ftrace_update_code
           - ftrace_code_disable
              - ftrace_make_nop
                this function expects the opcode, to be valid bl (branch op). Converting it to nop manually in the patched module will result it crash. As it perform the check for valid branch using (op & 0xfc000003) == 0x48000001

kpatch-build/kpatch-build
kamalesh-babulal: fixed in the v2.

kpatch-build/kpatch-build
kamalesh-babulal: fixed in v2.

kpatch-build/create-diff-object.c
jpoimboe: Actually `kpatch_create_mcount_sections()` converts it *from* a nop to a branch to fentry.  I'm assuming this function isn't needed because the instruction is already a branch on powerpc.

jpoimboe:
Tried to run kpatch-build on a ppc64le box and got an error:
```
./kpatch-build -s ~/linux ../test/integration/fedora-25/meminfo-string.patch -t vmlinux
Testing patch file
Reading special section data
ERROR: can't find special struct size.
```
So if special section support doesn't work yet for powerpc, that part of kpatch-build needs to be disabled.

jpoimboe:
Never mind about the last comment, I used the wrong branch...

kamalesh-babulal:
v2 changes:
- Modified Makefile to build core module only for x86_64 (Josh Poimboeuf)
- Build create-diff-object with insn support on x86_64. 
- Addressed the review comments.

kpatch-build/create-diff-object.c
```diff
@@ -872,6 +886,9 @@ static void rela_insn(struct section *sec, struct rela *rela, struct insn *insn)
 			return;
 	}
 }
+#else
+void rela_insn(struct section *sec, struct rela *rela, struct insn *insn) { }
+#endif
```
jpoimboe: I think this empty prototype isn't needed since it's only called from x86-specific code.

kamalesh-babulal: build will fail without the #ifdef guard:
```
gcc -MMD -MP -I../kmod/patch -Iinsn -Wall -g -Werror -c -o create-diff-object.o create-diff-object.c
create-diff-object.c:871:13: error: ‘rela_insn’ defined but not used [-Werror=unused-function]
 static void rela_insn(struct section *sec, struct rela *rela, struct insn *insn)
             ^
```

jpoimboe: Yes, I think the '#ifdef' is still needed, but not the '#else'.

kamalesh-babulal: Yes, Fixed in v3.

jpoimboe:
👍

jpoimboe:
@kamalesh-babulal The code looks good to me.  One question, how did you test it?

I attempted to test it with along with a manual merge of #607 on Linux 4.9. However, I had a problem.

```
index 8a42849..d8cf031 100644
--- a/fs/proc/meminfo.c
+++ b/fs/proc/meminfo.c
@@ -132,7 +132,7 @@ static int meminfo_proc_show(struct seq_file *m, void *v)
 	seq_printf(m, "VmallocTotal:   %8lu kB\n",
 		   (unsigned long)VMALLOC_TOTAL >> 10);
 	show_val_kb(m, "VmallocUsed:    ", 0ul);
-	show_val_kb(m, "VmallocChunk:   ", 0ul);
+	show_val_kb(m, "VMALLOCCHUNK:   ", 0ul);
 
 #ifdef CONFIG_MEMORY_FAILURE
 	seq_printf(m, "HardwareCorrupted: %5lu kB\n",
```

The module built ok, but the load failed with:
```
[   71.407000] module_64: kpatch_meminfo: REL24 -1152921504897399800 out of range!
```
Not sure if the bug is in kpatch-build or in upstream livepatch.  But I notice the patch module has several R_PPC64_TOC16_HA and R_PPC64_TOC16_LO entries.  Does kpatch-build need to process those somehow?

flaming-toast:
So this issue is more complicated than I thought, and I hope someone who is more familiar with ppc64le + TOCs can chime in about this (as I have zero experience with ppc64le...). I think the problem has something to do with incorrect TOC manipulation.

The error
```
   module_64: kpatch_meminfo: REL24 -1152921504897399800 out of range!
```
comes from `apply_relocate_add()` from arch/powerpc/kernel/module_64.c:

```
              case R_PPC_REL24:
                         /* FIXME: Handle weak symbols here --RR */
                         if (sym->st_shndx == SHN_UNDEF) {
                                 /* External: go via stub */
                                 value = stub_for_addr(sechdrs, value, me);
                                 if (!value)
                                         return -ENOENT;
                                 if (!restore_r2((u32 *)location + 1, me))
                                         return -ENOEXEC;

                                 squash_toc_save_inst(strtab + sym->st_name, value);
                         } else
                                 value += local_entry_offset(sym);

                         /* Convert value to relative */
                         value -= (unsigned long)location;
                         if (value + 0x2000000 > 0x3ffffff || (value & 3) != 0){
                                 pr_err("%s: REL24 %li out of range!\n",
                                        me->name, (long int)value);
                                 return -ENOEXEC;
                         }

```

Notice that SHN_LIVEPATCH (essentially UNDEF) symbols are unaccounted for, so *all* relas of type `R_PPC_REL24` that refer to livepatch symbols end up caling `local_entry_offset()` instead, which is incorrect behavior.

So I tried the naive solution which is to change the if statement to `if (sym->st_shndx == SHN_UNDEF || sym->st_shndx == SHN_LIVEPATCH)`, since in my head they're essentially `UNDEF` symbols. But then you run into another error:
```
   module_64: kpatch_meminfo: Expect noop after relocate, got e8a10088
```
Which comes from `restore_r2()`, which looks for a noop instruction right after the target rela location. The noops are placeholders and are intended to be overwritten with an instruction to restore the TOC pointer to r2. This instruction is executed after calls to global funcs/symbols that were SHN_UNDEF to begin with, and calls to these functions are entered through their "global entry points" [1], which basically clobbers r2 to set up the TOC pointer, so it makes sense that you would have to restore r2 after the function call. You can see the noops here, for symbols like `si_mem_available`, and `seq_printf`:

   http://paste.fedoraproject.org/531131/48486818/

For the meminfo patch, we have local symbols that were not originally SHN_UNDEF, but became SHN_UNDEF through create-diff-object (its sym->sec was not included in the final object). An example would be `show_val_kb`, and notice there are no noops after calls to this function, as seen in the disassembly. Moreover these functions are not entered through "global entry points", but rather through their "local entry points", which skips r2 clobbering/TOC pointer setup. So relas to `show_val_kb` for example would have called `local_entry_offset()`, as `show_val_kb` would not have been SHN_UNDEF in meminfo.o.

But even calling `local_entry_offset()` for the would-be non-UNDEF symbols gives me the same "REL24 out of rage!" error. Basically, I think this has something to do with "Local calls that become global" (see: [1]), so I'm stumped at the moment..

[1] http://mpe.github.io/posts/2016/05/23/kernel-live-patching-for-ppc64le/

kamalesh-babulal:
> Not sure if the bug is in kpatch-build or in upstream livepatch. But I notice the patch module has several R_PPC64_TOC16_HA and R_PPC64_TOC16_LO entries. Does kpatch-build need to process those somehow?

I remember Michael Ellerman, had a testcase where toc was completely omitted and that function was patched. I have too clearly, try all the patch(s) in /work/trees/userspace/kpatch/test/integration/f22. I will try them.

kamalesh-babulal:
> So this issue is more complicated than I thought, and I hope someone who is more familiar with ppc64le + TOCs can chime in about this (as I have zero experience with ppc64le...). I think the problem has something to do with incorrect TOC manipulation.

Thanks a lots for summarizing the issue. As you pointed out and as per my understanding, need to figure out how to resolve the issue of forcing the TOC setting/restoring.

kamalesh-babulal:
I have not tired rebased my patchset on #607. As seen in the below kernel warning, the bl to show_val_kb is stalled.
```
Jan 20 08:28:47 ubuntu1610 kernel: [13445.280205] livepatch: enabling patch 'kpatch_mem'
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498086] INFO: rcu_sched self-detected stall on CPU
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498855]       3-...: (5249 ticks this GP) idle=3e3/140000000000001/0 softirq=562478/562478 fqs=2623
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498910]        (t=5250 jiffies g=115783 c=115782 q=22)
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498945] Task dump for CPU 3:
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498946] cat             R  running task        0 15990  15963 0x00040004
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498949] Call Trace:
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498957] [c0000000638833a0] [c000000000106cd0] sched_show_task+0xd0/0x150 (unreliable)
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498961] [c000000063883410] [c000000000b00f3c] rcu_dump_cpu_stacks+0xfc/0x148
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498964] [c000000063883460] [c000000000154f4c] rcu_check_callbacks+0x91c/0xb20
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498966] [c000000063883590] [c00000000015ee3c] update_process_times+0x4c/0xa0
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498969] [c0000000638835c0] [c000000000176dc4] tick_sched_handle.isra.7+0x34/0xc0
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498970] [c0000000638835f0] [c000000000176eb8] tick_sched_timer+0x68/0xd0
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498972] [c000000063883630] [c00000000015fb24] __hrtimer_run_queues+0x124/0x420
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498974] [c0000000638836c0] [c000000000160adc] hrtimer_interrupt+0xfc/0x350
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498978] [c000000063883790] [c00000000002415c] __timer_interrupt+0x8c/0x270
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498980] [c0000000638837e0] [c000000000024570] timer_interrupt+0xa0/0xe0
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498982] [c000000063883810] [c000000000009064] decrementer_common+0x114/0x120
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498989] --- interrupt: 901 at meminfo_proc_show+0x50/0xa18 [kpatch_mem]
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498989]     LR = meminfo_proc_show+0x54/0xa18 [kpatch_mem]
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498990] [c000000063883b00] [d0000000018e06d8] meminfo_proc_show+0x48/0xa18 [kpatch_mem] (unreliable)
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498993] [c000000063883c40] [c00000000000bda0] livepatch_handler+0x38/0x74
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498996] [c000000063883ce0] [c0000000003c128c] proc_reg_read+0x8c/0xd0
Jan 20 08:29:24 ubuntu1610 kernel: [13482.498998] [c000000063883d10] [c00000000031eda8] __vfs_read+0x48/0x1b0
Jan 20 08:29:24 ubuntu1610 kernel: [13482.499000] [c000000063883da0] [c000000000320a98] vfs_read+0xb8/0x1a0
Jan 20 08:29:24 ubuntu1610 kernel: [13482.499001] [c000000063883de0] [c0000000003227ac] SyS_read+0x6c/0x110
Jan 20 08:29:24 ubuntu1610 kernel: [13482.499003] [c000000063883e30] [c00000000000b104] system_call+0x38/0xe0
Jan 20 08:29:30 ubuntu1610 kernel: [13488.213973] NMI watchdog: BUG: soft lockup - CPU#3 stuck for 22s! [cat:15990]
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214550] Modules linked in: kpatch_mem(OEK) autofs4 ibmvscsi rtc_generic
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214564] CPU: 3 PID: 15990 Comm: cat Tainted: G           OELK 4.10.0-rc3+ #1
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214566] task: c000000064c00000 task.stack: c000000063880000
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214567] NIP: d0000000018e06e0 LR: d0000000018e06e4 CTR: c000000000256630
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214568] REGS: c000000063883880 TRAP: 0901   Tainted: G           OELK  (4.10.0-rc3+)
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214569] MSR: 8000000000009033 <SF,EE,ME,IR,DR,RI,LE>
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214574]   CR: 22222842  XER: 20000000
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214574] CFAR: d0000000018e06e0 SOFTE: 1
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214574] GPR00: d0000000018e06d8 c000000063883b00 d0000000018e9fc0 c000000063883b60
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214574] GPR04: 0000000000000001 0000000000000001 0000000000001c29 000000007eec0000
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214574] GPR08: c000000001292fb0 0000000000010000 c00000000139f200 d0000000018e0f18
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214574] GPR12: c000000000256630 c00000000fe80900
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214586] NIP [d0000000018e06e0] meminfo_proc_show+0x50/0xa18 [kpatch_mem]
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214588] LR [d0000000018e06e4] meminfo_proc_show+0x54/0xa18 [kpatch_mem]
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214588] Call Trace:
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214590] [c000000063883b00] [d0000000018e06d8] meminfo_proc_show+0x48/0xa18 [kpatch_mem] (unreliable)
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214595] [c000000063883c40] [c00000000000bda0] livepatch_handler+0x38/0x74
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214598] [c000000063883ce0] [c0000000003c128c] proc_reg_read+0x8c/0xd0
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214600] [c000000063883d10] [c00000000031eda8] __vfs_read+0x48/0x1b0
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214602] [c000000063883da0] [c000000000320a98] vfs_read+0xb8/0x1a0
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214603] [c000000063883de0] [c0000000003227ac] SyS_read+0x6c/0x110
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214605] [c000000063883e30] [c00000000000b104] system_call+0x38/0xe0
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214606] Instruction dump:
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214608] fb61ffd8 fb81ffe0 fba1ffe8 fbc1fff0 fbe1fff8 f8010010 f821fec1 7c7f1b78
Jan 20 08:29:30 ubuntu1610 kernel: [13488.214612] 38610060 48000845 e8410018 38610060 <48000001> 60000000 3d020000 e9288048
```

jpoimboe:
@kamalesh-babulal I think that error you saw is because no relocations have been applied to meminfo_proc_show(), because #607 is missing.  FYI my manual merge of your branch with #607 is at https://github.com/jpoimboe/kpatch/tree/tmp

flaming-toast:
FWIW, I believe the generated klp relocations are correct...they are identical to the relocs to the original object (meminfo.o)

Inspecting relas up to offset `e0` for `meminfo_proc_show` -
```
[ patched meminfo.o ]
Relocation section '.rela.text.meminfo_proc_show' at offset 0x1ebc8 contains 142 entries:                                                                                                     
  Offset          Info           Type           Sym. Value    Sym. Name + Addend
000000000000  0022000000fc R_PPC64_REL16_HA  0000000000000000 .TOC. + 0
000000000004  0022000000fa R_PPC64_REL16_LO  0000000000000000 .TOC. + 4
00000000000c  00230000000a R_PPC64_REL24     0000000000000000 _mcount + 0
000000000040  00300000000a R_PPC64_REL24     0000000000000000 si_meminfo + 0
00000000004c  00310000000a R_PPC64_REL24     0000000000000000 si_swapinfo + 0
000000000054  000d00000032 R_PPC64_TOC16_HA  0000000000000000 .toc + 0
000000000058  000d00000040 R_PPC64_TOC16_LO_ 0000000000000000 .toc + 0
000000000068  000d00000032 R_PPC64_TOC16_HA  0000000000000000 .toc + 8
00000000006c  000d00000040 R_PPC64_TOC16_LO_ 0000000000000000 .toc + 8
000000000074  00320000000a R_PPC64_REL24     0000000000000000 total_swapcache_pages + 0
0000000000c4  00330000000a R_PPC64_REL24     0000000000000000 si_mem_available + 0
0000000000d0  001500000032 R_PPC64_TOC16_HA  0000000000000000 .rodata.meminfo_proc_s + 0
0000000000d4  001500000030 R_PPC64_TOC16_LO  0000000000000000 .rodata.meminfo_proc_s + 0
0000000000e0  00090000000a R_PPC64_REL24     0000000000000000 show_val_kb + 0

[ kpatch-meminfo.ko ]
Relocation section '.rela.text.meminfo_proc_show' at offset 0x1188 contains 101 entries:
0000000000000000  0000003b000000fc R_PPC64_REL16_HA       0000000000000000 .TOC. + 0
0000000000000004  0000003b000000fa R_PPC64_REL16_LO       0000000000000000 .TOC. + 4
000000000000000c  000000490000000a R_PPC64_REL24          0000000000000000 _mcount + 0
0000000000000040  000000440000000a R_PPC64_REL24          0000000000000000 si_meminfo + 0
0000000000000054  0000001000000032 R_PPC64_TOC16_HA       0000000000000000 .toc + 38
0000000000000058  0000001000000040 R_PPC64_TOC16_LO_DS    0000000000000000 .toc + 38
0000000000000068  0000001000000032 R_PPC64_TOC16_HA       0000000000000000 .toc + 40
000000000000006c  0000001000000040 R_PPC64_TOC16_LO_DS    0000000000000000 .toc + 40
00000000000000c4  000000480000000a R_PPC64_REL24          0000000000000000 si_mem_available + 0
00000000000000d0  0000000b00000032 R_PPC64_TOC16_HA       0000000000000000 .rodata.meminfo_proc_show.str1.8 + 0
00000000000000d4  0000000b00000030 R_PPC64_TOC16_LO       0000000000000000 .rodata.meminfo_proc_show.str1.8 + 0

Relocation section '.klp.rela.vmlinux..text.meminfo_proc_show' at offset 0x7f960 contains 41 entries:
000000000000004c  0000004c0000000a R_PPC64_REL24          0000000000000000 .klp.sym.vmlinux.si_swapinfo,0 + 0
0000000000000074  0000004d0000000a R_PPC64_REL24          0000000000000000 .klp.sym.vmlinux.total_swapcache_pages,0 + 0
00000000000000e0  0000004e0000000a R_PPC64_REL24          0000000000000000 .klp.sym.vmlinux.show_val_kb,1 + 0
```
The relas for `si_swapinfo`, `total_swapcache_pages`, and `show_val_kb` (that would originally be dynrelas) are now in `.klp.rela.vmlinux..text.meminfo_proc_show`, and they are the same as the relas from the original object, except they refer to new klp symbols.

kamalesh-babulal:
toc manipulation also required kernel changes  (https://gist.github.com/kamalesh-babulal/113b63b524b71ca583823f003afea8b0) and my branch [ppc64le_v2](https://github.com/kamalesh-babulal/kpatch/tree/ppc64le_v2) ( has the toc manipulation and special struct support). I could like to run it through all the available test case before posting. Any pointers on how to setup and run the test cases for klp modules on latest kernel will be helpful.

kamalesh-babulal:
>Any pointers on how to setup and run the test cases for klp modules on latest kernel will be helpful.

adding -s kernel-source to KPATCHBUILD helps.
```
--- a/test/integration/kpatch-test
+++ b/test/integration/kpatch-test
@@ -48,7 +48,7 @@ ROOTDIR="$(readlink -f $SCRIPTDIR/../..)"
 KPATCH="sudo $ROOTDIR/kpatch/kpatch"
 RMMOD="sudo rmmod"
 unset CCACHE_HASHDIR
-KPATCHBUILD="$ROOTDIR"/kpatch-build/kpatch-build
+KPATCHBUILD="$ROOTDIR""/kpatch-build/kpatch-build -s /root/latest/linux"
 ERROR=0
 LOG=test.log
 rm -f $LOG
```

joe-lawrence:
@kamalesh-babulal 
>> Any pointers on how to setup and run the test cases for klp modules on latest kernel will be helpful.

> adding -s kernel-source to KPATCHBUILD helps.

Looks like you're on your way to getting the tests off the ground by manually setting `KPATCHBUILD`.  Let us know is not the case, or if other testing issues pop up.

In instances like this, it would be handy if `kpatch-test` had a `-s, --sourcedir` option that would be passed long to `kpatch-build`.  This could be implemented like the `-d, --directory` option, adding ` -s $2` to `KPATCHBUILD`.

On a related note, it's been a while since I've run `kpatch-test` against livepatches.  You'll probably end up seeing a lot of "kpatch unload failed" complaints.  The script could be more intelligent about kpatch vs. livepatch features here, or at least handle failed unloads more gracefully.  Suggestions, patches welcome :)

PS - There is a `test/integration/rebase-patches` hack-script that might be useful if the patches don't apply to your kernel.  It assumes a kernel source in `$HOME/.kpatch/src` and then tries to recreate the `$distro-$release/` pile of patches.  The Fedora 25 patches are probably the most up to date, so hopefully they will still apply to your kernel.

kamalesh-babulal:
@joe-lawrence Thank you. Adding ```-s, --sourcedir``` and ```-d, --directory``` should be easy. I will extend ```kpatch-test``` with those options.

jpoimboe:
Hi @kamalesh-babulal, any luck with testing?

kamalesh-babulal:
Hi @flaming-toast, while discussing toc relocation issues for data read only patch on powerpc, @jpoimboe pointed out that, some toc related issues were encountered on s390 and a new gcc flag was introduced to force all relocation to do PLT/GOT instead of relative (assuming here). Could please share the details about the flag. Thanks.

flaming-toast:
@kamalesh-babulal Yes, the s390x patch for gcc is still a WIP and hasn't been posted up on public mailing lists yet. However, ARM currently has the same option, called -mpic-data-is-text-relative. From the gcc man page:
```
-mpic-data-is-text-relative
           Assume that each data segments are relative to text segment at load time. 
           Therefore, it permits addressing data using PC-relative operations.
```
So with the s390x option, when -m**no**-pic-data-is-text-relative is used in conjunction with -fPIC, gcc generates position-independent code even for references to local, static data, not just global data (so everything goes through the GOT). We needed to avoid using relative relocations on s390x due to the fact that modules are loaded too far away from the kernel core image and 32-bit offsets were not going to be sufficient. When all function calls and both local and global data references go through the PLT/GOT respectively, we achieve position independence (at the cost of performance), and the issues with relative relocations go away.

kamalesh-babulal:
> @flaming-toast Thank you for the information on -mno-pic-data-is-text-relative. Just curious, on approaches you tired before concluding on the gcc flag as the best approach.

flaming-toast:
@kamalesh-babulal I think the only other option would be to MODULES_VADDR closer to the core kernel image, so that everything (core kernel + modules) remained addressable within 4GB. If that were the case, the 32-bit relative relocations would work. However, the s390x maintainer was not receptive to the idea, and explained that it would nontrivially change the way they (plan) to implement KASLR on s390x, among other reasons. He was more supportive of the gcc option to avoid one-off changes to the s390x kernel code. Does ppc64le have the same relative-addressing problems?

kamalesh-babulal:
> @flaming-toast ppc64le has exactly the same issue. I asked the question about other approaches, just to make sure, I didn't miss other approaches. So I can try convincing the powerpc toolchains developers to implement something similar. I have been bounced different ideas from other developers internally, who worked on similar project previously like changing the relocation types in the kpatch, from TOC relative relocation to something like immediate load or manually creating a special section in module generated and for every TOC relative load and jump to a location, where all the manipulation is done. I am not really convinced on the approaches, as it looks more hackish.

kamalesh-babulal:
The major change from v2 is supporting ```-mcmodel=large``` and gcc-5, ggc-6 function prologues. The idea behind using ```-mcmodel=large``` is to avoid manually manipulating .toc section for DS-form of loads, which is and should be ideally done by the linker.

With ```-mcmodel=medium``` (which is the defacto standard for the powerpc kernel build) assumes the load to always be toc base + 32-bit offset range. This assumption breaks when the non-module kernel code is copied into a module, where the expectation is to have 64bit relocations.

Whereas ```-mcmodel=large```, makes no such assumptions about the range at which the data is available and generates a 64bit relocation a.k.a ```R_PPC64_ADDR``` for every load in the function section by generating an entry into the .toc section, which resolves to 64bit address.

gcc-5 and gcc-6 differs in function prologue, for initializing the toc base. Both the format are handled in this version. For example with gcc-5, the function entry is:
```
.globl my_func
.type my_func, @function
my_func:
addis r2, r12, (.TOC.-my_func)@ha
addi r2, r2, (.TOC.-my_func)@l
.localentry my_func, .-my_func
... ; function definition
blr
```
and for gcc-6 the function entry is:
```
.globl my_func
.type my_func, @function
.quad .TOC.-my_func
my_func:
.reloc ., R_PPC64_ENTRY ; optional
ld r2,-8(r12)
add r2,r2,r12
.localentry my_func, .-my_func
... ; function definition
blr
```
The last patch in the series is a bug fix. I have another series with testcases rebased for ubuntu-17.04, along with a test case for powerpc fixup. Which I will request for merge sooner.

jpoimboe:
@kamalesh-babulal I haven't reviewed it in-depth yet, but so far it looks really good!  Nice work!!!

One question about the -mcmodel approach.  Do you know if that flag has side effects which might affect gcc's optimizations and inlining decisions?  Or does it *only* change the relocations?

If it changes more than just the relocations, then the recompiled original kernel might be different from the *real* original kernel.  In that case, we may need to rethink how we do the compile passes.

kamalesh-babulal:
> @jpoimboe Thank you. AFAIK ```-mcmodel=large``` removes the assumption about the data availability within the 2^32 offset of the toc base (4GB range). It does not affect any other GCC optimization, from the testing I have done using the test cases from ```kpatch/test/integration/fedora-25```.

kamalesh-babulal:
Fixed a minor bug in ```kmod/patch/Makefile```  file. Replaced ```ARCH``` with ```PROCESSOR``` as overruling ```ARCH``` variable set by kbuild will trigger kernel module build failure. For ppc* the value is set to powerpc.

@jpoimboe Any feedback/suggestions on the patchset ?

jpoimboe:
@kamalesh-babulal Sorry for the delay, I'll try to review it this week.

kamalesh-babulal:
> @jpoimboe Thank you

kpatch-build/kpatch-elf.c
```diff
@@ -76,6 +76,19 @@ int is_debug_section(struct section *sec)
...
+ * gcc5 and gcc6, this check is required for ppc64le
+ * only.
+ */
+int gcc6_check(void)
+{
+#ifdef __powerpc__
+	return (GCC_VERSION >= 60100);
+#endif
+	return 0;
+}
```
jpoimboe: One potential problem with this GCC 6 check is that it checks the version which was used to compile create-diff-object itself, but that could have been a different version than was used to compile the kernel.  For example, kpatch-build could be delivered by the distro's RPM, but the user might be making a patch for another kernel.

However it's ok with me to leave this check for now and fix it later when somebody has a problem with it.

kamalesh-babulal:
> @jpoimboe Thanks for the review. I have tested it over gcc-5 mostly and gcc-6 to an extent. I am able to reproduce it on gcc-6 not consistency though with data-read-mostly.patch from ```test/integration/fedora-25/data-read-mostly.patch```. Some runs pass and some fails. One idea is to loop through the local symbols, to avoid the assumption of ordering.
```
diff --git a/kpatch-build/lookup.c b/kpatch-build/lookup.c
index 7f9da34..d377e29 100644
--- a/kpatch-build/lookup.c
+++ b/kpatch-build/lookup.c
@@ -83,21 +83,28 @@ static void find_local_syms(struct lookup_table *table, char *hint,
                            struct sym_compare_type *child_locals)
 {
        struct object_symbol *sym, *file_sym = NULL;
-       int i, in_file = 0;
-       struct sym_compare_type *child_sym = NULL;
+       int i, found_local_sym = 0, nr_local_sym = 0, in_file = 0;
+       struct sym_compare_type *tmp_child_sym, *child_sym = NULL;
 
        if (!child_locals)
                return;
 
+       tmp_child_sym = child_locals;
+       while(tmp_child_sym->name) {
+               nr_local_sym++;
+               tmp_child_sym++;
+       }
+
        for_each_obj_symbol(i, sym, table) {
                if (sym->type == STT_FILE) {
-                       if (in_file && !child_sym->name) {
+                       if (in_file && found_local_sym == nr_local_sym) {
                                if (table->local_syms)
                                        ERROR("find_local_syms for %s: found_dup", hint);
                                table->local_syms = file_sym;
                        }
 
                        if (!strcmp(hint, sym->name)) {
+                               found_local_sym = 0;
                                in_file = 1;
                                file_sym = sym;
                                child_sym = child_locals;
@@ -113,22 +120,28 @@ static void find_local_syms(struct lookup_table *table, char *hint,
                if (sym->bind != STB_LOCAL || (sym->type != STT_FUNC && sym->type != STT_OBJECT))
                        continue;
 
-               /*
-                * Symbols which get discarded at link time are missing from
-                * the lookup table, so skip them.
-                */
-               while (discarded_sym(table, child_sym))
-                       child_sym++;
-
-               /* make sure the child symbol and parent symbol match */
-               if (child_sym->name && child_sym->type == sym->type &&
-                   !strcmp(child_sym->name, sym->name))
-                       child_sym++;
-               else if (child_sym->name && child_sym->type == sym->type &&
-                        !kpatch_mangled_strcmp(child_sym->name, sym->name))
-                       child_sym++;
-               else
-                       in_file = 0;
+               tmp_child_sym = child_locals;
+               while(tmp_child_sym->name) {
+                       /*
+                        * Symbols which get discarded at link time are missing from
+                        * the lookup table, so skip them.
+                        */
+                       if (discarded_sym(table, tmp_child_sym))
+                               found_local_sym++;
+
+                       /* make sure the child symbol and parent symbol match */
+                       if (tmp_child_sym->name && tmp_child_sym->type == sym->type &&
+                                       !strcmp(tmp_child_sym->name, sym->name)) {
+                               found_local_sym++;
+                               break;
+                       }
+                       else if (tmp_child_sym->name && tmp_child_sym->type == sym->type &&
+                                !kpatch_mangled_strcmp(tmp_child_sym->name, sym->name)) {
+                                       found_local_sym++;
+                                       break;
+                       }
+                       tmp_child_sym++;
+               }
        }
```

kmod/patch/livepatch-patch-hook.c
```diff
@@ -248,6 +248,11 @@ static int __init patch_init(void)
 		goto out;
 	lpatch->mod = THIS_MODULE;
 	lpatch->objs = lobjects;
+#if __powerpc__
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(4, 11, 0))
+	lpatch->immediate = true;
+#endif
```
kamalesh-babulal: > Yes, Will change it the kernel version check.

kpatch-build/Makefile
```diff
@@ -7,7 +7,13 @@ TARGETS = create-diff-object create-klp-module create-kpatch-module
 SOURCES = create-diff-object.c kpatch-elf.c \
 		  create-klp-module.c \
 		  create-kpatch-module.c \
-		  lookup.c insn/insn.c insn/inat.c
+		  create-kpatch-module.c lookup.c
+
+ifeq ($(ARCH),x86_64)
+SOURCES += lookup.c insn/insn.c insn/inat.c
```
kamalesh-babulal: > will fix it in the next iteration.

kpatch-build/kpatch-build
```diff
@@ -145,6 +147,21 @@ find_core_symvers() {
...
+	# for gcc support checks.
+	echo "int func() { return 0; }" | \
+		gcc -S -x c -O2 -p -mprofile-kernel - -o - 2> /dev/null | \
+		grep -q "_mcount" || return 1
+
+	echo -e "#include <linux/compiler.h>\nnotrace int func() { return 0; }" | \
+		gcc -S -x c -O2 -p -mprofile-kernel - -o - 2> /dev/null | \
+		grep -q "_mcount" && return 1
+
+	return 0
```
kamalesh-babulal: > good point, the idea was to check early before doing an expensive readelf of the kernel. Yes, it can dropped.

kamalesh-babulal:
Fixed the review comments, expect the gcc version check in ```kpatch-build/kpatch-elf.c``` and dropped the  patch ```0009-create-diff-lookup-Fix-special-symbols-comparison.patch``` from the previous iteration, the issue of re-ordering of the symbol names have been addressed by #717.

kpatch-build/create-diff-object.c
```diff
@@ -146,13 +152,78 @@ static int kpatch_mangled_strcmp(char *s1, char *s2)
...
+ * with -mcmodel=large on ppc64le, gcc might generate entries in .toc section for symbol
+ * reference in relocation sections. There are chances of symbol referenced by toc + offset
+ * may point to different symbols in original and patched .o. Compare the symbols by
+ * reading its corresponding toc + offset rela from .toc section.
+ */
+	if (!strncmp(rela1->sym->name, ".toc", 4) &&
+	    !strncmp(rela2->sym->name, ".toc", 4)) {
+
+		elf1 = container_of(&rela1->sym->sec->list, struct kpatch_elf, sections);
+		toc1 = find_section_by_name(&elf1->sections, ".rela.toc");
```
jpoimboe: Instead I think you can just do: `toc1 = rela1->sym->sec->rela`.

kpatch-build/create-diff-object.c
```diff
@@ -146,13 +152,78 @@ static int kpatch_mangled_strcmp(char *s1, char *s2)
...
+		if (!toc1)
+			ERROR("cannot find .rela.toc");
+
+		rela3 = find_rela_by_offset(toc1, rela1->addend);
```
jpoimboe: I think it would be clearer to rename `rela3` to `rela1_toc_rela`

kpatch-build/create-diff-object.c
jpoimboe: Also maybe rename `toc1` to `toc_relasec1` or something?  It's longer, but it much better describes what it is.

kpatch-build/create-diff-object.c
```diff
@@ -200,8 +281,16 @@ static void kpatch_compare_correlated_section(struct section *sec)
...
+	    sec1->sh.sh_entsize != sec2->sh.sh_entsize) {
+		/*
+		 * .rodata section sometimes have sh_addr generated
+		 * and it might differ in the patched *.o file. Skip,
+		 * this case.
+		 */
+		if (!strncmp(sec1->name, ".rodata", 7) &&
+		    sec1->sh.sh_addr == sec2->sh.sh_addr)
+			DIFF_FATAL("%s section header details differ", sec1->name);
+	}
```
jpoimboe: Instead of doing this, it would be better to just remove the `sec1->sh.sh_addr != sec2->sh.sh_addr ||` line above.

It doesn't really make sense to compare the section addresses anyway, since those will change if a previous section changes, not the current section.  The only reason that worked on x86 is because the address is always zero for some reason.

kpatch-build/create-diff-object.c
```diff
@@ -2109,9 +2291,27 @@ static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
...
+				 * fs/proc/proc_sysctl::sysctl_head_grab:
+				 * 166: 0000000000000000   256 FUNC    GLOBAL DEFAULT [<localentry>: 8]    42 unregister_sysctl_table
+				 * 167: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND .TOC.
+				 *
+				 * These type of function have the sym->type == 2, treat them as local
+				 * symbol. This creates an entry under klp.relocations and relocation
+				 * is handled by livepatch.
+				 */
+				if (lookup_is_exported_symbol(table, rela->sym->name)) {
+					if (rela->sym->type != 2)
```
jpoimboe: This should use STT_FUNC instead of 2 (and the same for the comment above it).

kpatch-build/kpatch-gcc
```diff
@@ -15,7 +15,10 @@ if [[ "$TOOLCHAINCMD" = "gcc" ]] ; then
 	while [ "$#" -gt 0 ]; do
 		if [ "$1" = "-o" ]; then
 			obj=$2
-			[[ $2 = */.tmp_*.o ]] && obj=${2/.tmp_/}
+			# skip copying the original .o files, when .tmp_mc_*.o
+			# is passed from recordmcount.pl.
+			[[ $2 = */.tmp_mc_*.o ]] && break;
+			[[ $obj = */.tmp_*.o ]] && obj=${2/.tmp_/}
```
jpoimboe: The 1st line seems to match the comment, but what's the purpose of the 2nd line?

Also it would be clearer to use the $obj variable everywhere instead of $2.

kpatch-build/create-diff-object.c
```diff
@@ -922,6 +1020,9 @@ static void kpatch_replace_sections_syms(struct kpatch_elf *kelf)
 				continue;
 			}
 
+#ifdef __powerpc__
+			add_off = 0;
+#else
```
jpoimboe: I don't see how this would work for relative relocations, like PPC64_REL24.  In those cases I think the logic would be similar to what x86 does in the R_X86_64_PC32 case, because the address offset would be relative to the instruction address.  Or am I missing something?

Thankfully you wouldn't need rela_insn(), because that's just needed for determining the instruction length.  ppc has a fixed instruction size so it wouldn't need to disassemble.

jpoimboe:
Have you tried meminfo-string.patch?  I just tried it (built with GCC 6 this time) and the load failed with:
```
module_64: kpatch_meminfo_string: REL24 -1152921504828326972 out of range!
```

kpatch-build/create-diff-object.c
kamalesh-babulal: with ```-mcmodel=large```, the relative relocation or R_PPC64_REL24 types are limited to functions only. With -ffunction-sections gcc flag every function is the section and the section is replaced with their symbols by
```
if (rela->sym->sec && rela->sym->sec->sym) {
        rela->sym = rela->sym->sec->sym;
        continue;
}
```
for every data access, the -mcmodel=large generates two relocations types
```R_PPC64_TOC16_HA``` and ```R_PPC64_TOC16_LO_DS``` . Which is DS-Form load.
```add_off=0``` should be the right thing to do here.

File: kpatch-build/create-klp-module.c [hunk 17]
```diff
@@ -208,8 +239,23 @@ static void create_klp_relasecs_and_syms(struct kpatch_elf *kelf, struct section
...
+		else
+			rela->offset = krelas[index].offset;
+
+		/*
+		 * gcc6 adds offset for 0x8 for every local function entry in
+		 * .toc section, for avoiding setup of toc but when the previously
+		 * local function becomes global, toc is anyways setup. So remove
+		 * the wrong addend.
+		 */
+		if (gcc6_check() && !strncmp(base->name, ".toc", 4)) {
```
kamalesh-babulal: Agree, will change them.

kpatch-build/create-diff-object.c
kamalesh-babulal: Sure, will rename it to something meaningful.

kpatch-build/create-diff-object.c
kamalesh-babulal: Will rename it to something meaningful.

kpatch-build/create-diff-object.c
kamalesh-babulal: Agree, will replace magical numbers with readable definitions.

kpatch-build/kpatch-gcc
kamalesh-babulal: The second line is the original code (interchanged the $2 and $obj), which is required for append the changed objects file to ```$KPATCH_GCC_TEMPDIR/changed_objs```. Sure, will change the usage of $2.

kamalesh-babulal:
For ```R_PPC64_REL24``` relocations, the idea is to allow the ```kernel/modues_64.c:apply_relocate_add()``` to handle the relocation. I had posted the [first version](https://lists.ozlabs.org/pipermail/linuxppc-dev/2017-June/159303.html) to the mailing list, cc'ing you on the patch for the creating a stub to handle livepatch symbols. I was pointed out that it will not work in the case of functions passing parameters over the stack. The reason being, there was an intermediate stack created while branching into the function and if the called function wanted to access the parameters from the back chain it's not possibe. I have re-worked on it, by adopting the ``entry_64.S::livepatch_handler`` ASM. Where a livepatch stack is created in the task stack to store and restore TOC/LR values.

Now, if there is a stack corruption while returning from the branched function, it will end up calling the generic trap instruction and continue without a BUG(). Unlike ASM where adding a bug entry is easy, here in the stub code it opcodes of the instruction. Working on the emitting bug part. For testing purpose, If you apply the first version of patch posted to the mailing list over 4.10 kernel. It will resolve this issue.

kpatch-build/create-diff-object.c
kamalesh-babulal: Was not sure, why it's always zero on x86 and did not want to break the assumption. So included the addr check. Will remove it.

kpatch-build/create-diff-object.c
jpoimboe: If I remember correctly, there are a few cases on x86 where `-ffunction-sections` doesn't work and certain functions aren't bundled into their own sections.  For example, functions in `.text.unlikely`.

So you might want to at least add a sanity check and ERROR() if the relocation type is `R_PPC64_REL24` and the if (rela->sym->sec && rela->sym->sec->sym) is false, to catch such a case.

jpoimboe:
I tested with your upstream patch against 4.12 and GCC 6 and it worked!  This is looking really good.  I just had the one comment about kpatch_replace_sections_syms().

kpatch-build/create-diff-object.c
kamalesh-babulal: Can you please point me to test case, which uncovered the issue on x86 ?. I have included the sanity check in the next version.

kamalesh-babulal:
> Thank you

kpatch-build/create-diff-object.c
jpoimboe: Sorry, I don't have a test case in mind.  If we ever get the error then we'll have one :-)

joe-lawrence:
👍  good work, @kamalesh-babulal !

kpatch-build/kpatch-elf.c
```diff
@@ -328,9 +341,29 @@ void kpatch_create_symbol_list(struct kpatch_elf *kelf)
...
+				 * where .localentry for the functions, will generated
+				 * st_value of 0x8 for the local functions. Skip, this
+				 * case.
+				 */
+				if (sym->sym.st_value != 0) {
+					if (!gcc6_check() ||
+					   (sym->type != STT_FUNC || sym->sym.st_value != 0x8))
+						ERROR("symbol %s at offset %lu within section %s, expected 0",
 					      sym->name, sym->sym.st_value, sym->sec->name);
+				}
```
jpoimboe: So I'm having trouble understanding this code.  It doesn't seem to match the comment.  It doesn't look like the symbol is skipped.  Instead, it's just ignored for GCC 6, and is otherwise reported as an error.  Is this code even needed?

Also I'm worried this check might produce some false positives.  Is there a more specific way to test for localentry symbols?

Looking at the binutils source for some clues:
```
  if (psym->st_other ^ ELF_ST_VISIBILITY (psym->st_other))
    printf (" [%s] ", get_symbol_other (psym->st_other ^ ELF_ST_VISIBILITY (psym->st_other)));

...

static const char *
get_ppc64_symbol_other (unsigned int other)
{
  if (PPC64_LOCAL_ENTRY_OFFSET (other) != 0)
    {
      static char buf[32];
      snprintf (buf, sizeof buf, _("<localentry>: %d"),
		PPC64_LOCAL_ENTRY_OFFSET (other));
      return buf;
    }
  return NULL;
}

...

#define PPC64_LOCAL_ENTRY_OFFSET(other)				\
  ppc64_decode_local_entry (((other) & STO_PPC64_LOCAL_MASK)	\
			    >> STO_PPC64_LOCAL_BIT)

...

/* The ELFv2 ABI uses three bits in the symbol st_other field of a
   function definition to specify the number of instructions between a
   function's global entry point and local entry point.
   The global entry point is used when it is necessary to set up the
   toc pointer (r2) for the function.  Callers must enter the global
   entry point with r12 set to the global entry point address.  On
   return from the function, r2 may have a different value to that
   which it had on entry.
   The local entry point is used when r2 is known to already be valid
   for the function.  There is no requirement on r12 when using the
   local entry point, and on return r2 will contain the same value as
   at entry.
   A value of zero in these bits means that the function has a single
   entry point with no requirement on r12 or r2, and that on return r2
   will contain the same value as at entry.
   Values of one and seven are reserved.  */
#define STO_PPC64_LOCAL_BIT		5
#define STO_PPC64_LOCAL_MASK		(7 << STO_PPC64_LOCAL_BIT)
```

So it looks like a specific check of the st_other field could be done to avoid false positives.  And then I think you could get rid of the gcc 6 check, which was problematic anyway for the case where kpatch-build was compiled with a different compiler as the kernel.

Or maybe this check isn't needed at all and this whole change can just be removed?

kpatch-build/create-diff-object.c
```diff
@@ -2094,6 +2253,27 @@ static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
...
+			 *	add r2,r2,r12
+			 * 	.localentry my_func, .-my_func
+			 *
+			 * first entry in function entry is optional and might
+			 * have symbol length of 0. We should not skip this entry
+			 * while building up the rela list, rather skip its'
+			 * lookup.
+			 */
+			if (!strlen(rela->sym->name))
+				continue;
```
kamalesh-babulal: Time to put on brown paper bag :(. Right, it should be ```rela->type == R_PPC64_ENTRY```

kpatch-build/create-klp-module.c
```diff
@@ -208,8 +239,23 @@ static void create_klp_relasecs_and_syms(struct kpatch_elf *kelf, struct section
 		ALLOC_LINK(rela, &klp_relasec->relas);
 		rela->sym = sym;
 		rela->type = krelas[index].type;
-		rela->offset = krelas[index].offset;
-		rela->addend = krelas[index].addend;
+		if (!strcmp(base->name, ".toc"))
+			rela->offset = toc_offset;
+		else
+			rela->offset = krelas[index].offset;
```
kamalesh-babulal: The reason, why it been handled here is. When more two or more objects files are linked together, their .toc enteries are added to the .toc section of kpatch-output object and there is no fool proof way to predictable the arrangement of entries in the combined .toc section. From ```kmod/patch/Makefile``` file ```kpatch-$(KPATCH_NAME)-objs += patch-hook.o kpatch.lds output.o``` . When .toc section entries from output.o and patch-hook.o are merged, the arrangement of the combined .toc is not guaranteed to map the index based offset. I have added a comment with an example, explaining the re-arrangement at line 177 explaining this re-arrangement. I guess, it better to move it here.

kpatch-build/kpatch-elf.c
kamalesh-babulal: With gcc 6 the local function symbol st_value is 0x8. Agree, I will re-work on a more reliable way for local entry detection and get rid of gcc6_check().

kamalesh-babulal:
> @joe-lawrence Thank you

kpatch-build/create-klp-module.c
jpoimboe: Ah, ok.  Makes sense, thanks.

jpoimboe: I saw the comment, I think I just didn't fully understand it.  Now it makes sense.  No need to move the comment unless you want to.

kpatch-build/kpatch-elf.c
kamalesh-babulal: The localentry or the local function can be guaranteed by 
1. if the ```symbol->st_shndx != SHN_UNDEF```
2. Decode the number of instruction between functions global and local entry point. This information is only available for local functions ```sym->bind == STB_LOCAL``` or is 0.

This will also remove the need for gcc6_check(). I am running a couple of test before posting the next version.

kamalesh-babulal:
This version of patchset addresses the review comment about reliably detecting local function. As per PowerPC64 ABI v2, the symbol those to be resolved dynamically has the st_shndx member set to SHN_UNDEF and to be sure, the number of instructions between the global and local entry of a function is also checked, along with st_shndx of the STT_FUNC symbol.

kpatch-build/kpatch-elf.c
```diff
@@ -76,6 +76,31 @@ int is_debug_section(struct section *sec)
...
+#ifdef __powerpc__
+/*
+ * Number of instructions between global and local entry point of a
+ * function is mostly 8 instructions. i.e st_other == 3. The count of
+ * instruction can be decoded only for local function symbols.
+ */
+unsigned int has_local_entry_offset(GElf_Sym sym)
+{
+	return !!PPC64_LOCAL_ENTRY_OFFSET(sym.st_other);
+}
```
jpoimboe: The function comment doesn't describe the intent of the function very well.  More on that below.

jpoimboe: Also, a couple of changes to make the function consistent with other code:

- make the argument `struct symbol *`
- make the return code `int`
- if the function isn't used outside of the file, make it `static`.

And similar comments for `is_local_func()` below.

kpatch-build/kpatch-elf.c
```diff
@@ -76,6 +76,31 @@ int is_debug_section(struct section *sec)
...
+
+/*
+ * As per ELF ABI v2 specification, any function those symbol reference
+ * is to be dynamically resolved. The st_shndx member of such symbol
+ * is set to SHN_UNDEF.
+ */
```
jpoimboe: I think the function comment is confusing because it doesn't describe the code.

kpatch-build/kpatch-elf.c
```diff
@@ -76,6 +76,31 @@ int is_debug_section(struct section *sec)
...
+unsigned int is_local_func(struct symbol *sym)
+{
+	if (sym->type != STT_FUNC || sym->sym.st_shndx == SHN_UNDEF)
+		return 0;
+
+	return 1;
```
jpoimboe: I think removing the double negative would be more readable:
```
return (sym->type == STT_FUNC && sym->sym.st_shndx != SHN_UNDEF);
```
Also, why not check `sym->bind != STB_GLOBAL` instead of `sym->sym.st_shndx != SHN_UNDEF`?

kpatch-build/kpatch-elf.c
```diff
@@ -328,9 +353,32 @@ void kpatch_create_symbol_list(struct kpatch_elf *kelf)
...
+				if (sym->sym.st_value != 0) {
+#ifdef __powerpc__
+					if (!is_local_func(sym) ||
+					    !has_local_entry_offset(sym->sym) ||
+					    (sym->sym.st_value != 0x8))
+#endif
+						ERROR("symbol %s at offset %lu within section %s, expected 0",
 					      sym->name, sym->sym.st_value, sym->sec->name);
+				}
 				sym->sec->sym = sym;
```
jpoimboe: I propose the following changes to make this code clearer:

- Move the long comment to above the `has_local_entry_offset()` function, since it describes the intent of that function.  It's too much information for this location, IMO.
- Remove the "Skip, this case" comment because nothing is skipped.
- Rename `has_local_entry_offset()` to `is_localentry_sym()` and do the `is_local_func()` and `st_value == 0x8` checks inside that function.
- Add a newline between the error checking clause and the `sym->sec->sym = sym;`.  Otherwise the most important line of code is visually obscured.

kpatch-build/kpatch-elf.h
```diff
@@ -131,6 +131,22 @@ struct rela *find_rela_by_offset(struct section *relasec, unsigned int offset);
...
+#ifdef __powerpc__
+#ifndef R_PPC64_ENTRY
+#define R_PPC64_ENTRY	118
+#endif
+
+/* Symbol st_others value for powerpc */
+#define STO_PPC64_LOCAL_BIT     5
+#define STO_PPC64_LOCAL_MASK    (7 << STO_PPC64_LOCAL_BIT)
+#define PPC64_LOCAL_ENTRY_OFFSET(other)                                 \
+	(((1 << (((other) & STO_PPC64_LOCAL_MASK) >> STO_PPC64_LOCAL_BIT)) >> 2) << 2)
```
jpoimboe: I'd rather have these defines for st_other to be in the .c file, right above the function which uses them.

kpatch-build/kpatch-elf.c
kamalesh-babulal: Consider the following case there the ``` sym->bind == STB_GLOBAL and sym->sym.st_shndx != SHN_UNDEF ```
```
51: 0000000000000008  1816 FUNC    LOCAL  DEFAULT [<localentry>: 8]     9 mmap_region
60: 0000000000000008    72 FUNC    GLOBAL DEFAULT [<localentry>: 8]     7 cleanup_module
65: 0000000000000008   916 FUNC    GLOBAL DEFAULT [<localentry>: 8]     5 init_module
```
just checking for ```sym->bind != STB_GLOBAL ```, will end in false positive.

kpatch-build/create-klp-module.c
```diff
@@ -208,8 +239,23 @@ static void create_klp_relasecs_and_syms(struct kpatch_elf *kelf, struct section
...
+
+		/*
+		 * gcc6 adds offset for 0x8 for every local function entry in
+		 * .toc section, for avoiding setup of toc but when the previously
+		 * local function becomes global, toc is anyways setup. So remove
+		 * the wrong addend.
+		 */
+		if (!strcmp(base->name, ".toc") &&
+			rela->sym->type == STT_FUNC && rela->sym->bind == STB_LOCAL) {
```
kamalesh-babulal: Here ```rela->addend``` check for the ```klp.rela.toc``` relocation entries, where the previously local functions are global now, but their ``` sym->bind == STB_LOCAL ``` and the ```sym.st_shndx == 0xff20 ``` a.k.a not equal to ``` sym.st_shndx != SHN_UNDEF```. The latter check for ```sym.st_shndx``` will return false positive and also given that they are not part of current livepatch module the instructions between global and local entry point of these function can not be decoded.

kamalesh-babulal:
@jpoimboe Thank you so much for the review comments.

kpatch-build/kpatch-elf.c
```diff
@@ -76,6 +76,65 @@ int is_debug_section(struct section *sec)
...
+/*
+ * As per ELF ABI v2 specification, any function those symbol reference
+ * is to be dynamically resolved. The st_shndx member of such symbol
+ * is set to SHN_UNDEF. This function return 1, if the function's
+ * is local by checking for it's sym.st_shndx value.
+ */
+static int is_local_func(struct symbol *sym)
+{
+	if (sym->type == STT_FUNC && sym->sym.st_shndx == SHN_UNDEF)
+		return 0;
```
jpoimboe: This condition changed from the last version.  Shouldn't it be
```
if (sym->type != STT_FUNC || sym->sym.st_shndx == SHN_UNDEF)
        return 0;
```
Or just
```
return (sym->type == STT_FUNC && sym->sym.st_shndx != SHN_UNDEF);
```
?

jpoimboe: Also, I think this function should be removed and this check should just be moved directly into `is_localentry_sym()`.  The function comment above doesn't seem to be very helpful anyway.

kamalesh-babulal: @jpoimboe Thanks for the review. Make sense, combining them into a single function. Order of the check should be

1. ```sym->type != STT_FUNC```
2. ```sym->sym.st_shndx == SHN_UNDEF```
3. ```sym->sym.st_value != 0x8``` and
4. No of instructions between global and local entry point.

jpoimboe:
This PR has been merged.  Thank you @kamalesh-babulal !!!

kamalesh-babulal:
@jpoimboe @flaming-toast @joe-lawrence Thank you !!

PR #651, opened 2016-12-23
support dup file+symbol

### LINKED ISSUE #604: Resolving duplicate local functions
The following patch fails to compile:

```
/usr/libexec/kpatch/create-diff-object: ERROR: dir.o: kpatch_create_patches_sections: 2528: lookup_local_symbol call_filldir (dir.c)
```

``` diff
diff --git a/fs/ext4/dir.c b/fs/ext4/dir.c
index f8d56e4..949aa45 100644
--- a/fs/ext4/dir.c
+++ b/fs/ext4/dir.c
@@ -115,6 +115,9 @@ static int ext4_readdir(struct file *filp,
    struct super_block *sb = inode->i_sb;
    int ret = 0;
    int dir_has_error = 0;
+   static unsigned int count;
+
+   count += 1;

    if (is_dx_dir(inode)) {
        err = ext4_dx_readdir(filp, dirent, filldir);
```

The problem is that both fs/ext3/dir.c and fs/ext4/dir.c define call_filldir. I haven't yet thought of a better way of handling this...

zhouchengming1:
This patch needs some explanation. When we want to resolve a dup
file+symbol symbol, we first find out our base object in lookup table,
then we can use this information to resolve ambiguous local symbols.

We use resolve_dup_file_symbol to find out the correct base object in
lookup table. If all local symbols of a file can be found in the
kelf_base, we think it's a matched file.

Fixes #604.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1836,6 +1836,43 @@ static void kpatch_process_special_sections(struct kpatch_elf *kelf)
 	}
 }
 
+static struct sym_compare_type *kpatch_elf_locals(struct kpatch_elf *kelf)
+{
+	struct symbol *sym;
+	int i = 0, sym_num = 0;
+	struct sym_compare_type *sym_array;
+
+	list_for_each_entry(sym, &kelf->symbols, list) {
+		if (sym->bind != STB_LOCAL)
+			continue;
+		if (sym->type != STT_FUNC && sym->type != STT_OBJECT)
+			continue;
+
+		sym_num++;
+	}
+
+	if (!sym_num)
+		return NULL;
+
+	sym_array = malloc((sym_num + 1) * sizeof(struct sym_compare_type));
+	if (!sym_array)
+		ERROR("malloc");
+
+	list_for_each_entry(sym, &kelf->symbols, list) {
+		if (sym->bind != STB_LOCAL)
+			continue;
+		if (sym->type != STT_FUNC && sym->type != STT_OBJECT)
+			continue;
+
+		sym_array[i].type = sym->type;
+		sym_array[i++].name = sym->name;
+	}
+	sym_array[i].type = 0;
+	sym_array[i].name = NULL;
+
+	return sym_array;
+}
+
 static void kpatch_create_patches_sections(struct kpatch_elf *kelf,
 					   struct lookup_table *table, char *hint,
 					   char *objname)
@@ -1872,7 +1909,7 @@ static void kpatch_create_patches_sections(struct kpatch_elf *kelf,
 		if (sym->type == STT_FUNC && sym->status == CHANGED) {
 			if (sym->bind == STB_LOCAL) {
 				if (lookup_local_symbol(table, sym->name,
-				                        hint, &result))
+				                        &result))
 					ERROR("lookup_local_symbol %s (%s)",
 					      sym->name, hint);
 			} else {
@@ -2034,11 +2071,8 @@ static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
 			if (rela->sym->bind == STB_LOCAL) {
 				/* An unchanged local symbol */
 				ret = lookup_local_symbol(table,
-					rela->sym->name, hint, &result);
-				if (ret == 2)
-					ERROR("lookup_local_symbol: ambiguous %s:%s relocation, needed for %s",
-			               hint, rela->sym->name, sec->base->name);
-				else if (ret)
+					rela->sym->name, &result);
+				if (ret)
 					ERROR("lookup_local_symbol %s:%s needed for %s",
 			               hint, rela->sym->name, sec->base->name);
 
@@ -2417,6 +2451,7 @@ int main(int argc, char *argv[])
 	struct symbol *sym;
 	char *hint = NULL, *objname, *pos;
 	char *mod_symvers_path, *pmod_name;
+	struct sym_compare_type *base_locals;
 
 	arguments.debug = 0;
 	argp_parse (&argp, argc, argv, 0, NULL, &arguments);
@@ -2436,6 +2471,20 @@ int main(int argc, char *argv[])
 	kpatch_compare_elf_headers(kelf_base->elf, kelf_patched->elf);
 	kpatch_check_program_headers(kelf_base->elf);
 	kpatch_check_program_headers(kelf_patched->elf);
+
+	list_for_each_entry(sym, &kelf_base->symbols, list) {
+		if (sym->type == STT_FILE) {
+			hint = sym->name;
+			break;
+		}
+	}
+	if (!hint)
+		ERROR("FILE symbol not found in base. Stripped?\n");
+
+	/* create symbol lookup table */
+	base_locals = kpatch_elf_locals(kelf_base);
+	lookup = lookup_open(arguments.args[2], mod_symvers_path, hint, base_locals);
+	free(base_locals);
 
 	kpatch_mark_grouped_sections(kelf_patched);
 	kpatch_replace_sections_syms(kelf_base);
@@ -2491,18 +2540,6 @@ int main(int argc, char *argv[])
 	 * kpatch_patched.
 	 */
 	kpatch_elf_teardown(kelf_patched);
-
-	list_for_each_entry(sym, &kelf_out->symbols, list) {
-		if (sym->type == STT_FILE) {
-			hint = sym->name;
-			break;
-		}
-	}
-	if (!hint)
-		ERROR("FILE symbol not found in output. Stripped?\n");
-
-	/* create symbol lookup table */
-	lookup = lookup_open(arguments.args[2], mod_symvers_path);
 
 	/* extract module name (destructive to arguments.modulefile) */
 	objname = basename(arguments.args[2]);
--- a/kpatch-build/lookup.c
+++ b/kpatch-build/lookup.c
@@ -55,6 +55,7 @@ struct lookup_table {
 	int obj_nr, exp_nr;
 	struct object_symbol *obj_syms;
 	struct export_symbol *exp_syms;
+	struct object_symbol *local_syms;
 };
 
 #define for_each_obj_symbol(ndx, iter, table) \
@@ -62,6 +63,55 @@ struct lookup_table {
 
 #define for_each_exp_symbol(ndx, iter, table) \
 	for (ndx = 0, iter = table->exp_syms; ndx < table->exp_nr; ndx++, iter++)
+
+static void find_local_syms(struct lookup_table *table, char *hint,
+			    struct sym_compare_type *locals)
+{
+	struct object_symbol *sym, *file_sym;
+	int i, in_file = 0;
+	struct sym_compare_type *local_index;
+
+	for_each_obj_symbol(i, sym, table) {
+		if (sym->type == STT_FILE) {
+			if (in_file && !local_index->name) {
+				if (table->local_syms)
+					ERROR("find_local_syms for %s: found_dup", hint);
+				table->local_syms = file_sym;
+			}
+
+			if (!strcmp(hint, sym->name)) {
+				in_file = 1;
+				file_sym = sym;
+				local_index = locals;
+			}
+			else
+				in_file = 0;
+
+			continue;
+		}
+
+		if (!in_file)
+			continue;
+		if (sym->bind != STB_LOCAL || (sym->type != STT_FUNC && sym->type != STT_OBJECT))
+			continue;
+
+		if (local_index->name &&
+		    local_index->type == sym->type &&
+		    !strcmp(local_index->name, sym->name))
+			local_index++;
+		else
+			in_file = 0;
+	}
+
+	if (in_file && !local_index->name) {
+		if (table->local_syms)
+			ERROR("find_local_syms for %s: found_dup", hint);
+		table->local_syms = file_sym;
+	}
+
+	if (!table->local_syms)
+		ERROR("find_local_syms for %s: found_none", hint);
+}
 
 static void obj_read(struct lookup_table *table, char *path)
 {
@@ -203,7 +253,8 @@ static void symvers_read(struct lookup_table *table, char *path)
 	fclose(file);
 }
 
-struct lookup_table *lookup_open(char *obj_path, char *symvers_path)
+struct lookup_table *lookup_open(char *obj_path, char *symvers_path,
+				 char *hint, struct sym_compare_type *locals)
 {
 	struct lookup_table *table;
 
@@ -215,6 +266,10 @@ struct lookup_table *lookup_open(char *obj_path, char *symvers_path)
 	obj_read(table, obj_path);
 	symvers_read(table, symvers_path);
 
+	table->local_syms = NULL;
+	if (locals)
+		find_local_syms(table, hint, locals);
+
 	return table;
 }
 
@@ -225,40 +280,38 @@ void lookup_close(struct lookup_table *table)
 	free(table);
 }
 
-int lookup_local_symbol(struct lookup_table *table, char *name, char *hint,
+int lookup_local_symbol(struct lookup_table *table, char *name,
                         struct lookup_result *result)
 {
-	struct object_symbol *sym, *match = NULL;
-	int i;
+	struct object_symbol *sym;
 	unsigned long pos = 0;
-	char *curfile = NULL;
+	int i, match = 0, in_file = 0;
+
+	if (!table->local_syms)
+		return 1;
 
 	memset(result, 0, sizeof(*result));
 	for_each_obj_symbol(i, sym, table) {
-		if (sym->type == STT_FILE) {
-			if (!strcmp(sym->name, hint)) {
-				curfile = sym->name;
-				continue; /* begin hint file symbols */
-			} else if (curfile)
-				curfile = NULL; /* end hint file symbols */
+		if (sym->skip)
+			continue;
+
+		if (sym->bind == STB_LOCAL && !strcmp(sym->name, name))
+			pos++;
+
+		if (table->local_syms == sym) {
+			in_file = 1;
+			continue;
 		}
-		if (sym->bind == STB_LOCAL) {
-			if (sym->name && !strcmp(sym->name, name)) {
-				/*
-				 * need to count any occurrence of the symbol
-				 * name, unless we've already found a match
-				 */
-				if (!match)
-					pos++;
 
-				if (!curfile)
-					continue;
+		if (!in_file)
+			continue;
 
-				if (match)
-					/* dup file+symbol, unresolvable ambiguity */
-					return 2;
-				match = sym;
-			}
+		if (sym->type == STT_FILE)
+			break;
+
+		if (sym->bind == STB_LOCAL && !strcmp(sym->name, name)) {
+			match = 1;
+			break;
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


jpoimboe:
@zhouchengming1 I like the approach, but all the global variables make my head spin a little bit :-)  Any ideas about restructuring the code a bit to get rid of the globals?

zhouchengming1:
@jpoimboe  Thanks for your review. I restruct the code to get rid of the globals in lookup.c, so left the only global saved_kelf_base in create-diff-object.c to remember the kelf_base.

zhouchengming1:
Or you think i should change the interface of kpatch_create_patches_sections() and kpatch_create_dynamic_rela_sections() to get rid of the global saved_kelf_base too ?

jpoimboe:
@zhouchengming1 Yeah, I was hoping to get rid of saved_kelf_base too, if you can figure out a good way to do that.

jpoimboe:
@zhouchengming1 BTW, I just noticed the new pull request.  It would be better to re-use this one, so all the conversation is in a single place.  To do so, please force push to your original iss604 branch.

zhouchengming1:
@jpoimboe  Ok, I just modify the code and force push to the original iss604 branch. Thanks for your review.

kpatch-build/create-diff-object.c
```diff
@@ -2381,8 +2445,6 @@ int main(int argc, char *argv[])
 	kpatch_mark_ignored_sections(kelf_patched);
 	kpatch_compare_correlated_elements(kelf_patched);
 	kpatch_check_fentry_calls(kelf_patched);
-	kpatch_elf_teardown(kelf_base);
-	kpatch_elf_free(kelf_base);
```
jpoimboe: This base elf teardown code is at this location for a good reason: by ensuring the 'base' structures aren't used in the rest of the file, it keeps things nicely separated, which helps keep the code simpler and makes it easier to flush out bugs.  Which gives me an idea for how to rearrange the code in this patch.

Since the local symbols are part of the lookup table, maybe we could pass kelf_base->symbols to lookup_open() at the beginning of the program, after calling kpatch_elf_open().  lookup_open() could find the local syms and setup table->local_syms.  If it can't find a unique matching FILE+locals combination, it could error out.

(BTW, I wrote a python script to find such cases, and looked for duplicate FILE+locals combinations in the vmlinux symbol table for an allyesconfig kernel, and I only found 4 instances in an allyesconfig kernel, and they were in drivers that are normally not compiled into vmlinux.)

Then, later, lookup_local_symbol() could do its lookup based on table->local_syms.  Then there wouldn't be the concept of an "ambiguous" symbol, and so we would no longer need kpatch_resolve_ambiguous_symbol().

I think that would simplify this code quite a bit.  Thoughts?

zhouchengming1: You are right, we shouldn't change the placement of teardown of the base elf. This approch is bad.
So I will modify the code as you suggested.
Thanks.

zhouchengming1:
@jpoimboe Hi, I force push to my original iss604 branch again, modified as you suggested. Thanks for your review.

kpatch-build/lookup.c
```diff
@@ -48,12 +48,77 @@ struct lookup_table {
...
 	Elf *elf;
 	struct symbol *syms;
+	struct symbol *local_syms;
 };
 
 #define for_each_symbol(ndx, iter, table) \
 	for (ndx = 0, iter = table->syms; ndx < table->nr; ndx++, iter++)
 
-struct lookup_table *lookup_open(char *path)
+static void resolve_dup_file_symbol(struct lookup_table *table, char *hint,
```
jpoimboe: This function should have a better name to describe what it does, maybe find_local_syms()?

kpatch-build/lookup.c
```diff
@@ -48,12 +48,77 @@ struct lookup_table {
...
+					multi_base = 1;
+					in_file = 0;
+					break;
+				}
+				table->local_syms = prev;
+			}
+
+			if (!strcmp(hint, sym->name)) {
+				in_file = 1;
+				prev = sym;
```
jpoimboe: Can you rename 'prev' to something more descriptive like 'file_sym'?

kpatch-build/lookup.c
```diff
@@ -48,12 +48,77 @@ struct lookup_table {
...
+
+		if (!in_file)
+			continue;
+		if (sym->bind != STB_LOCAL || (sym->type != STT_FUNC && sym->type != STT_OBJECT))
+			continue;
+
+		for ( ; local_index->name; local_index++)
+			if (local_index->type == sym->type &&
+			    !strcmp(local_index->name, sym->name))
+				break;
```
jpoimboe: Why is there a loop here?  Shouldn't the symbols match up one-to-one between locals and sym?  For example, wouldn't something like this work instead of the for loop?
```
        if (local_index->type == sym->type && !strcmp(local_index->name, sym->name))
                local_index++;
        else
                in_file = 0;
   ```

File: kpatch-build/lookup.c [hunk 5]
```diff
@@ -48,12 +48,77 @@ struct lookup_table {
...
+	struct symbol *sym, *prev;
+	int i, in_file = 0;
+	int multi_base = 0;
+	struct sym_compare_type *local_index;
+
+	for_each_symbol(i, sym, table) {
+		if (sym->type == STT_FILE) {
+			if (in_file) {
+				if (table->local_syms) {
+					multi_base = 1;
```
jpoimboe: I found the multi_base variable confusing.  It might be clearer to either rename it to something like 'found_dup'.  Or, even better, get rid of it and instead use a goto to the error path, as that would make the code flow clearer IMO.

kpatch-build/lookup.c
```diff
@@ -48,12 +48,77 @@ struct lookup_table {
...
+		if (sym->bind != STB_LOCAL || (sym->type != STT_FUNC && sym->type != STT_OBJECT))
+			continue;
+
+		for ( ; local_index->name; local_index++)
+			if (local_index->type == sym->type &&
+			    !strcmp(local_index->name, sym->name))
+				break;
+
+		if (!local_index->name) {
+			printf("sym: %s can't be found, skip.\n", sym->name);
```
jpoimboe: No need to printf here.

kpatch-build/lookup.c
```diff
@@ -48,12 +48,77 @@ struct lookup_table {
...
+			table->local_syms = prev;
+	}
+
+	if (multi_base) {
+		table->local_syms = NULL;
+		printf("find multiple matching base.\n");
+	} else if (!table->local_syms)
+		printf("can't find matching base.\n");
+	else
+		printf("find unique maching base.\n");
```
jpoimboe: No need for all these printfs.  For the error cases -- when we can't find the local syms, or when a duplicate was found -- the ERROR macro can be used.

kpatch-build/lookup.c
zhouchengming1: Yes, I am not good at taking names. :)

kpatch-build/lookup.c
zhouchengming1: Ok, I will refactor the code to make it clearer. Thanks.

kpatch-build/lookup.c
zhouchengming1: Yes, error out for these two error cases maybe clearer. It makes sense to error out when we can't find the local syms. But if we error out when a duplicate was found, we can't patch global functions for these files too. (you found 4 drivers of such case)

kpatch-build/lookup.c
```diff
@@ -143,49 +212,45 @@ void lookup_close(struct lookup_table *table)
...
 
-	memset(result, 0, sizeof(*result));
 	for_each_symbol(i, sym, table) {
-		if (sym->type == STT_FILE) {
-			if (!strcmp(sym->name, hint)) {
-				curfile = sym->name;
-				continue; /* begin hint file symbols */
-			} else if (curfile)
-				curfile = NULL; /* end hint file symbols */
+		if (sym->bind == STB_LOCAL && !strcmp(sym->name, name))
```
zhouchengming1: I will include this fix. Thanks.

kpatch-build/lookup.c
jpoimboe: True, I did find 4 driver files which had duplicates, but that was out of thousands of .c files in the kernel.  Patching one of those files would be quite rare.  And they would *only* be unpatchable when compiled into vmlinux, and not when compiled as modules, which is probably the common case.

So I would say, consider duplicates to be an error for now.

If we ever got to the point in the future where we realistically wanted to patch one of those files, then we could set a "dup" boolean variable in the lookup table and then we could error out later if the patch tried to access local symbols when "dup" is set.  But I get the feeling that such a scenario is highly unlikely, and not worth worrying about.

kpatch-build/lookup.c
zhouchengming1: Yes, that's what I did before. But later we encountered a fail when find_local_syms().

Reproducer:
kernel: 3.10.34
gcc: gcc version 4.8.5 20150623 (Red Hat 4.8.5-4) (GCC)
file: fs/hugetlbfs/inode.c

local symbols in vmlinux of this inode.c :
 23718: ffffffff8126da80    16 FUNC    LOCAL  DEFAULT    1 hugetlbfs_write_begin
 23719: ffffffff8126da90   251 FUNC    LOCAL  DEFAULT    1 hugetlbfs_file_mmap
 23720: ffffffff8126db90    44 FUNC    LOCAL  DEFAULT    1 hugetlbfs_set_page_dirty
 23721: ffffffff8126dbc0   263 FUNC    LOCAL  DEFAULT    1 hugetlbfs_read_actor
 23722: ffffffff8126dcd0   480 FUNC    LOCAL  DEFAULT    1 hugetlbfs_read
 23723: ffffffff8126deb0   189 FUNC    LOCAL  DEFAULT    1 hugetlbfs_statfs
 23724: ffffffff8126df70    11 FUNC    LOCAL  DEFAULT    1 hugetlbfs_write_end
 23725: ffffffff8126df80    57 FUNC    LOCAL  DEFAULT    1 hugetlbfs_migrate_page
 23726: ffffffff81c2fec9   114 FUNC    LOCAL  DEFAULT   26 exit_hugetlbfs_fs
 23727: ffffffff81f2f950    16 OBJECT  LOCAL  DEFAULT   29 hugetlbfs_vfsmount
 23728: ffffffff81f2f960     8 OBJECT  LOCAL  DEFAULT   29 hugetlbfs_inode_cachep
 23729: ffffffff819fe700    56 OBJECT  LOCAL  DEFAULT   15 hugetlbfs_fs_type
 23730: ffffffff819fe740   680 OBJECT  LOCAL  DEFAULT   15 hugetlbfs_backing_dev_inf
 23731: ffffffff8126dfc0    26 FUNC    LOCAL  DEFAULT    1 hugetlbfs_mount
 23732: ffffffff8126e100  1170 FUNC    LOCAL  DEFAULT    1 hugetlbfs_fill_super
 23733: ffffffff8126dfe0    58 FUNC    LOCAL  DEFAULT    1 hugetlbfs_put_super
 23734: ffffffff8126e020    30 FUNC    LOCAL  DEFAULT    1 hugetlbfs_i_callback
 23735: ffffffff8126e040    20 FUNC    LOCAL  DEFAULT    1 init_once
 23736: ffffffff8126e060   158 FUNC    LOCAL  DEFAULT    1 hugetlbfs_alloc_inode
 23737: ffffffff81ae8546   368 FUNC    LOCAL  DEFAULT   18 init_hugetlbfs_fs
 23738: ffffffff8169db40   112 OBJECT  LOCAL  DEFAULT    4 tokens
 23739: ffffffff8169dee0   176 OBJECT  LOCAL  DEFAULT    4 hugetlbfs_ops
 23740: ffffffff8169dc80   192 OBJECT  LOCAL  DEFAULT    4 hugetlbfs_dir_inode_opera
 23741: ffffffff8126e5a0   581 FUNC    LOCAL  DEFAULT    1 hugetlb_get_unmapped_area
 23742: ffffffff8126e7f0   404 FUNC    LOCAL  DEFAULT    1 hugetlbfs_get_inode
 23743: ffffffff8169de20   168 OBJECT  LOCAL  DEFAULT    4 hugetlbfs_aops
 23744: ffffffff8169dbc0   192 OBJECT  LOCAL  DEFAULT    4 hugetlbfs_inode_operation
 23745: ffffffff8126e990   119 FUNC    LOCAL  DEFAULT    1 hugetlbfs_mknod
 23746: ffffffff8126ea10    55 FUNC    LOCAL  DEFAULT    1 hugetlbfs_mkdir
 23747: ffffffff8126ea50    26 FUNC    LOCAL  DEFAULT    1 hugetlbfs_create
 23748: ffffffff8126ea70   168 FUNC    LOCAL  DEFAULT    1 hugetlbfs_symlink
 23749: ffffffff8126eb20   482 FUNC    LOCAL  DEFAULT    1 truncate_hugepages
 23750: ffffffff8126ed10   347 FUNC    LOCAL  DEFAULT    1 hugetlbfs_setattr
 23751: ffffffff8126ee70    52 FUNC    LOCAL  DEFAULT    1 hugetlbfs_evict_inode
 23752: ffffffff8126eeb0    98 FUNC    LOCAL  DEFAULT    1 hugetlbfs_destroy_inode
 23753: ffffffff819fe680   128 OBJECT  LOCAL  DEFAULT   15 anon_ops
 23754: ffffffff81f2f944     1 OBJECT  LOCAL  DEFAULT   29 __print_once.30939
 23755: ffffffff81c18d58     8 OBJECT  LOCAL  DEFAULT   19 __initcall_init_hugetlbfs

local symbols of inode.o :
     5: 0000000000000000    16 FUNC    LOCAL  DEFAULT    1 hugetlbfs_write_begin
     6: 0000000000000010   251 FUNC    LOCAL  DEFAULT    1 hugetlbfs_file_mmap
     7: 0000000000000110    44 FUNC    LOCAL  DEFAULT    1 hugetlbfs_set_page_dirty
     9: 0000000000000140   263 FUNC    LOCAL  DEFAULT    1 hugetlbfs_read_actor
    13: 0000000000000250   480 FUNC    LOCAL  DEFAULT    1 hugetlbfs_read
    14: 0000000000000430   189 FUNC    LOCAL  DEFAULT    1 hugetlbfs_statfs
    16: 00000000000004f0    11 FUNC    LOCAL  DEFAULT    1 hugetlbfs_write_end
    18: 0000000000000500    57 FUNC    LOCAL  DEFAULT    1 hugetlbfs_migrate_page
    20: 0000000000000000   114 FUNC    LOCAL  DEFAULT   16 exit_hugetlbfs_fs
    21: 0000000000000010    16 OBJECT  LOCAL  DEFAULT    5 hugetlbfs_vfsmount
    22: 0000000000000020     8 OBJECT  LOCAL  DEFAULT    5 hugetlbfs_inode_cachep
    23: 0000000000000080    56 OBJECT  LOCAL  DEFAULT    3 hugetlbfs_fs_type
    24: 00000000000000c0   680 OBJECT  LOCAL  DEFAULT    3 hugetlbfs_backing_dev_inf
    25: 0000000000000540    26 FUNC    LOCAL  DEFAULT    1 hugetlbfs_mount
    26: 0000000000000680  1170 FUNC    LOCAL  DEFAULT    1 hugetlbfs_fill_super
    27: 0000000000000560    58 FUNC    LOCAL  DEFAULT    1 hugetlbfs_put_super
    28: 00000000000005a0    30 FUNC    LOCAL  DEFAULT    1 hugetlbfs_i_callback
    29: 00000000000005c0    20 FUNC    LOCAL  DEFAULT    1 init_once
    30: 00000000000005e0   158 FUNC    LOCAL  DEFAULT    1 hugetlbfs_alloc_inode
    33: 0000000000000000   368 FUNC    LOCAL  DEFAULT   19 init_hugetlbfs_fs
    34: 0000000000000040   112 OBJECT  LOCAL  DEFAULT   21 tokens
    36: 00000000000003e0   176 OBJECT  LOCAL  DEFAULT   21 hugetlbfs_ops
    37: 0000000000000180   192 OBJECT  LOCAL  DEFAULT   21 hugetlbfs_dir_inode_opera
    38: 0000000000000b20   581 FUNC    LOCAL  DEFAULT    1 hugetlb_get_unmapped_area
    39: 0000000000000d70   404 FUNC    LOCAL  DEFAULT    1 hugetlbfs_get_inode
    40: 0000000000000320   168 OBJECT  LOCAL  DEFAULT   21 hugetlbfs_aops
    41: 00000000000000c0   192 OBJECT  LOCAL  DEFAULT   21 hugetlbfs_inode_operation
    42: 0000000000000f10   119 FUNC    LOCAL  DEFAULT    1 hugetlbfs_mknod
    43: 0000000000000f90    55 FUNC    LOCAL  DEFAULT    1 hugetlbfs_mkdir
    44: 0000000000000fd0    26 FUNC    LOCAL  DEFAULT    1 hugetlbfs_create
    45: 0000000000000ff0   168 FUNC    LOCAL  DEFAULT    1 hugetlbfs_symlink
    46: 00000000000010a0   482 FUNC    LOCAL  DEFAULT    1 truncate_hugepages
    47: 0000000000001290   347 FUNC    LOCAL  DEFAULT    1 hugetlbfs_setattr
    48: 00000000000013f0    52 FUNC    LOCAL  DEFAULT    1 hugetlbfs_evict_inode
    49: 0000000000001430    98 FUNC    LOCAL  DEFAULT    1 hugetlbfs_destroy_inode
    50: 0000000000000000   128 OBJECT  LOCAL  DEFAULT    3 anon_ops
    51: 0000000000000004     1 OBJECT  LOCAL  DEFAULT    5 __print_once.30939
    53: 0000000000000000     8 OBJECT  LOCAL  DEFAULT   23 __exitcall_exit_hugetlbfs
    55: 0000000000000000     8 OBJECT  LOCAL  DEFAULT   25 __initcall_init_hugetlbfs

We can see there is a difference: inode.o includes '__exitcall_exit_hugetlbfs' , but vmlinux doesn't.
So the one-to-one match will fail. I don't know why the Linker drop this local symbol from inode.o.

zhouchengming1: I just used the same gcc, test on kernel-4.9 , found inode.o also didn't have local symbol '__exitcall_exit_hugetlbfs', so one-to-one match will succeed on kernel-4.9.

zhouchengming1:
@jpoimboe  Ok, I modified the code again and rebase on the new master branch. Thanks for your review and suggestions.

kpatch-build/lookup.c
```diff
@@ -65,6 +66,58 @@ struct lookup_table {
...
+static void find_local_syms(struct lookup_table *table, char *hint,
+			    struct sym_compare_type *locals)
+{
+	struct object_symbol *sym, *file_sym;
+	int i, in_file = 0;
+	struct sym_compare_type *local_index;
+
+	for_each_obj_symbol(i, sym, table) {
+		if (sym->type == STT_FILE) {
+			if (in_file) {
```
jpoimboe: I found a bug here, when trying to patch fs/proc/cmdline.c. It gives the "find_local_syms found_dup" error, because there's another cmdline.c which has no local symbols. It treats an empty file as a match.

It should instead make sure that local_index->name is zero before checking the error and/or assigning table->local_syms.  So I think this line should be `if (in_file && !local_index->name) {` and a similar check below at line 108.

zhouchengming1: Yes, it's a bug... thanks.

kpatch-build/lookup.c
```diff
@@ -65,6 +66,58 @@ struct lookup_table {
...
+				if (table->local_syms)
+					goto error;
```
zhouchengming1: Yeah, either way is fine :)

zhouchengming1:
@jpoimboe   hello, I modified the code again, thanks again for your review.

jpoimboe:
👍

PR #652, opened 2016-12-23
compile original module for create-diff-object

### LINKED ISSUE #612: Binary patch for CVE-2016-3134 fails to load (CentOS 7, kernel 3.10.0-327.4.4)
I encounered the following problem while experimenting with Kpatch on CentOS 7 with kernel 3.10.0-327.4.4.

kpatch-build reports no errors when building a binary patch from the fix for CVE-2016-3134,
[netfilter-fix-unconditional-helper.patch](https://github.com/dynup/kpatch/files/536079/netfilter-fix-unconditional-helper.patch.txt), but the resulting binary patch fails to load with the following errors in dmesg:

<pre>
kpatch: symbol 'uncond.48011' not found in symbol table
kpatch: unable to find symbol 'uncond.48011'
</pre>

The modules to be patched have similar symbols but with the different numeric suffixes added by GCC:

<pre>
# grep -F uncond. /proc/kallsyms 
ffffffffa03914a0 b uncond.48012 [ip6_tables]
ffffffffa00c8580 b uncond.48784 [ip_tables]
</pre>

It seems that either these static locals were not correlated properly by create-diff-object or they cannot be correlated at all. If the latter is the case, create-diff-object should report an error like it usually does in such cases.

Another question is, why GCC gave these variables the different numeric suffixes compared to what it did when building the kernel packages for CentOS. But this is beyond the scope of this issue, perhaps. I use the same GCC version as was used for the kernel packages, GCC 4.8.3 20140911 (Red Hat 4.8.3-9).

Interestingly enough, the same binary patch builds and loads just fine on the "neighbor" kernel versions, 3.10.0-327.3.1 and 3.10.0-327.4.5.

zhouchengming1:
Normally, the symbols in the freshly built object file that
create-diff-object used are the same with our original module.

But the name of the symbols may change if use --skip-gcc-check
or the influence of KCFLAGS. This patch compile original module
for create-diff-object, like original vmlinux, to detect this
problem earlier, before we need to insmod klp.ko.

Fixes #612.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -532,6 +532,7 @@ do
 	cp -f "$OBJDIR/$i" "$TEMPDIR/patched/$i" || die
 done
 
+unset KCFLAGS
 echo "Extracting new and modified ELF sections"
 FILES="$(cat "$TEMPDIR/changed_objs")"
 cd "$TEMPDIR"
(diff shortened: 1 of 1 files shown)


jpoimboe:
Thanks @zhouchengming1 but I don't think this is the right fix.  I'll explain on #612.

PR #653, opened 2017-01-10
Small kpatch-test fixups

joe-lawrence:
This is a set of really small fixes for kpatch-test:

* **testing: handle empty glob cases** - found when running ```kpatch-test``` against an empty patch directory.  I could have modified the globbing behavior via ```shopt -s failglob```, but I thought that was a more esoteric fix (and changes default behavior for the rest of the script).

* **testing: skip gcc check** - installed gcc vs kernel build gcc can drift on Vagrant VM boxes.  We could alternatively add a command line switch to ```kpatch-test``` to optionally pass this option along.

* **testing: skip combined test for a single patch** - this cleans up the test output a bit.

* **testing: add full path for /sbin executables** - as seen on a Fedora Vagrant VM box when running a command via ```ssh ... <command>```.  I verified the user PATH with:
```
% ssh vagrant@192.168.121.4 echo \$PATH
/usr/lib64/ccache:/usr/local/bin:/usr/bin
```
Note that ```/etc/sudoers``` on most distributions includes a line like ```/sbin:/bin:/usr/sbin:/usr/bin```, so any commands that ```kpatch-test``` is invoking via ```sudo``` should be fine.

### DIFF
--- a/test/integration/kpatch-test
+++ b/test/integration/kpatch-test
@@ -267,6 +267,10 @@ build_combined_module() {
 }
 
 run_combined_test() {
+	if [[ ! -e COMBINED.patch ]]; then
+		return
+	fi
+
 	if [[ ! -e kpatch-COMBINED.ko ]]; then
 		log "can't find kpatch-COMBINED.ko, skipping"
 		return
--- a/test/integration/kpatch-test
+++ b/test/integration/kpatch-test
@@ -40,6 +40,8 @@
 # built from the *.patch files.  They can be used for more custom tests above
 # and beyond the simple loading and unloading tests.
 
+shopt -s nullglob
+
 SCRIPTDIR="$(readlink -f $(dirname $(type -p $0)))"
 ROOTDIR="$(readlink -f $SCRIPTDIR/../..)"
 # TODO: option to use system-installed binaries instead
@@ -95,6 +97,10 @@ done
 if [[ ${#PATCH_LIST[@]} = 0 ]]; then
 	PATCH_LIST=($PATCHDIR/*.patch)
 	TEST_LIST=($PATCHDIR/*.test)
+	if [[ ${#PATCH_LIST[@]} = 0 ]]; then
+		echo "No patches found!"
+		exit 1
+	fi
 else
 	for file in "${PATCH_LIST[@]}"; do
 		prefix=${file%%.patch}
--- a/test/integration/kpatch-test
+++ b/test/integration/kpatch-test
@@ -120,12 +120,12 @@ log() {
 }
 
 unload_all() {
-	for i in `lsmod |egrep '^kpatch' |awk '{print $1}'`; do
+	for i in `/sbin/lsmod |egrep '^kpatch' |awk '{print $1}'`; do
 		if [[ $i != kpatch ]]; then
 			$KPATCH unload $i >> $LOG 2>&1 || error "\"kpatch unload $i\" failed"
 		fi
 	done
-	if lsmod |egrep -q '^kpatch'; then
+	if /sbin/lsmod |egrep -q '^kpatch'; then
 		$RMMOD kpatch >> $LOG 2>&1 || error "\"rmmod kpatch\" failed"
 	fi
 }


jpoimboe:
> testing: handle empty glob cases - found when running kpatch-test against an empty patch directory. I could have modified the globbing behavior via shopt -s failglob, but I thought that was a more esoteric fix (and changes default behavior for the rest of the script).

I think I'd prefer that instead because the compgen stuff isn't very readable (for me, at least), and we already rely on bash, and I doubt the rest of the script cares anyway.  Or alternatively you could do `shopt -s nullgob`.

> testing: skip gcc check - installed gcc vs kernel build gcc can drift on Vagrant VM boxes. We could alternatively add a command line switch to kpatch-test to optionally pass this option along.

Why not update the VM with the right version of gcc?  They shouldn't need to change once they're set up for integration testing.

joe-lawrence:
>I think I'd prefer that instead because the compgen stuff isn't very readable (for me, at least), and we already rely on bash, and I doubt the rest of the script cares anyway. Or alternatively you could do shopt -s nullgob.

I'll test out ```shopt -s nullgob```.  (My feeling is that it may obfuscate some future bug, but you're definitely right about improving readability.)

>Why not update the VM with the right version of gcc? They shouldn't need to change once they're set up for integration testing.

What is the "right version" of gcc?  For kpatch, it wants the version that built the kernel.  For an off-the-shelf VM or cloud image, it's usually whatever happens to be the latest that the package manger installs at the time of VM image creation.  At the moment (```dnf update``` this morning), Fedora 24 ships gcc "6.3.1 20161221 (Red Hat 6.3.1-1)", but kernel 4.8.6-300.fc25.x86_64 is built with gcc "6.2.1 20160916 (Red Hat 6.2.1-2)".

I guess this leads to another testing question, are we interested in verifying against a particular OS release as it was GA'd (presumably gcc == kernel gcc), or a more current, updated state?  If it is the former, then perhaps the gcc version assumption holds.

jpoimboe:
> What is the "right version" of gcc? For kpatch, it wants the version that built the kernel. For an off-the-shelf VM or cloud image, it's usually whatever happens to be the latest that the package manger installs at the time of VM image creation. At the moment (dnf update this morning), Fedora 24 ships gcc "6.3.1 20161221 (Red Hat 6.3.1-1)", but kernel 4.8.6-300.fc25.x86_64 is built with gcc "6.2.1 20160916 (Red Hat 6.2.1-2)".

The "right version" is the version that built the kernel.  Otherwise it could (in theory) introduce subtle bugs with kpatch.  Even when starting from an off the shelf VM, vagrant and ansible allow you to upgrade (or downgrade) the gcc package to a specific version, no?

> I guess this leads to another testing question, are we interested in verifying against a particular OS release as it was GA'd (presumably gcc == kernel gcc), or a more current, updated state? If it is the former, then perhaps the gcc version assumption holds.

Ideally it would be something close to the GA version.  But the most important thing is that, once setup, the VM doesn't change, so our tests have a stable base.

One exception: we might also want to have a special "rolling" VM which tests for upstream kernel changes.  In that case, it could change often, e.g. every 3 months for a new Linux release.  But even then the kernel would be self-built so we wouldn't have a gcc version mismatch.

joe-lawrence:
> Even when starting from an off the shelf VM, vagrant and ansible allow you to upgrade (or downgrade) the gcc package to a specific version, no?

Yes, a provisioning or post-install script could inspect vmlinux and then attempt to adjust the gcc package version accordingly.

> Ideally it would be something close to the GA version.

Official hashicorp Vagrant box images are periodically updated, and the expectation is that they contain current, updated package versions.  According to the [documentation](https://www.vagrantup.com/docs/boxes/versioning.html#automatic-update-checking), Vagrant will check for updates by default.  However the actual update is manual.  I don't know if there are any official GA-versioned boxes out there... so this might be homebrew territory.

> But the most important thing is that, once setup, the VM doesn't change, so our tests have a stable base.  One exception: we might also want to have a special "rolling" VM which tests for upstream kernel changes.

Would "rolling" distro VMs be interesting as well?  We can't test every version combination, but it might be prudent to know when a distro update kernel (or gcc) breaks something.


To bring this back on topic, for v2 I'll modify + test with ```shopt -s nullgob``` and drop the ```--skip-gcc-check```

jpoimboe:
> Official hashicorp Vagrant box images are periodically updated, and the expectation is that they contain current, updated package versions. According to the documentation, Vagrant will check for updates by default. However the actual update is manual. I don't know if there are any official GA-versioned boxes out there... so this might be homebrew territory.

I think homebrew is fine, there are tools to help with that, like virt-builder.

> Would "rolling" distro VMs be interesting as well? We can't test every version combination, but it might be prudent to know when a distro update kernel (or gcc) breaks something.

I don't know, it's seems pretty rare that we run into distro-specific issues.  The vast majority of new issues seem to be related to upstream kernel changes.  I get the feeling it would be good enough to stick to major distro releases.

joe-lawrence:
v2 - use ```shopt -s nullgob``` and drop the ```--skip-gcc-check``` patch
testing: ```$PWD``` containing no patches, one patch, then multiple patches.  Also verified no ```vagrant ssh ... -c``` path problems.

jpoimboe:
👍

PR #654, opened 2017-01-12
kmod: let kernel apply TAINT_LIVEPATCH

joe-lawrence:
Upstream commit [2992ef29ae01](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=2992ef29ae01) ("livepatch/module: make TAINT_LIVEPATCH module-specific") v4.9+ modified the kernel to add the TAINT_LIVEPATCH flag on module load.  To support this feature, add the "livepatch" module info in the {k,live}patch modules and drop the add_taint() in the core module.

Testing: ran on a RHEL7 kernel with and without a [2992ef29ae01](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=2992ef29ae01) backport and observed the expected results with the backport: ```cat /sys/module/<kpatch_module>/taint``` = ```OEK```.

I did not see this kernel feature backported yet to Ubuntu 16 LTS or Fedora 24, so no kernel version checks are needed for them.

### DIFF
--- a/kmod/core/core.c
+++ b/kmod/core/core.c
@@ -1008,9 +1008,22 @@ int kpatch_register(struct kpatch_module *kpmod, bool replace)
 		func->op = KPATCH_OP_NONE;
 	} while_for_each_linked_func();
 
+/* HAS_MODULE_TAINT - upstream 2992ef29ae01 "livepatch/module: make TAINT_LIVEPATCH module-specific" */
+#ifdef RHEL_RELEASE_CODE
+# if RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(7, 4)
+#  define HAS_MODULE_TAINT
+# endif
+#elif LINUX_VERSION_CODE >= KERNEL_VERSION(4, 9, 0)
+# define HAS_MODULE_TAINT
+#endif
+
 #ifdef TAINT_LIVEPATCH
+# ifdef HAS_MODULE_TAINT
+	/* kernel will add TAINT_LIVEPATCH on module load. */
+# else
 	pr_notice_once("tainting kernel with TAINT_LIVEPATCH\n");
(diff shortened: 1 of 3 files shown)


jpoimboe:
:+1:

kmod/core/core.c
```diff
@@ -1008,9 +1008,22 @@ int kpatch_register(struct kpatch_module *kpmod, bool replace)
 		func->op = KPATCH_OP_NONE;
 	} while_for_each_linked_func();
 
+/* HAS_KERNEL_TAINT - upstream 2992ef29ae01 "livepatch/module: make TAINT_LIVEPATCH module-specific" */
+#ifdef RHEL_RELEASE_CODE
+# if RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(7, 4)
+#  define HAS_KERNEL_TAINT
```
flaming-toast: Not to be picky, but can you change this to HAS_MODULE_TAINT instead? Since with HAS_KERNEL_TAINT I don't immediately understand that we're actually talking about the new per-module taint referenced in that commit.

joe-lawrence: I couldn't think of a really good name, so I thought ```HAS_KERNEL_TAINT``` would imply that the kernel was the actor applying the taint flag, not the module itself.  If ```HAS_MODULE_TAINT``` is easier to remember (ie, module _specific_ taint), then I can certainly change it.

jpoimboe: I could go either way, at this very moment I slightly prefer HAS_MODULE_TAINT :-)

joe-lawrence:
v2 - ```%s/HAS_KERNEL_TAINT/HAS_MODULE_TAINT/g```

flaming-toast:
Thanks @joe-lawrence!

PR #655, opened 2017-01-13
support dup file+symbol

### LINKED ISSUE #604: Resolving duplicate local functions

zhouchengming1:
This patch needs some explanation. When we want to resolve a dup
file+symbol symbol, we first find out our base object in lookup table,
then we can use this information to resolve ambiguous local symbols.

We use resolve_dup_file_symbol to find out the correct base object in
lookup table. If all local symbols of a file can be found in the
kelf_base, we think it's a matched file.

Fixes #604.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -62,6 +62,8 @@ char *childobj;
 
 enum loglevel loglevel = NORMAL;
 
+static struct kpatch_elf *saved_kelf_base;
+
 /*******************
  * Data structures
  * ****************/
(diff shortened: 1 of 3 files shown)


jpoimboe:
Please force push to your original branch so we can reuse #651.  Thanks!

Issue #81, opened 2014-03-23
Unified patch management interface required


mhiramat:
Current simple module load/unload interface has some problems.
- There is no way to describe/manage the dependency among patches. Wish to have a way to describe patch dependency (e.g. patch B requires the kernel with patch A). The kernel version dependency can be solved with module versioning.
- To avoid a kernel panic when failing to unpatch, patching/unpatching should be done via sysfs or ioctl which can return errors. module unloading is very critical one. We can lock the patch modules with increment refcount from kpatch module until the unpatching was done successfully.
- We have no way to check or list-up what functions are already patched, and which patch module has the patched function.

PR #656, opened 2017-01-19
create dynamic relocs for changed functions in self object

zhouchengming1:
Currently, we only create dynamic relocs for changed functions of
other objects, but not this object. It will cause a problem like:

original:	funcA and funcB (funcA calls funcB)
patch-1:	funcA-1 and funcB-1 (funcA-1 calls funcB-1)
patch-2:	funcB-2 (funcA-1 should call new funcB-2, but it calls funB-1)

As we don't create dynrela for funcB-1 in funcA-1, it will call
funcB-1 directly (not by jumping from funcB). So the new funcB-2
will not get called by funcA-1.
This patch will create dynamic relocs for all the normal relas of
changed functions, including changed ones in this object.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1918,6 +1918,7 @@ static void kpatch_create_dynamic_rela_sections(struct kpatch_elf *kelf,
 	struct lookup_result result;
 	struct kpatch_patch_dynrela *dynrelas;
 	int vmlinux, external, ret;
+	int not_strip;
 
 	vmlinux = !strcmp(objname, "vmlinux");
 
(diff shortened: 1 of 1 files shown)


jpoimboe:
@zhouchengming1 To prevent this type of problem, we recommend that you use cumulative updates.  In this case, patch-2 should be a superset of patch-1.  So patch-2 should have funcB-2 and funcA-2 (which is a duplicate of funcA-1).

If you don't use cumulative updates, there are other subtle problems you could have, which kpatch-build would not be able to detect.  See #81 for more discussion about this.

zhouchengming1:
@jpoimboe  Thank you, I learned a lot from #81 . 

>So there are at least two possible solutions:

>Only allow one patch module to be loaded at any time. If the patch module needs to be updated with a >replacement, then we can atomically unload the old one and load the new one.

Maybe this approch is the best, always let the newest patch module in effective.

>Or we could try to put some more intelligence in kpatch-build to try to detect dependencies between >patch modules and always do the right thing. I haven't really given much thought to the feasibility of this >option yet.

This way also can do, if we let kpatch-build know what functions have been patched.
eg. first load patch1 (func-a, func-b), then when kpatch-build create patch2, make sure func-a and func-b are included into patch2-module, even if they are unchanged. Since we know what functions changed from patch1.ko itself, so this method is not hard too.

At last, I think method 1 is better, can solve my problem too.  Thanks.

PR #658, opened 2017-01-20
rename changed vmlinux exported functions

zhouchengming1:
Some vmlinux exported functions are used in kmod/patch/kpatch-patch-hook.c
or kmod/patch/livepatch-patch-hook.c, if we patch these functions,
klp.ko will include them, so load_module will find these new functions
instead of those in kernel.

But the problem is that these new functions need to be dynamic relocated
by livepatch or kpatch-core first, so they are not ready when be called,
will cause kernel crash.

So we rename these vmlinux exported functions in kpatch-build, then
load_module can't find them when do normal relocations for klp.ko.


jpoimboe:
Thanks for reporting this.  I found another way to fix it which is less invasive, see #661.

PR #663, opened 2017-01-25
lookup: use log.h version of ERROR macro

jpoimboe:
Use the log.h version of the ERROR macro so the childobj gets printed.


joe-lawrence:
:+1:

Issue #657, opened 2017-01-19
detect __might_sleep()-related line number changes with CONFIG_DEBUG_ATOMIC_SLEEP


jpoimboe:
Similar to our WARN change detection, we also need detection for __might_sleep() calls when CONFIG_DEBUG_ATOMIC_SLEEP is enabled.  I saw this when making a patch which added a line to poll_schedule_timeout().

Before:
```
   50f:   be f7 01 00 00          mov    $0x1f7,%esi                                                                                                                                                               
   514:   48 c7 c7 00 00 00 00    mov    $0x0,%rdi                                                       
                          517: R_X86_64_32S       .rodata.do_select.str1.8+0x98                          
   51b:   e8 00 00 00 00          callq  520 <do_select+0x520>                                           
                          51c: R_X86_64_PC32      ___might_sleep-0x4                  
```
After:
```
 50f:   be fb 01 00 00          mov    $0x1fb,%esi
 514:   48 c7 c7 00 00 00 00    mov    $0x0,%rdi                                                         
                        517: R_X86_64_32S       .rodata.do_select.str1.8+0x98                            
 51b:   e8 00 00 00 00          callq  520 <do_select+0x520>                                             
                        51c: R_X86_64_PC32      ___might_sleep-0x4  
```
The only change was the line number provided as an argument to __might_sleep().  It would be nice for kpatch-build to ignore such changes.

joe-lawrence:
Trying to reproduce this and kpatch-build succeeds, despite the line number discrepancy: 

```
% uname -r
4.10.0-rc6+

% grep CONFIG_DEBUG_ATOMIC_SLEEP ~/linux/.config 
CONFIG_DEBUG_ATOMIC_SLEEP=y

% cd ~/kpatch
% git show
commit 8e1aef2893e651fc369645a7966f00182692050d
Merge: 471f1fd 586feb4
Author: Jessica Yu <jeyu@redhat.com>
Date:   Mon Jan 30 09:57:04 2017 -0800

    Merge pull request #659 from joe-lawrence/4.9-unwinder
    
    RFC - 4.9 unwinder

% cat poll_schedule_timeout.patch 
diff -Nupr src.orig/fs/select.c src/fs/select.c
--- src.orig/fs/select.c        2017-02-01 17:08:12.496686370 -0500
+++ src/fs/select.c     2017-02-01 17:08:53.348686370 -0500
@@ -237,9 +237,12 @@ int poll_schedule_timeout(struct poll_wq
[...]

PR #669, opened 2017-02-07
detect and ignore __might_sleep line number changes

joe-lawrence:
When CONFIG_DEBUG_ATOMIC_SLEEP is enabled, __might_sleep calls will add
the line number to the instruction stream.  Detect and revert any such
changes.

Fixes: #657.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -226,13 +226,13 @@ out:
 }
 
 /*
- * Determine if a section has changed only due to a WARN* macro call's
- * embedding of the line number into an instruction operand.
+ * Determine if a section has changed only due to a WARN* or might_sleep
+ * macro call's embedding of the line number into an instruction operand.
  *
  * Warning: Hackery lies herein.  It's hopefully justified by the fact that
  * this issue is very common.
  *
- * Example:
+ * Example WARN*:
  *
  *  938:   be 70 00 00 00          mov    $0x70,%esi
  *  93d:   48 c7 c7 00 00 00 00    mov    $0x0,%rdi
@@ -242,18 +242,26 @@ out:
  *  94b:   e8 00 00 00 00          callq  950 <tcp_conn_request+0x950>
  *                         94c: R_X86_64_PC32      warn_slowpath_null-0x4
  *
+ * Example might_sleep:
+ *
+ *  50f:   be f7 01 00 00          mov    $0x1f7,%esi
+ *  514:   48 c7 c7 00 00 00 00    mov    $0x0,%rdi
+ *                         517: R_X86_64_32S       .rodata.do_select.str1.8+0x98
+ *  51b:   e8 00 00 00 00          callq  520 <do_select+0x520>
+ *                         51c: R_X86_64_PC32      ___might_sleep-0x4
+ *
  * The pattern which applies to all cases:
  * 1) immediate move of the line number to %esi
  * 2) (optional) string section rela
  * 3) (optional) __warned.xxxxx static local rela
- * 4) warn_slowpath_* rela
+ * 4) warn_slowpath_* or __might_sleep or ___might_sleep rela
  */
-static int kpatch_warn_only_change(struct section *sec)
+static int kpatch_line_macro_change_only(struct section *sec)
 {
 	struct insn insn1, insn2;
 	unsigned long start1, start2, size, offset, length;
 	struct rela *rela;
-	int warnonly = 0, found;
+	int lineonly = 0, found;
 
 	if (sec->status != CHANGED ||
 	    is_rela_section(sec) ||
@@ -292,7 +300,7 @@ static int kpatch_warn_only_change(struct section *sec)
 
 		/*
 		 * Verify zero or more string relas followed by a
-		 * warn_slowpath_* rela.
+		 * warn_slowpath_* or __might_sleep or ___might_sleep rela.
 		 */
 		found = 0;
 		list_for_each_entry(rela, &sec->rela->relas, list) {
@@ -302,7 +310,9 @@ static int kpatch_warn_only_change(struct section *sec)
 				continue;
 			if (!strncmp(rela->sym->name, "__warned.", 9))
 				continue;
-			if (!strncmp(rela->sym->name, "warn_slowpath_", 14)) {
+			if (!strncmp(rela->sym->name, "warn_slowpath_", 14) ||
+			   (!strcmp(rela->sym->name, "__might_sleep")) ||
+			   (!strcmp(rela->sym->name, "___might_sleep"))) {
 				found = 1;
 				break;
 			}
@@ -311,10 +321,10 @@ static int kpatch_warn_only_change(struct section *sec)
 		if (!found)
 			return 0;
 
-		warnonly = 1;
+		lineonly = 1;
 	}
 
-	if (!warnonly)
+	if (!lineonly)
 		ERROR("no instruction changes detected for changed section %s",
 		      sec->name);
 
@@ -333,10 +343,10 @@ static void kpatch_compare_sections(struct list_head *seclist)
 			sec->status = NEW;
 	}
 
-	/* exclude WARN-only changes */
+	/* exclude WARN-only, might_sleep changes */
 	list_for_each_entry(sec, seclist, list) {
-		if (kpatch_warn_only_change(sec)) {
-			log_debug("reverting WARN-only section %s status to SAME\n",
+		if (kpatch_line_macro_change_only(sec)) {
+			log_debug("reverting macro / line number section %s status to SAME\n",
 				  sec->name);
 			sec->status = SAME;
 		}


joe-lawrence:
Lightly tested on fedora/4.10.  I can throw the test suite at this if it seems like a reasonable approach.  If we get more of these line-number bumps, perhaps we refactor these routines with a generic pattern description parsing?  Didn't seem worth it just yet, so I copy/pasted ```kpatch_warn_only_change()```

kpatch-build/create-diff-object.c
```diff
@@ -321,6 +321,93 @@ static int kpatch_warn_only_change(struct section *sec)
...
+
+		/* verify it's a mov immediate to %esi */
+		insn_get_opcode(&insn1);
+		insn_get_opcode(&insn2);
+		if (insn1.opcode.value != 0xbe || insn2.opcode.value != 0xbe)
+			return 0;
+
+		/* verify the next instruction is mov $0x0,%rdi */
+		if (*(unsigned int *)(start1 + offset + length) != 0xc7c748)
+			return 0;
```
jpoimboe: Setting CONFIG_DEBUG_ATOMIC_SLEEP and looking through the compiled code at calls to ___might_sleep(), this pattern doesn't always hold.  Sometimes it sets %esi first, sometimes it sets %rdi first, sometimes there are instructions in-between.  So I would remove this restriction.

After doing that, I think the only difference between this function and kpatch_warn_only_change() is the string compares inside the rela loop.  So I think it would be worthwhile to combine the two functions into a single `kpatch_line_macro_change_only()` function.  It might be as simple as adding the "__might_sleep" check at the end of the loop.

joe-lawrence: Thanks for double checking, you're right about the code pattern.
**block/blk-core.c :: generic_make_request_checks()** is an interesting example of this:
```
0000000000000000 <generic_make_request_checks>:
   0:   0f 1f 44 00 00          nopl   0x0(%rax,%rax,1)
                        1: R_X86_64_NONE        __fentry__-0x4
   5:   55                      push   %rbp
   6:   31 d2                   xor    %edx,%edx
>  8:   be 60 07 00 00          mov    $0x760,%esi
   d:   48 89 e5                mov    %rsp,%rbp
  10:   41 57                   push   %r15
  12:   41 56                   push   %r14
  14:   41 55                   push   %r13
  16:   41 54                   push   %r12
  18:   49 89 ff                mov    %rdi,%r15
  1b:   53                      push   %rbx
  1c:   48 83 ec 38             sub    $0x38,%rsp
  20:   8b 5f 28                mov    0x28(%rdi),%ebx
> 23:   48 c7 c7 00 00 00 00    mov    $0x0,%rdi
                        26: R_X86_64_32S        .rodata.blk_start_queue.str1.1
  2a:   65 48 8b 04 25 28 00    mov    %gs:0x28,%rax
  31:   00 00
  33:   48 89 45 d0             mov    %rax,-0x30(%rbp)
  37:   31 c0                   xor    %eax,%eax
> 39:   e8 00 00 00 00          callq  3e <generic_make_request_checks+0x3e>
                        3a: R_X86_64_PC32       __might_sleep-0x4
```
where the line number is stored before saving the stack frame, then the filename relocation is saved, more instructions, then finally the callq + __might_sleep rela.

If I understand **kpatch_warn_only_change()** correctly, it's verifying for a given section:
- before and after versions must match up until...
- an immediate mov to %esi.
- a "warn_slowpath_" non-string relocation somewhere later in this section
(BTW, does checking for the optional relocations matter?  I guess it saves a couple extra strcmp's)

Adding "__might_sleep" to the relocation check works out for **generic_make_request_checks()** but I'm wondering if this algorithm is a bit optimistic.  I'm trying to think of an example patch that could inadvertently pass the test, but it's getting late on Friday afternoon and I'm stumped :)

I'd be curious to see other patterns that you may had noticed.

jpoimboe:
> before and after versions must match up until...
> an immediate mov to %esi.

To be clear, it also ensures that the mov to esi is the *only* change in the entire function (unless there are other pattern matches).

> (BTW, does checking for the optional relocations matter? I guess it saves a couple extra strcmp's)

Yes, because it ensures that there are no other relocations in-between the mov to %esi and the call to might_sleep/warn.  Otherwise, for example, the relocation could be for a call to another function, for which %esi would be a function argument.

> Adding "__might_sleep" to the relocation check works out for generic_make_request_checks() but I'm wondering if this algorithm is a bit optimistic.

It's a valid concern.  But IMO it would be highly unlikely for the algorithm to come up with a false positive.  And anyway, as a safeguard, we expect the user to always check the output of kpatch-build to ensure the right functions were detected as changed.

I suppose we could make it a little safer by separating the relocation loop into two separate loops -- one for warn and one for might_sleep.  Not sure if that's really necessary.

> I'd be curious to see other patterns that you may had noticed.

Looking around with `objdump -dr vmlinux.o` and searching for `might_sleep`, I didn't see any other patterns.  Sometimes the esi mov comes first, sometimes the rdi file name string mov comes first.  That's the same reason `kpatch_warn_only_change()` treats the string relocation as optional, since it could have come before the detected esi mov.


> Edit: just noticed that ___might_sleep (three underscores) might need to be checked as well.

Ah, right, both the two- and three-underscore versions should be checked for.

joe-lawrence:
Okay, one more time through **kpatch_warn_only_change()** and I understand now.  I wasn't considering the continuation of checks through the rest of the section after a potential WARN macro match. 

I'm running the long kpatch-test to verify no inadvertent hiccups in the test cases.  I'll post the results once that's finished, but in the mean time I'll push v2 as you may have comments on the inline comments or naming conventions.

kpatch-build/create-diff-object.c
```diff
@@ -302,7 +310,9 @@ static int kpatch_warn_only_change(struct section *sec)
 				continue;
 			if (!strncmp(rela->sym->name, "__warned.", 9))
 				continue;
-			if (!strncmp(rela->sym->name, "warn_slowpath_", 14)) {
+			if (!strncmp(rela->sym->name, "warn_slowpath_", 14) ||
+			   (!strncmp(rela->sym->name, "__might_sleep", 14)) ||
+			   (!strncmp(rela->sym->name, "___might_sleep", 15))) {
```
jpoimboe: I think just a strcmp compare will suffice since they're not substrings.  Otherwise the patch looks good to me.

joe-lawrence:
v3
changes: use **strcmp** instead of **strncmp** as suggested by @jpoimboe 
testing on Fedora 25 + 4.9 +  ```CONFIG_DEBUG_ATOMIC_SLEEP```:

- The long test finished and aside from meminfo-cmdline-rebuild-SLOW.patch, all were happy.  (I was getting a bunch of ```find_local_syms: 113: find_local_syms for xyz.c: found_none``` errors from this patch, but I don't think its related this this PR.)

- From the integration tests, gcc-static-local-var-5.patch contains a few instances of the compact might_sleep pattern.  I verified that the line numbers in **kauditd_thread()** moved, but that create-diff-object ignored the change to that function:

```
Disassembly of section .text.kauditd_thread:

0000000000000000 <kauditd_thread>:
...
 121:	be 23 02 00 00       	mov    $0x223,%esi
 126:	48 c7 c7 00 00 00 00 	mov    $0x0,%rdi
			129: R_X86_64_32S	.rodata.kauditd_thread.str1.1
 12d:	e8 00 00 00 00       	callq  132 <kauditd_thread+0x132>
			12e: R_X86_64_PC32	__might_sleep-0x4

...
checking file kernel/audit.c
Reading special section data
Building original kernel
Building patched kernel
audit.o: changed function: audit_log_config_change
audit.o: new function: kpatch_audit_foo
audit.o: new function: kpatch_audit_check
audit.o: changed function: audit_log_end
audit.o: changed function: audit_log_start
Patched objects: vmlinux
Building patch module: kpatch-gcc-static-local-var-5.ko
SUCCESS
```

- Using a custom patch below, I tested the longer version with similar results:

```
diff -Nupr src.orig/block/blk-core.c src/block/blk-core.c
--- src.orig/block/blk-core.c   2017-02-10 14:25:47.844626220 -0500
+++ src/block/blk-core.c        2017-02-10 16:51:11.475236308 -0500
@@ -1885,6 +1885,10 @@ generic_make_request_checks(struct bio *
        char b[BDEVNAME_SIZE];
        struct hd_struct *part;
 
+
+
+
+
        might_sleep();
 
        if (bio_check_eod(bio, nr_sectors))


0000000000000000 <generic_make_request_checks>:
...
   8:   be 60 07 00 00          mov    $0x760,%esi
...
  23:   48 c7 c7 00 00 00 00    mov    $0x0,%rdi
                        26: R_X86_64_32S        .rodata.blk_start_queue.str1.1
...
  39:   e8 00 00 00 00          callq  3e <generic_make_request_checks+0x3e>
                        3a: R_X86_64_PC32       __might_sleep-0x4

...
checking file block/blk-core.c
Reading special section data
Building original kernel
Building patched kernel
ERROR: no functional changes found. Check /root/.kpatch/build.log for more details.
```

jpoimboe:
Forget to push it? :-)

joe-lawrence:
Yup :)  should be up now.

jpoimboe:
Thanks @joe-lawrence !

👍

joe-lawrence:
>The long test finished and aside from meminfo-cmdline-rebuild-SLOW.patch, all were happy. (I was getting a bunch of find_local_syms: 113: find_local_syms for xyz.c: found_none errors from this patch, but I don't think its related this this PR.)

Hmm, I'm seeing this on a stock RHEL7 box with the same patch from this branch...  It's blowing up on a ton of .o files.  I don't remember ever seeing these error messages before.   Before we merge this PR, this afternoon I'll see if I can repro with kpatch master and figure out what's going on :(

jpoimboe:
Ok, thanks for testing it.  You never know what's going to break where.  We really need to get that CI testing going.

joe-lawrence:
I'll open a new issue, but the find_local_syms errors I was getting are **not** related to this patchset, so feel to merge.

Another observation: regarding the relatively new ```convert-global-local.patch```, is it expected to see this warning on load:
```
kpatch: Skipped dynrela for this_cpu_cmpxchg16b_emu (0xffffffffc058d527 <- 0xffffffff8f40fec0): the instruction has been changed already.
kpatch: This is not necessarily a bug but it may indicate in some cases that the binary patch does not handle paravirt operations, alternatives or the like properly.
```

jpoimboe:
That warnings looks like a known limitation of kpatch.ko, related to the handling of .altinstructions sections.  I think there's nothing we can do about it.  It should be fixed in livepatch though.

👍

flaming-toast:
@joe-lawrence Yeah, that's a known issue with applying dynrelas to sections that have been paravirt patched (see arch/x86/kernel/alternative.c) It's something that the kpatch core module cannot handle, because the module loader typically applies these paravirt/alternative sections to the patch module before the kpatch core module gets a chance to apply dynrelas to it. We skip applying the dynrela in this case, because we don't want to overwrite the paravirt patched instruction that had been applied by the module loader.  With livepatch, we got around this issue by renaming these sections with a ".klp.arch." prefix and applying them manually with `arch_klp_init_module_loaded()`.

PR #673, opened 2017-02-09
kpatch-build fixups for various kernel configs

joe-lawrence:
This patchset addresses a few issues with ```kpatch-build``` and various kernel config settings:

1. Some versions of gawk do not support the whitespace ```'\s'``` operator.  Drop back to using the ```[[:space:]]``` expression as it's supported in older verisons, too.
2. @marcin-github reported issues building on Gentoo when the kernel was configured with ```CONFIG_DEBUG_INFO_REDUCED```.  This reorders ```readelf -wi``` output a bit, and requires a smarter awk script to parse special structure sizings.
3. @marcin-github also reported problems when ```CONFIG_DEBUG_INFO_SPLIT``` was set.  This configuration stores debuginfo in separate ```.dwo``` files.  Since no major distribution is doing this (yet), just detect and note that this isn't supported.
4. @marcin-github also reported problems when ```CONFIG_PARAVIRT``` was not set.  This is technically not required, so adjust ```kpatch-build``` to handle cases when it is not set.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -516,10 +516,10 @@ echo "Reading special section data"
 SPECIAL_VARS=$(readelf -wi "$VMLINUX" |
 	       gawk --non-decimal-data '
 	       BEGIN { a = b = p = e = 0 }
-	       a == 0 && /DW_AT_name.* alt_instr\s*$/ {a = 1; next}
-	       b == 0 && /DW_AT_name.* bug_entry\s*$/ {b = 1; next}
-	       p == 0 && /DW_AT_name.* paravirt_patch_site\s*$/ {p = 1; next}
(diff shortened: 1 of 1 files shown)


joe-lawrence:
Testing:
- RHEL6
  - awk script ```[[:space:]]``` usage
- On Fedora 25
  - with ```CONFIG_DEBUG_INFO_REDUCED``` set, verified kpatch patch module build and load
  - with ```CONFIG_DEBUG_INFO_SPLIT``` set, verified kpatch-build kicked back an error message
  - with ```CONFIG_PARAVIRT``` not set, verified kpatch patch module build and load, also loaded combined kpatch-test patch

jpoimboe:
👍

PR #674, opened 2017-02-09
testing - rebase for integration tests for Fedora 25

joe-lawrence:
Fedora moved to the 4.9 kernel in early January (now up to 4.9.7-201.fc25.x86_64) and that broke a few of the integration tests.  Rebase and update only those that need adjusting (including fuzz).


joe-lawrence:
Testing - ran the kpatch-test script and all patches build and test programs were happy.

jpoimboe:
I have some mixed feelings about rebasing the tests to a newer kernel.  It's hard to test our existing functionality against a moving baseline.  For example, if a PR fails the integration tests, we don't know if it happened because of something wrong with the PR, or if something changed in the kernel.  But this is probably a minor issue since these tests will probably only be rebased every 3 months or so, and we should eventually have automated tests to catch such issues.

Another issue is the more we rebase these tests against newer kernels, the more likely it is that they no longer test what we think they're testing.  I'd bet that several of them don't test what they were originally created to test, and that problem will get worse as we go on.

Related to that, it's on my TODO list to a) go back and make sure they're still testing what they're supposed to; and b) come up with a way to ensure that each test continues to test what it's supposed to test.  If we have that, combined with automated testing of the latest "rolling" Fedora kernel plus other select "stable" distros, we'll be in good shape.

Anyway, the GA version of the F25 kernel is already obsolete, and so I think it makes sense to treat Fedora as a rolling release and keep testing the latest kernels.

👍

joe-lawrence:
I forgot about potentially testing the GA kernels only, instead of the latest and greatest.  (I had reserved a Fedora 25 Beaker machine wanting to test another PR and ran into this issue.)  It's definitely a PITA to continuously rebase.  Feel free to merge this one or not.

jpoimboe:
I do think there is value in rebasing for every version of the kernel, as it often uncovers new issues.  So I stand by the 👍

joe-lawrence:
Ok -- I think Fedora is probably tracking newer kernels more aggressively than RHEL or Ubuntu anyway.

jpoimboe:
Right, the stable distros might rarely (or even never) need a rebase anyway.

PR #680, opened 2017-02-28
create-diff-object: Mark correlated static local variables the SAME i…

rosslagerwall:
…nitially

Normal correlated symbols are marked the SAME initially but static local
variables are correlated in a separate function. Also mark these the
SAME.

This fixes an issue where patching a function which called printk_once
(which uses a static local variable) would fail to build because the
static local variable was considered new and thus introduced a new data
member into .data..read_mostly which is not allowed to change.


jpoimboe:
👍

Issue #676, opened 2017-02-15
find_local_syms errors and meminfo-cmdline-rebuild-SLOW.patch


joe-lawrence:
I'm hitting a lot of errors like "ERROR: md.o: find_local_syms: 113: find_local_syms for md.c: found_none" when building ```meminfo-cmdline-rebuild-SLOW.patch```.  I've seen these errors building against a Fedora 4.9 debug kernel as well as RHEL7 3.10.0-514.10.1.el7.x86_64.

I bisected kpatch back to aa2907df297:
```
aa2907df297fdfa57e608c110702579303193641 is the first bad commit
commit aa2907df297fdfa57e608c110702579303193641
Author: Zhou Chengming <zhouchengming1@huawei.com>
Date:   Sat Feb 4 09:31:25 2017 +0800

    support dup file+symbol

    We use kelf_base->symbols to find a unique matching FILE+locals combination
    when we call lookup_open(). If we can't find one matching or we find more
    than one matching, we error out.

    If we find a unique one, we setup table->local_syms in lookup_open(),
    so later lookup_local_symbol() could do its lookup based on table->local_syms.

    Fixes #604.

    Suggested-by: Josh Poimboeuf <jpoimboe@redhat.com>
    Signed-off-by: Zhou Chengming <zhouchengming1@huawei.com>
```

jpoimboe:
I'm able to recreate on Fedora 24.  Will work up a fix.

PR #684, opened 2017-03-03
lookup: skip discarded symbols in local symbol comparison

jpoimboe:
A few symbols are discarded in the kernel linking phase, which means
they won't be in the lookup table.  Skip their comparison.
    
This fixes a bunch of warnings seen when building a patch which triggers
a tree-wide rebuild:
    
    create-diff-object: ERROR: aes_glue.o: find_local_syms: 112: find_local_syms for aes_glue.c: found_none
    create-diff-object: ERROR: aesni-intel_glue.o: find_local_syms: 112: find_local_syms for aesni-intel_glue.c: found_none
    create-diff-object: ERROR: init.o: find_local_syms: 112: find_local_syms for init.c: found_none
    create-diff-object: ERROR: iosf_mbi.o: find_local_syms: 112: find_local_syms for iosf_mbi.c: found_none
    create-diff-object: ERROR: setup.o: find_local_syms: 112: find_local_syms for setup.c: found_none
    ...
    
After this patch, there's still one warning remaining:
   
    create-diff-object: ERROR: dynamic_debug.o: find_local_syms: 133: find_local_syms for dynamic_debug.c: found_none
    
That one has a completely different cause, which I'll fix in another
pull request (coming soon).

Fixes: #676

### DIFF
--- a/kpatch-build/lookup.c
+++ b/kpatch-build/lookup.c
@@ -84,6 +84,9 @@ static void find_local_syms(struct lookup_table *table, char *hint,
 	int i, in_file = 0;
 	struct sym_compare_type *child_sym;
 
+	if (!child_locals)
+		return;
+
 	for_each_obj_symbol(i, sym, table) {
 		if (sym->type == STT_FILE) {
 			if (in_file && !child_sym->name) {
@@ -287,10 +290,7 @@ struct lookup_table *lookup_open(char *obj_path, char *symvers_path,
 
 	obj_read(table, obj_path);
 	symvers_read(table, symvers_path);
-
-	table->local_syms = NULL;
-	if (locals)
-		find_local_syms(table, hint, locals);
+	find_local_syms(table, hint, locals);
 
 	return table;
 }
--- a/kpatch-build/lookup.c
+++ b/kpatch-build/lookup.c
@@ -65,15 +65,15 @@ struct lookup_table {
 	for (ndx = 0, iter = table->exp_syms; ndx < table->exp_nr; ndx++, iter++)
 
 static void find_local_syms(struct lookup_table *table, char *hint,
-			    struct sym_compare_type *locals)
+			    struct sym_compare_type *child_locals)
 {
 	struct object_symbol *sym, *file_sym;
 	int i, in_file = 0;
-	struct sym_compare_type *local_index;
+	struct sym_compare_type *child_sym;
 
 	for_each_obj_symbol(i, sym, table) {
 		if (sym->type == STT_FILE) {
-			if (in_file && !local_index->name) {
+			if (in_file && !child_sym->name) {
 				if (table->local_syms)
 					ERROR("find_local_syms for %s: found_dup", hint);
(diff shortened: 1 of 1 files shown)


kpatch-build/lookup.c
```diff
@@ -262,6 +283,8 @@ struct lookup_table *lookup_open(char *obj_path, char *symvers_path,
 		ERROR("malloc table");
 	memset(table, 0, sizeof(*table));
 
+	table->vmlinux = !strcmp(basename(obj_path), "vmlinux");
+
```
joe-lawrence: It probably doesn't matter, but is it any better/worse to push this assignment down to lookup_open() just before it calls find_local_syms() ?

joe-lawrence:
I'd like to try out on a few kernels on Monday, but this looks promising.  Thanks for fixing!

kpatch-build/lookup.c
jpoimboe: I don't have an opinion either way, though I would at least argue that it shouldn't be in the `if (locals)` condition block since it might end up being useful in other places someday.

BTW, looking at the code, the `table->local_syms = NULL` and `if (locals)` check should probably be moved to inside `find_local_syms()` because they're low-level details involved in finding the locals.  I'll tack on a commit to this PR to do that.

jpoimboe:
Added a commit with a minor cleanup to lookup_open().

joe-lawrence:
Hi Josh, I'm trying out the **find_local_syms-fix** branch on Fedora + upstream kernel tree and get the following error:
```
% kpatch-build --skip-cleanup -s /root/linux fedora-25/meminfo-cmdline-rebuild-SLOW.patch
Skipping cleanup
Testing patch file
Reading special section data
Building original kernel
Building patched kernel
cp: cannot stat '/root/.kpatch/obj/lib/.lib_exports.o': No such file or directory
```

That file is listed in the changed objects file, but doesn't exist when kpatch-build tries to make copies.
```
% grep '\.lib_exports.o' /root/.kpatch/tmp/changed_objs 
lib/.lib_exports.o
arch/x86/lib/.lib_exports.o
% find /root/.kpatch/ -name .lib_exports.o
(nothing)
```

jpoimboe:
That's weird, though I don't see how my branch could have caused that... Can you confirm whether it's related to my branch of if it's a pre-existing issue?

joe-lawrence:
Hmmm, same result when building that patch against kpatch master.  With respect to this PR, I guess ignore this for now while I try and figure out why this is failing for me :(

jpoimboe:
I don't know the root cause, but the fix might be to add that file to the list of ignored object files in kpatch-build/kpatch-gcc.

joe-lawrence:
Looks to be kernel-version specific (4.10 and up) as I just built successfully against an upstream 4.9 kernel.   :+1: to this PR  and I'll open a new issue regarding .lib_exports.o.

jpoimboe:
Thanks a lot for the testing and review @joe-lawrence !

PR #685, opened 2017-03-03
kpatch-build: build the kernel in ~/.kpatch/src again

jpoimboe:
Once upon a time, kpatch-build did the kernel build in three passes.
The extra pass was done without '-ffunction-sections -fdata-sections',
so it could produce the original vmlinux file.

At that time, there was no ~/.kpatch/obj directory.  The kernel was
built directly in ~/.kpatch/src.  Because the same directory was used
for both the original kernel build and the '-ffunction-sections
-fdata-sections' build, the entire tree had to be rebuilt twice for
every kpatch-build incantation, making it very slow.

That situation was improved with the following commit:

  5352d8b01aa1 ("build objects in separate directory to fix caching")

That built the regular and special binaries in ~/.kpatch/obj and
~/.kpatch/obj2, respectively.

Since then we've simplified things so that it only does two build
passes: original and patched, both with '-ffunction-sections
-fdata-sections', and ~/.kpatch/obj2 was removed.  However,
~/.kpatch/obj still remained.  That's because we never had a reason to
change it, until now.

Recent commit aa2907df297f ("support dup file+symbol")
triggers a new warning:

  create-diff-object: ERROR: dynamic_debug.o: find_local_syms: 124: find_local_syms for dynamic_debug.c: found_none

This was actually a preexisting issue which that commit helped uncover.
The root issue is that dynamic_debug.c has some creative uses of the
`__FILE__` macro.  When building the kernel objects outside the source
tree, the macro results in a absolute path like:

  /home/jpoimboe/.kpatch/src/lib/dynamic_debug.c

But when building inside the source tree it's a relative path:

  lib/dynamic_debug.c

The Fedora kernel is built in-tree, and I would imagine most other
distros are also built that way.  So the way kpatch builds can result in
a slightly different 'original' object than the distro version, thanks
to the `__FILE__` macro.

In this case, the order of the symbol table changed slightly between
vmlinux and the 'orig' object.  Presumably, the difference in string
lengths was enough to convince the compiler to shuffle things around a
bit.

So considering that bug, and the possibility of other mismatches, go
back to building the kernel in the source tree.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -40,7 +40,6 @@ ARCHVERSION="$(uname -r)"
 CPUS="$(getconf _NPROCESSORS_ONLN)"
 CACHEDIR="${CACHEDIR:-$HOME/.kpatch}"
 SRCDIR="$CACHEDIR/src"
-OBJDIR="$CACHEDIR/obj"
 RPMTOPDIR="$CACHEDIR/buildroot"
 VERSIONFILE="$CACHEDIR/version"
 TEMPDIR="$CACHEDIR/tmp"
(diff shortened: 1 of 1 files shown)


joe-lawrence:
Oh man, the ```__MACROs__``` strike again!  From a code inspection POV, these look good.  (I also like the ```grep -q``` clean ups.)   Looks like you tested with successive ```kpatch-build -s``` invocations as well...  :+1:

PR #687, opened 2017-03-10
kpatch-gcc: update ignorelist to avoid foo/.lib_exports.o files

### LINKED ISSUE #686: .lib_exports.o: No such file or directory
Building meminfo-cmdline-rebuild-SLOW.patch against an upstream 4.10 kernel tree, I get the following error:
```
% kpatch-build --skip-cleanup -s /root/linux fedora-25/meminfo-cmdline-rebuild-SLOW.patch
Skipping cleanup
Testing patch file
Reading special section data
Building original kernel
Building patched kernel
cp: cannot stat '/root/.kpatch/obj/lib/.lib_exports.o': No such file or directory
```
That file is listed in the changed objects file, but doesn't exist when kpatch-build tries to make copies.
```
% grep '\.lib_exports.o' /root/.kpatch/tmp/changed_objs 
lib/.lib_exports.o
arch/x86/lib/.lib_exports.o
% find /root/.kpatch/ -name .lib_exports.o
(nothing)
```

joe-lawrence:
Upstream kernel commit 7f2084fa55e6 ("[kbuild] handle exports in lib-y
objects reliably") (v4.9+) added temporary dummy .lib_exports.o objects
to the kernel build.  As these ephemeral files don't contain any code,
update the kpatch-gcc glob pattern to ignore them.

(glob pattern suggested by flaming-toast)

Fixes #686.

### DIFF
--- a/kpatch-build/kpatch-gcc
+++ b/kpatch-build/kpatch-gcc
@@ -32,7 +32,7 @@ if [[ "$TOOLCHAINCMD" = "gcc" ]] ; then
 				arch/x86/vdso/*|\
 				arch/x86/entry/vdso/*|\
 				drivers/firmware/efi/libstub/*|\
-				.*.o)
+				*.*.o)
 					break
 					;;
 				*.o)


jpoimboe:
👍

PR #688, opened 2017-03-10
README.md updates

joe-lawrence:
A small collection of documentation updates.

Tacking on part of the running kernel release is a little ugly, but I kept running into the situation where dnf/yum wants to install the lastest kernel-xyz package in Vagrant boxes if I omit a version spec.  The debuginfo lines for Oracle already do something this, so using the current kver throughout would be more consistent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -443,8 +443,7 @@ kpatch-build already works with both livepatch and kpatch.  If your kernel has
 CONFIG\_LIVEPATCH enabled, it detects that and builds a patch module in the
 livepatch format.  Otherwise it builds a kpatch patch module.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,27 +25,31 @@ Installation
 
 ###Prerequisites
 
-####Fedora 23
+####Fedora 25
```
jpoimboe: Maybe we should just remove the version here so this line doesn't become obsolete every 6 months :-)

joe-lawrence: That's probably the easiest going forward, as I would think most of the commands would remain constant.

jpoimboe:
👍

PR #690, opened 2017-03-14
Add author guide examples

joe-lawrence:
Since I was researching how kpatch implemented these features, I figured I could document them along the way in the patch authoring guide.

(Un)load hooks - stop_machine context -- is it worth mentioning the `kpatch_module_config` case?  In that case, are the hooks called in process context (ie, interrupts may occur) under some module mutex?

Shadow variables - I didn't see any kpatch examples outside of the integration tests.  Are there any interesting pitfalls that we should point out?

### DIFF
--- a/doc/patch-author-guide.md
+++ b/doc/patch-author-guide.md
@@ -108,8 +108,36 @@ functions.
 ### Use a kpatch load hook
 
 If you need to change the contents of an existing variable in-place, you can
-use the KPATCH_LOAD_HOOK macro to specify a function to be called when the
+use the `KPATCH_LOAD_HOOK` macro to specify a function to be called when the
 patch module is loaded.
+
+`kpatch-macros.h` provides `KPATCH_LOAD_HOOK` and `KPATCH_UNLOAD_HOOK` macros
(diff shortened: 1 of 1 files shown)


doc/patch-author-guide.md
```diff
@@ -108,23 +108,104 @@ functions.
...
++		newpid = kpatch_shadow_alloc(p, "newpid", sizeof(*newpid),
++					     GFP_KERNEL);
++		if (newpid)
++			*newpid = ctr++;
+```
+
+A shadow variable may also be accessed via `kpatch_shadow_get`.  Here the
+patch modifies `task_context_switch_counts()` to fetch the shadow variable
+associated with the current `struct task_struct *p` object and a "newpid" tag.
+As in the previous patch chunk, a the shadow variable pointer may be accessed
```
jpoimboe: s/a the/the/

joe-lawrence: Fixed, thanks.

jpoimboe:
This documentation is important, thanks @joe-lawrence !

👍

PR #692, opened 2017-03-24
Author guide fixups

joe-lawrence:
Two small tweaks for the patch author guide.  The first clarifies the context for load hooks (stop_machine and module notifier).  The second adds a shadow variable - data semantic example.

I left the locking example todo in place -- @jpoimboe -- would `0010-CVE-2014-2706-2.patch` that you mailed me last week work here?  If so I can include that as well.

### DIFF
--- a/doc/patch-author-guide.md
+++ b/doc/patch-author-guide.md
@@ -218,10 +218,56 @@ safe to access:
 Data semantic changes
 ---------------------
 
-Sometimes, the data itself remains the same, but how it's used is changed.  A
-common example is locking semantic changes.
+Part of the stable-tree [backport](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/commit/fs/aio.c?h=linux-3.10.y&id=6745cb91b5ec93a1b34221279863926fba43d0d7)
+to fix CVE-2014-0206 changed the reference count semantic of `struct
+kioctx.reqs_active`. Associating a shadow variable to new instances of this
+structure can be used by patched code to handle both new (post-patch) and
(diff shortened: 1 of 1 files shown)


doc/patch-author-guide.md
```diff
@@ -113,8 +113,15 @@ patch module is loaded.
 
 `kpatch-macros.h` provides `KPATCH_LOAD_HOOK` and `KPATCH_UNLOAD_HOOK` macros
 to define such functions.  The signature of both hook functions is `void
-foo(void)` and and they may run in `stop_machine` context (so they must not
-sleep).
+foo(void)`.  Their execution context is as follows:
+
+* For objects that are already loaded, hook functions will be run by
```
jpoimboe: The term 'object' has a specific meaning in our code, but some readers of this doc might not grok what it means.  So maybe this should say something like "For patches to vmlinux or for kernel modules which have already been loaded ..."

doc/patch-author-guide.md
```diff
@@ -113,8 +113,15 @@ patch module is loaded.
...
+`stop_machine` as part of applying or removing a patch. (Therefore the hooks
+must not block or sleep.)
+
+* For objects not yet loaded, a module-notifier will execute load hooks when
```
jpoimboe: Similar change here: "For patches to kernel modules which haven't been loaded yet..."

joe-lawrence:
v2
  - incorporated 'object' terminology suggestions
  - added locking semantic text

jpoimboe:
Good stuff!

👍

PR #694, opened 2017-04-20
Author guide notes

joe-lawrence:
More odds and ends for the patch writer's guide (and one for the top level readme).

### DIFF
--- a/README.md
+++ b/README.md
@@ -514,7 +514,8 @@ so it looks like a normal function to the kernel.
 - **oops stack traces**: Yes.  If the replacement function is involved in an
   oops, the stack trace will show the function and kernel module name of the
   replacement function, just like any other kernel module function.  The oops
-  message will also show the taint flag (currently `TAINT_USER`).
+  message will also show the taint flag (see the FAQ "How can I detect if
+  somebody has patched the kernel" for specifics).
 - **kdump/crash**: Yes.  Replacement functions are normal functions, so crash
   will have no issues.
 - **ftrace**: Yes, but certain uses of ftrace which involve opening the
--- a/doc/patch-author-guide.md
+++ b/doc/patch-author-guide.md
@@ -342,7 +342,10 @@ When changing header files, be extra careful.  If data is being changed, you
 probably need to modify the patch.  See "Data struct changes" above.
 
 If a function prototype is being changed, make sure it's not an exported
-function.  Otherwise it could break out-of-tree modules.
+function.  Otherwise it could break out-of-tree modules.  One way to
+workaround this is to define an entirely new copy of the function (with
+updated code) and patch in-tree callers to invoke it rather than the
+deprecated version.
 
 Many header file changes result in a complete rebuild of the kernel tree, which
 makes kpatch-build have to compare every .o file in the kernel.  It slows the
--- a/doc/patch-author-guide.md
+++ b/doc/patch-author-guide.md
@@ -387,3 +387,71 @@ to zero; instead the variable keeps its old value.
 To work around this limitation one needs to retain the reference to the static local.
 This might be as simple as adding the variable back in the patched function in a 
 non-functional way and ensuring the compiler doesn't optimize it away.
+
+Code removal
+------------
+
+Some fixes may replace or completely remove functions and references
+to them. Remember that kpatch modules can only add new functions and
+redirect existing functions, so "removed" functions will continue to exist in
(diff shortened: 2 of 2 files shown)


doc/patch-author-guide.md
```diff
@@ -365,13 +377,26 @@ Some examples:
   function will also change.  In this case there's nothing you can do to
   prevent the extra changes.
 
+* If a changed function was originally inlined, but turned into a callable
+  function after patching, consider adding `inline` to the function
+  definition.  Likewise, if a function is only inlined after patching,
+  consider `noinline`.  Remember though, these keywords are only suggestions
+  to the compiler.  YMMV.
```
jpoimboe: If you use `__always_inline` it will always inline.  Ditto with `noinline`.

README.md
```diff
@@ -514,7 +514,8 @@ so it looks like a normal function to the kernel.
 - **oops stack traces**: Yes.  If the replacement function is involved in an
   oops, the stack trace will show the function and kernel module name of the
   replacement function, just like any other kernel module function.  The oops
-  message will also show the taint flag (currently `TAINT_USER`).
+  message will also show the taint flag (see the FAQ  "How can I detect if
```
jpoimboe: nit: there are two spaces after "FAQ"

doc/patch-author-guide.md
```diff
@@ -214,6 +214,15 @@ safe to access:
...
 ```
+Notes:
+* `kpatch_shadow_alloc` initializes only shadow variable metadata. It
+  allocates variable storage via `kmalloc` with the `gfp_t` flags it is
+  given, but otherwise leaves the area untouched. Initialization of a shadow
+  variable is the responsibility of the caller.
+* As soon as `kpatch_shadow_alloc` creates a shadow variable, its presence
+  will be reported by `kpatch_shadow_get`. Care should be taken to avoid any
+  potential race conditions between a kernel thread that allocates a shadow
+  variable and concurrent threads that may attempt to use it.
```
jpoimboe: Here it might be good to describe our solution of splitting it up into two patches.  (The second patch could still be a superset of the first so it would still be compatible with our recommendation of cumulative patches.)

doc/patch-author-guide.md
joe-lawrence: I've never actually needed to use these before, thanks for reminding me about this.

doc/patch-author-guide.md
joe-lawrence: With this note, I was targeting the use-before-initialization race:
```
sv = kpatch_shadow_alloc(obj, "var", sizeof(int), GFP_KERNEL);
... registered but not initialized ...
*sv = 123;
```
where another thread could get a hit on kpatch_shadow_get and then use it before this one has initialized it.  (Applicable to both kpatch and livepatch models.)

I think the two-patch approach you proposed was addressing a potential race under the livepatch consistency model only:
* task1 - successfully patched allocation path, makes new shadow variables
* task2 - free path not patched yet, leaks those new shadow variables

Since the patch author guide is kpatch-specific right now, I only mentioned this first race.  I think that one is best fixed by modifying the shadow variable API to accept an already allocated (and initialized!) variable space.

Let me know if I'm mis-remembering that conversation :)

jpoimboe: Ah, you're right, thanks for clarifying.  Once we have a livepatch shadow data feature upstream, the new API (and corresponding race) will need to be documented as well.

joe-lawrence:
v2:
- incorporate @jpoimboe 's `__always_inline` suggestion and remove YMMV disclaimer
- remove double space from README.md update

doc/patch-author-guide.md
```diff
@@ -365,13 +377,25 @@ Some examples:
   function will also change.  In this case there's nothing you can do to
   prevent the extra changes.
 
+* If a changed function was originally inlined, but turned into a callable
+  function after patching, consider adding `__always_inline` to the function
+  definition.  Likewise, if a function is only inlined after patching,
+  consider using`noinline` to prevent the compiler from doing so.
```
jpoimboe: Missing space before `noinline`

joe-lawrence: Ooops, fixed.

jpoimboe:
👍

flaming-toast:
Thanks for improving the author guide, @joe-lawrence !

PR #696, opened 2017-04-20
create-diff-object: don't create dynrelas for exported vmlinux symbol…

### LINKED ISSUE #695: don't create dynrelas for exported symbols used by patched modules
@joe-lawrence saw the following errors when loading upstream commit `128394eff343fc6d2f32172f03e24829539c5835` on RHEL7 with kpatch.ko:

```
Skipped dynrela for copy_user_generic_unrolled (0xffffffffa0475942 <- 0xffffffff813211e0): the instruction has been changed already.
Skipped dynrela for copy_user_generic_unrolled (0xffffffffa0475a57 <- 0xffffffff813211e0): the instruction has been changed already.
```

It's running into issue #580, but I think that can be avoided by leaving `copy_user_generic_unrolled` as a normal relocation instead of converting it to a dynrela, because it's an exported symbol.  @flaming-toast would you agree, or am I missing an edge case when it comes to patching a module?  Is there any reason why a patch to a module would need a dynrela for an exported symbol?

jpoimboe:
…s used by modules

@joe-lawrence  saw the following errors when loading Linux commit 128394eff343
("sg_write()/bsg_write() is not fit to be called under KERNEL_DS"):
```
  Skipped dynrela for copy_user_generic_unrolled (0xffffffffa0475942 <- 0xffffffff813211e0): the instruction has been changed already.
  Skipped dynrela for copy_user_generic_unrolled (0xffffffffa0475a57 <- 0xffffffff813211e0): the instruction has been changed already.
```
That is known issue #580, but it can be avoided by leaving
`copy_user_generic_unrolled` as a normal relocation instead of
converting it to a dynrela, because it's an exported symbol.

Also remove the manual check for `__fentry__` because it's covered by
the exported symbol check.

Also remove a duplicate comment about unexported global object symbols
being in another .o in the patch object.

Fixes #695.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2123,22 +2123,8 @@ static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
 			} else {
 				/*
 				 * We have a patch to a module which references
-				 * a global symbol.
-				 */
-
-				/*
-				 * __fentry__ relas can't be converted to
-				 * dynrelas because the ftrace module init code
-				 * runs before the dynrela code can initialize
-				 * them.  __fentry__ is exported by the kernel,
-				 * so leave it as a normal rela.
-				 */
-				if (!strcmp(rela->sym->name, "__fentry__"))
-					continue;
-
-				/*
-				 * Try to find the symbol in the module being
-				 * patched.
+				 * a global symbol.  Try to find the symbol in
+				 * the module being patched.
 				 */
 				if (lookup_global_symbol(table, rela->sym->name,
 							 &result)) {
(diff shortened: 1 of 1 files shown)


joe-lawrence:
testing:  fwiw these changes successfully built/loaded fixes for CVE-2016-2384, CVE-2016-6480, CVE-2016-9588 and CVE-2016-9576, all which had complained that they "Skipped dynrela for ... the instruction has been changed already."

flaming-toast:
I'm still thinking about whether we can turn dynrelas for exported symbols from modules into normal relas too.

For example, if a kpatch module X patches a module Y that uses an exported symbol from module Z (i.e., module Y would have a dynrela for a symbol from module Z), wouldn't module Z already be loaded by the time module Y is loaded (which then gets by patched by kpatch module X)? I don't see how module Y can possibly load without module Z being loaded first, so I feel like those exported symbols should always be available by the time kpatch writes its dynrelas.

But that's a question for another day, this commit fixes #695, so 👍

Issue #703, opened 2017-05-11
section mismatch .eh_frame


profihost:
While trying to build patches for CVE-2017-7308 and CVE-2017-8890.  I'm always getting section mismatch errors for eh_frame.

Example patch:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=657831ffc38e30092a2d5f03d385d710eb88b09a

I was able to "workaround" this by adding:
#include "kpatch-macros.h"
KPATCH_IGNORE_SECTION(".eh_frame");

jpoimboe:
Can you share your distro, kernel version and config?  Any chance your kernel has a patch to enable `-fasynchronous-unwind-tables`?

disaster123:
I'm using an opensuse 42.2 kernel build from https://github.com/openSUSE/kernel-source/commits/openSUSE-42.2 which has unwinder dwarf2 support enabled.

jpoimboe:
Does the following patch fix it?

jpoimboe@67b3bd0f5e507d2ff66a2c18f42ce51619ede2fc

disaster123:
!strcmp(name, ".eh_frame", 9);
=>
!strncmp(name, ".eh_frame", 9);

But yes that works / fixes it! Thanks!

jpoimboe:
@disaster123 I think It's generally better to build everything in a single module.

disaster123:
OK thanks!

PR #704, opened 2017-05-11
elf: add .eh_frame* sections to debug section list

jpoimboe:
SUSE-based kernels have a DWARF unwinder, so they build with the gcc
'-fasynchronous-unwind-tables' flag, which adds .eh_frame and
.eh_frame_hdr sections.  Treat those sections like the other debug
sections.

Fixes: #703

### DIFF
--- a/kpatch-build/kpatch-elf.c
+++ b/kpatch-build/kpatch-elf.c
@@ -71,7 +71,9 @@ int is_debug_section(struct section *sec)
 		name = sec->base->name;
 	else
 		name = sec->name;
-	return !strncmp(name, ".debug_", 7);
+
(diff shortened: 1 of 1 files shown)


joe-lawrence:
Simple enough, though is there any easy way to confirm that this is safe for kpatch_include_debug_sections()?  Ie, do we need to test the "asynchronous unwind table" for patched code?

jpoimboe:
@joe-lawrence That would probably be a good idea, though it shouldn't block this PR.

@disaster123 You might want to make sure the in-kernel DWARF unwinder can unwind through patched functions.

joe-lawrence:
👍

PR #707, opened 2017-05-18
create-diff-object: only check if kernel object starts with "vmlinux"

vincentbernat:
On Debian/Ubuntu, the `vmlinux` from `-dbg` package has a version number
appended to it. For example:
`/usr/lib/debug/boot/vmlinux-3.13.0-117-generic`. Make it work
nonetheless.


flaming-toast:
We already account for the `vmlinux-*` case in create-diff-object as well, so I'm fine with this change 👍

Issue #675, opened 2017-02-10
build error: lookup.c:109:21: error: ‘file_sym’ may be used uninitialized in this function


marcin-github:
Hi!
First questions, is it safe to add optimization CFLAGS to build process of tools? I'm trying to prepare ebuild form my purposes, and build process pass system CFLAGS.
With -O1 or -O2 I'm getting:

cc -MMD -MP -O2 -I../kmod/patch -Iinsn -Wall -Werror -c -o lookup.o lookup.c
lookup.c: In function ‘lookup_open’:
lookup.c:109:21: error: ‘file_sym’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
   table->local_syms = file_sym;
                     ^
lookup.c:70:30: note: ‘file_sym’ was declared here
  struct object_symbol *sym, *file_sym;
                              ^
lookup.c:101:15: error: ‘local_index’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
    local_index++;
               ^
lookup.c:72:27: note: ‘local_index’ was declared here
  struct sym_compare_type *local_index;
                           ^
cc1: all warnings being treated as errors
make[1]: *** [Makefile:17: lookup.o] Error 1

joe-lawrence:
@marcin-github - I think it should be safe to build with -O1 or -O2.  If you initialize ```file_sym``` and ```local_index``` at the beginning of ```find_local_syms()``` does that make the compiler happy?  If so, I would be curious to see a timing comparison with/without those flags.  Patches welcome :)

aliceinwire:
some problem here:
lookup.c: In function ‘lookup_open’:
lookup.c:132:21: error: ‘file_sym’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
   table->local_syms = file_sym;
                     ^
lookup.c:83:30: note: ‘file_sym’ was declared here
  struct object_symbol *sym, *file_sym;
                              ^
lookup.c:129:27: error: ‘child_sym’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  if (in_file && !child_sym->name) {
                           ^
lookup.c:85:27: note: ‘child_sym’ was declared here
  struct sym_compare_type *child_sym;
                           ^
cc1: all warnings being treated as errors
Makefile:17: recipe for target 'lookup.o' failed
make[1]: *** [lookup.o] Error 1

aliceinwire:
works with -O1
not works with -O2

PR #712, opened 2017-06-05
lookup: silence maybe-uninitialized warnings for -O2

jpoimboe:
When compiling with -O2, it fails with:
```
  gcc -MMD -MP -O2 -I../kmod/patch -Iinsn -Wall -g -Werror -c -o lookup.o lookup.c
  lookup.c: In function ‘lookup_open’:
  lookup.c:132:21: error: ‘file_sym’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
     table->local_syms = file_sym;
     ~~~~~~~~~~~~~~~~~~^~~~~~~~~~
  lookup.c:83:30: note: ‘file_sym’ was declared here
    struct object_symbol *sym, *file_sym;
                                ^~~~~~~~
  lookup.c:129:27: error: ‘child_sym’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
    if (in_file && !child_sym->name) {
                    ~~~~~~~~~^~~~~~
  lookup.c:85:27: note: ‘child_sym’ was declared here
    struct sym_compare_type *child_sym;
                             ^~~~~~~~~
  Makefile:17: recipe for target 'lookup.o' failed
  make[1]: *** [lookup.o] Error 1
  make[1]: Leaving directory '/home/jpoimboe/git/kpatch/kpatch-build'
  Makefile:14: recipe for target 'build-kpatch-build' failed
  make: *** [build-kpatch-build] Error 2
```
As far as I can tell, these are false positive warnings.  When
in_file is 1, file_sym and child_sym are properly initialized.  But
silence the warnings anyway so Gentoo users can build with -O2.

Fixes: #675

### DIFF
--- a/kpatch-build/lookup.c
+++ b/kpatch-build/lookup.c
@@ -80,9 +80,9 @@ static int discarded_sym(struct lookup_table *table,
 static void find_local_syms(struct lookup_table *table, char *hint,
 			    struct sym_compare_type *child_locals)
 {
-	struct object_symbol *sym, *file_sym;
+	struct object_symbol *sym, *file_sym = NULL;
 	int i, in_file = 0;
-	struct sym_compare_type *child_sym;
+	struct sym_compare_type *child_sym = NULL;
 
 	if (!child_locals)
 		return;


joe-lawrence:
👍

PR #715, opened 2017-06-30
macros: add KPATCH_REVERT_FUNCTION macro

jpoimboe:
Add a KPATCH_REVERT_FUNCTION macro for forcing a function to be patched
with its original contents.  This is needed for the case where the patch
is a cumulative upgrade to a previous patch which modified the function,
and now the current patch needs to revert that change.

This is basically a copy/paste/modify job from KPATCH_IGNORE_FUNCTION
and kpatch_mark_ignored_functions_same().

### DIFF
--- a/kmod/patch/kpatch-macros.h
+++ b/kmod/patch/kpatch-macros.h
@@ -40,6 +40,17 @@ struct kpatch_unload {
 	void *__kpatch_ignore_func_##_fn __section(.kpatch.ignore.functions) = _fn;
 
 /*
+ * KPATCH_REVERT_FUNCTION macro
+ *
+ * This macro is for forcing a function to be patched with its original
+ * contents.  This is needed for the case where the patch is a cumulative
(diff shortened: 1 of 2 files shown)


jpoimboe:
This probably needs a patch author guide update as well, to document the steps:


1) compile v1 and note the list of changed functions

2) compile v2 (without KPATCH_REVERT_FUNCTION macro) and note the
   list of changed functions

3) for those functions which changed in v1 and did not change in v2,
   annotate them with KPATCH_REVERT_FUNCTION

jpoimboe:
We had problems testing this macro in a case where the function symbol had an isra suffix.  Adding a reference to the function name caused gcc to change its inlining decision for the function, causing the isra suffix to get removed, and then a seg fault happened.

So I'm going to drop this patch for now.  We can use #716 to track the fact that we'll need to bring back the replace feature, or something comparable, to make cumulative update reverts more straightforward.

PR #717, opened 2017-07-06
lookup: allow local symbols to be in any order

jpoimboe:
With #650, we found that using -ffunction-sections and -fdata-sections
sometimes causes GCC to output the local symbols in a different order in
the symbol table.  So don't assume they're in the same order, and
instead search all the locals.

This requires two passes: once going through the lookup table symbols
and once going through the .o symbols.  This is needed to make sure
there aren't any extra symbols in one of the files.

I also reorganized the code a bit to simplify it.

/cc @kamalesh-babulal


kamalesh-babulal:
> @jpoimboe Thanks for the patch. It resolves the re-ordering issue seen with testcase data-read-mostly.patch, I will try and run with other testcases also.

joe-lawrence:
👍 Thanks for testing @kamalesh-babulal !

PR #723, opened 2017-07-18
kpatch-build: fix errors messages for missing files/dirs

jpoimboe:
If you give kpatch-build a bad argument for the '-s' option, it shows
the following error:

  $ kpatch-build/kpatch-build -s foo
  ERROR: source dir  not found.

The supplied 'foo' argument isn't printed as intended.

Also fix some other options which have a similar issue.


joe-lawrence:
👍

PR #726, opened 2017-08-16
kpatch-build: check if DEBUG_INFO is enabled

rudis:
Without this option kpatch-build dies with "ERROR: can't find special
struct alt_instr size.".

I just noticed this when trying to build a patch for a custom kernel which didn't have this option enabled. This error message would've helped.


jpoimboe:
Thanks.  I'm pretty sure the previous CONFIG_DEBUG_KERNEL check was a typo, and was meant to be CONFIG_DEBUG_INFO, so you should be able to s/DEBUG_KERNEL/DEBUG_INFO/.

rudis:
Thanks, updated.

jpoimboe:
👍

Issue #714, opened 2017-06-21
memory.o: symbol changed sections: .LC39


disaster123:
Hello,

i'm trying to apply this simply looking patch on top of my 4.4 kernel via kpatch but build fails with:

...
Reading special section data
Building original kernel
Building patched kernel
Extracting new and modified ELF sections
ERROR: memory.o: symbol changed sections: .LC39
/usr/lib/kpatch/create-diff-object: unreconcilable difference
mmap.o: changed function: __split_vma.isra.38
mmap.o: changed function: copy_vma
mmap.o: changed function: unmapped_area_topdown
mmap.o: changed function: SyS_remap_file_pages
mmap.o: changed function: unmapped_area
ERROR: 1 error(s) encountered. Check /root/.kpatch/build.log for more details.

Patch:
```
 mm/memory.c | 7 ++++---
 1 file changed, 4 insertions(+), 3 deletions(-)

diff --git a/mm/memory.c b/mm/memory.c
index 380b731..581cc73 100644
--- a/mm/memory.c
+++ b/mm/memory.c
@@ -2684,9 +2684,10 @@ static int do_anonymous_page(struct mm_struct *mm, struct vm_area_struct *vma,
 		return VM_FAULT_SIGBUS;
 
 	/* Check if we need to add a guard page to the stack */
-	if ((vma->vm_flags & (VM_GROWSDOWN|VM_GROWSUP)) &&
-			expand_stack(vma, address) < 0)
-		return VM_FAULT_SIGSEGV;
+	if (stack_guard_area(vma, address)) {
+		if (expand_stack(vma, address) < 0)
+			return VM_FAULT_SIGSEGV;
+	}
 
 	/* Use the zero-page for reads */
 	if (!(flags & FAULT_FLAG_WRITE) && !mm_forbids_zeropage(mm)) {
```

Greets,
Stefan

rudis:
I came across a similar issue and the referenced patch fixes it for me (I think). Could you test my patch to see if it helps. I tried to find the kernel you used but couldn't find this hunk.

disaster123:
It's a custom kernel based on the opensuse 42.3 kernel ;-)

jpoimboe:
@disaster123 I checked the opensuse 42.3 kernel (kernel-default-4.4.79-19.1) but this patch doesn't apply to it either.

disaster123:
[...]

PR #727, opened 2017-08-18
Consider local constants as twins if the sections contain the same data

rudis:
While building a local patch is stumbled over the same (or a similar issue) than #714. The problem seems to be that the section for the local constant (.LC1 in my case) is renamed (similar to gcc's name mangling for functions) and thus not considered as twin section by create-diff-object. The attached patch compares the content of the sections of the twin symbol and marks the section as twin if the content is the same.

Note that I'm not sure if this really is the correct solution as I'm not entirely sure what this local constant is actually used for (it seems related to the kernels __bug_table) and why the section sometimes changes.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -940,6 +940,52 @@ static void kpatch_correlate_static_local_variables(struct kpatch_elf *base,
 	}
 }
 
+/*
+ * At least section .rela__bug_table may contain local constants.  They may
+ * end up in different sections (e.g. .rodata.$FUNC.isra.37.str1.1 and
+ * .rodata.$FUNC.isra.36.str1.1) due to gcc's function name mangling (see
+ * kpatch_rename_mangled_functions()) and are therefore not considered as
+ * twins.  If the sections have the same content then consider them as twins.
+ */
+static void kpatch_correlate_local_constants(struct kpatch_elf *base,
+					     struct kpatch_elf *patched)
+{
+	struct symbol *sym;
+	struct symbol *sym1, *sym2;
+	Elf_Data *dat1, *dat2;
+
+	list_for_each_entry(sym, &base->symbols, list) {
+		if (sym->type != STT_NOTYPE)
+			continue;
+		if (strncmp(".LC", sym->name, 3))
+			continue;
+
+		/*
+		 * Only consider the case if the symbols are twins but the
+		 * their sections are not.
+		 */
+		sym1 = sym;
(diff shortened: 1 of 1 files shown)


jpoimboe:
Thanks.  The .LC1 symbols are just strings.  GCC assigns all strings a .LCx label, but most of them end up being relocated by their section (e.g., rodata.*.str.1.1) and so they don't end up in the symbol table.  You're right that the bug table is involved here, it refers to them by their .LCx names instead of by their section, so they *do* end up in the symbol table.

I'm not quite sure this is the right approach.  One issue is that the numbering of the .LC symbols might also change.  For example, if a BUG() is inserted towards the beginning of the object file, all the other .LC symbols will be offset by 1.

I'll need to think about it a little more.

jpoimboe:
@rudis do you have a patch which recreates this issue?

disaster123:
@jpoimboe check the patch from my issue: https://github.com/dynup/kpatch/issues/714

rudis:
[test.patch](https://github.com/dynup/kpatch/files/1235719/test.txt)

Yes. The attached patch reproduces it with the current Debian Stretch kernel. To get it either install the packages `linux-image-4.9.0-3-amd64` (for the config), `linux-image-4.9.0-3-amd64-dbg` (for vmlinux) and `linux-source-4.9` (for the source) on a Debian Stretch or download them from `https://packages.debian.org` and unpack them manually. Then unpack the kernel with `tar xf /usr/src/linux-source-4.9.tar.xz` and build it with `kpatch-build -s ./linux-source-4.9 -c /boot/config-4.9.0-3-amd64 -v /usr/lib/debug/vmlinux-4.9.0-3-amd64 test.patch`. This reproduces the issue for me:

```
Using source directory at /home/kpatch/linux-source-4.9
Testing patch file
checking file net/ipv6/ip6_output.c
checking file net/ipv4/ip_output.c
checking file net/ipv4/udp.c
Reading special section data
Building original kernel
Building patched kernel
ERROR: ip_output.o: symbol changed sections: .LC0
/usr/lib/kpatch/create-diff-object: unreconcilable difference
udp.o: changed function: udp_send_skb
ERROR: ip6_output.o: symbol changed sections: .LC1
/usr/lib/kpatch/create-diff-object: unreconcilable difference
ERROR: 2 error(s) encountered. Check /home/kpatch/.kpatch/build.log for more details.
```

jpoimboe:
Thanks @rudis , I was able to recreate.  I have a fix coming.

disaster123:
works fine - thanks

PR #728, opened 2017-08-18
kpatch-build: respect CPPFLAGS

rudis:
Helps applying compiler flags from distributions, e.g. Debian.

### DIFF
--- a/kpatch-build/Makefile
+++ b/kpatch-build/Makefile
@@ -1,7 +1,7 @@
 include ../Makefile.inc
 
-CFLAGS  += -I../kmod/patch -Iinsn -Wall -g -Werror
-LDFLAGS += -lelf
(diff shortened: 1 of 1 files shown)


jpoimboe:
👍

rudis:
Good idea, I've updated the patch to use make's implicit rules and also remove the compile rule which is also handled by make.

kpatch-build/Makefile
```diff
@@ -1,7 +1,7 @@
 include ../Makefile.inc
 
-CFLAGS  += -I../kmod/patch -Iinsn -Wall -g -Werror
-LDFLAGS += -lelf
+CFLAGS += -MMD -MP -I../kmod/patch -Iinsn -Wall -g
```
rudis: Sorry, no idea how that happened. Fixed version pushed.

jpoimboe:
👍

joe-lawrence:
Looks good, thanks @rudis

PR #729, opened 2017-08-18
Makefile.inc: replace uname -p with uname -m

rudis:
uname -p returns "unknown" on many systems effectively breaking the
build. Replace it with uname -m which yields the correct architecture.

I didn't touch kmod/patch/Makefile because I don't know what values uname -m will return on non-x86 architectures.

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -3,7 +3,7 @@ CC    = gcc
 
 INSTALL = /usr/bin/install
 
-ARCH       = $(shell uname -p)
+ARCH       = $(shell uname -m)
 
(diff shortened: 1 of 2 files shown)


jpoimboe:
It should be fine to make the same change to kmod/patch/Makefile.  The only other arch we support right now is ppc64le, and I verified that it has the same output with uname -m and uname -p.

rudis:
Updated.

jpoimboe:
👍

joe-lawrence:
Looks good. Thanks, @rudis

PR #730, opened 2017-08-18
create-diff-object: rename mangled string sections

### LINKED ISSUE #714: memory.o: symbol changed sections: .LC39

jpoimboe:
When building the patch from PR #727, kpatch-build thought a .LC0 string
symbol moved to a different section.  In fact, its section got renamed
due to the GCC IPA SRA optimization.

Original object:
```
  [12] .rodata.__ip_append_data.isra.46.str1.1 PROGBITS     0000000000000000 000002e7 0000002e  1 AMS    0   0  1
  [13] .rodata.__ip_append_data.isra.46.str1.8 PROGBITS     0000000000000000 00000318 0000002a  1 AMS    0   0  8
  [14] .text.__ip_append_data.isra.46 PROGBITS     0000000000000000 00000350 00000a6d  0 AX     0   0 16
  [15] .rela.text.__ip_append_data.isra.46 RELA         0000000000000000 00053598 000001b0 24 I     110  14  8
```
Patched object:
```
  [12] .rodata.__ip_append_data.isra.47.str1.1 PROGBITS     0000000000000000 000002e7 0000002e  1 AMS    0   0  1
  [13] .rodata.__ip_append_data.isra.47.str1.8 PROGBITS     0000000000000000 00000318 0000002a  1 AMS    0   0  8
  [14] .text.__ip_append_data.isra.47 PROGBITS     0000000000000000 00000350 00000a5b  0 AX     0   0 16
  [15] .rela.text.__ip_append_data.isra.47 RELA         0000000000000000 00053820 000001b0 24 I     110  14  8
```
So we need to rename the .rodata.foo.isra.x.str1.* sections just like we
already do with the corresponding .text.foo.isra.x and
.rodata.foo.isra.x sections.

There's some ugly copy/paste code here that might be slightly improved
with a macro, but I don't feel inclined to do so :-)

Fixes #714.
Fixes #727.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -546,6 +546,24 @@ static void kpatch_correlate_symbols(struct list_head *symlist1, struct list_hea
 			if (is_special_static(sym1))
 				continue;
 
+			/*
+			 * The .LCx symbols point to strings, usually used for
+			 * the bug table.  Don't correlate and compare the
+			 * symbols themselves, because the suffix number might
+			 * change.
+			 *
+			 * If the symbol is used by the bug table (usual case),
+			 * it may get pulled in by
+			 * kpatch_regenerate_special_section().
+			 *
+			 * If the symbol is used outside of the bug table (not
+			 * sure if this actually happens anywhere), any string
+			 * changes will be detected elsewhere in rela_equal().
+			 */
+			if (sym1->type == STT_NOTYPE &&
+			    !strncmp(sym1->name, ".LC", 3))
+				continue;
+
 			/* group section symbols must have correlated sections */
 			if (sym1->sec &&
 			    sym1->sec->sh.sh_type == SHT_GROUP &&


jpoimboe:
Good idea.

rudis:
Thank you for the fast response and patch.

However, I'm not sure if this fixes the issue you noticed in my patch:
> One issue is that the numbering of the .LC symbols might also change. For example, if a BUG() is inserted towards the beginning of the object file, all the other .LC symbols will be offset by 1.

What happens if the symbol name for the constant changes? The sections should be marked as twins but does that also work for the (possibly renamed) symbols? I didn't test it yet, but wanted to mention it.

jpoimboe:
@rudis Thanks, you're right.  I added another commit to address that issue, though I wasn't able to figure out a good way to test it.

I also rebased the original commit with @joe-lawrence 's suggestion.

kpatch-build/create-diff-object.c
```diff
@@ -546,6 +546,16 @@ static void kpatch_correlate_symbols(struct list_head *symlist1, struct list_hea
 			if (is_special_static(sym1))
 				continue;
 
+			/*
+			 * The .LCx symbols are NOTYPE symbols which point to
```
rudis: Maybe make that explicit by checking the type of the symbol?

rudis:
Thanks. I'm still not entirely sure how this is supposed to work. So here's my current understanding, please correct me if it's wrong: The sections are considered as twins with the first patch (which is necessary as all sections must have a twin, the original bug) and then when .rela__bug_table is compared in kpatch_compare_correlated_rela_sections() we check that all relocations are equal (for string symbols such as local constants by comparing their value) and in the same order.

Now if any .LCx symbol changes we've already correlated the section so that's fine (Is the section content itself used in any comparison? Is there just one section for all local constants of one function?) and we skip the symbol during symbol correlation (as the name may change we can't find a twin). Then during relocation comparison we detect that .rela__bug_table has changed (or to be more precise the string (our .LCx string) referenced by its relocation) and then we will add a patched version of the symbol and table?

Is that assumption correct? If so then based on my limited knowledge the approach should work fine.

jpoimboe:
> Thanks. I'm still not entirely sure how this is supposed to work.

Yes, it's complicated :-)

> So here's my current understanding, please correct me if it's wrong: The sections are considered as twins with the first patch (which is necessary as all sections must have a twin, the original bug) and then when .rela__bug_table is compared in kpatch_compare_correlated_rela_sections() we check that all relocations are equal (for string symbols such as local constants by comparing their value) and in the same order.

Right, although the bug table section (and its rela section) is treated specially.  Regardless of whether it has changed, only a subset of the bug table is included in the new object.  So the bug table section comparison doesn't really matter.

However, it's possible that the .LCx symbols may also be used by functions.  In that case the comparison would matter.  A change in the string constant would result in the function being detected as changed due to the rela_equal() string check.

> Now if any .LCx symbol changes we've already correlated the section so that's fine (Is the section content itself used in any comparison?

Yes.

> Is there just one section for all local constants of one function?

Well, each function can have a str1.1 and a str1.8 section.  And also, IIRC, in certain cases, if two functions use the same string (e.g.., `__FILE__`), the string will be in the str1.1 section of function A, but function B will still reference it.

> and we skip the symbol during symbol correlation (as the name may change we can't find a twin). Then during relocation comparison we detect that .rela__bug_table has changed (or to be more precise the string (our .LCx string) referenced by its relocation) and then we will add a patched version of the symbol and table?

As I mentioned above, whether the bug table has changed or not, only part of the bug table is included in the new object file.  Only those bug table entries which are needed by changed or new functions are included.

And the str1.x sections are also treated specially.  They are always included in the new object file, regardless of whether they have changed.

So that leads to the question: why are we correlating and comparing these sections in the first place, if they're just going to be included anyway?  Need to think about that one.

jpoimboe:
Pushed a new version.  I dropped the first commit because I think renaming the .rodata.func.str1.x sections isn't actually necessary.  Those sections don't need to be compared, and they get included anyway.

rudis:
Thank you for the detailed explanation, it helped a lot! Last patch looks good and works fine here.

joe-lawrence:
👍

PR #732, opened 2017-08-29
kpatch-build: allow the number of make jobs to be specified using CPUS

gktrk:
Instead of always using the maximum number of CPUs available, allow
user to tune the number of make jobs using the environment variable
CPUS.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -326,6 +326,7 @@ usage() {
 	echo "		-s, --sourcedir	   Specify kernel source directory" >&2
 	echo "		-c, --config	   Specify kernel config file" >&2
 	echo "		-v, --vmlinux	   Specify original vmlinux" >&2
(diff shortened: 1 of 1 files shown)


gktrk:
Though I wonder if we should leave `CPUS` alone and use `JOBS` instead, defaulting to `CPUS` if not given

jpoimboe:
I think I'd prefer an optional `-j` command-line option, which can be passed on to make.  If the option isn't specified then we can keep the default.  What do you think?

gktrk:
Agreed. How about the newly pushed code?

jpoimboe:
Looks good, thanks @gktrk!

👍

gktrk:
Commit message updated

jpoimboe:
👍

joe-lawrence:
Thanks @gktrk ...  (btw, since there is a corresponding commit in the elivepatch-server project, I wonder if the make "--load-average" option is also interesting for your use case.)

PR #733, opened 2017-09-08
kpatch-build: support multiple .patch files

joe-lawrence:
This a lightly tested, RFC for teaching kpatch-build to operate on multiple input patches.

The main motivation behind this is to avoid using tools like combinediff to aggregate a bunch of patch files into a single input file for kpatch-build.  I have seen combinediff create some odd output, especially when multiple patches modify the same file/functions.

This implementation applies each patch in order to the kernel tree.  I ditched the early patch test (I don't know how to --dry-run multiple patches), but I kept the APPLIEDPATCH logic (the program seems to want to restore the SRCDIR to an unpatched state).

Maybe it would be cleaner to move this functionality out to a helper utility?  Basically copy combinediff, but have it work against a real tree that it can apply against, gaining full file context.  Suggestions welcome... this RFC was just a way to kick off the conversation.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -319,6 +319,13 @@ find_kobj() {
 	done
 }
 
+# Only allow alphanumerics and '_' and '-' in the module name.  Everything else
+# is replaced with '-'.  Also truncate to 48 chars so the full name fits in the
+# kernel's 56-byte module name array.
+module_name_string() {
+	echo ${1//[^a-zA-Z0-9_-]/-} |cut -c 1-48
+}
+
 usage() {
 	echo "usage: $(basename $0) [options] <patch file>" >&2
 	echo "		-h, --help	   Show this help message" >&2
@@ -335,7 +342,7 @@ usage() {
 	echo "		                   (not recommended)" >&2
 }
 
-options=$(getopt -o hr:s:c:v:j:t:o:d -l "help,sourcerpm:,sourcedir:,config:,vmlinux:,jobs:,target:,output:,debug,skip-gcc-check,skip-cleanup" -- "$@") || die "getopt failed"
+options=$(getopt -o hr:s:c:v:j:t:n:o:d -l "help,sourcerpm:,sourcedir:,config:,vmlinux:,jobs:,target:,name:,output:,debug,skip-gcc-check,skip-cleanup" -- "$@") || die "getopt failed"
 
 eval set -- "$options"
 
@@ -375,6 +382,10 @@ while [[ $# -gt 0 ]]; do
 		;;
 	-t|--target)
 		TARGETS="$TARGETS $2"
+		shift
+		;;
+	-n|--name)
+		MODNAME=$(module_name_string "$2")
 		shift
 		;;
 	-o|--output)
@@ -430,15 +441,16 @@ fi
 
 [[ -z $TARGETS ]] && TARGETS="vmlinux modules"
 
-PATCHNAME=$(basename "$PATCHFILE")
-if [[ "$PATCHNAME" =~ \.patch$ ]] || [[ "$PATCHNAME" =~ \.diff$ ]]; then
-	PATCHNAME="${PATCHNAME%.*}"
-fi
+# If no kpatch module name was provided on the command line, derive one
+# from the .patch filename
+if [[ -z $MODNAME ]] ; then
+	MODNAME=$(basename "$PATCHFILE")
+	if [[ "$MODNAME" =~ \.patch$ ]] || [[ "$MODNAME" =~ \.diff$ ]]; then
+		MODNAME="${MODNAME%.*}"
+	fi
 
-# Only allow alphanumerics and '_' and '-' in the module name.  Everything else
-# is replaced with '-'.  Also truncate to 48 chars so the full name fits in the
-# kernel's 56-byte module name array.
-PATCHNAME=$(echo ${PATCHNAME//[^a-zA-Z0-9_-]/-} |cut -c 1-48)
+	MODNAME=$(module_name_string "$MODNAME")
+fi
 
 source /etc/os-release
 DISTRO=$ID
@@ -664,7 +676,7 @@ for i in $FILES; do
 	if [[ -e "orig/$i" ]]; then
 		# create-diff-object orig.o patched.o kernel-object output.o Module.symvers patch-mod-name
 		"$TOOLSDIR"/create-diff-object "orig/$i" "patched/$i" "$KOBJFILE" \
-			"output/$i" "$SRCDIR/Module.symvers" "kpatch_${PATCHNAME//-/_}" 2>&1 |tee -a "$LOGFILE"
+			"output/$i" "$SRCDIR/Module.symvers" "kpatch_${MODNAME//-/_}" 2>&1 |tee -a "$LOGFILE"
 		check_pipe_status create-diff-object
 		# create-diff-object returns 3 if no functional change is found
 		[[ $rc -eq 0 ]] || [[ $rc -eq 3 ]] || ERROR=$(expr $ERROR "+" 1)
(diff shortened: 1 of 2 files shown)


kpatch-build/kpatch-build
```diff
@@ -592,10 +617,6 @@ fi
 # unsupported kernel option checking: CONFIG_DEBUG_INFO_SPLIT
 grep -q "CONFIG_DEBUG_INFO_SPLIT=y" "$CONFIGFILE" && die "kernel option 'CONFIG_DEBUG_INFO_SPLIT' not supported"
 
-echo "Testing patch file"
-cd "$SRCDIR" || die
-patch -N -p1 --dry-run < "$PATCHFILE" || die "source patch file failed to apply"
-cp "$PATCHFILE" "$APPLIEDPATCHFILE" || die
```
jpoimboe: It would be nice to keep this test.

What if we have an apply_patches() function along with a remove_patches() function.  They could use a global numeric variable to track the number of patches currently applied.  apply_patches() and remove_patches() could be used here, and then apply_patches() could be used again in the "Building original kernel" section and remove_patches() could be used again in the cleanup.

Then we could get rid of APPLIEDPATCHFILE, as well as the SRC_LIST stuff.  Thoughts?

kpatch-build/kpatch-build
```diff
@@ -395,16 +408,27 @@ while [[ $# -gt 0 ]]; do
...
+		[[ "$1" = "--" ]] && shift && continue
+		[[ ! -f "$1" ]] && die "patch file '$1' not found"
+		PATCH_LIST+=($(readlink -f "$1"))
 		;;
 	esac
 	shift
 done
 
+if [[ ${#PATCH_LIST[@]} = 0 ]]; then
+        PATCH_LIST=(*.patch)
```
jpoimboe: So if no patches are specified, I think this just uses the patches in the current directory?  If so, I'd rather not go there, it could create some "surprises".

kpatch-build/kpatch-build
```diff
@@ -430,15 +454,16 @@ fi
...
 [[ -z $TARGETS ]] && TARGETS="vmlinux modules"
 
-PATCHNAME=$(basename "$PATCHFILE")
-if [[ "$PATCHNAME" =~ \.patch$ ]] || [[ "$PATCHNAME" =~ \.diff$ ]]; then
-	PATCHNAME="${PATCHNAME%.*}"
-fi
+# If no kpatch module name was provided on the command line, derive one
+# from the first .patch filename
+if [[ -z $MODNAME ]] ; then
+	MODNAME=$(basename "${PATCH_LIST[0]}")
```
jpoimboe: Naming the patch based on the first .patch name could be confusing.  Maybe just call it kpatch-patch?

joe-lawrence:
Thanks for taking a look.  What do you think about a separate utility that we could use instead?  (ie, `combinediff` with a reference tree)  I would think that might be more re-usable than retrofitting `kpatch-build` to handle multiple patches.

We could still spin off a few of the things noticed here: module naming (by user), `{k,live}patch` auto prefixing, etc.

jpoimboe:
Unless there would be another user of the multiple patch functionality outside of kpatch-build, I'd rather keep the functionality in the script because it makes maintenance easier IMO.

joe-lawrence:
I can write it either way, it's just that this script does a lot of interesting stuff.  Downloading distro kernels, combining patches, etc. that could be considered stand alone tools, at least in the traditional unix do-one-thing sense.  This feature is mostly implemented by this RFC, so completing it won't be a big deal, but just thought I toss the external tool idea out there.

kpatch-build/kpatch-build
joe-lawrence: Agreed, I'll dump a "no patches specified" msg instead.

kpatch-build/kpatch-build
joe-lawrence: That seems like a reasonable default.

jpoimboe:
True, it is a monolithic frankenscript, but I like it that way.  Less moving parts ;-)

kpatch-build/kpatch-build
joe-lawrence: Yeah, the `APPLIEDPATCHFILE` thing is a little wart, so creating an `apply_patches()` function with some kind of memory of which ones were applied sounds nicer.

kpatch-build/kpatch-build
```diff
@@ -664,7 +701,7 @@ for i in $FILES; do
 	if [[ -e "orig/$i" ]]; then
 		# create-diff-object orig.o patched.o kernel-object output.o Module.symvers patch-mod-name
 		"$TOOLSDIR"/create-diff-object "orig/$i" "patched/$i" "$KOBJFILE" \
-			"output/$i" "$SRCDIR/Module.symvers" "kpatch_${PATCHNAME//-/_}" 2>&1 |tee -a "$LOGFILE"
+			"output/$i" "$SRCDIR/Module.symvers" "kpatch_${MODNAME//-/_}" 2>&1 |tee -a "$LOGFILE"
```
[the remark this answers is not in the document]
joe-lawrence: Good point, I'll fix `{k,live}patch` auto naming into its own commit, then remove the `kpatch_` prefix from a user named module.

joe-lawrence:
Still a work in progress, but the patches-applied list/counter works.  On the topic of module prefixes, any preferences for "livepatch" vs. "klp"?

jpoimboe:
I prefer livepatch because that's the "external" name of the component, and "klp" is an internal naming detail (which, personally, I regret).

From a quick glance, the code looks great.  Let me know when it's ready for review.

joe-lawrence:
This should be ready for review.  Below is the testing I did on 4.13, so you'll see "livepatch" prefixes where appropriate.  (Similar results for "kpatch" generated modules for those kernels.)  All generated .ko's load and function successfully.  
```
# No patches on cmdline
% kpatch-build -s ~/linux -t vmlinux
ERROR: no patch file(s) specified
usage: kpatch-build [options] <patch1 ... patchN>
                patchN             Input patchfile(s)
                -h, --help         Show this help message
                -r, --sourcerpm    Specify kernel source RPM
                -s, --sourcedir    Specify kernel source directory
                -c, --config       Specify kernel config file
                -v, --vmlinux      Specify original vmlinux
                -j, --jobs         Specify the number of make jobs
                -t, --target       Specify custom kernel build targets
                -o, --output       Specify output folder
                -d, --debug        Keep scratch files in /tmp
                --skip-cleanup     Skip post-build cleanup
                --skip-gcc-check   Skip gcc version matching check
                                   (not recommended)

# Force a patch problem
% kpatch-build -s ~/linux -n joe cmdline-string.patch cmdline-string.patch 
Testing patch file(s)
ERROR: /root/kpatch.joe/test/integration/fedora-25/cmdline-string.patch file
failed to apply. Check /root/.kpatch/build.log for more details.
% tail /root/.kpatch/build.log 
patching file fs/proc/cmdline.c
Reversed (or previously applied) patch detected!  Skipping patch.
1 out of 1 hunk ignored

# Multiple input patches, provide a module name
% kpatch-build -s ~/linux -n joe cmdline-string.patch meminfo-string.patch -t vmlinux
Testing patch file(s)
Reading special section data
Building original kernel
Building patched kernel
cmdline.o: changed function: cmdline_proc_show
Patched objects: vmlinux
Building patch module: joe.ko
SUCCESS

# Multiple input patches, use default module name
% kpatch-build -s ~/linux cmdline-string.patch meminfo-string.patch -t vmlinux
Testing patch file(s)
Reading special section data
Building original kernel
Building patched kernel
cmdline.o: changed function: cmdline_proc_show
Patched objects: vmlinux
Building patch module: livepatch-patch.ko
SUCCESS

# Single input patch, use default module name
% kpatch-build -s ~/linux cmdline-string.patch -t vmlinux
Testing patch file(s)
Reading special section data
Building original kernel
Building patched kernel
cmdline.o: changed function: cmdline_proc_show
Patched objects: vmlinux
Building patch module: livepatch-cmdline-string.ko
SUCCESS

% ls -l *.ko
-rw-r--r--. 1 root root 715024 Sep 20 16:07 joe.ko
-rw-r--r--. 1 root root 461344 Sep 20 16:23 livepatch-cmdline-string.ko
-rw-r--r--. 1 root root 715080 Sep 20 16:15 livepatch-patch.ko
```

PR #737, opened 2017-09-20
kpatch-build: do symbol/section bundling in create-diff-object

### LINKED ISSUE #700: symbol load_elf_binary at offset 3536 within section .text.load_elf_binary, expected 0
[patch.txt](https://github.com/dynup/kpatch/files/973473/patch.txt)

I'm testing a patch (see attached) for CVE-2015-1593 against RHEL7.3 GA + kernel-3.10.0-200.el7.x86_64 (rebuilt by RHEL7.3 GA toolset) and was running into the following kpatch-build error:
```
Skipping cleanup
Fedora/Red Hat distribution detected
Downloading kernel source for 3.10.0-200.el7.x86_64
Unpacking kernel source
Testing patch file
checking file arch/x86/mm/mmap.c
checking file fs/binfmt_elf.c
Reading special section data
Building original kernel
Building patched kernel
mmap.o: changed function: stack_maxrandom_size
mmap.o: changed function: arch_pick_mmap_layout
binfmt_elf.o: changed function: load_elf_binary
compat_binfmt_elf.o: changed function: load_elf_binary
Patched objects: vmlinux
Building patch module: kpatch-klp.ko
/usr/local/libexec/kpatch/create-kpatch-module: ERROR: tmp_output.o: kpatch_create_symbol_list: 331: symbol load_elf_binary at offset 3536 within section .text.load_elf_binary, expected 0
```
It's interesting to note that kpatch-build reports that both `binfmt_elf.o` and `compat_binfmt_elf.o` modify a function called `load_elf_binary`.  (Check out [fs/compat_binfmt_elf.c](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/compat_binfmt_elf.c#n131) for it #include's [binfmt_elf.c](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/binfmt_elf.c#n682)!)  I think kpatch-build is getting tripped up on these duplicate functions as:
```
TMP=~/.kpatch/tmp/patch/tmp_output.o
readelf --wide --symbols $TMP | grep load_elf_binary
Symbol table '.symtab' contains 52 entries:
   Num:    Value          Size Type    Bind   Vis      Ndx Name
   ...
    24: 0000000000000000  3523 FUNC    LOCAL  DEFAULT    5 load_elf_binary
    26: 0000000000000dd0  3656 FUNC    LOCAL  DEFAULT    5 load_elf_binary
```
Note that the first `load_elf_binary` symbol value is 0 and the second is 0xdd0 = 3536, the found and expected values in the kpatch-build complaint.

jpoimboe:
kpatch-elf.c is used by binaries other than create-diff-object, but
create-diff-object is the only one that cares about "bundling".  Move
the bundling to create-diff-object.

Fixes #700.


kpatch-build/create-diff-object.c
```diff
@@ -81,6 +81,60 @@ struct special_section {
...
+ */
+static void kpatch_bundle_symbols(struct kpatch_elf *kelf)
+{
+	struct symbol *sym;
+
+	list_for_each_entry(sym, &kelf->symbols, list) {
+		if (is_bundleable(sym)) {
+			if (sym->sym.st_value != 0) {
+#ifdef __powerpc__
+				if (!is_localentry_sym(sym))
```
joe-lawrence: For powerpc, I think you also need to move is_localentry_sym() implementation into this file (or prototype in a header file).  Can you do a quick build on power to verify?

jpoimboe:
NAK, apparently the bundling assumption is still being relied on somewhere I didn't expect.  This fix allows the kpatch to be built, but when it loads, it produces warnings because it tries to apply the second function's dynrelas at the wrong addresses (using the first function's base address but the second function's dynrela offsets).

jpoimboe:
I found the issue, it turns out that create-kpatch-module was relying on bundling, though I believe it shouldn't have been.  So I added a commit to remove its reliance on bundling.  I think this code is good now.  It fixes #700, and also passed the centos7 quick integration tests.

jpoimboe:
Oops, I forgot to fix the ppc64le issue and test it on a ppc64le box.

jpoimboe:
Ok, it builds on ppc64le now.  I also tested with livepatch and found a similar issue there, so I added a corresponding fix for create-klp-module.

joe-lawrence:
👍

PR #740, opened 2017-09-25
Testing updates

joe-lawrence:
This updates the test patches for CentOS 7.4 GA release kernel (there were a few patches that needed rebasing) and drops the combine-patch logic out of kpatch-test.


joe-lawrence:
[quick.log](https://github.com/dynup/kpatch/files/1329604/quick.log)

Attached is a log of the "quick" integration test, which tries to build as many test patches into a single combined module as possible.  After this change, it feeds all the patches to kpatch-build on the command line and lets it combine them (instead of running combinediff).

PR #743, opened 2017-10-04
kpatch-build: create-klp-module: set default arguments.no_klp_arch

joe-lawrence:
Valgrind complains about an uninitialized variable in
create-klp-module.c:

  ==4412== Conditional jump or move depends on uninitialised value(s)
  ==4412==    at 0x402846: main (create-klp-module.c:497)

This warning refers to main()'s struct arguments stack variable,
precisely its .no_klp_arch member.  Initialize it to zero to avoid the
complaint.

### DIFF
--- a/kpatch-build/create-klp-module.c
+++ b/kpatch-build/create-klp-module.c
@@ -449,7 +449,7 @@ int main(int argc, char *argv[])
 	char *strings;
 	int ksyms_nr, krelas_nr;
 
-	arguments.debug = 0;
+	memset(&arguments, 0, sizeof(arguments));
 	argp_parse (&argp, argc, argv, 0, 0, &arguments);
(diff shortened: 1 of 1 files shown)


joe-lawrence:
I was about to add this one to PR #742 when it got merged.  After this change, create-diff-object, create-klp-module, and create-kpatch-module should have clean valgrind output (except for the memory leaks) for the sample test cmdline-string.patch, both kpatch and livepatch versions.

jpoimboe:
Nice find.  How about memset'ing the struct to zero so we don't make the same mistake in the future?

joe-lawrence:
Sure, memset is probably a better future proof fix... will re-bake and send out tomorrow.

joe-lawrence:
v2 using memset.

PR #744, opened 2017-10-07
Shell quoting improvements and additional cleanup

rudis:
Hi,

I tried to load a kpatch from a path containing spaces (it contained the output of `uname -v` to match only  a specific kernel) and it failed due to invalid quoting.

The first patch should correct this issue by quoting all necessary variables. The second is additional cleanup found be shellcheck. I'm not sure about the last one, as I don't know if the macro is a string (in that case I don't know how this could ever work) or if it's a number (then the quotes could be removed instead).

I didn't try every possible invocation of kpatch so I've might have missed some issues with this patch (shell quoting is quite delicate), please review carefully.

Regards
Simon

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -286,16 +286,16 @@ find_parent_obj() {
 		num=0
 		if [[ -n "$last_deep_find" ]]; then
 			parent="$(grep -l "$grepname" "$last_deep_find"/.*.cmd | grep -v "$pdir/.${file}.cmd" |head -n1)"
-			num="$(grep -l "$grepname" "$last_deep_find"/.*.cmd | grep -v "$pdir/.${file}.cmd" |wc -l)"
+			num="$(grep -l "$grepname" "$last_deep_find"/.*.cmd | grep -vc "$pdir/.${file}.cmd")"
 		fi
 		if [[ "$num" -eq 0 ]]; then
-			parent="$(find * -name ".*.cmd" | xargs grep -l "$grepname" | grep -v "$pdir/.${file}.cmd" |head -n1)"
-			num="$(find * -name ".*.cmd" | xargs grep -l "$grepname" | grep -v "$pdir/.${file}.cmd" | wc -l)"
+			parent="$(find * -name ".*.cmd" -print0 | xargs -0 grep -l "$grepname" | grep -v "$pdir/.${file}.cmd" |head -n1)"
+			num="$(find * -name ".*.cmd" -print0 | xargs -0 grep -l "$grepname" | grep -vc "$pdir/.${file}.cmd")"
 			[[ "$num" -eq 1 ]] && last_deep_find="$(dirname "$parent")"
 		fi
 	else
 		parent="$(grep -l "$grepname" "$dir"/.*.cmd | grep -v "$dir/.${file}.cmd" |head -n1)"
-		num="$(grep -l "$grepname" "$dir"/.*.cmd | grep -v "$dir/.${file}.cmd" | wc -l)"
+		num="$(grep -l "$grepname" "$dir"/.*.cmd | grep -vc "$dir/.${file}.cmd")"
 	fi
 
 	[[ "$num" -eq 0 ]] && PARENT="" && return
@@ -717,7 +717,7 @@ for i in $FILES; do
 			"output/$i" "$SRCDIR/Module.symvers" "${MODNAME//-/_}" 2>&1 |tee -a "$LOGFILE"
 		check_pipe_status create-diff-object
 		# create-diff-object returns 3 if no functional change is found
-		[[ "$rc" -eq 0 ]] || [[ "$rc" -eq 3 ]] || ERROR="$(expr $ERROR "+" 1)"
+		[[ "$rc" -eq 0 ]] || [[ "$rc" -eq 3 ]] || ERROR="$((ERROR + 1))"
 		if [[ "$rc" -eq 0 ]]; then
 			[[ -n "$ERROR_IF_DIFF" ]] && die "$ERROR_IF_DIFF"
 			CHANGED=1
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -286,17 +286,17 @@ find_parent_obj() {
 	if [[ "$DEEP_FIND" -eq 1 ]]; then
 		num=0
 		if [[ -n "$last_deep_find" ]]; then
-			parent="$(grep -l "$grepname" "$last_deep_find"/.*.cmd | grep -v "$pdir/.${file}.cmd" |head -n1)"
-			num="$(grep -l "$grepname" "$last_deep_find"/.*.cmd | grep -vc "$pdir/.${file}.cmd")"
+			parent="$(grep -l "$grepname" "$last_deep_find"/.*.cmd | grep -Fv "$pdir/.${file}.cmd" |head -n1)"
+			num="$(grep -l "$grepname" "$last_deep_find"/.*.cmd | grep -Fvc "$pdir/.${file}.cmd")"
 		fi
 		if [[ "$num" -eq 0 ]]; then
-			parent="$(find ./* -name ".*.cmd" -print0 | xargs -0 grep -l "$grepname" | grep -v "$pdir/.${file}.cmd" | cut -c3- | head -n1)"
-			num="$(find ./* -name ".*.cmd" -print0 | xargs -0 grep -l "$grepname" | grep -vc "$pdir/.${file}.cmd")"
+			parent="$(find ./* -name ".*.cmd" -print0 | xargs -0 grep -l "$grepname" | grep -Fv "$pdir/.${file}.cmd" | cut -c3- | head -n1)"
+			num="$(find ./* -name ".*.cmd" -print0 | xargs -0 grep -l "$grepname" | grep -Fvc "$pdir/.${file}.cmd")"
 			[[ "$num" -eq 1 ]] && last_deep_find="$(dirname "$parent")"
 		fi
 	else
-		parent="$(grep -l "$grepname" "$dir"/.*.cmd | grep -v "$dir/.${file}.cmd" |head -n1)"
-		num="$(grep -l "$grepname" "$dir"/.*.cmd | grep -vc "$dir/.${file}.cmd")"
+		parent="$(grep -l "$grepname" "$dir"/.*.cmd | grep -Fv "$dir/.${file}.cmd" |head -n1)"
+		num="$(grep -l "$grepname" "$dir"/.*.cmd | grep -Fvc "$dir/.${file}.cmd")"
 	fi
 
 	[[ "$num" -eq 0 ]] && PARENT="" && return
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -290,8 +290,8 @@ find_parent_obj() {
 			num="$(grep -l "$grepname" "$last_deep_find"/.*.cmd | grep -vc "$pdir/.${file}.cmd")"
 		fi
 		if [[ "$num" -eq 0 ]]; then
-			parent="$(find * -name ".*.cmd" -print0 | xargs -0 grep -l "$grepname" | grep -v "$pdir/.${file}.cmd" |head -n1)"
-			num="$(find * -name ".*.cmd" -print0 | xargs -0 grep -l "$grepname" | grep -vc "$pdir/.${file}.cmd")"
+			parent="$(find ./* -name ".*.cmd" -print0 | xargs -0 grep -l "$grepname" | grep -v "$pdir/.${file}.cmd" | cut -c3- | head -n1)"
+			num="$(find ./* -name ".*.cmd" -print0 | xargs -0 grep -l "$grepname" | grep -vc "$pdir/.${file}.cmd")"
 			[[ "$num" -eq 1 ]] && last_deep_find="$(dirname "$parent")"
 		fi
 	else
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -167,10 +167,28 @@ find_core_symvers() {
 	[[ -e "$SYMVERSFILE" ]]
 }
 
+gcc_version_from_file() {
+	readelf -p .comment "$1" | grep -o 'GCC:.*'
+}
 gcc_version_check() {
+	local c="$TEMPDIR/test.c" o="$TEMPDIR/test.o"
+	local out gccver kgccver
+
+	# gcc --version varies between distributions therefore extract version
+	# by compiling a test file and compare it to vmlinux's version.
+	echo 'void main(void) {}' > "$c"
+	out="$(gcc -c -pg -ffunction-sections -o "$o" "$c" 2>&1)"
+	gccver="$(gcc_version_from_file "$o")"
+	kgccver="$(gcc_version_from_file "$VMLINUX")"
+	rm -f "$c" "$o"
+
+	if [[ -n "$out" ]]; then
+		warn "gcc >= 4.8 required for -pg -ffunction-settings"
+		echo "gcc output: $out"
+		return 1
+	fi
+
 	# ensure gcc version matches that used to build the kernel
-	local gccver="$(gcc --version |head -n1 |cut -d' ' -f3-)"
-	local kgccver="$(readelf -p .comment "$VMLINUX" |grep GCC: | tr -s ' ' | cut -d ' ' -f6-)"
 	if [[ "$gccver" != "$kgccver" ]]; then
 		warn "gcc/kernel version mismatch"
 		echo "gcc version:    $gccver"
@@ -178,23 +196,6 @@ gcc_version_check() {
 		echo "Install the matching gcc version (recommended) or use --skip-gcc-check"
 		echo "to skip the version matching enforcement (not recommended)"
 		return 1
-	fi
-
-	# On ppc64le, livepatch needs the GCC -mprofile-kernel flag.  A script
-	# is run during the kernel build to ensure the flag is present:
-	#
-	#   <linux>/arch/powerpc/tools/gcc-check-mprofile-kernel.sh
-	#
-	# Postpone the check to the kernel build instead of duplicating it here.
-	if [[ "$ARCH" = "ppc64le" ]]; then
-		return
-	else
-		# ensure gcc version is >= 4.8
-		gccver="$(echo "$gccver" |cut -d'.' -f1,2)"
-		if [[ "$gccver" < 4.8 ]]; then
-			warn "gcc >= 4.8 required"
-			return 1
-		fi
 	fi
 
 	return
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -286,7 +286,7 @@ find_parent_obj() {
 	if [[ "$DEEP_FIND" -eq 1 ]]; then
 		num=0
 		if [[ -n "$last_deep_find" ]]; then
-			parent="$(grep -l "$grepname" "$last_deep_find"/.*.cmd | grep -Fv "$pdir/.${file}.cmd" |head -n1)"
+			parent="$(grep -l "$grepname" "$last_deep_find"/.*.cmd | grep -Fv "$pdir/.${file}.cmd" | head -n1)"
 			num="$(grep -l "$grepname" "$last_deep_find"/.*.cmd | grep -Fvc "$pdir/.${file}.cmd")"
 		fi
 		if [[ "$num" -eq 0 ]]; then
@@ -295,7 +295,7 @@ find_parent_obj() {
 			[[ "$num" -eq 1 ]] && last_deep_find="$(dirname "$parent")"
 		fi
 	else
-		parent="$(grep -l "$grepname" "$dir"/.*.cmd | grep -Fv "$dir/.${file}.cmd" |head -n1)"
+		parent="$(grep -l "$grepname" "$dir"/.*.cmd | grep -Fv "$dir/.${file}.cmd" | head -n1)"
 		num="$(grep -l "$grepname" "$dir"/.*.cmd | grep -Fvc "$dir/.${file}.cmd")"
 	fi
 
@@ -344,7 +344,7 @@ find_kobj() {
 # is replaced with '-'.  Also truncate to 48 chars so the full name fits in the
 # kernel's 56-byte module name array.
 module_name_string() {
-	echo "${1//[^a-zA-Z0-9_-]/-}" |cut -c 1-48
+	echo "${1//[^a-zA-Z0-9_-]/-}" | cut -c 1-48
 }
 
 usage() {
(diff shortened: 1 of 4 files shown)


kpatch-build/kpatch-build
```diff
@@ -262,64 +262,64 @@ find_special_section_data() {
...
+	grepname="${1%.o}"
+	grepname="$grepname\.o"
+	if [[ "$DEEP_FIND" -eq 1 ]]; then
 		num=0
-		if [[ -n $last_deep_find ]]; then
-			parent=$(grep -l $grepname $last_deep_find/.*.cmd | grep -v $pdir/.${file}.cmd |head -n1)
-			num=$(grep -l $grepname $last_deep_find/.*.cmd | grep -v $pdir/.${file}.cmd |wc -l)
+		if [[ -n "$last_deep_find" ]]; then
+			parent="$(grep -l "$grepname" "$last_deep_find"/.*.cmd | grep -v "$pdir/.${file}.cmd" |head -n1)"
+			num="$(grep -l "$grepname" "$last_deep_find"/.*.cmd | grep -v "$pdir/.${file}.cmd" |wc -l)"
```
jpoimboe: Should `${file}` be quoted here (and below) in case it has spaces (probably won't ever happen)?

kpatch-build/kpatch-build
```diff
@@ -262,64 +262,64 @@ find_special_section_data() {
...
-	PARENT=$(basename $parent)
-	PARENT=${PARENT#.}
-	PARENT=${PARENT%.cmd}
-	PARENT=$dir/$PARENT
-	[[ ! -e $PARENT ]] && die "ERROR: can't find parent $PARENT for $1"
+	dir="$(dirname "$parent")"
+	PARENT="$(basename "$parent")"
+	PARENT="${PARENT#.}"
+	PARENT="${PARENT%.cmd}"
+	PARENT="$dir/$PARENT"
```
jpoimboe: Should `$PARENT` be quoted?

kpatch/kpatch
```diff
@@ -74,14 +74,14 @@ __find_module () {
 	MODULE="$1"
 	[[ -f "$MODULE" ]] && return
 
-	MODULE=$INSTALLDIR/$(uname -r)/"$1"
+	MODULE="$INSTALLDIR/$(uname -r)/$1"
```
jpoimboe: The `$1` needs to be quoted.

kpatch-build/kpatch-build
```diff
@@ -118,7 +118,7 @@ clean_cache() {
 
 check_pipe_status() {
 	rc="${PIPESTATUS[0]}"
-	if [[ $rc = 139 ]]; then
+	if [[ "$rc" = 139 ]]; then
```
jpoimboe: I notice several cases like this, where the quotes aren't necessary because it's comparing a variable which isn't based on a filename.  I tried to comment on many of them below.

kpatch-build/kpatch-build
```diff
@@ -157,10 +157,10 @@ find_core_symvers() {
 		SYMVERSFILE="$DATADIR/core/Module.symvers"
 	elif [[ -e "$SCRIPTDIR/../libexec/kpatch/create-diff-object" ]]; then
 		# installation path
-		if [[ -e $SCRIPTDIR/../lib/kpatch/$ARCHVERSION/Module.symvers ]]; then
-			SYMVERSFILE="$(readlink -f $SCRIPTDIR/../lib/kpatch/$ARCHVERSION/Module.symvers)"
+		if [[ -e "$SCRIPTDIR/../lib/kpatch/$ARCHVERSION/Module.symvers" ]]; then
+			SYMVERSFILE="$(readlink -f "$SCRIPTDIR/../lib/kpatch/$ARCHVERSION/Module.symvers")"
 		elif [[ -e /lib/modules/$ARCHVERSION/extra/kpatch/Module.symvers ]]; then
-			SYMVERSFILE="$(readlink -f /lib/modules/$ARCHVERSION/extra/kpatch/Module.symvers)"
+			SYMVERSFILE="$(readlink -f "/lib/modules/$ARCHVERSION/extra/kpatch/Module.symvers")"
```
jpoimboe: I think we can assume that ARCHVERSION doesn't have spaces (here and elsewhere).

kpatch-build/kpatch-build
```diff
@@ -221,25 +221,25 @@ find_special_section_data_ppc64le() {
...
 
-	[[ -n $SPECIAL_VARS ]] && eval "$SPECIAL_VARS"
+	[[ -n "$SPECIAL_VARS" ]] && eval "$SPECIAL_VARS"
 
-	[[ -z $FIXUP_STRUCT_SIZE ]] && die "can't find special struct fixup_entry size"
-	[[ -z $BUG_STRUCT_SIZE ]] && die "can't find special struct bug_entry size"
-	[[ -z $EX_STRUCT_SIZE ]]  && die "can't find special struct exception_table_entry size"
+	[[ -z "$FIXUP_STRUCT_SIZE" ]] && die "can't find special struct fixup_entry size"
+	[[ -z "$BUG_STRUCT_SIZE" ]] && die "can't find special struct bug_entry size"
+	[[ -z "$EX_STRUCT_SIZE" ]]  && die "can't find special struct exception_table_entry size"
```
jpoimboe: I think all the quotes above (and below) for `SPECIAL_VARS` and `*_STRUCT_SIZE` aren't needed.

kpatch-build/kpatch-build
```diff
@@ -221,25 +221,25 @@ find_special_section_data_ppc64le() {
...
 
 	return
 }
 
 find_special_section_data() {
-	if [[ $ARCH = "ppc64le" ]]; then
+	if [[ "$ARCH" = "ppc64le" ]]; then
```
jpoimboe: ARCH should never have spaces so it doesn't need quoting (here and elsewhere)

kpatch-build/kpatch-build
```diff
@@ -329,25 +329,25 @@ find_kobj() {
 					KOBJFILE=vmlinux
 					return
 			esac
-			if [[ $DEEP_FIND -eq 0 ]]; then
+			if [[ "$DEEP_FIND" -eq 0 ]]; then
```
jpoimboe: `DEEP_FIND` doesn't need to be quoted

kpatch-build/kpatch-build
```diff
@@ -447,64 +447,64 @@ mkdir -p "$TEMPDIR" || die "Couldn't create $TEMPDIR"
...
+	[[ -z "$VMLINUX" ]] && VMLINUX="$SRCDIR"/vmlinux
 	[[ ! -e "$VMLINUX" ]] && die "can't find vmlinux"
 
 	# Extract the target kernel version from vmlinux in this case.
-	ARCHVERSION=$(strings "$VMLINUX" | grep -e "^Linux version" | awk '{ print($3); }')
+	ARCHVERSION="$(strings "$VMLINUX" | grep -e "^Linux version" | awk '{ print($3); }')"
 fi
 
-[[ $SKIPCLEANUP -eq 0 ]] && trap cleanup EXIT INT TERM HUP
+[[ "$SKIPCLEANUP" -eq 0 ]] && trap cleanup EXIT INT TERM HUP
```
jpoimboe: `SKIPCLEANUP` doesn't need to be quoted

kpatch-build/kpatch-build
```diff
@@ -447,64 +447,64 @@ mkdir -p "$TEMPDIR" || die "Couldn't create $TEMPDIR"
...
 
-KVER=${ARCHVERSION%%-*}
-if [[ $ARCHVERSION =~ - ]]; then
-	KREL=${ARCHVERSION##*-}
-	KREL=${KREL%.*}
+KVER="${ARCHVERSION%%-*}"
+if [[ "$ARCHVERSION" =~ - ]]; then
+	KREL="${ARCHVERSION##*-}"
+	KREL="${KREL%.*}"
```
jpoimboe: Pretty sure these don't need quoting

kpatch-build/kpatch-build
```diff
@@ -447,64 +447,64 @@ mkdir -p "$TEMPDIR" || die "Couldn't create $TEMPDIR"
...
+[[ -z "$TARGETS" ]] && TARGETS="vmlinux modules"
 
 source /etc/os-release
-DISTRO=$ID
-if [[ $DISTRO = fedora ]] || [[ $DISTRO = rhel ]] || [[ $DISTRO = ol ]] || [[ $DISTRO = centos ]]; then
-	[[ -z $VMLINUX ]] && VMLINUX=/usr/lib/debug/lib/modules/$ARCHVERSION/vmlinux
+DISTRO="$ID"
+if [[ "$DISTRO" = fedora ]] || [[ "$DISTRO" = rhel ]] || [[ "$DISTRO" = ol ]] || [[ "$DISTRO" = centos ]]; then
+	[[ -z "$VMLINUX" ]] && VMLINUX="/usr/lib/debug/lib/modules/$ARCHVERSION/vmlinux"
 	[[ -e "$VMLINUX" ]] || die "kernel-debuginfo-$ARCHVERSION not installed"
```
jpoimboe: `DISTRO` doesn't need quoting (here and below)

kpatch-build/kpatch-build
```diff
@@ -447,64 +447,64 @@ mkdir -p "$TEMPDIR" || die "Couldn't create $TEMPDIR"
...
 	fi
 
-	export PATH=/usr/lib/ccache:$PATH
+	export PATH="/usr/lib/ccache:$PATH"
 fi
 
 find_dirs || die "can't find supporting tools"
 
-if [[ $SKIPGCCCHECK -eq 0 ]]; then
+if [[ "$SKIPGCCCHECK" -eq 0 ]]; then
```
jpoimboe: `SKIPGCCHECK` doesn't need quoting

kpatch-build/kpatch-build
```diff
@@ -669,23 +669,23 @@ echo "Extracting new and modified ELF sections"
...
+		MODNAME="$(basename "${PATCH_LIST[0]}")"
 		if [[ "$MODNAME" =~ \.patch$ ]] || [[ "$MODNAME" =~ \.diff$ ]]; then
 			MODNAME="${MODNAME%.*}"
 		fi
 	else
 		MODNAME="patch"
 	fi
 
-	if $KPATCH_MODULE; then
+	if "$KPATCH_MODULE"; then
```
jpoimboe: `KPATCH_MODULE` doesn't need quoting

kpatch-build/kpatch-build
```diff
@@ -697,81 +697,81 @@ for i in $FILES; do
...
 		"$TOOLSDIR"/create-diff-object "orig/$i" "patched/$i" "$KOBJFILE" \
 			"output/$i" "$SRCDIR/Module.symvers" "${MODNAME//-/_}" 2>&1 |tee -a "$LOGFILE"
 		check_pipe_status create-diff-object
 		# create-diff-object returns 3 if no functional change is found
-		[[ $rc -eq 0 ]] || [[ $rc -eq 3 ]] || ERROR=$(expr $ERROR "+" 1)
-		if [[ $rc -eq 0 ]]; then
-			[[ -n $ERROR_IF_DIFF ]] && die $ERROR_IF_DIFF
+		[[ "$rc" -eq 0 ]] || [[ "$rc" -eq 3 ]] || ERROR="$(expr $ERROR "+" 1)"
+		if [[ "$rc" -eq 0 ]]; then
+			[[ -n "$ERROR_IF_DIFF" ]] && die "$ERROR_IF_DIFF"
```
jpoimboe: `$rc`, `$ERROR`, `$ERROR_IF_DIFF` don't need quoting, here and below

kpatch-build/kpatch-build
```diff
@@ -697,81 +697,81 @@ for i in $FILES; do
...
 	fi
 done
 
-if [[ $ERROR -ne 0 ]]; then
+if [[ "$ERROR" -ne 0 ]]; then
 	die "$ERROR error(s) encountered"
 fi
 
-if [[ $CHANGED -eq 0 ]]; then
+if [[ "$CHANGED" -eq 0 ]]; then
```
jpoimboe: `CHANGED` doesn't need quoting

kpatch-build/kpatch-build
```diff
@@ -697,81 +697,81 @@ for i in $FILES; do
...
 cd "$SRCDIR"
 make prepare >> "$LOGFILE" 2>&1 || die
 
-if [[ ! -z $UBUNTU_KERNEL ]]; then
+if [[ ! -z "$UBUNTU_KERNEL" ]]; then
 	# UBUNTU: add UTS_UBUNTU_RELEASE_ABI to utsrelease.h after regenerating it
-	UBUNTU_ABI=${ARCHVERSION#*-}
-	UBUNTU_ABI=${UBUNTU_ABI%-*}
+	UBUNTU_ABI="${ARCHVERSION#*-}"
+	UBUNTU_ABI="${UBUNTU_ABI%-*}"
```
jpoimboe: `UBUNTU_KERNEL` and `UBUNTU_ABI` don't need quoting

kpatch/kpatch
```diff
@@ -129,19 +129,19 @@ core_loaded () {
...
-	checksum=$(readelf -p .kpatch.checksum $1 | grep '\[.*\]' | awk '{print $3}')
+	checksum="$(readelf -p .kpatch.checksum "$1" | grep '\[.*\]' | awk '{print $3}')"
 
 	# Fail checksum match only if both exist and diverge
-	if [[ ! -z $checksum ]] && [[ -e "$SYSFS/${modname}/checksum" ]] ; then
-		sysfs_checksum=$(cat $SYSFS/${modname}/checksum)
-		[[ $checksum == $sysfs_checksum ]] || return 1
+	if [[ ! -z "$checksum" ]] && [[ -e "$SYSFS/${modname}/checksum" ]] ; then
+		sysfs_checksum="$(cat "$SYSFS/${modname}/checksum")"
+		[[ "$checksum" == "$sysfs_checksum" ]] || return 1
```
jpoimboe: Here I think `modname` needs quoting, and `checksum` and `sysfs_checksum` don't need quoting

kpatch/kpatch
```diff
@@ -205,20 +205,20 @@ unload_module () {
 	PATCH="${PATCH%.ko}"
 	ENABLED="$SYSFS/$PATCH/enabled"
```
jpoimboe: Does `PATCH` need quoting?

kpatch/kpatch
```diff
@@ -276,75 +276,75 @@ case "$1" in
...
 			PATCH="$2"
-			[[ "$PATCH" != $(basename "$PATCH") ]] && die "please supply patch module name without path"
+			[[ "$PATCH" != "$(basename "$PATCH")" ]] && die "please supply patch module name without path"
 			;;
 		esac
 		shift
 	done
 
-	MODULE=$INSTALLDIR/$KVER/"$PATCH"
+	MODULE="$INSTALLDIR/$KVER/$PATCH"
```
jpoimboe: I think `PATCH` needs quoting

kpatch-build/kpatch-build
rudis: As I mentioned in the commit message, I think the best way is to quote all variables even if they don't require quoting (at the moment) to prevent possible confusion by the reader and to future proof the code.

This means adding a few more quote changes in this commit but shouldn't hurt in the long run and ensures that anybody copy&pasting/adapting existing code won't introduce quoting issues. As shell's quote handling is a little fragile it's best practice to just quote everything to prevent mistakes.

kpatch-build/kpatch-build
rudis: As mentioned above, I'd like to keep the quotes for consistency and future proofing.

kpatch-build/kpatch-build
rudis: Dito.

kpatch-build/kpatch-build
rudis: Dito.

kpatch-build/kpatch-build
rudis: I'm confused here as `${file}` is quoted by the surrounding quotes: `"$pdir/.${file}.cmd"`

kpatch-build/kpatch-build
rudis: Same here, it's quoted.

kpatch-build/kpatch-build
rudis: See above.

kpatch-build/kpatch-build
rudis: Dito.

kpatch-build/kpatch-build
rudis: Dito.

kpatch-build/kpatch-build
rudis: Dito.

kpatch-build/kpatch-build
rudis: Dito.

kpatch-build/kpatch-build
rudis: Dito.

kpatch-build/kpatch-build
rudis: Dito.

kpatch/kpatch
rudis: As above, quoted by the surrounding quotes.

kpatch/kpatch
rudis: Dito.

kpatch/kpatch
rudis: Dito.

kpatch/kpatch
rudis: Dito.

rudis:
Thank you for the review. Will push a new version with proper handling for UTS_UBUNTU_RELEASE_ABI as number when you're happy with this round.

jpoimboe:
Bash quotes are confusing.  :-) Please ignore all my comments and go ahead with the UTS_UBUNTU_RELEASE_ABI fix.

rudis:
They are. Took me a while to get a correct patch as well. Don't worry ;-)

Just pushed the fix for UTS_UBUNTU_RELEASE_ABI and another commit which cleanups a few more minor issues found by shellcheck.

rudis:
There's one remaining issue which will break sooner or later:
```
if [[ "$gcver" < 4.8 ]]; then
```
The shell doesn't support version compares so this is a lexicographically comparsion which will break if gcc ever gets to version 10 (or similar issues). I don't know if this is worth fixing though, just wanted to mention it.

jpoimboe:
Thanks for fixing all these issues.  The code looks good to me.

Regarding the gccver comparison issue, we should definitely get that fixed, as I believe GCC 10 will be released in a few years (2020).

rudis:
New commit pushed which fixes the version check by implementing the last comment of #565 (compiling with gcc -ffunction-sections -pg and a test file for the .comment section).

kpatch-build/kpatch-build
```diff
@@ -167,10 +167,28 @@ find_core_symvers() {
...
+	readelf -p .comment "$1" | grep -o 'GCC:.*'
+}
 gcc_version_check() {
+	local c="$TEMPDIR/test.c" aout="$TEMPDIR/test"
+	local out gccver kgccver
+
+	# gcc --version varies between distributions therefore extract version
+	# by compiling a test file and compare it to vmlinux's version.
+	echo 'void main(void) {}' > "$c"
+	out="$(gcc -pg -ffunction-sections -o "$aout" "$c" 2>&1)"
```
jpoimboe: I think it needs the `-c` option, so that it doesn't link in any of the GCC code for calling `main()`.  Otherwise on my F26 system I get two different versions of GCC in the `.comment` section:
```$ readelf -p .comment /tmp/a | grep -o 'GCC:.*'
GCC: (GNU) 7.1.1 20170622 (Red Hat 7.1.1-3)
GCC: (GNU) 7.2.1 20170915 (Red Hat 7.2.1-2)
```

kpatch-build/kpatch-build
```diff
@@ -186,16 +204,6 @@ gcc_version_check() {
 	#   <linux>/arch/powerpc/tools/gcc-check-mprofile-kernel.sh
 	#
 	# Postpone the check to the kernel build instead of duplicating it here.
```
jpoimboe: Now that we're doing the check for all arches, these ppc64le specific comments can be removed.

kpatch-build/kpatch-build
rudis: Interesting, I get only one version here (Debian). Will the version with `-c` match the version in vmlinux (which is also linked I guess)?

kpatch-build/kpatch-build
rudis: I wanted to keep as reference that `-mprofile-kernel` is required on ppc64le. But if you think it's not necessary I'll remove it.

rudis:
Pushed new version which handles the mentioned issues.

kpatch-build/kpatch-build
jpoimboe: Yes, compiling with `-c` matches the vmlinux GCC version.  Presumably because vmlinux is 100% compiled from source, whereas linking an executable brings in the `main()` boilerplate object code from somewhere.

jpoimboe:
Looks great to me @rudis , thanks!

👍

joe-lawrence:
@rudis - can you recommend any linting tools that could automate some of this checking for the future?

rudis:
@joe-lawrence I used `shellcheck` (https://www.shellcheck.net/).

The remaining warnings are false-positives or not relevant for kpatch. They could be disabled with a shellcheck annotation, but I'm not sure if that's worth it. On the one hand it documents the code (for example missing quotes because the value should be expanded), on the other hand it clutters the code. Tell me what you think.

joe-lawrence:
@rudis - interesting tool... as far as automating it to detect new errors, too bad it doesn't have something smatch's new_bugs that filters out existing complaints (even with slight code shift).

rudis:
@joe-lawrence We could disable all existing (false positive) complaints (via `# shellcheck disable=` comments) and add a Makefile target to check if there any (new) complaints. That should work.

joe-lawrence:
@rudis - How many false positives are left after your changes?  If there are only a few, maybe they're worth modifying to avoid the complaints altogether.

rudis:
@joe-lawrence Just pushed the necessary changes to silence shellcheck. I fixed the remaining relevant issues and ignored the false positives. With this `make check` (tell me if you want a different name) will detect new issues introduced in the future.

rudis:
Added two more commits. Please review the "grep -F" one carefully, as I'm not 100% sure if it's correct. It seems to work fine here though.

jpoimboe:
The only thing I'm not crazy about is the noise created by all the shellcheck annotations.  Is it possible to put the reason for the annotation on the same line as the annotation itself?  e.g., something like
```
# shellcheck disable=SC1091 (Don't check external file.)
```
If not, then I guess the current approach is fine.

Otherwise all the changes look good to me, assuming they pass our regression tests.

rudis:
Sadly no, everything on the same line is considered a syntax error. We could drop the explaining comment, but I didn't want to add the shellcheck annotation without explaining why it's there.

I don't know how to run the test suite, please run the regression tests and report any failures.

jpoimboe:
> Sadly no, everything on the same line is considered a syntax error. We could drop the explaining comment, but I didn't want to add the shellcheck annotation without explaining why it's there.

That's too bad.  The patches look good to me then.

> I don't know how to run the test suite, please run the regression tests and report any failures.

Yeah, we're still working on making the regression tests more easily runnable (or even better, automated).  Either @joe-lawrence or I will run them.

joe-lawrence:
I'll try to kick off a long regression test over the weekend with these changes.  The only thing that sticks out to me is the Makefile check rule.  On the side, I've been playing with smatch and valgrind and have been executing them from an externally stored project of sorts (ie, no integration into the kpatch build).  @jpoimboe - do you have a preference for these sort of tools, in tree or out of tree?

jpoimboe:
@joe-lawrence Do you mean copying the tools in-tree?  Or do you mean adding smatch/valgrind makefile targets with the assumption that the user installs the tools on their system?

joe-lawrence:
@jpoimboe : I was referring to adding Makefile targets and other such hooks for these tools (ie, not bringing the tools themselves into the tree).

jpoimboe:
@joe-lawrence I think it sounds like a good idea.

joe-lawrence:
👍   the long integration test finished this morning (clocked in at 35 hours on my virtual machine) with only four failures... but these are due to running in Fedora 26 without rebasing the Fedora 25 patches which failed.  I think the patchset here is good to go.

jpoimboe:
Thanks @joe-lawrence, and thanks @rudis for all the cleanups!

PR #745, opened 2017-10-09
kpatch: add ABI backwards compatibility

jpoimboe:
When running a kernel for a long period of time without rebooting, it's
possible that newer versions of the kpatch script may get installed.  So
new versions of the kpatch script need to support old versions of
kpatch.ko.

### DIFF
--- a/kpatch/kpatch
+++ b/kpatch/kpatch
@@ -27,14 +27,6 @@ INSTALLDIR=/var/lib/kpatch
 SCRIPTDIR="$(readlink -f $(dirname $(type -p $0)))"
 VERSION="0.4.0"
 
-# Livepatch is built into the kernel, if it's not present
-# we must use kpatch core module.
-if [[ -e /sys/kernel/livepatch ]] ; then
-	SYSFS="/sys/kernel/livepatch"
-else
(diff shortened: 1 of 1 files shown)


kpatch/kpatch
```diff
@@ -158,6 +165,8 @@ load_module () {
 			echo "loading core module: $COREMOD"
 			insmod "$COREMOD" || die "failed to load core module"
 		fi
+
+		init_sysfs_var
```
joe-lawrence: I think this call to `init_sysfs_var` is redundant as the one below should have already been called via `load_module` -> `verify_module_checksum`, right?

jpoimboe: The `init_sysfs_var()` call on line 234 is called first, no matter what the subcommand is.  However, at that point the core module (kpatch.ko) might not have been loaded yet and SYSFS might have the wrong value.  That's why `init_sysfs_var()` is called again after loading the core module.

Does that answer your question?  Would a comment help?

jpoimboe: @joe-lawrence ping?

jpoimboe:
I added some comments, do they look ok?

joe-lawrence:
:+1  thanks for adding the comments

PR #746, opened 2017-10-11
Documentation updates

rudis:
Fixes a few minor typos/inconsistencies and adds installation instructions for Stretch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ may occur!**
 
 Here's a video of kpatch in action:
 
-[![kpatch video](http://img.youtube.com/vi/juyQ5TsJRTA/0.jpg)](http://www.youtube.com/watch?v=juyQ5TsJRTA)
+[![kpatch video](https://img.youtube.com/vi/juyQ5TsJRTA/0.jpg)](https://www.youtube.com/watch?v=juyQ5TsJRTA)
 
 And a few more:
 
--- a/README.md
+++ b/README.md
@@ -195,7 +195,7 @@ apt update may report a "404 Not Found" error, as well as a complaint about
 disabling the repository by default.  This message may be ignored (see issue
 #710).
 
-#### Debian 8.0
+#### Debian 8 (Jessie)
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -508,7 +510,7 @@ updating the instruction directly.  This approach also ensures that the code
 modification path is reliable, since ftrace has been doing it successfully for
 years.
 
-**Q Is kpatch compatible with \<insert kernel debugging subsystem here\>?**
+**Q. Is kpatch compatible with \<insert kernel debugging subsystem here\>?**
```
joe-lawrence: Can you also correct the line starting with "**Q: Is it possible to register a function that..." a few lines down.  (ie, `s/:/.` )

README.md
```diff
@@ -351,6 +351,7 @@ can have some major pitfalls if you're not careful.  To learn more about how to
 properly create live patches, see the [Patch Author
 Guide](doc/patch-author-guide.md).
 
+
```
joe-lawrence: I don't think these spaces show up in the Markdown rendering, but if we want to be consistent with two blank lines, then can you also update the spaces preceding the Installation and License sections.

README.md
```diff
@@ -405,6 +406,7 @@ The primary steps in kpatch-build are:
 - Link all the output objects into a cumulative object
 - Generate the patch module
 
+
 ### Patching
```
joe-lawrence: I guess the pattern is to double-space the ### sub-sections?  If so, does the Debian 7.x Install subsection need an extra line, too?

rudis: I tried to follow the pattern in the document and that seemed to be the most consistent way. The Debian 7.x section has four levels (####) so I added only one empty line there.

README.md
rudis: Will do.

README.md
rudis: Will do.

rudis:
Pushed updated version which addresses the mentioned issues. Also added another commit to use https:// for youtube links.

joe-lawrence:
👍  looks good, thanks @rudis

PR #749, opened 2017-10-17
create-diff-object: handle missing padding at end of special section

### LINKED ISSUE #747: testing: parainstructions-section.patch read/write past end of buffer on Fedora 26
When building `parainstructions-section.patch` on Fedora 26 running:

```
% uname -r
4.11.8-300.fc26.x86_64

% gcc --version |head -n1 |cut -d' ' -f3-
7.1.1 20170622 (Red Hat 7.1.1-3)

% readelf -p .comment usr/lib/debug/lib/modules/4.11.8-300.fc26.x86_64/vmlinux |grep GCC: | tr -s ' ' | cut -d ' ' -f6-
7.1.1 20170622 (Red Hat 7.1.1-3)
```

valgrind complains about an invalid write (past end of allocated memory):

```
==23044== Invalid write of size 8
==23044==    at 0x4C32FF3: memcpy@GLIBC_2.2.5 (vg_replace_strmem.c:1021)
==23044==    by 0x405D05: kpatch_regenerate_special_section (create-diff-object.c:1852)
==23044==    by 0x4069FA: kpatch_process_special_sections (create-diff-object.c:2090)
==23044==    by 0x409216: main (create-diff-object.c:2874)
==23044==  Address 0x5ad8fe8 is 24 bytes inside a block of size 28 alloc'd
==23044==    at 0x4C2EB6B: malloc (vg_replace_malloc.c:299)
==23044==    by 0x405A7F: kpatch_regenerate_special_section (create-diff-object.c:1802)
==23044==    by 0x4069FA: kpatch_process_special_sections (create-diff-object.c:2090)
==23044==    by 0x409216: main (create-diff-object.c:2874)
==23044== 
==23044== Syscall param pwrite64(buf) points to unaddressable byte(s)
==23044==    at 0x5151953: __pwrite64_nocancel (in /usr/lib64/libc-2.25.so)
==23044==    by 0x4E49454: ??? (in /usr/lib64/libelf-0.169.so)
==23044==    by 0x4E458E0: elf_update (in /usr/lib64/libelf-0.169.so)
==23044==    by 0x40C568: kpatch_write_output_elf (kpatch-elf.c:813)
==23044==    by 0x4094ED: main (create-diff-object.c:2950)
==23044==  Address 0x5ad8fec is 0 bytes after a block of size 28 alloc'd
==23044==    at 0x4C2EB6B: malloc (vg_replace_malloc.c:299)
==23044==    by 0x405A7F: kpatch_regenerate_special_section (create-diff-object.c:1802)
==23044==    by 0x4069FA: kpatch_process_special_sections (create-diff-object.c:2090)
==23044==    by 0x409216: main (create-diff-object.c:2874)
```
I threw some debugging printf's into `kpatch_regenerate_special_section()` for the problematic call and I see:
```
>>> special->name=.parainstructions sec->name=.rela.parainstructions
>>>>>> dest = malloc(28) = 0x5ad83d0
>>>>>> memcpy(dest+0, src+0, group_size=16) 
>>>>>> memcpy(dest+16, src+16, group_size=16)
```
which indicates that the function is trying to read past the end of src and write past the end of dest.  The logic is a bit curious as the buffer size is determined by `sec->base->sh.sh_size` (28 bytes) and confirmed by readelf:
```
% readelf --wide --sections orig/fs/proc/generic.o

There are 126 section headers, starting at offset 0x46410:

Section Headers:
  [Nr] Name                   Type            Address          Off    Size   ES Flg Lk Inf Al
  ...
  [29] .parainstructions      PROGBITS        0000000000000000 0006a8 00001c 00   A  0   0  8
  [30] .rela.parainstructions RELA            0000000000000000 023ec0 000030 18   I 123  29  8
```
but the group_size is 16 bytes, read from the PARA_STRUCT_SIZE (confirmed `struct paravirt_patch_site` size in crash utility).  `kpatch_regenerate_special_section()`'s for-loop iterates over the source buffer, copying data to the destination buffer by incrementing offsets by `group_size`.

In the case that valgrind complains about and my last debugging printf shows, it increments offsets from 0 to 16... then tries copying another 16 bytes.  Too bad the size of both buffers were only 28 bytes.

jpoimboe:
The paravirt_patch_site struct has 12 bytes of data and 4 bytes of
padding, for a total of 16 bytes.  However, when laying out the structs
in the .parainstructions section, the vmlinux script only aligns before
each struct's data, not after.  So the last entry doesn't have the
4-byte padding, which breaks kpatch_regenerate_special_section()'s
assumption of a 16-byte struct, resulting in a memcpy past the end of
the section.

Fixes #747.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1793,7 +1793,7 @@ static void kpatch_regenerate_special_section(struct kpatch_elf *kelf,
 {
 	struct rela *rela, *safe;
 	char *src, *dest;
-	int group_size, src_offset, dest_offset, include, align, aligned_size;
+	int group_size, src_offset, dest_offset, include;
 
 	LIST_HEAD(newrelas);
 
@@ -1817,6 +1817,18 @@ static void kpatch_regenerate_special_section(struct kpatch_elf *kelf,
 	for ( ; src_offset < sec->base->sh.sh_size; src_offset += group_size) {
 
 		group_size = special->group_size(kelf, src_offset);
+
+		/*
+		 * In some cases the struct has padding at the end to ensure
+		 * that all structs after it are properly aligned.  But the
+		 * last struct in the section may not be padded.  In that case,
+		 * shrink the group_size such that it still (hopefully)
+		 * contains the data but doesn't go past the end of the
+		 * section.
+		 */
+		if (src_offset + group_size > sec->base->sh.sh_size)
+			group_size = sec->base->sh.sh_size - src_offset;
+
 		include = should_keep_rela_group(sec, src_offset, group_size);
 
 		if (!include)
@@ -1852,12 +1864,6 @@ static void kpatch_regenerate_special_section(struct kpatch_elf *kelf,
 		memcpy(dest + dest_offset, src + src_offset, group_size);
 		dest_offset += group_size;
 	}
-
-	/* verify that group_size is a divisor of aligned section size */
-	align = sec->base->sh.sh_addralign;
-	aligned_size = ((sec->base->sh.sh_size + align - 1) / align) * align;
-	if (src_offset != aligned_size)
-		ERROR("group size mismatch for section %s\n", sec->base->name);
 
 	if (!dest_offset) {
 		/* no changed or global functions referenced */


joe-lawrence:
👍 looks good

PR #750, opened 2017-10-18
Fix livepatch version checks for RHEL

jpoimboe:
Clean up the version checking in the livepatch-patch-hook, and then add version checking for RHEL for when it gets CONFIG_LIVEPATCH.


kmod/patch/livepatch-patch-hook.c
```diff
@@ -248,11 +258,9 @@ static int __init patch_init(void)
 		goto out;
 	lpatch->mod = THIS_MODULE;
 	lpatch->objs = lobjects;
-#if __powerpc__
-#if (LINUX_VERSION_CODE >= KERNEL_VERSION(4, 12, 0))
+#if defined(__powerpc__) && defined(HAVE_IMMEDIATE)
```
joe-lawrence: BTW there are a bunch of more `#ifdef __powerpc__` instances in create-diff-object.c and kpatch-elf.c.

joe-lawrence:
:+1  Changes look sane, I like the version check cleanup.  I'll let you reply to my earlier comment before merging.

kmod/patch/livepatch-patch-hook.c
jpoimboe: The ones in create-diff-object.c and kpatch-elf.c aren't specific to kernel versions so I think they're fine as they are.

PR #753, opened 2017-11-01
kpatch script - livepatch fixups

joe-lawrence:


joe-lawrence:
This patchset implements most of the features listed in #753.  If Miroslav's in-kernel "livepatch: Introduce signal and force sysfs attributes" patchset is merged soon, we can add support for that, too.

This is lightly tested with upstream livepatch modules and RHEL kpatch modules.  I used various combinations of loading / disabling / etc. of the livepatch "busymod" sample modules to simulate a stalled task.  (In this case, it's a kworker, so the signal had no effect and it never completed the transition.)

We probably want to adjust the transition sleep duration / frequency, I chose every 10 seconds for up to a full minute for testing.  User messages could be clarified a bit as well.  I figured I'd open a PR to generate some feedback on the patchset before spending more time on the details.

kpatch/kpatch
```diff
@@ -211,6 +271,11 @@ load_module () {
 		fi
 	done
 
+	if ! wait_for_patch_transition "$modname" ; then
+		unload_module "$modname"
+		die "failed to load module $module (stalled)"
```
jpoimboe: Here I think it should disable the patch, similar to the re-enable path above.

kpatch/kpatch
```diff
@@ -156,6 +160,56 @@ verify_module_checksum () {
...
+			echo "$tid $(cat "$proc_task/comm")"
+		fi
+	done
+}
+
+signal_process() {
+	local pid="$1"
+	echo "signaling pid $pid $(cat /proc/"$pid"/comm)"
+	kill -SIGSTOP "$pid"
+	sleep .1
```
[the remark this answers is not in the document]
joe-lawrence: That was something that kgraft used in its kgr script ("give kernel time to distribute the signal to all threads").  I assumed that there was some kind of multiple-signals-multiple-threads race condition that this avoided.

kpatch/kpatch
```diff
@@ -156,6 +160,56 @@ verify_module_checksum () {
...
+	kill -SIGCONT "$pid"
+}
+
+signal_stalled_processes() {
+	local proc_task
+	local tid
+	for proc_task in /proc/[0-9]*/task/[0-9]*; do
+		if [[ $(cat "$proc_task/patch_state" 2>/dev/null) == "0" ]]; then
```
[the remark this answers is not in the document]
joe-lawrence: You're right.  I'll fix up for both cases.

kpatch/kpatch
```diff
@@ -156,6 +160,56 @@ verify_module_checksum () {
...
+			tid=${proc_task#*/task/}
+			signal_process "$tid"
+		fi
+	done
```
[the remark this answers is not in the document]
joe-lawrence: Ok.  I think he settled on `/sys/kernel/livepatch/<patch>/signal` in v3 of his patch.  I can check for and use that if it exists.

kpatch/kpatch
```diff
@@ -156,6 +160,56 @@ verify_module_checksum () {
...
+}
+
+wait_for_patch_transition() {
+	local module="$1"
+	local i
+	for (( i=0; i<SIGNAL_RETRIES; i++ )); do
+		in_transition "$module" || return 0
+		sleep "$SIGNAL_FREQ"
+		signal_stalled_processes
+	done
```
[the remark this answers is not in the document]
joe-lawrence: Sure, I can modify the signal and delay algorithm to do this.

kpatch/kpatch
joe-lawrence: unload_module() called in the error path will first disable the patch, then unload it.  Are you suggesting that we should only disable and not unload the module here?

kpatch/kpatch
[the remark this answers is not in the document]
jpoimboe: Ok.

kpatch/kpatch
jpoimboe: Ah, I see.  It would still be good to have an error message before calling unload_module(), like "failed to enable module $module (stalled), unloading"

joe-lawrence:
v2
- reworked the `is_stalled()` code to handle both enabling and disabling transitions in progress
- added a few more status messages to keep the user informed
- differentiated between signal and transition delays
- on a VM running a kernel build, loaded a livepatch that modified `ep_poll()`, I saw up to 33 second transition delay even after signaling, so I bumped the post-signal wait time up to 1 full minute

Sample output:
```
% kpatch/kpatch load ./livepatch-callbacks-demo.ko
loading patch module: ./livepatch-callbacks-demo.ko
waiting (up to 5 seconds) for patch transition to complete...
patch transition has stalled, signaling stalled process(es):
signaling pid 1 systemd
signaling pid 502 systemd-udevd
signaling pid 572 auditd
signaling pid 595 systemd-logind
signaling pid 596 dbus-daemon
signaling pid 693 systemd
waiting (up to 60 seconds) for patch transition to complete...
transition complete (15 seconds)
```
I pushed v2 up as a soft work-in-progress.  I'll be away for the next week and wanted to post the current progress.  I'd like to do a bit more testing of scenarios, including kpatch modules, to make sure they all work as expected.

v3 : should we skip the kpatch script PID `$$` when (manually) signaling?  I can't imagine any good would come of that.

kpatch/kpatch
```diff
@@ -156,6 +160,106 @@ verify_module_checksum () {
...
+	[[ "$patch_enabled" != "$patch_state" ]] && return 0
+	return 1
+}
+
+get_transition_patch() {
+	local module
+	local modname
+	for module in "$SYSFS"/*; do
+		modname=$(basename "$module")
+		in_transition "$modname" && echo "$modname"; return
```
jpoimboe: This return will always happen, even if not in_transition, so the loop will always only have one iteration.

kpatch/kpatch
```diff
@@ -156,6 +160,106 @@ verify_module_checksum () {
...
+
+wait_for_patch_transition() {
+	local module="$1"
+	local i
+
+	in_transition "$module" || return 0
+
+	echo "waiting (up to $POST_ENABLE_WAIT seconds) for patch transition to complete..."
+	for (( i=0; i<POST_ENABLE_WAIT; i++ )); do
+		in_transition "$module" || return 0
```
jpoimboe: It would be good to have a "transition complete ($i seconds)" message here, like below.

jpoimboe:
> on a VM running a kernel build, loaded a livepatch that modified ep_poll(), I saw up to 33 second transition delay even after signaling, so I bumped the post-signal wait time up to 1 full minute

This sounds fishy to me.  It would be good to understand why there was such a long delay after signaling.  If I'm not mistaken, it should always only take about 1 second after signaling to complete the transition, unless there's a kthread or idle task sleeping on the patched function, which seems unlikely for ep_poll().

joe-lawrence:
> This sounds fishy to me. It would be good to understand why there was such a long delay after signaling. If I'm not mistaken, it should always only take about 1 second after signaling to complete the transition, unless there's a kthread or idle task sleeping on the patched function, which seems unlikely for ep_poll().

It looks like it is systemd (pid 1) that is holding up the show.  If I manually do a `kill -s SIGSTOP 1` and then watch `cat /proc/1/stack`, I never see it enter `do_signal_stop()`... so I'm guessing that the init process is somehow special with respect to (at least these) signals.  To confirm, when I run the same livepatch load test and watch the output of `kpatch list`, the list of stalled processes always shrinks down to `1 systemd`.  Perhaps the poll finally times out and that is what unwedges the transition.

jpoimboe:
> v3 : should we skip the kpatch script PID $$ when (manually) signaling? I can't imagine any good would come of that.

Yes, good idea.  Also the 5+60 wait looks good, that hopefully gives pid 1 enough time to wake itself up.

joe-lawrence:
v4
- address v3 issues (loops, etc)
- skip signalling our own pid
- updated `is_stalled()` to ignore `KLP_UNDEFINED` state

### Additional usability thoughts:
- Should the `kpatch load` automatically signal threads, or just report the stalled tasks to the user?
- Are stack traces for stalled tasks interesting?
- these follow from Jiri's [Re: [PATCH v3 2/2] livepatch: force transition process to finish](https://marc.info/?l=linux-kernel&m=150996982023085&w=2):
>The process that is actually being suggested here (and towards which the implementation is heading) is that in case of the convergence being stuck for unknown reason, the admin of the system collects all the necessary data (namely stacktraces of the relevant tasks), and requests a clearance from the livepatch distributor to force the transition.

### User output examples
I modified the livepatch-callbacks-demo to modify `ep_poll()` and inserted a 10 second call to `msleep()`, forcing a scenario where some PIDs are sitting in that function, holding off the (un)patching transitions:
```
% kpatch/kpatch load ~/linux/samples/livepatch/livepatch-callbacks-demo.ko
waiting (up to 5 seconds) for patch transition to complete...
patch transition has stalled, signaling stalled process(es):
signaling pid 19366 systemd-logind
signaling pid 19367 systemd-udevd
signaling pid 1 systemd
signaling pid 572 auditd
signaling pid 596 dbus-daemon
signaling pid 693 systemd
waiting (up to 60 seconds) for patch transition to complete...
transition complete (11 seconds)
```
Meanwhile in another window, a `kpatch list` looks like:
```
% kpatch/kpatch list
Loaded patch modules:
livepatch_callbacks_demo [enabling...]

Stalled processes:
19366 systemd-logind
19367 systemd-udevd
1 systemd
572 auditd
596 dbus-daemon
693 systemd

Installed patch modules:
```
Once the `kpatch load` is finally complete, `kpatch list` looks like:
```
% kpatch/kpatch list
Loaded patch modules:
livepatch_callbacks_demo [enabled]

Installed patch modules:
```

Finally unloading the demo livepatch:
```
% kpatch/kpatch unload livepatch-callbacks-demo
disabling patch module: livepatch_callbacks_demo
waiting (up to 5 seconds) for patch transition to complete...
patch transition has stalled, signaling stalled process(es):
signaling pid 19348 systemd-journal
waiting (up to 60 seconds) for patch transition to complete...
transition complete (4 seconds)
unloading patch module: livepatch_callbacks_demo
```
during the transition, `kpatch list`:
```
% kpatch/kpatch list
Loaded patch modules:
livepatch_callbacks_demo [disabling...]

Stalled processes:
19348 systemd-journal

Installed patch modules:
```
and after the transition:
```
% kpatch/kpatch list
Loaded patch modules:

Installed patch modules:
```

jpoimboe:
> Should the kpatch load automatically signal threads, or just report the stalled tasks to the user?

IMO, "kpatch load" is a curated do-everything-for-the-user subcommand.  So I think it should signal the threads.  If the user wants more control then they should do the steps manually, either completely manually or using other kpatch subcommands.

>     Are stack traces for stalled tasks interesting?

Perhaps, but not as part of "kpatch load", for the same reason.

>     these follow from Jiri's Re: [PATCH v3 2/2] livepatch: force transition process to finish:
> 
>     The process that is actually being suggested here (and towards which the implementation is heading) is that in case of the convergence being stuck for unknown reason, the admin of the system collects all the necessary data (namely stacktraces of the relevant tasks), and requests a clearance from the livepatch distributor to force the transition.

In such a case I think it would be appropriate to give the admin special steps to follow.  Perhaps with the use of kpatch subcommands to make it easier.  But I don't think we have to worry about that for now, at least I think it's outside the scope of this PR.

Those are my opinions, at least.

kpatch/kpatch
```diff
@@ -156,6 +160,119 @@ verify_module_checksum () {
...
+		sleep 1s
+	done
+
+	echo "patch transition has stalled, signaling stalled process(es):"
+	signal_stalled_processes
+
+	echo "waiting (up to $POST_SIGNAL_WAIT seconds) for patch transition to complete..."
+	for (( i=0; i<POST_SIGNAL_WAIT; i++ )); do
+		if ! in_transition "$module" ; then
+			echo -e "transition complete ($i seconds)"
```
jpoimboe: I think '-e' isn't needed (here and above)

joe-lawrence:
v4
- Removed unneeded "-e" from echo statements

PR #755, opened 2017-11-07
kpatch-build, x86: do not use the patched functions as callbacks directly

euspectre:
A kernel crash happened in __do_softirq() in very rare cases when the
binary patch created from mainline commit be82485fbcbb
("netlink: fix an use-after-free issue for nlk groups") was unloaded.

Kernel: 3.10.0-514.16.1.vz7.30.10, Virtuozzo 7, x86_64.

Investigation has shown that the kernel tried to execute an RCU callback,
deferred_put_nlk_sk(), defined in the patch module after the module had
been unloaded.

The callback was set by the patched variant of netlink_release(), that is
why the address of the patched deferred_put_nlk_sk() rather than of the
original one was used.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2302,6 +2302,55 @@ static int kpatch_is_core_module_symbol(char *name)
 		!strcmp(name, "kpatch_shadow_get"));
 }
 
+/*
+ * If the patched code refers to a symbol, for example, calls a function
+ * or stores a pointer to a function somewhere, the address of that symbol
+ * must be resolved somehow before the patch is applied. The symbol may be
+ * present in the original code too, so the patch may refer either to that
+ * version of the symbol (dynrela is used for that) or to its patched
+ * version directly (with a normal relocation).
+ *
+ * Dynrelas may be needed for the symbols not present in this object file
+ * (rela->sym->sec is NULL), because it is unknown if the patched versions
+ * of these symbols exist and where they are.
+ *
+ * The patched code can usually refer to a symbol from this object file
+ * directly. If it is a function, this may also improve performance because
+ * it will not be needed to call the original function first, find the
+ * patched one and then use Ftrace to pass control to it.
+ *
+ * There is an exception though, at least on x86. It is safer to use
+ * a dynrela if the patched code stores a pointer to a function somewhere
+ * (relocation of type R_X86_64_32S). The function could be used as
+ * a callback and some kinds of callbacks are called asynchronously. If
+ * the patch module sets such callback and is unloaded shortly after,
+ * the kernel could try to call the function via an invalid pointer and
+ * would crash. With dynrela, the kernel would call the original function
+ * in that case.
+ *
+ * Nested functions used as callbacks are a special case. They are not
+ * supposed to be visible outside of the function that defines them.
+ * Their names may differ in the original and the patched kernels which
+ * makes it difficult to use dynrelas. Fortunately, nested functions are
+ * rare and are unlikely to be used as asynchronous callbacks, so the
+ * patched code can refer to them directly. It seems, one can only
+ * distinguish such functions by their names containing a dot. Other
+ * kinds of functions with such names (e.g. optimized copies of functions)
+ * are unlikely to be used as callbacks.
+ */
+static int function_ptr_rela(const struct rela *rela)
+{
+	return (rela->sym->type == STT_FUNC && rela->type == R_X86_64_32S);
+}
+
+static int may_need_dynrela(const struct rela *rela)
+{
+	if (!rela->sym->sec)
+		return 1;
+
+	return (function_ptr_rela(rela) && !strchr(rela->sym->name, '.'));
+}
+
 static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
 						struct lookup_table *table,
 						char *hint, char *objname,
@@ -2360,8 +2409,9 @@ static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
 		    !strcmp(sec->name, ".rela.kpatch.dynrelas"))
 			continue;
 		list_for_each_entry_safe(rela, safe, &sec->relas, list) {
-			if (rela->sym->sec)
+			if (!may_need_dynrela(rela))
 				continue;
+
 			/*
 			 * Allow references to core module symbols to remain as
 			 * normal relas, since the core module may not be
@@ -2547,7 +2597,15 @@ static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
 			rela2->offset = index * sizeof(*krelas) + \
 				offsetof(struct kpatch_relocation, ksym);
 
-			rela->sym->strip = 1;
+			/*
+			 * Mark the referred to symbol for removal but
+			 * only if it is not from this object file.
+			 * The symbols from this object file may be needed
+			 * later (for example, they may have relocations
+			 * of their own which should be processed).
+			 */
+			if (!rela->sym->sec)
+				rela->sym->strip = 1;
 			list_del(&rela->list);
 			free(rela);
 


euspectre:
Not sure if livepatch-patch-hook.c needs a similar fix. It might but I am not very familiar with that code yet.

jpoimboe:
Thanks, this is an interesting issue.  However I'm not convinced that this RCU synchronize really belongs in kpatch.  Isn't the issue specific to the patch itself?  I think a more appropriate fix would be to add a patch unload hook which calls rcu_barrier().

I think this is an example of a broader issue, which should be documented in the patch author guide:  You have to be very careful if there are any function pointers to a patched function.

In fact, I wonder if we could detect such an issue (a pointer to a patched function exists) and warn about it in kpatch-build.

euspectre:
Well, if KPatch did not optimize the references to the patched functions when loading the patch module, only the addresses of the original functions were used and this problem would not happen. 

Because of how KPatch core loads the patch module, it is perfectly OK to have the pointers to the patched functions. Such optimization of function references seems valuable.

I think, even with unload hooks available, having rcu_barrier() as a "safety net" in the exit function is worth it. It makes sure no RCU callbacks, not only the one from this particular patch, could harm the system when the patch module unloads.

Timer callbacks, workqueue functions and some other callbacks that might be set from the patched code could still be problematic - but unlike RCU, these should be waited for separately. There is no way to wait for all such pending callbacks at once, so one has to use hooks for these anyway.

jpoimboe:
> Well, if KPatch did not optimize the references to the patched functions when loading the patch module, only the addresses of the original functions were used and this problem would not happen.

Hm.  So maybe a better fix would be to ensure that function pointers always refer to the original function instead of the patched function?

euspectre:
Perhaps, it would be more reliable, yes.

jpoimboe:
> However, I cannot see if it is doable for the callbacks only rather than for all functions. In an arch-independent way, I mean.

Here's how I think it can be detected on x86:
```
    4154:       48 c7 c6 00 00 00 00    mov    $0x0,%rsi
                        4157: R_X86_64_32S      .text+0xb50
    415b:       e8 00 00 00 00          callq  4160 <netlink_release+0x3a0>
                        415c: R_X86_64_PC32     call_rcu-0x4
```
We can look for a `R_X86_64_32S` relocation which references a symbol to a replacement function.  In such a case we can convert it to a klp rela for the original function.

> Perhaps, some performance measurements could show if using the pointers to the patched functions directly is really worth it. It looks so in the code, esp. for the "old" kpatch which does a table lookup in the common Ftrace filter each time the function is called. Livepatch does things differently, IIRC.

Performance is important, but safety is more important.  I would much rather be safe and take a small performance hit.  And anyway, kpatch.ko does a hash lookup which should be pretty fast.

euspectre:
So, we can detect the assignment of callbacks by checking the relocation type and without decoding instructions and such, at least, on x86. This way we would not have to disable the pointer optimization everywhere but rather, for the callbacks only. Thanks for the explanation!

I'll experiment with that when I have time.

euspectre:
BTW, I have reproduced the issue using  workqueue functions instead of RCU callbacks. I created a patch for create_user_ns() and free_user_ns() in a VZ7 kernel (based on version 3.10.0-693.1.1 from RHEL). free_user_ns() is a work function set by create_user_ns().

Loading/unloading the patch module in a loop in parallel with 'unshare --map-root-user --user sh -c whoami' triggered the crash in a couple of minutes.

So, yes, a more general solution rather than a fix for just RCU callbacks is needed.

I'll try to prepare a patch based on your suggestion.

jpoimboe:
Ok, thanks for looking at it!  I haven't looked at what's needed, I'm guessing it's a change to kpatch_create_intermediate_sections().  Let me know if you need any help.

euspectre:
Here it is, please take a look.

The patches created by kpatch-build with create-diff-object changed this way have passed my tests with RCU and workqueue so far.

euspectre:
Comments?

jpoimboe:
Sorry, I've been swamped with other work and haven't had a chance to look at it.  I'll try to review it soon...

disaster123:
Just a note. This one gives me a segfault in create-diff-obj[2592]: segfault at 7f3fa3faea81 ip 00007f3fa364edcc sp 00007ffdf4c064a0 error 4 in libc-2.19.so[7f3fa3604000+1a1000]

euspectre:
@disaster123: Strange. Which source patch kpatch-build was processing when the segfault happened?

And - for which kernel?

disaster123:
Patch is this one:
https://pastebin.com/zNUK5C2U

Kernel is an "older" OpenSuSE 42.3 Kernel - source / base:
https://github.com/openSUSE/kernel-source/tree/9c032968bf05aaff3eb55048440c07249e99db33

euspectre:
Thanks for the info. I'll try to reproduce the issue.

disaster123:
Were you able to reproduce?

euspectre:
Unfortunately, I have my hands full right now. Perhaps, I'll find time for that later this week.

euspectre:
I have reproduced the problem, investigating...

euspectre:
Nested function in bch_cache_show(). No surprise. They already caused problems in the past.

The problem happens when create-diff-object processes drivers/md/bcache/sysfs.o.

<pre>
(gdb) bt
<...>
#5  0x0000000000407aef in error (__format=0x410468 "ERROR: %s: %s: %d: lookup_local_symbol %s:%s needed for %s", __errnum=0, __status=1)
    at /usr/include/bits/error.h:42
#6  kpatch_create_intermediate_sections (kelf=0x623cc0, table=0x75f860, 
    hint=0x7ffff7fb4541 <error: Cannot access memory at address 0x7ffff7fb4541>, objname=0x7fffffffe752 "bcache", 
    pmod_name=0x7fffffffe774 "test_patch") at create-diff-object.c:2426
#7  0x00000000004093f6 in main (argc=7, argv=0x7fffffffe318) at create-diff-object.c:2943
</pre>

create-diff-object.c:
```
        if (rela->sym->bind == STB_LOCAL) {
	/* An unchanged local symbol */
	ret = lookup_local_symbol(table,
		rela->sym->name, &result);
	if (ret)
		ERROR("lookup_local_symbol %s:%s needed for %s",
		hint, rela->sym->name, sec->base->name); // <<< Here
```
create-diff-object was trying to process the reference to the function `cmp.36204` in `bch_cache_show()`. From the source code of bch_cache_show() in drivers/md/bcache/sysfs.c:

```
        if (attr == &sysfs_priority_stats) {
		int cmp(const void *l, const void *r)
		{	return *((uint16_t *) r) - *((uint16_t *) l); }
	<...>
	sort(p, n, sizeof(uint16_t), cmp, NULL);
```
Technically, cmp() is a callback but I'd rather skip references to such nested functions in kpatch_create_intermediate_sections() somehow. I doubt they are used as asynchronous callbacks anywhere in the kernel. Therefore it should be OK to process the references to such callbacks the same way as before, without dynrelas.

The question is how to detect such functions in create-diff-object.

euspectre:
Assuming GCC creates the names for the nested functions using the same rules as for static locals (xxxx.nnnn), we could probably check if the name of the referred-to functions contains a dot.

euspectre:
By the way, the actual crash happens because the source file name ("sysfs.c", referred to by `hint` variable) has been freed by `kpatch_elf_free(kelf_base)` by that time.

This is a separate issue. The original issue with the nested function only helped reveal it.
Moving the call `kpatch_elf_free(kelf_base)` to the similar calls at the end of `main()` should fix it.

euspectre:
I have updated the fixes, please review.

It seems, there is no way to recognize nested functions in the binary code except by their names. On the other hand, I checked a few kernels from different distros and all callbacks with '.' in their names were nested functions. So, my patch uses this to distinguish nested functions and the ordinary ones.

As for the crashes due to 'hint' pointing to nowhere, I think it is better to copy the relevant string rather than to defer freeing of `kelf_base`. This is what the second patch does.

kpatch-build/create-diff-object.c
```diff
@@ -2866,6 +2866,10 @@ int main(int argc, char *argv[])
 	if (!hint)
 		ERROR("FILE symbol not found in base. Stripped?\n");
 
+	hint = strdup(hint);
+	if (!hint)
+		ERROR("failed to copy file name.");
+
```
joe-lawrence: This is _really_ minor, but if we move these lines up into the search for STT_FILE, there would only be a single assignment to hint, like `hint = strdup(sym->name);`

kpatch-build/create-diff-object.c
```diff
@@ -2303,6 +2303,23 @@ static int kpatch_is_core_module_symbol(char *name)
...
+#ifdef __x86_64__
+static int may_need_dynrela(const struct rela *rela)
+{
+	if (!rela->sym->sec)
+		return 1;
+
+	return (rela->sym->type == STT_FUNC &&
+		rela->type == R_X86_64_32S &&
+		!strchr(rela->sym->name, '.'));
+}
```
joe-lawrence: Will this inadvertently affect function symbols like `aio_setup_rw.isra.17` or `percpu_ref_get_many.constprop.21`?  I don't have any examples off hand, but I know I've seen the compiler generate names like that in the past.  (I haven't followed this PR closely, so apologies if such names don't apply here.)

euspectre: I had the same question some time ago, so I checked a few kernels from RHEL, VZ, OpenSUSE and Ubuntu - all functions with dots in their names that were used in the relocations of such kind (R_X86_64_32S) were nested ones. GCC seems currently not to use the optimized copies of the functions (*isra*, *constprop*, ...) when an address of a callback is needed.

Of course, I cannot be 100% sure this is the case for all kernels and all GCC versions. Ingenuity of the compilers should not be underestimated.

Still, this patch covers all problematic cases I have encountered so far. Even if some non-nested function with a dot is used as a callback, the patch does not make things worse than they were before it.

joe-lawrence: Thanks for the explanation and due diligence research.  One more high level question: if kpatch patches a static local callback named "foo" and there are several other "foo" function symbols scattered across vmlinux/modules, would this change resolve "foo" to the appropriate object file?  (ie, the original implementation of the instance that kpatch modified.)

Also, this shouldn't hold up the patch, but is it possible to convert any of the input patch repros into a non-crashing integration test?

euspectre: >  One more high level question: if kpatch patches a static local callback named "foo" and there are several other "foo" function symbols scattered across vmlinux/modules, would this change resolve "foo" to the appropriate object file?

If I understand the dynrela handling code correctly, it should be OK.
There are 2 cases here:

1. There are 2 or more 'foo' functions and all these are changed by the patch. KPatch/Livepatch core will use both function name and its index ('sympos') in the list of such symbols in the patch module to find the right 'foo'.
2. One 'foo' function is changed by the patch and another one is not. The relocations for the former will be resolved the same way as in (1). For the second 'foo' function, KPatch/Livepatch core will assume it is exported by the kernel or a module and will search for it appropriately.

euspectre: > Also, this shouldn't hold up the patch, but is it possible to convert any of the input patch repros into a non-crashing integration test?

It might be possible but, perhaps, quite cumbersome.

One could prepare a patch that adds nops in `create_user_ns()` and `free_user_ns()` (kernel/user_namespace.c), for example. Build the binary patch and check its binary code, e.g. with objdump or libelf-based app. Find the place where the address or `free_user_ns()` is loaded somewhere (this may be tricky) and check if there is a direct relocation of type R_X86_64_32S there. If there is one, the test fails.

If there is no direct relocations there, the test may try to load the patch module, do some unshare / clone(NEW_USER) operations, to make sure the patched code is at least called.

Something like that. I am not familiar with the integration test suite of KPatch yet, so  I cannot say if this can be implemented here.

euspectre: > if we move these lines up into the search for STT_FILE, there would only be a single assignment to hint, like hint = strdup(sym->name);
Good point. Will do it.

euspectre: > how about the attached .patch file? The idea here being that adding the function pointer should be enough to generate the relocation. (Ie, I don't think we even need to call it.)

You are right.

At least on my system, the same kind of relocation is created for `fp = &cmdline_proc_show;` in your test as for the callback assignements in my patches.

I have checked that the different addresses are found and the test fails if the patch is built using the original create-diff-object,  without the fix. The test passes if the fixed create-diff-object is used.

joe-lawrence: 👍 seems like a reasonable change, at least for x86 (I'm not very knowledgeable about other arches we'd (like to) support, though.)  We can add integration tests in a follow-up commit.

I'll give @jpoimboe some time to review in case he still wants to comment.

jpoimboe: Apologies, spectre/meltdown has me pulled in other directions.  I hope to get to this soonish.

kpatch-build/create-diff-object.c
```diff
@@ -2361,8 +2378,9 @@ static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
 		    !strcmp(sec->name, ".rela.kpatch.dynrelas"))
 			continue;
 		list_for_each_entry_safe(rela, safe, &sec->relas, list) {
-			if (rela->sym->sec)
+			if (!may_need_dynrela(rela))
```
jpoimboe: This would be more readable if it were restructured slightly:
```
if (rela->sym->sec && !function_ptr_rela(rela))
```
In other words, keep the original check, but put the new check in a dedicated `function_ptr_rela()` function.

And also `function_ptr_rela()` needs some comments, specifically a) why we do dynrelas for function pointers; and b) why functions with '.' are skipped.

kpatch-build/create-diff-object.c
```diff
@@ -2833,7 +2859,9 @@ int main(int argc, char *argv[])
 
 	list_for_each_entry(sym, &kelf_base->symbols, list) {
 		if (sym->type == STT_FILE) {
-			hint = sym->name;
+			hint = strdup(sym->name);
+			if (!hint)
+				ERROR("failed to copy file name.");
```
jpoimboe: I think we can take a different approach here.  The hint is actually no longer needed in kpatch_create_patches_sections() and kpatch_create_intermediate_sections().  It _used_ to be needed for the lookup_local_symbol() calls, but that code was restructured and we forgot to update the error messages accordingly.  The error messages no longer need to show the hint.

kpatch-build/create-diff-object.c
euspectre: OK, will do it.

kpatch-build/create-diff-object.c
euspectre: So, we can get rid of that 'hint' altogether. Good.

euspectre: > So, we can get rid of that 'hint' altogether. Good.

Ah, I see. `hint` is still needed for `lookup_open()` and such but it is no longer necessary after `kelf_base` is freed.

euspectre:
I have updated the patches.

In the first one, I kept `may_need_dynrela` because it seems more readable in my opinion but moved the relevant conditions into `function_ptr_rela` like you suggested.

I also added a comment describing what this all is about. Quite lengthy, I admit, but it may help other people understand this part of the code better. Well, in several months, I would also appreciate that explanation myself after I have forgotten the details ;-)

euspectre:
Corrected the comment and pushed the patches again.

jpoimboe:
Awesome comment.  The code looks good, my only concern is that this only fixes x86, and ppc64le will still have the same issue.  We should try to figure out how to detect function pointer relocations on ppc for may_need_dynrela().

@kamalesh-babulal

jpoimboe:
I looked at a ppc64le version of net/netlink/af_netlink.o, and it looks like it should be easy to detect function pointer assignment.  Here's an example:
```
    51dc:       00 00 82 3c     addis   r4,r2,0
                        51dc: R_PPC64_TOC16_HA  .text+0x20b0
    51e0:       b8 03 7d 38     addi    r3,r29,952
    51e4:       00 00 84 38     addi    r4,r4,0
                        51e4: R_PPC64_TOC16_LO  .text+0x20b0
    51e8:       01 00 00 48     bl      51e8 <netlink_release+0x328>
                        51e8: R_PPC64_REL24     call_rcu_sched
```
And in kpatch's case, the above .text+0x20b0 relocations will be converted to their function symbols by kpatch_replace_sections_syms(), so they will actually be:
```
    51dc:       00 00 82 3c     addis   r4,r2,0
                        51dc: R_PPC64_TOC16_HA  deferred_put_nlk_sk
    51e0:       b8 03 7d 38     addi    r3,r29,952
    51e4:       00 00 84 38     addi    r4,r4,0
                        51e4: R_PPC64_TOC16_LO  deferred_put_nlk_sk
    51e8:       01 00 00 48     bl      51e8 <netlink_release+0x328>
                        51e8: R_PPC64_REL24     call_rcu_sched
```
So I think the code would be something like
```
static int function_ptr_rela(const struct rela *rela)
{
       return (rela->sym->type == STT_FUNC &&
	       (rela->type == R_X86_64_32S ||
		rela->type == R_PPC64_TOC16_HA ||
		rela->type == R_PPC64_TOC_LO));
}
```
And then we could get rid of the `#ifdef __x86_64__`.

kamalesh-babulal:
> static int function_ptr_rela(const struct rela *rela)
{
       return (rela->sym->type == STT_FUNC &&
	       (rela->type == R_X86_64_32S ||
		rela->type == R_PPC64_TOC16_HA ||
		rela->type == R_PPC64_TOC_LO));
}

A minor nitpick, ` R_PPC64_TOC_LO` relocation type should be `R_PPC64_TOC16_LO_DS`.  I will try review and test the patch on PowerPC by tomorrow.

euspectre:
@kamalesh-babulal, @jpoimboe: I have updated the patch based on your suggestions to make testing easier.

kamalesh-babulal:
Thank you @euspectre. I managed to try your patch yesterday and observed that, the way the relocation entries of function relies on `.toc section` for resolving the address of the function, make it's more complex to resolve to original function for certain entries and to patched function for rest of the relocation entries.

On PowerPC, the kernel for `kpatch` is compiled with `-mcmodel=large`, which translates below
the code (compiled with `-mcmodel=medium`)
```
    51dc:       00 00 82 3c     addis   r4,r2,0
                        51dc: R_PPC64_TOC16_HA  deferred_put_nlk_sk
    51e0:       b8 03 7d 38     addi    r3,r29,952
    51e4:       00 00 84 38     addi    r4,r4,0
                        51e4: R_PPC64_TOC16_LO  deferred_put_nlk_sk
    51e8:       01 00 00 48     bl      51e8 <netlink_release+0x328>
                        51e8: R_PPC64_REL24     call_rcu_sched
```
into:
```
 3a8:   00 00 02 3d     addis   r8,r2,0
                        3a8: R_PPC64_TOC16_HA   .toc+0x138
 3ac:   c0 03 7d 38     addi    r3,r29,960
 3b0:   00 00 88 e8     ld      r4,0(r8)
                        3b0: R_PPC64_TOC16_LO_DS        .toc+0x138
 3b4:   01 00 00 48     bl      3b4 <netlink_release+0x3ac>
                        3b4: R_PPC64_REL24      call_rcu_sched
 3b8:   00 00 00 60     nop
```
and the respective `.toc` section entry will be

```  
Relocation section '.rela.toc' at offset 0xf38 contains 24 entries:
...
0000000000000138  0000000200000026 R_PPC64_ADDR64         0000000000000008 deferred_put_nlk_sk + 0
```
If the `rela->sym`  is `.toc + addend`, then the loop should be reading the symbol at `.toc section + offset (offset == rela->addend)`. The loop can be written as:
```
        if (!strcmp(sec->name, ".rela.toc")) { /* Fallback for older method for .toc section */
                if (rela->sym->sec)
                        continue;
        } else {
                if (rela->sym && !strncmp(rela->sym->name, ".toc", 4)) {
                        if (rela->sym->sec && rela->sym->sec->base) {
                            r = find_rela_by_offset(rela->sym->sec->base, rela->addend);
                            if (r) {
                                if (!may_need_dynrela(r))
                                    continue;
                                else
                                {
                                /*
                                 * Create a relocation entry in .toc
                                 * with the r->sym and change the binding
                                 * to STB_GLOBAL.
                                 * Also the change the rela->addend to this
                                 * new entry ?
                                 */
                                continue;
                                }
                            }
                    }
                }
                if (!may_need_dynrela(rela))
                        continue;
        }
```

`.toc+offset` symbol cannot be just converted into dynrela, which would convert all the call sites to
resolve to the original function address instead of patched function address. The reason being, the function entry might be `.toc + offset`, might different relocation entries loading it. To replicate the current approach done on x86, we may need to duplicate rela entry in `.toc section` with `STB_GLOBAL` version of function symbol and also convert `.toc+0x138` to new `.toc+offset` in `netlink_release`.

Another question is about reliable detection of rcu_callback symbols. There might be additional check needed. For example
```
 37c:   00 00 22 3d     addis   r9,r2,0
                        37c: R_PPC64_TOC16_HA   .toc+0x70
 380:   30 00 7d e8     ld      r3,48(r29)
 384:   ff ff a0 38     li      r5,-1
 388:   00 00 89 e8     ld      r4,0(r9)
                        388: R_PPC64_TOC16_LO_DS        .toc+0x70
 38c:   01 00 00 48     bl      38c <netlink_release+0x384>
                        38c: R_PPC64_REL24      sock_prot_inuse_add
 390:   00 00 00 60     nop
 394:   00 00 42 3d     addis   r10,r2,0
                        394: R_PPC64_TOC16_HA   .toc+0x130
 398:   00 02 80 38     li      r4,512
 39c:   00 00 6a e8     ld      r3,0(r10)
                        39c: R_PPC64_TOC16_LO_DS        .toc+0x130
 3a0:   01 00 00 48     bl      3a0 <netlink_release+0x398>
                        3a0: R_PPC64_REL24      __local_bh_enable_ip
 3a4:   00 00 00 60     nop
 3a8:   00 00 02 3d     addis   r8,r2,0
                        3a8: R_PPC64_TOC16_HA   .toc+0x138
 3ac:   c0 03 7d 38     addi    r3,r29,960
 3b0:   00 00 88 e8     ld      r4,0(r8)
                        3b0: R_PPC64_TOC16_LO_DS        .toc+0x138
 3b4:   01 00 00 48     bl      3b4 <netlink_release+0x3ac>
                        3b4: R_PPC64_REL24      call_rcu_sched
```
like `call_rcu_sched`, both `sock_prot_inuse_add` and `__local_bh_enable_ip` follow the same sequence of loading `R_PPC64_TOC16_HA/R_PPC64_TOC16_LO_DS`. It seems non-trivial on PowerPC to reliably detecting the relocation entries those should be dynrela.

euspectre:
I see. If I understand you correctly, it is currently unclear how to fix things for PowerPC properly.

If so, I'd suggest to add the fix for x86 first if there are no visible problems with it and handle PowerPC in some later patch on top of this one. The story of this PR is not extremely long yet but it is already decent ;-)

I'd revert function_ptr_rela to
```
static int function_ptr_rela(const struct rela *rela)
{
	return (rela->sym->type == STT_FUNC &&
		rela->type == R_X86_64_32S;
}
```
for now.

jpoimboe:
Thanks @kamalesh-babulal , I always forget about `-mcmodel=large` and the pain it causes.

> Another question is about reliable detection of rcu_callback symbols. There might be additional check needed. For example
>  ...
> like call_rcu_sched, both sock_prot_inuse_add and __local_bh_enable_ip follow the same sequence of loading R_PPC64_TOC16_HA/R_PPC64_TOC16_LO_DS. It seems non-trivial on PowerPC to reliably detecting the relocation entries those should be dynrela.

We're not trying to detect rcu_callback symbols.  We want to detect *all* function pointers.

For powerpc `-mcmodel=large`, the formula to detect them should still be relatively simple:

```
if ((rela->type == R_PPC64_TOC16_HA || rela->type == R_PPC64_TOC_LO) &&
    toc_rela(rela)->sym->type == STT_FUNC))
```

Where toc_rela() is a helper which follows the .toc rela (e.g. `.toc+0x138`) to obtain the real symbol rela (e.g. `deferred_put_nlk_sk + 0`).

On x86, toc_rela(rela) would just return rela.  I think the toc_rela() helper would be useful in other places as well, like in rela_equal().

jpoimboe:
@euspectre It's ok with me if we do the ppc64le changes in another PR.  However, we should still wait until they're both ready, and merge them at the same time.  Otherwise we would have subtly different behavior between the arches.  A patch that would be safe on x86 could be dangerous on ppc64le.

Right now it's the patch author's responsibility to handle function pointers appropriately.  Once we change that to make it kpatch's responsibility, we should do it for both arches simultaneously.

euspectre:
> However, we should still wait until they're both ready, and merge them at the same time.

@jpoimboe That is fine with me. We may otherwise forget the fix for PowerPC anyway. There are quite a few other bugs to fix and tasks to do, as usual. Hopefully, there will be no new  meltdown/spectre-like bugs in the near future ;-)

I have updated the patch, so a fix for PowerPC could be done on top of it. I cannot be of much help there, ufortunately, becase I know only a little about that architecture. That is why it is better to do that in a separate patch/PR.

kamalesh-babulal:
> For powerpc -mcmodel=large, the formula to detect them should still be relatively simple:
 Thanks @jpoimboe . If the PowerPC changes are be tracked in another PR. I can open one, once we have a patch. I will build upon this approach, I have an idea on how to implement other bits on PowerPC. Need to implement the idea and see what more is needed. Thank you @euspectre.

euspectre:
@kamalesh-babulal
> If the PowerPC changes are be tracked in another PR. I can open one, once we have a patch.

Thanks! Looking forward to it.

joe-lawrence:
If I read the latest commentary correctly, we're going to leave this PR open until all supported arches have equivalent behavior?

If so, can we spin "kpatch-build: 'hint' is not needed in kpatch_create_*_sections()" as an independent PR ... I just hit this bug doing while working on s390x, so it would be nice to get out of the way :)

euspectre:
> If so, can we spin "kpatch-build: 'hint' is not needed in kpatch_create_*_sections()" as an independent > PR ... I just hit this bug doing while working on s390x, so it would be nice to get out of the way :)

@joe-lawrence Makes sense. I have moved that hint-related fix into https://github.com/dynup/kpatch/pull/788 and updated this PR.

jpoimboe:
This is superseded by #789 which has both x86 and ppc versions of the change.

PR #756, opened 2017-11-07
kpatch-build: Fix clean rule (remove insn/*.o)

juergh:
Might result in build failures or unexpected behaviour otherwise.


jpoimboe:
Thanks @juergh !

PR #758, opened 2017-11-09
kpatch-build cleanups

juergh:
Misc kpatch-build fixes and enhancements.


jpoimboe:
Thanks!

One comment: I think `--archversion` is incompatible with `--sourcedir` and `--sourcerpm`, so it would be a good idea to print an error if somebody tries to combine them.

PR #759, opened 2017-11-10
kpatch-build cleanups

juergh:
Added --archversion commandline option conflict detection.

PR #760, opened 2017-11-13
Improve logging

juergh:


jpoimboe:
This is a nice improvement, but as somebody who uses the -d flag a lot, the thing I most often care about is the kpatch-build xtrace (along with the preserving of ~/.kpatch/tmp).

So what do you think about the following?

* -d: xtrace in kpatch-build only (not kpatch-gcc)
* -d -d: all the other stuff (xtrace everywhere, log to stdout)

juergh:
I usually want all the debug output on stdout but without the xtrace noise xtrace :-) How about we keep -d as is (without xtrace in kpatch-gcc) and add another flag for the new behaviour?

juergh:
How about this: https://github.com/juergh/kpatch/tree/improve-logging-v2
And specifically commit https://github.com/juergh/kpatch/commit/9cfb5cb3ce85df284eeb6dcd34d38a70bd338073
which introduces a new commandline option to enable the alternate debug mode.

jpoimboe:
That could work, but I'd like to try to keep the debug options as simple as we can.  I'm ok with running kpatch-build as `bash -x kpatch-build` to get an xtrace.  But I still need the ~/.kpatch/tmp from being deleted.  So how about

* -d: don't delete ~/.kpatch/tmp
* -d -d: ... and log to stdout

Then I can run `bash -x kpatch-build -d` and you can run `kpatch-build -d -d`.  That sound ok?

jpoimboe:
Ok, that sounds good.

juergh:
Force pushed to branch improve-logging.
@jpoimboe Do you have an estimate for when this might get merged? I don't want to be too far ahead of current master with my other changes.

jpoimboe:
Sorry, for the delay, we'll try to merge #759 soon.

This new commit isn't working for me.  getopt is interpreting "-dd" as "-d -d", so it just enables xtrace but nothing else.

jpoimboe:
👍

joe-lawrence:
The latest version works as expected.  Thanks @juergh

PR #762, opened 2017-11-27
testing fedora 27: rebase integration test patches

joe-lawrence:
This patchset only rebases the code and removes a few unsupported test patches.  Running the entire integration test suite renders a few build errors in which I will open up separate issues for.


jpoimboe:
Assuming the tests still work, looks good to me.

@joe-lawrence in the future it's fine with me if you just push these test updates through without a review.

jpoimboe:
Actually I have a question about the second commit.  Fedora doesn't enable CONFIG_LIVEPATCH, does it?  So aren't we still using the kpatch core module for Fedora?

joe-lawrence:
> Assuming the tests still work, looks good to me.

I haven't tested with the latest Fedora 27 kernel, only the one it GA'ed with.  Any build or load problems from these were opened as separate issues.

> Actually I have a question about the second commit. Fedora doesn't enable CONFIG_LIVEPATCH, does it? So aren't we still using the kpatch core module for Fedora?

Oh yeah, that's right.  It's been a while since I opened this PR, so I don't know what I was thinking. :D 
I can drop that second commit when pushing.

joe-lawrence:
There's work on reforming our tests (see #798), but while that's in progress, I just wanted to clear out this old PR and push the rebased Fedora integration tests out.  Fedora 28 will probably be out soon and render all of this out of date.  Again.

PR #772, opened 2017-12-03
Add Gentoo to README

aliceinwire:


README.md
```diff
@@ -270,6 +270,28 @@ Configure ccache (installed by kpatch package):
...
+#### Gentoo
+
+*NOTE: You'll need about 15GB of free disk space for the kpatch-build cache in
+`~/.kpatch` and for ccache.*
+
+Install Kpatch and Kpatch dependencies:
+
+```bash
+emerge --ask sys-kernel/kpatch
+```
```
joe-lawrence: This is a little different than the instructions for the other distros, which install build and runtime dependencies rather than the kpatch package itself.  (Please forgive my lack of Gentoo knowledge here.)  I suppose the net effect might be equivalent for a source-based distros like Gentoo, but with the added side effect of clean kpatch binaries in their installed paths.  Could that lead to confusion if the user decides to then build kpatch.git but not `make install` ?

README.md
```diff
@@ -270,6 +270,28 @@ Configure ccache (installed by kpatch package):
...
+
+```bash
+emerge --ask dev-util/ccache
+```
+
+Configure ccache:
+
+```bash
+ccache --max-size=5G
+```
```
joe-lawrence: Does `kpatch-build` work out of the box for Gentoo?  There's a bunch of distro-specific code there to fetch proper kernel sources and whatnot.

aliceinwire: yes, it work out of the box but only by using -s for specifying the source directory.
Of course -r dosen't work.

README.md
aliceinwire: actually i'm not sure what the other distro are doing.
with emerge --ask sys-kernel/kpatch it will install all the needed dependency for having a working kpatch in the system that is recommended instead of installing it from source.
user also have a live package that can be enabled using the live flag for download and install the latest kpatch from github.

README.md
joe-lawrence: Okay, good.  How about the kernel sources on the user's machine?  Are kernel sources always available on a Gentoo install somewhere?

aliceinwire: yes, usually they are always present under /usr/src/linux/ 
the user than install the own kernel by make && make install
Gentoo is only providing the kernel sources, we don't provide any kernel binary.

README.md
joe-lawrence: The Installation section for the other distros could have simply said `yum/dnf/apt-get install kpatch` and let the package manager do all of the heavy lifting.  However, the way it's currently written, it is as if the user has a fairly minimal (or default) distro install and needs to build and use kpatch from these git sources.  IOW, I read this section of  `README.md` as a user-readable translation of a kpatch specfile.  I suspect it was originally written this way before distributions were providing kpatch packages -- maybe it's time we update to suggest the user installs official builds unless the they know what they're doing.

For the time being, I think the instructions are targeting developers looking to get started on kpatch.git development work on their given distro (without installing an official kpatch package).

aliceinwire: I think for Gentoo is still better if the user use emerge for installing Kpatch, for not breaking the distribution FHS.
Is maybe better writing both way ?

README.md
joe-lawrence: Once the user has built their kernel, are they free to remove /usr/src/linux (or via the Gentoo package manager?)  If so, providing the emerge line(s) could be helpful.

On a related note, there was a recent enhancement: 1950c193 ("kpatch-build: Add -a, --archversion commandline option") that added the ability to build a kpatch module for a non-running kernel.  It isn't compatible with the `-s|--sourcedir` option, but maybe could be made useful for Gentoo if specific kernel version sources are emerge-able.  This is outside the scope for this PR, but I thought it maybe of interest for the kpatch build-server tool that you are developing.

aliceinwire: the source code under /usr/src/linux is linked by eselect kernel and is suggested not to remove. But it could have been clean by emerge if the user is moving to a new kernel and after calling the cleaning function. 
elivepatch have a special function for emerge old kernel, because we are keeping only recent kernel in the Gentoo-sources repository tree. (https://github.com/gentoo/gentoo/tree/master/sys-kernel/gentoo-sources)
Maybe I could try adding it to kpatch-build, in a later pull request. That function is using a overlay for getting old kernel. (https://github.com/aliceinwire/gentoo-sources_overlay/tree/master/sys-kernel/gentoo-sources)
something like -g for emerging old Gentoo sources.

joe-lawrence:
👍 ... let's go with this, thanks for adding the Gentoo instructions!

PR #777, opened 2017-12-27
version update

aliceinwire:


jpoimboe:
Oops, sorry about that.  I do releases so rarely that I sometimes forget to update these files.

PR #782, opened 2018-01-11
dnf to yum

bmcculley:
Switching dnf to yum under RHEL, CentOS, and oracle.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ sudo yum install ccache
 ccache --max-size=5G
 
 # optional, for kpatch-test
-sudo dnf install patchutils
+sudo yum install patchutils
 ```
 
 #### Oracle Linux 7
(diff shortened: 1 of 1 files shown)


joe-lawrence:
Hi @bmcculley, I don't believe that dnf is available by default on RHEL7 or CentOS7.  At least I didn't see it in any official package list:
* http://mirror.centos.org/centos/7/os/x86_64/Packages/
* http://mirror.centos.org/centos/7/extras/x86_64/Packages/
* http://mirror.centos.org/centos/7/updates/x86_64/Packages/

If you do have dnf installed on such systems, what does `yum info $(rpm -q --whatprovides $(which dnf))` output?  It should include a "From repo" line listing its repository source.

bmcculley:
Hi @joe-lawrence, yes that's what I am saying that dnf is not available by default on those systems which is why I am suggesting that the last command under those operating systems that say to use **dnf** be _switched_ to **yum** instead  (these seem to be copy pasted from fedora). 

Please look at the [changes](https://github.com/dynup/kpatch/pull/782/files), thank you.

joe-lawrence:
Oops, my bad --  you are correct here, we did had inadvertent `dnf` invocations where they shouldn't be.  Good eye and thanks for fixing! 👍

PR #784, opened 2018-02-08
kpatch: a couple small enhancements in retry logic

euspectre:
The first of the changes is to makes sure we get "Device or resource busy" message in English when it is output by insmod. kpatch does not expect it in any other language, so let us force "C" locale here.

The other change is rather small too, it just makes it easier to adjust the number of attempts to load the patches, as well as the interval between retries by putting them into easy to find variables. This is especially helpful for the "old-style" (not livepatch-based) patches when doing stress testing, etc. In such conditions, I often had to increase both the number of attempts and the interval.


jpoimboe:
👍

PR #786, opened 2018-02-16
kmod/core: fix definition of KERNELRELEASE

euspectre:
A cosmetic fix.

If KPATCH_BUILD ending with 'build/' is passed to 'make', KERNELRELEASE
will become 'build' and the error message will look like:

"<...> doesn't exist. Try installing the kernel-devel-build RPM or
linux-headers-build DEB."

Let us fix that.

### DIFF
--- a/kmod/core/Makefile
+++ b/kmod/core/Makefile
@@ -1,6 +1,6 @@
 # make rules
 KPATCH_BUILD ?= /lib/modules/$(shell uname -r)/build
-KERNELRELEASE := $(lastword $(subst /, , $(dir $(KPATCH_BUILD))))
+KERNELRELEASE := $(lastword $(subst /, , $(dir $(patsubst %/,%,$(KPATCH_BUILD)))))
 THISDIR := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
 
 ifeq ($(wildcard $(KPATCH_BUILD)),)


kmod/core/Makefile
```diff
@@ -1,6 +1,6 @@
 # make rules
 KPATCH_BUILD ?= /lib/modules/$(shell uname -r)/build
-KERNELRELEASE := $(lastword $(subst /, , $(dir $(KPATCH_BUILD))))
+KERNELRELEASE := $(shell echo "$(KPATCH_BUILD)" | sed -r 's,.*/lib/modules/([^/]+)/build.*,\1,')
```
jpoimboe: Shelling out can be fragile, how about something like this:
```
diff --git a/kmod/core/Makefile b/kmod/core/Makefile
index 3a01339..8aaa3f5 100644
--- a/kmod/core/Makefile
+++ b/kmod/core/Makefile
@@ -1,6 +1,6 @@
 # make rules
-KERNELRELEASE := $(lastword $(subst /, , $(dir $(KPATCH_BUILD))))
+KERNELRELEASE := $(lastword $(subst /, , $(dir $(abspath $(KPATCH_BUILD)))))
 THISDIR := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
 
 ifeq ($(wildcard $(KPATCH_BUILD)),)
```

euspectre: > -KERNELRELEASE := $(lastword $(subst /, , $(dir $(KPATCH_BUILD))))
> +KERNELRELEASE := $(lastword $(subst /, , $(dir $(abspath $(KPATCH_BUILD)))))

It works on my machine. But can we rely on $abspath to always remove the trailing slash? The documentation does not guarantee that explicitly:

    $(abspath names…)
    For each file name in names return an absolute name that does not
    contain any . or .. components, nor any repeated path separators (/).
    Note that, in contrast to realpath function, abspath does not resolve
    symlinks and does not require the file names to refer to an existing
    file or directory. Use the wildcard function to test for existence. 

It might be an implementation-specific behaviour.

jpoimboe: You're right, it doesn't explicitly say that it removes the trailing slash, but it's at least implied, IMO.  Which is good enough for me.  Or if you have some other way to do it without shelling out, that would also be fine.

joe-lawrence: Dunno if this is any more elegent than `abspath`, but there's always pattern substitution:
```
KERNELRELEASE := $(lastword $(subst /, , $(dir $(patsubst %/,%,$(KPATCH_BUILD)))))
```

euspectre: @joe-lawrence 
> Dunno if this is any more elegent than abspath, but there's always pattern substitution:

Good idea. That will do. I'll update the patch.

PR #787, opened 2018-02-20
Do not mention initrd in the help messages and in the man page

euspectre:
Kpatch no longer uses initrd to make sure the patch modules are loaded
at boot. The users could either install the provided systemd service
for that or come up with some other solution.

The messages mentioning initrd could confuse the users.


euspectre:
Another cosmetic fix.

jpoimboe:
:+1:

PR #788, opened 2018-02-22
kpatch-build: 'hint' is not needed in kpatch_create_*_sections()

euspectre:
Found in the scope of https://github.com/dynup/kpatch/pull/755 but not
related to the main problem discussed there.

kpatch_create_patches_sections() and kpatch_create_intermediate_sections()
used 'hint' in error messages.

However, the string 'hint' refers to is owned by 'kelf_base' and is
freed before kpatch_create_*_sections() are called. As a result, if
these functions try to output errors and print 'hint',
create-diff-object will crash.

As suggested in the mentioned PR, 'hint' is actually no longer needed at
that stage, so I have removed it from kpatch_create_*_sections().


joe-lawrence:
:thumbsup:    Thanks, @euspectre !

PR #789, opened 2018-02-26
[RFC] PC64le - do not use the patched functions as callbacks directly

kamalesh-babulal:
It was observed by Evgenii Shatokhin in #755, that when the RCU callback was called on the patched function, from unloaded livepatch module triggered a kernel crash.

This patch implements the approach on Powerpc outlined in #755. With -mcmodel=large, like any other data, function pointers are also loaded relative to the current TOC base and are populated as relocation entries in .toc section. Every function call/passing such function as pointers needs to access them through .toc section + offset.

The simplest approach would be to convert the .toc + offset relocation into a dynamic rela referring to the original function address but the same function is also a patched function. Thus, on PowerPC we need to append a new relocation entry into .rela.toc, duplicating the entry used as a function pointer and converting the newly added rela as dynamic rela.

Also, modify the function to load the new entry by modifying the addend value to newly created dynamic rela.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2328,6 +2328,55 @@ static int kpatch_is_core_module_symbol(char *name)
 		!strcmp(name, "kpatch_shadow_get"));
 }
 
+/*
+ * If the patched code refers to a symbol, for example, calls a function
+ * or stores a pointer to a function somewhere, the address of that symbol
+ * must be resolved somehow before the patch is applied. The symbol may be
+ * present in the original code too, so the patch may refer either to that
+ * version of the symbol (dynrela is used for that) or to its patched
+ * version directly (with a normal relocation).
+ *
+ * Dynrelas may be needed for the symbols not present in this object file
+ * (rela->sym->sec is NULL), because it is unknown if the patched versions
+ * of these symbols exist and where they are.
+ *
+ * The patched code can usually refer to a symbol from this object file
+ * directly. If it is a function, this may also improve performance because
+ * it will not be needed to call the original function first, find the
+ * patched one and then use Ftrace to pass control to it.
+ *
+ * There is an exception though, at least on x86. It is safer to use
+ * a dynrela if the patched code stores a pointer to a function somewhere
+ * (relocation of type R_X86_64_32S). The function could be used as
+ * a callback and some kinds of callbacks are called asynchronously. If
+ * the patch module sets such callback and is unloaded shortly after,
+ * the kernel could try to call the function via an invalid pointer and
+ * would crash. With dynrela, the kernel would call the original function
+ * in that case.
+ *
+ * Nested functions used as callbacks are a special case. They are not
+ * supposed to be visible outside of the function that defines them.
+ * Their names may differ in the original and the patched kernels which
+ * makes it difficult to use dynrelas. Fortunately, nested functions are
+ * rare and are unlikely to be used as asynchronous callbacks, so the
+ * patched code can refer to them directly. It seems, one can only
+ * distinguish such functions by their names containing a dot. Other
+ * kinds of functions with such names (e.g. optimized copies of functions)
+ * are unlikely to be used as callbacks.
+ */
+static int function_ptr_rela(const struct rela *rela)
+{
+	return (rela->sym->type == STT_FUNC && rela->type == R_X86_64_32S);
+}
+
+static int may_need_dynrela(const struct rela *rela)
+{
+	if (!rela->sym->sec)
+		return 1;
+
+	return (function_ptr_rela(rela) && !strchr(rela->sym->name, '.'));
+}
+
 static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
 						struct lookup_table *table,
 						char *objname,
@@ -2386,8 +2435,9 @@ static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
 		    !strcmp(sec->name, ".rela.kpatch.dynrelas"))
 			continue;
 		list_for_each_entry_safe(rela, safe, &sec->relas, list) {
-			if (rela->sym->sec)
+			if (!may_need_dynrela(rela))
 				continue;
+
 			/*
 			 * Allow references to core module symbols to remain as
 			 * normal relas, since the core module may not be
@@ -2576,7 +2626,15 @@ static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
 			rela2->offset = index * sizeof(*krelas) + \
 				offsetof(struct kpatch_relocation, ksym);
 
-			rela->sym->strip = 1;
+			/*
+			 * Mark the referred to symbol for removal but
+			 * only if it is not from this object file.
+			 * The symbols from this object file may be needed
+			 * later (for example, they may have relocations
+			 * of their own which should be processed).
+			 */
+			if (!rela->sym->sec)
+				rela->sym->strip = 1;
 			list_del(&rela->list);
 			free(rela);
 
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2328,6 +2328,16 @@ static int kpatch_is_core_module_symbol(char *name)
 		!strcmp(name, "kpatch_shadow_get"));
 }
 
+static struct rela *toc_rela(const struct rela *rela)
+{
+	if (rela->type != R_PPC64_TOC16_HA &&
+	    rela->type != R_PPC64_TOC16_LO_DS)
+		return (struct rela *)rela;
+
+	/* Will return NULL for .toc constant entries */
+	return find_rela_by_offset(rela->sym->sec->rela, rela->addend);
+}
+
 /*
  * If the patched code refers to a symbol, for example, calls a function
  * or stores a pointer to a function somewhere, the address of that symbol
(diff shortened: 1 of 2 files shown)


kamalesh-babulal:
The current patch applies on top of @euspectre patch. I have included the patch from @euspectre tree in this pull request for readability and testing purpose. I will redo the pull request, with my patch alone, in the final version.

kpatch-build/create-diff-object.c
```diff
@@ -2351,17 +2415,83 @@ static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
 	if (!strsym)
 		ERROR("can't find .kpatch.strings symbol");
 
+	/* Allocate toc dynrela array to half of upper bound (nr/2) */
+	toc_dynrela_off = malloc(sizeof(int) * (nr/2));
+	memset(toc_dynrela_off, 0, sizeof(*toc_dynrela_off));
+	off_idx = 0;
+
```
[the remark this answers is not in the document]
kamalesh-babulal: Thanks for the review. Will include the suggestions in next iteration.

kamalesh-babulal:
> How much of the added code in kpatch_create_intermediate_sections() should be enclosed in #ifdef __powerpc__ blocks? (Or was this just a prototype for comment?)

It's a prototype for comments. Changes introduced to `kpatch_create_intermediate_sections` are `powerpc` specific and should be guarded within `#ifdef __powerpc__`.

jpoimboe:
I'm confused about why we have to add a new .toc entry.  Why can't we just convert those .toc.rela function ptr relas to dynrelas?

kamalesh-babulal:
> I'm confused about why we have to add a new .toc entry. Why can't we just convert those .toc.rela function ptr relas to dynrelas?

I got confused here. Yes, we do not need the duplicate entry and just converting the `.toc` to dynrela should be sufficient. I was testing the v2 of this patch (which just converts the .toc to dynrela), against integration test cases and debugging an issue with v2. Will update the finding soon.

kamalesh-babulal:
With #791, `function_ptr_rela` will need be modified, to something like:
```
static int function_ptr_rela(const struct rela *rela)
{
        const struct rela *rela1 = toc_rela(rela);
        return (rela1->sym->type == STT_FUNC &&
                   /* Skip switch table on PowerPC */
                   rela1->addend == rela1->sym->sym.st_value &&
                   (rela->type == R_X86_64_32S ||
                   rela->type == R_PPC64_TOC16_HA ||
                   rela->type == R_PPC64_TOC16_LO_DS));
}
```
Will wait for the comments on #791 and on this pull-request, before re-spinning v3.

kpatch-build/create-diff-object.c
```diff
@@ -2408,10 +2436,45 @@ static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
...
+					 * It's ok to have duplicate .tocoffset,
+					 * instead of linear search to eliminate
+					 * duplicates.
+					 */
+					if (rela->type == R_PPC64_TOC16_HA)
+						toc_dynrela_off[off_idx++] = rela->addend;
+					continue;
+				}
+			}
+#else
```
jpoimboe: The control flow is way too complex and hard to follow.

What if we add a `rela->needs_dynrela` bool, which gets set in the first loop which increments the `nr` variable?  Then we could move the `may_need_dynrela()` call there, and that would hopefully vastly simplify this code (e.g., no need for the `toc_dynrela_off` array or the spaghetti code flow).

kamalesh-babulal: clever :). Will re-do the patch based on your suggestion.

kamalesh-babulal:
**v3 Changes:**
 * Drop the complex looping and array for tracking `.toc` offsets.
 * Introduce `need_dynrela` in relocation struct to decide on dynrela.
 * Don't carry const for `toc_rela()`, returned rela from `toc_rela()`, will be modified.

kpatch-build/create-diff-object.c
```diff
@@ -2326,8 +2392,16 @@ static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
...
+		list_for_each_entry(rela, &sec->relas, list) {
 			nr++; /* upper bound on number of kpatch relas and symbols */
+			if (!strcmp(sec->name, ".rela.toc")) {
+				rela->need_dynrela =
+					(!rela->sym->sec || rela->need_dynrela);
+				continue;
+			}
+			if (may_need_dynrela(rela))
+				toc_rela(rela)->need_dynrela = 1;
+		}
```
jpoimboe: This is a lot better than before, but I'm still confused by this code.  Why are all the toc special cases needed in function_ptr_rela(), may_need_rela(), and here?  This code loops through all the sections, including `.rela.toc.`, so can we just do something like
```
			nr++; /* upper bound on number of kpatch relas and symbols */
			if (may_need_dynrela(rela))
				rela->need_dynrela = 1;
		}
```
along with removing all the toc_rela() calls in function_ptr_rela() and may_need_rela()?

kpatch-build/kpatch-elf.h
```diff
@@ -90,6 +90,7 @@ struct rela {
 	int addend;
 	int offset;
 	char *string;
+	int need_dynrela;
```
jpoimboe: Can you make this a bool?

kpatch-build/kpatch-elf.c
```diff
@@ -186,6 +186,7 @@ void kpatch_create_rela_list(struct kpatch_elf *kelf, struct section *sec)
 		rela->type = GELF_R_TYPE(rela->rela.r_info);
 		rela->addend = rela->rela.r_addend;
 		rela->offset = rela->rela.r_offset;
+		rela->need_dynrela = 0;
```
jpoimboe: No need for this explicit initialization, we already memset the rela struct to zero in ALLOC_LINK.

kpatch-build/create-diff-object.c
kamalesh-babulal:
```
Relocation section '.rela.toc' at offset 0xcc6b0 contains 46 entries:
...
0000000000000128  0000353400000026 R_PPC64_ADDR64         0000000000000000 genl_sk_destructing_waitq + 0
0000000000000130  0000003e00000026 R_PPC64_ADDR64         0000000000000000 .text.netlink_release + 35c
0000000000000138  0000002a00000026 R_PPC64_ADDR64         0000000000000000 .text.deferred_put_nlk_sk + 8

Relocation section '.rela.text.netlink_release' at offset 0xcadf0 contains 44 entries:
...
0000000000000390  000035580000000a R_PPC64_REL24          0000000000000000 __local_bh_enable_ip + 0
0000000000000398  0000007300000032 R_PPC64_TOC16_HA       0000000000000000 .toc + 138
00000000000003a0  0000007300000040 R_PPC64_TOC16_LO_DS    0000000000000000 .toc + 138
```
When looping with:
```
			nr++; /* upper bound on number of kpatch relas and symbols */
			if (may_need_dynrela(rela))
				rela->need_dynrela = 1;
		}
```
`netlink_release + 0x3a0` is not a function symbol, but of relocation type `(R_PPC64_TOC16_HA/R_PPC64_TOC16_LO_DS)`
referring to symbol in `.rela.toc + 0x138`, We need to operate on both `rela->type` and `rela->sym->sec->rela + rela->addend`, where `netlink_release + 0x3a0` is checked for `rela-type` and  `.text.deferred_put_nlk_sk + 8` for STT_FUNC/switch label.

Without `toc_rela(netlink_release + 0x3a0)` the `.toc + 0x138` need_dynrela will not be set and while parsing `.rela.toc_rela` section it's not obvious on if a rela requires, `need_dynrela` to be set, other than for `!rela->sym->sec case`.

Alternative above loop can be re-written as:
```
            nr++; /* upper bound on number of kpatch relas and symbols */
            if (may_need_dynrela(rela))
                toc_rela(rela)->need_dynrela = 1;
        }
```
With an additional check in may_need_dynrela():
```
static int may_need_dynrela(const struct rela *rela)
{
        /* Don't over-ride need_dynrela flag if set, required for .toc sym*/
        if (rela->need_dynrela || !rela->sym->sec)
                return 1;
    ...
}
```
`may_need_dynrela()` and `function_ptr_rela()` operates on both `rela` and `rela->sym->sec->rela`. We might need to keep `toc_rela()` in both.

jpoimboe: Thanks for the explanation.  Sorry, I keep forgetting that the R_PPC64_TOC16_HA relocation is for the original section, not the .toc.  Can you add the above readelf snippet to a comment above the `if (may_need_dynrela(rela))` line explaining why the toc_rela() is getting need_dynrela set?  Otherwise I'll keep getting confused :-)

Can you also explain why you need to check rela->need_dynrela before setting toc_rela(rela)->need_dynrela?  I'm trying to understand how it would be possible for a ".toc + xxx) relocation to be converted to a dynrela.

Assuming we do really need that, how about doing the loop like this?

```
            nr++; /* upper bound on number of kpatch relas and symbols */
            if (!rela->need_dynrela && may_need_dynrela(rela))
                toc_rela(rela)->need_dynrela = 1;
        }
```

Also a comment for the `!strchr(toc_rela(rela)->sym->name, '.')` check in may_need_dynrela() would also be helpful.

Another thing that would make it easier for me to follow would be to rename rela1 to rela_toc in function_ptr_rela().

Thanks!

kamalesh-babulal: > the if (may_need_dynrela(rela)) line explaining why the toc_rela() is getting need_dynrela set? Otherwise I'll keep getting confused :-)   
> Can you also explain why you need to check rela->need_dynrela before setting toc_rela(rela)->need_dynrela? I'm trying to understand how it would be possible for a ".toc + xxx)" relocation to be converted to a dynrela.

:) `.rela.toc` is superseded by `.rela.text.*` sections, those might have set the need_dynrela                         for entries in `.rela.toc.` Without the !rela->need_dynrela check, may_need_dynrela() for `.rela.toc` rela* will unset the need_dynrela flag, failing to rela->type check in function_ptr_rela().
`.rela.toc` rela type is always `R_PPC64_ADDR64`.

> check in may_need_dynrela() would also be helpful.

function_ptr_rela(rela) will return 0, if the rela returned by toc_rela() is NULL. Duplicate check for NULL in may_need_dynrela(), can be avoided.

jpoimboe: > Without the !rela->need_dynrela check, may_need_dynrela() for .rela.toc rela* will unset the need_dynrela flag, failing to rela->type check in function_ptr_rela().

Hm, but if it's already set, how would it get unset?  To clarify, I'm talking about your rewritten version of the loop:
```
            nr++; /* upper bound on number of kpatch relas and symbols */
            if (may_need_dynrela(rela))
                toc_rela(rela)->need_dynrela = 1;
        }
```

kamalesh-babulal: Sorry got little mixed up here, true it will not get unset.  Having the check for need_dynrela, 
`if (!rela->need_dynrela && may_need_dynrela(rela))`  will speed the handling of `.rela.toc` entries but not sure, if its worth the optimization. Arguably `.rela.toc` is only on PPC64le and an additional check for rela entries for non-toc sections might be an overkill. I will re-spin the patch with 
```
            nr++; /* upper bound on number of kpatch relas and symbols */
            if (may_need_dynrela(rela))
                toc_rela(rela)->need_dynrela = 1;
        }
```
Thanks.

kamalesh-babulal:
**v4 Changes:**
* Incorporated review comments from Josh. 
   - Remove the check for `.rela.toc` section relas.
   - Add PowerPC specific comments for toc_rela() usage.

kpatch-build/create-diff-object.c
```diff
@@ -2302,6 +2302,79 @@ static int kpatch_is_core_module_symbol(char *name)
...
+{
+	if (!rela->sym->sec)
+		return 1;
+
+	/*
+	 * On PowerPC, use rela referenced sym->name in .rela.toc + off
+	 * if exist, to check for nested functions instead of rela->sym.
+	 * Referenced sym holds the function name and rela->name holds
+	 * reference section name.
+	 */
```
jpoimboe: I don't think this comment is really needed.  What I really wanted was a comment describing why it's doing the strchr() call with a check for the '.' character in the symbol name.

kamalesh-babulal: > I don't think this comment is really needed. What I really wanted was a comment describing why it's doing the strchr() call with a check for the '.' character in the symbol name.

Sorry if I have got it wrong, last paragraph from Evgenii comments above function_ptr_rela() explains why symbol name strchr() is checked for '.', Should that paragraph be moved here?

jpoimboe: Ah.  Yes, I believe that last paragraph belongs here.

kamalesh-babulal:
**v5 Changes:**
* Replaced PowerPC specific comment related to nested function, with the original comments on why the check is required, as suggested by Josh.

joe-lawrence:
:+1: 

@kamalesh-babulal , I went reading back through @euspectre 's PR #755 , and remembered that I had posted this integration patch and test:

[function-pointer.patch.txt](https://github.com/dynup/kpatch/files/1841950/function-pointer.patch.txt)
[function-pointer.test.txt](https://github.com/dynup/kpatch/files/1841956/function-pointer.test.txt)

This should be applicable to PowerPC implementation as well, right?  If so, I can add this test in a follow-up PR.

jpoimboe:
I'm going to be out next week.  Here's my (rough, untested) WIP for refactoring the dynrela code in case anybody's interested: https://github.com/jpoimboe/kpatch/tree/dynrela .  I'll try to pick it up again in April.

kamalesh-babulal:
> This should be applicable to PowerPC implementation as well, right? If so, I can add this test in a follow-up PR.

@joe-lawrence Yes, It will test the functionality on PowerPC too. Please add this test case to the integration tests. I have been testing this PR with the kernel commit be82485fbcb mentioned in the first comment of PR #755. Can you please merge this PR, so I can rebase and push the patch for PR #793 and other refactoring code too.

> I'm going to be out next week. Here's my (rough, untested) WIP for refactoring the dynrela code in case anybody's interested: https://github.com/jpoimboe/kpatch/tree/dynrela . I'll try to pick it up again in April.

@jpoimboe I will try and review/test the patch next week.

joe-lawrence:
@kamalesh-babulal - okay, sounds good

PR #790, opened 2018-02-27
If kpatch fails to disable a patch, retry a few times

euspectre:
This is similar to how loading of the patches works now. Needed mostly
for the "old" kpatch, i.e. for the kernels that do not support livepatch.

If the patched functions are currently used, loading of the patch fails
with "Device or resource busy" error. kpatch script then retries the
operation several times.

In some cases, it could be convenient to do the same thing when
unloading or simply disabling the patches. One of the use cases is when
it is needed to replace a loaded cumulative patch with its previous
version, esp. if the patches have patch/unpatch hooks. It is often more
reliable to disable the loaded patches first and then load the new
patch. Disable operation may fail due to activeness safety check - so
let us retry it a few times.

### DIFF
--- a/kpatch/kpatch
+++ b/kpatch/kpatch
@@ -348,19 +348,44 @@ load_module () {
 	return 0
 }
 
+disable_patch () {
+	local modname="$1"
+
+	local enabled="$SYSFS/$modname/enabled"
+	[[ -e "$enabled" ]] || die "patch module $1 is not loaded"
+
+	if [[ "$(cat "$enabled")" -eq 1 ]]; then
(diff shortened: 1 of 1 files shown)


jpoimboe:
Looks good.  One minor request: instead of the function returning an error, just `die` at the point of error, for consistency with the rest of the script's error handling.  Thanks!

euspectre:
OK, will do it in v2. Thanks for a prompt review!

joe-lawrence:
Good idea, thanks @euspectre

PR #794, opened 2018-03-13
Fix resulting module vermagic when source dir is a git tree

sm00th:
Sometimes git doesn't see that the patches have been reverted, if that
happens during ./scripts/setlocalversion call the resulting patch module
is built with a wrong vermagic because the tree is still considered
dirty.

Fix by moving git update-index call into remove_patches function so that
it is called every time the patches are reverted, not only on cleanup.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -104,15 +104,16 @@ remove_patches() {
 		patch="${PATCH_LIST[$idx]}"
 		patch -p1 -R -d "$SRCDIR" < "$patch" &> /dev/null
 	done
+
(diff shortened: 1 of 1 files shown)


joe-lawrence:
I think this change makes sense, did you have a particular kpatch-build invocation that provoked this?  (like `--skip-cleanup`)

sm00th:
> I think this change makes sense, did you have a particular kpatch-build invocation that provoked this?

I've hit it with both no additional options (apart from `--sourcedir`) and `--debug`, it wasn't 100% reproducible though.

joe-lawrence:
:+1:

PR #801, opened 2018-03-19
Undefined reference failure logic rework

### LINKED ISSUE #783: kpatch-build fails when patched module calls new vmlinux function
7770c18f93cc breaks patches which add calls from patched modules to new functions in a patched vmlinux.  That change enabled undefined symbol checking of the build log by kpatch-build, previously these checks were silently failing.  The `module-call-external.patch` integration test now fails to build as it adds a new symbol to the kernel and calls it from the nfsd module.

There's a little bit of git history behind (allowing) undefined symbols:

- 2014-08-29 - 4dee89269c83 ("add support for shadow variables")
-- Adds "undefined reference" kbuild.log check
-- Adds `LDFLAGS_vmlinux="--warn-unresolved-symbols"` and `LDFLAGS_MODULE="--warn-unresolved-symbols"`

- 2014-10-01 - ea445c03338a ("allow shadow functions to be called from modules")
-- Replaces `LDFLAGS_MODULE` with `KBUILD_MODPOST_WARN=1`
-- Adds "undefined!" kbuild.log check

- 2015-10-03 - f5de932b8dcf ("allow patched modules to call external functions")
-- Modifies the kmod core module to look for "external" symbols, which may be exported (as previously checked before this change), or provided by another .o within the patch module
-- Adds `module-call-external.patch` to the integration tests

- 2016-12-05 - 1330b4a3e5e9 ("kpatch-build: fix 'undefined reference to kpatch_shadow_*' errors")
-- Moves `--warn-unresolved-symbols` linker option to kpatch-gcc

- 2017-11-09 - 7770c18f93cc ("kpatch-build: Fix the logfile checks for 'undefined'")
-- Makes the undefined symbol checks actually work

So I think the intention was to tolerate undefined symbols to support the shadow variable API as provided by the kpatch core module, as well as calling new vmlinux functions from patched modules.

Shadow variable API references from patched modules and patched vmlinux tested as OK.  However, I think we need to be a little more clever about how we handle cases like `module-call-external.patch`.  Perhaps we should collect the list of reported undefined symbols and compare to symbols provided by the patch before considering it a build failure?

edit: added info about f5de932 ("allow patched modules to call external functions")

sm00th:
Don't die outright when encountering an 'undefined reference' error,
instead write those down and check if needed symbols are provided by the
resulting module or core kpatch module.

Fixes #783.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -713,8 +713,13 @@ export KPATCH_GCC_TEMPDIR
 CROSS_COMPILE="$TOOLSDIR/kpatch-gcc " \
 	KBUILD_MODPOST_WARN=1 \
 	make "-j$CPUS" $TARGETS 2>&1 | logger || die
-grep "undefined reference" "$LOGFILE" | grep -qv kpatch_shadow && die
-grep "undefined!" "$LOGFILE" | grep -qv kpatch_shadow && die
+
+# source.c:(.section+0xFF): undefined reference to `symbol'
+grep "undefined reference" "$LOGFILE" | sed -r "s/^.*\`(.*)'$/\1/" \
+	>${TEMPDIR}/undefined_references
+
+# WARNING: "symbol" [path/to/module.ko] undefined!
+grep "undefined!" "$LOGFILE" | cut -d\" -f2 >>${TEMPDIR}/undefined_references
 
 if [[ ! -e "$TEMPDIR/changed_objs" ]]; then
 	die "no changed objects found"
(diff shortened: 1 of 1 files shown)


jpoimboe:
I added a call to an undefined foo() function, and got:
```
ERROR: Undefined symbols: foo foo foo . Check /home/jpoimboe/.kpatch/build.log for more details.
```
The three "foo"s was a bit curious.  Looking at the build log, I see
```
$ grep undefined build.log 
/home/jpoimboe/.kpatch/src/fs/proc/meminfo.c:57: warning: undefined reference to `foo'
/home/jpoimboe/.kpatch/src/fs/proc/meminfo.c:57: warning: undefined reference to `foo'
/home/jpoimboe/.kpatch/src/fs/proc/meminfo.c:57: warning: undefined reference to `foo'
WARNING: "foo" [/home/jpoimboe/.kpatch/tmp/patch/kpatch-foo.ko] undefined!
```
So there were three warnings in the vmlinux link and one warning in the module link.

Why did the error message show three "foo"s?  In this case I would have expected either one or four, but not three.

Also I wonder... do we need to check *both* vmlinux and module?  Would it be sufficient to only check for the module link warning?  I can't think of any scenarios where the vmlinux link would show an undefined reference but the module link wouldn't.

jpoimboe:
> I can't think of any scenarios where the vmlinux link would show an undefined reference but the module link wouldn't.

Well, maybe I can, but I that's irrelevant because the module link is the one that matters, since we'll be loading it.

sm00th:
> Why did the error message show three "foo"s? In this case I would have expected either one or four, but not three.
The last message happens during livepatch build, but undefined reference grep happens before that - right after patched kernel build. I didn't move the check, so it always happened there.

I guess this answers your second question as well - we don't check both, but we can switch to checking just the module if there are benefits in doing so.

I'll also definitely have to make the lists unique before comparison.

kpatch-build/kpatch-build
```diff
@@ -854,6 +859,27 @@ if ! "$KPATCH_MODULE"; then
 	check_pipe_status create-klp-module
 fi
 
+readelf --wide --symbols $KMOD_PATH "$TEMPDIR/patch/$MODNAME.ko" 2>/dev/null | \
```
joe-lawrence: Do we need $KMOD_PATH if we have its static API list down below?

sm00th: Ah, right, this is left over.

joe-lawrence:
Looking good, but did we reach a consensus on checking only the module link step?  I think that check is more interesting to kpatch-build than the vmlinux link complaints.

Also by inspection I'm pretty sure this will work out for a livepatch build, but did you test that scenario?

sm00th:
> Looking good, but did we reach a consensus on checking only  the module link step?  I think that check is more interesting to  kpatch-build than the vmlinux link complaints.

Hm, do I understand correctly that you mean the patch-module link step? I tried moving "undefined" greps to only check logs from "Building patch module" step and it fails to detect an issue with static function patch I have: [module-call-external-static.patch](https://github.com/dynup/kpatch/files/1837048/module-call-external-static.patch.txt)

There is only one "undefined" warning during the whole build process and it is when we are linking patched nfsd.ko

> Also by inspection I'm pretty sure this will work out for a livepatch build, but did you test that scenario?

I mostly test with livepatch, I only tested kpatch when we read symbols from kpatch.ko

joe-lawrence:
> There is only one "undefined" warning during the whole build process and it is when we are linking patched nfsd.ko

Ah ok then.

Thanks for working on this, I think it came out nicely. :+1:

PR #803, opened 2018-03-21
kpatch-build: include .rela.rodata section

### LINKED ISSUE #702: changed section .rela.rodata not selected for inclusion
[patch.txt](https://github.com/dynup/kpatch/files/977481/patch.txt)

I'm testing a patch (see attached) for CVE-2014-3690 against RHEL7.3 GA + kernel-3.10.0-200.el7.x86_64 (rebuilt by RHEL7.3 GA toolset) and was running into the following kpatch-build error:
```
Using cache at /root/.kpatch/src
Testing patch file
checking file arch/x86/kvm/vmx.c
Reading special section data
Building original kernel
Building patched kernel
ERROR: vmx.o: 1 unsupported section change(s)
vmx.o: new function: vmx_set_constant_host_state
vmx.o: changed function: prepare_vmcs02.constprop.94
vmx.o: changed function: vmx_create_vcpu
vmx.o: changed function: vmx_vcpu_run
vmx.o: changed section .rela.rodata not selected for inclusion
/usr/local/libexec/kpatch/create-diff-object: unreconcilable difference
```
readelf does indeed show a difference in this section between the original and
patched object files:
```
% readelf -x 395 $ORIG

Hex dump of section '.rela.rodata':
  0x00000000 00000000 00000000 01000000 c8010000 ................
  0x00000010 b9010000 00000000                   ........


% readelf -x 395 $PATCHED

Hex dump of section '.rela.rodata':
  0x00000000 00000000 00000000 01000000 c8010000 ................
  0x00000010 e9010000 00000000                   ........
             ^
```

gogoxiaoxiao:
Fixes #702.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1346,6 +1346,14 @@ static void kpatch_include_standard_elements(struct kpatch_elf *kelf)
 			list_for_each_entry(rela, &sec->relas, list)
 				kpatch_include_symbol(rela->sym, 0);
 		}
+
+		/* include .rela.rodata section for vmx_vcpu_run */
+		if (!strcmp(sec->name, ".rela.rodata"))
+		{
+			sec->include = 1;
+			list_for_each_entry(rela, &sec->relas, list)
+				kpatch_include_symbol(rela->sym, 0);
+		}
 	}
 
 	/* include the NULL symbol */
@@ -1531,7 +1539,7 @@ static void kpatch_migrate_included_elements(struct kpatch_elf *kelf, struct kpa
 		if (sym->sec && !sym->sec->include)
 			/* break link to non-included section */
 			sym->sec = NULL;
-			
+
 	}
 
 	*kelfout = out;


jpoimboe:
This patch is a bit mysterious, as it doesn't have any explanation.  I went ahead and recreated the issue so I could understand the problem.

The problem is this nasty inline asm in vmx_cpu_run():
```
		".pushsection .rodata \n\t"
		".global vmx_return \n\t"
		"vmx_return: " _ASM_PTR " 2b \n\t"
		".popsection"
```
That stores a pointer to an internal vmx_cpu_run() address in .rodata (resulting in the relocation in .rela.rodata).  When vmx_cpu_run() is modified, it shifts the code addresses, causing that pointer (and the resulting relocation addend) to change.

The fix is indeed to include .rela.rodata.  However, I think including all the other symbols in .rela.rodata isn't quite right, because it means vmx_cpu_run() will always get pulled into the patch, even if it didn't change.  In fact, I think we have a similar issue with .rela.toc for ppc64le (opened #807 for that).

jpoimboe:
> The fix is indeed to include .rela.rodata. However, I think including all the other symbols in .rela.rodata isn't quite right, because it means vmx_cpu_run() will always get pulled into the patch, even if it didn't change. In fact, I think we have a similar issue with .rela.toc for ppc64le (opened #807 for that).

To clarify, if vmx_cpu_run() doesn't change, rela.rodata will still get pulled in, meaning that the vmx_cpu_run rela will get converted to an (unnecessary) dynrela.  It's a minor issue, but it should be fine.

The inclusion code is quite confusing, so I'm going to fix this issue and #807 at the same time with a cleanup.

jpoimboe:
This is being addressed with #808.

gogoxiaoxiao:
Thanks for reviewing, I have tested the patch and it works ok 👍

PR #805, opened 2018-03-21
gitignore: ignore cache.mk from newer kernel builds

jpoimboe:


joe-lawrence:
:+1:

PR #806, opened 2018-03-21
drop unused rpmdevtools dependency

cardoe:
It appears that since 0dec5136eef945aa086a0290b3a8de65393fdf5f nothing
has been used from the rpmdevtools package by kpatch-build.

Signed-off-by: Doug Goldstein <cardoe@cardoe.com>


joe-lawrence:
yup as far as I can tell there's nothing left that calls any of the programs provided by `rpmdevtools`  :+1:

jpoimboe:
Thanks @cardoe !

PR #811, opened 2018-03-26
Cross compilation fixes

rbilovol:
These two patches allow cross-compiled LivePatch generation.
They fix two cross-compile scenarios:
 - building LivePatch for another architecture
 - building LivePatch for another distro

This has been tested on OpenEmbedded build system.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -387,12 +387,13 @@ usage() {
 	echo "		                   (can be specified multiple times)" >&2
 	echo "		--cross-compile    Specify the prefix used for all executables" >&2
 	echo "		                   used during compilation" >&2
+	echo "		--distro           Override distro name" >&2
 	echo "		--skip-cleanup     Skip post-build cleanup" >&2
 	echo "		--skip-gcc-check   Skip gcc version matching check" >&2
 	echo "		                   (not recommended)" >&2
 }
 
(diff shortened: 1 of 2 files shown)


sm00th:
What is your usecase for the `--distro` option? $DISTRO in kpatch-build is mostly used for autodetection of source/vmlinux locations which is not useful when you are building for a different target. The only check that should have any effect would be the `copy_user_64.o` workaround for rhel7.

rbilovol:
Currently I use --distro option just to disable existing (and future) distro-specific hacks/tunings when build LivePatches for another distro. In the future, though, we may have OpenEmbedded/Yocto-specific tunings which we want to specify.

joe-lawrence:
Hi @rbilovol , are you currently cross compiling kpatches and if so, from which arch to which arch?  I see where this is PR is coming from by adding the prefix to invoke the other compiler, but I'm not sure if this is enough:  What about the `ARCH` variable in `kpatch-build`?  How would the gcc plugins (see `kpatch-build/gcc-plugins` directory) interact with cross-compiling?  Does this cross-build `kmod/patch/kpatch.ko` etc.

BTW, this would be a really nice feature to have as I have an s390 cross compiler installed on my machine and it would be handy to test a few builds once we have s390 kpatch support.

jpoimboe:
Hi @rbilovol.  In addition to @joe-lawrence 's questions...

For `--cross-compile`, I don't think a new cmdline argument is necessary.  Instead I would suggest that the user use an environment variable.  In fact, we should probably just use the same variable which is used to cross-compile the kernel: `CROSS_COMPILE`.

I also don't see the point of `--distro`, since, as @sm00th mentioned, those mostly just affect the defaults for finding files.  Can you be more specific about why it's needed?

PR #816, opened 2018-04-04
kmod/core: account for failing modules in notifier

### LINKED ISSUE #699: removing patched module left kpatch patch module stuck
[patch.txt](https://github.com/dynup/kpatch/files/971897/patch.txt)

With the attached kpatch, I didn't have hardware suitable to successfully load the `aesni_intel` driver.  A modprobe of that driver did engage kpatch though, and that left the kpatch patch module loaded, but in an unloadable state (as well as dropping a WARN):
```
% kpatch load kpatch-klp.ko 

[21388.055840] kpatch: delaying patch of unloaded module 'aesni_intel'
[21388.073346] kpatch: loaded patch module 'kpatch_klp'

% modprobe aesni_intel
modprobe: ERROR: could not insert 'aesni_intel': No such device

[21394.026565] kpatch: patching newly loaded module 'aesni_intel'

% lsmod | grep aesni_intel        # no matches

% kpatch unload kpatch-klp
disabling patch module: kpatch_klp
/usr/local/sbin/kpatch: line 210: echo: write error: Invalid argument
kpatch: can't disable kpatch_klp

[21411.922922] kpatch: can't remove ftrace filter at address 0xffffffffa0671ab0
[21411.929970] ------------[ cut here ]------------
[21411.934615] WARNING: at /root/kpatch/kmod/core/core.c:721 kpatch_unlink_object+0x70/0xa0 [kpatch]()
[21411.943653] can't unregister ftrace for address 0xffffffffa0671ab0
[21412.021367]  bnx2 mdio dm_mirror dm_region_hash dm_log dm_mod [last unloaded: kpatch_klp]
[21412.028310] CPU: 2 PID: 2286 Comm: kpatch Tainted: GF    U    IO--------------   3.10.0-200.el7.x86_64 #1
[21412.037868] Hardware name: Dell Inc. PowerEdge R410/0N051F, BIOS 1.11.0 07/20/2012
[21412.045430]  ffff8802f3633da8 00000000cc6b74f7 ffff8802f3633d60 ffffffff81603cd3
[21412.052863]  ffff8802f3633d98 ffffffff8106e231 ffff8802f34a6480 ffff8802f34a6618
[21412.060295]  ffff8802f34a6600 00000000ffffffea 0000000000000008 ffff8802f3633e00
[21412.067728] Call Trace:
[21412.070178]  [<ffffffff81603cd3>] dump_stack+0x19/0x1b
[21412.075317]  [<ffffffff8106e231>] warn_slowpath_common+0x61/0x80
[21412.081321]  [<ffffffff8106e2ac>] warn_slowpath_fmt+0x5c/0x80
[21412.087064]  [<ffffffffa0652f40>] kpatch_unlink_object+0x70/0xa0 [kpatch]
[21412.093848]  [<ffffffffa06537b8>] kpatch_unregister+0x1a8/0x270 [kpatch]
[21412.100544]  [<ffffffffa03970b4>] patch_enabled_store+0x54/0xb0 [kpatch_klp]
[21412.107591]  [<ffffffff812d5fdf>] kobj_attr_store+0xf/0x20
[21412.113074]  [<ffffffff8123c236>] sysfs_write_file+0xc6/0x140
[21412.118820]  [<ffffffff811c61dd>] vfs_write+0xbd/0x1e0
[21412.123954]  [<ffffffff811c6c28>] SyS_write+0x58/0xb0
[21412.129005]  [<ffffffff81613ce9>] system_call_fastpath+0x16/0x1b
[21412.135009] ---[ end trace cc33c1e1d7715c72 ]---

% lsmod | grep kpatch
kpatch_klp             14046  1 
kpatch                 52025  1 kpatch_klp
```

joe-lawrence:
The module notifier currently only handles newly loaded modules in the
MODULE_STATE_COMING state.  If target modules need to be unloaded, the
any kpatch module that patches it must first be disabled, releasing
module references held against the target module.  When the kpatch
modules are disabled, the target module is unpatched and the kpatch
core's data structures updated accordingly.

If a loading module happens to fail its init routine (missing hardware
for example), that module will not complete loading.  The kpatch core
doesn't properly account for this "phantom" target module, so when the
kpatch patch module is removed, it spews out an ugly warning when
attempting to remove a non-existing ftrace filter on the target module.

In the problematic case above, the module handler is notified of the
failing module in its MODULE_STATE_GOING state.  Add code to the handler
to clear any ftrace filters, call pre and post unpatch callbacks, etc.

Fixes #699.

### DIFF
--- a/kmod/core/core.c
+++ b/kmod/core/core.c
@@ -330,7 +330,7 @@ static inline void post_patch_callback(struct kpatch_object *object)
 		(*object->post_patch_callback)(object);
 }
 
-static inline void pre_unpatch_callbacks(struct kpatch_object *object)
+static inline void pre_unpatch_callback(struct kpatch_object *object)
 {
 	if (kpatch_object_linked(object) &&
 	    object->pre_unpatch_callback &&
@@ -424,7 +424,7 @@ static int kpatch_remove_patch(void *data)
 
 	/* run any user-defined pre-unpatch callbacks */
 	list_for_each_entry(object, &kpmod->objects, list)
-		pre_unpatch_callbacks(object);
+		pre_unpatch_callback(object);
 
 	/* Check if any inconsistent NMI has happened while updating */
 	ret = kpatch_state_finish(KPATCH_STATE_SUCCESS);
@@ -783,8 +783,10 @@ static int kpatch_unlink_object(struct kpatch_object *object)
 		}
 	}
 
-	if (object->mod)
+	if (object->mod) {
 		module_put(object->mod);
+		object->mod = NULL;
+	}
 
 	return 0;
 }
@@ -861,8 +863,8 @@ err_put:
 	return ret;
 }
 
-static int kpatch_module_notify(struct notifier_block *nb, unsigned long action,
-				void *data)
+static int kpatch_module_notify_coming(struct notifier_block *nb,
+				       unsigned long action, void *data)
 {
 	struct module *mod = data;
 	struct kpatch_module *kpmod;
@@ -917,6 +919,53 @@ out:
 	/* no way to stop the module load on error */
 	WARN(ret, "error (%d) patching newly loaded module '%s'\n", ret,
 	     object->name);
+
+	return 0;
+}
+
+static int kpatch_module_notify_going(struct notifier_block *nb,
+				      unsigned long action, void *data)
+{
+	struct module *mod = data;
+	struct kpatch_module *kpmod;
+	struct kpatch_object *object;
+	struct kpatch_func *func;
+	bool found = false;
+
+	if (action != MODULE_STATE_GOING)
+		return 0;
+
+	down(&kpatch_mutex);
+
+	list_for_each_entry(kpmod, &kpmod_list, list) {
+		list_for_each_entry(object, &kpmod->objects, list) {
+			if (!kpatch_object_linked(object))
+				continue;
+			if (!strcmp(object->name, mod->name)) {
+				found = true;
+				goto done;
+			}
+		}
+	}
+done:
+	if (!found)
+		goto out;
+
+	/* run user-defined pre-unpatch callback */
+	pre_unpatch_callback(object);
+
+	/* remove from the global func hash */
+	list_for_each_entry(func, &object->funcs, list)
+		hash_del_rcu(&func->node);
+
+	/* run user-defined pre-unpatch callback */
+	post_unpatch_callback(object);
+
+	kpatch_unlink_object(object);
+
+out:
+	up(&kpatch_mutex);
+
 	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


joe-lawrence:
[test3.patch.txt](https://github.com/dynup/kpatch/files/1875910/test3.patch.txt)
[module-probe-failed-testing.txt](https://github.com/dynup/kpatch/files/1875909/module-probe-failed-testing.txt)

Testing includes a case in which the aesni_intel driver fails to load, but a loaded kpatch module has pre/post (un)patch callbacks defined against it.  The kpatch core doesn't properly account for the target module failing its init, and when unloading the kpatch module, it can crash the machine.  See the  attached module-probe-failed-testing.txt for before/after details.

kmod/core/core.c
```diff
@@ -890,33 +892,63 @@ static int kpatch_module_notify(struct notifier_block *nb, unsigned long action,
...
+		/* run user-defined pre-unpatch callback */
+		post_unpatch_callback(object);
 
-	/* run user-defined post-patch callback */
-	post_patch_callback(object);
+		/*
+		 * can't unlink an object that has been removed from the
+		 * global hash, so just clear its mod pointer.
+		 */
+		object->mod = NULL;
```
joe-lawrence: To be perfectly symmetric, I wanted to call `kpatch_link_object()` here, but that fails with "can't remove ftrace filter at address".  Am I missing something?  Since the module isn't going to be loaded anyway, clearing the `mod` pointer was good enough to trip up any future `kpatch_object_linked()` checks.

joe-lawrence: Hmm, if I move this teardown code up into the module coming case (right after it runs the post-patch callback), it all runs w/o incident.  I suspect you're right that something external to kpatch is getting in the way here.  I'm looking through the `load_module()` and `do_init_module()` code now, though nothing is jumping out yet.

joe-lawrence: Bingo, here it is: `ftrace_module_notify_exit() -> ftrace_release_mod(mod)`   (And renamed in newer RHEL7 kernels as `ftrace_module_notify()`.  Looking that the `module_notify_list` in crash, I can see that it's before the kpatch module notifier.

jpoimboe: Ah, then we may end up having to register *two* notifiers: the coming notifier with priority INT_MIN so it's called last, and the going notifier with priority INT_MAX, so it's called first.

joe-lawrence:
v2:
- clear object->mod in kpatch_unlink_object() so that future kpatch_object_linked() checks will fail
- split into separate coming and going notifiers so we can register last and first (respectively) on the notifier list,  h/t Josh for the suggestion

kmod/core/core.c
```diff
@@ -1189,7 +1242,10 @@ static int kpatch_init(void)
 	if (!kpatch_root_kobj)
 		return -ENOMEM;
 
-	ret = register_module_notifier(&kpatch_module_nb);
+	ret = register_module_notifier(&kpatch_module_nb_coming);
+	if (ret)
+		goto err_root_kobj;
+	ret = register_module_notifier(&kpatch_module_nb_going);
 	if (ret)
 		goto err_root_kobj;
```
jpoimboe: I think we need to unregister the coming notifier here on error.

joe-lawrence:
v3:
- If the `kpatch_module_nb_going` handler fails to register, be sure to unregister the `kpatch_module_nb_coming` handler in the `kpatch_init()` error path.

PR #817, opened 2018-04-04
create-diff-object: fix ppc64le static local variable correlation

jpoimboe:
On ppc64le, the static local variable correlation doesn't take into
account the .toc rela indirection for data references, meaning that it's
basically broken in many cases.

Fix it by making the code .toc-aware.

Fixes #793.

Reported-by: Kamalesh Babulal <kamalesh@linux.vnet.ibm.com>


jpoimboe:
@kamalesh-babulal What do you think about this approach to fix #793?

Issue #798, opened 2018-03-16
Create unit tests


jpoimboe:
Our current integration tests are a nightmare to maintain:

- They require writing test cases (patches)
- They require constant porting of those test cases to other distros and kernel versions
- They take forever to run
- Each group of tests only runs on a specific distro
- Due to ever-changing kernel sources, each patch may no longer test what it was originally designed to test
- They're not compatible with Travis or other CIs
- I could go on...

Because the integration tests are so unwieldy, we don't have a reasonable way for contributors or maintainers to run regression tests, which is negatively affecting our code quality.  That needs to change ASAP.

A key insight is that 95+% of our bugs are in create-diff-object.  So I propose the following:

- Create a new unit test framework in `test/unit` which provides pre-built .o files to create-diff-object.
- Most tests will probably just be "does it cause create-diff-object to spit out an error or seg fault".  So it would be nice if we could add a new basic test to the suite just by adding a pair of appropriately named .o files to the directory.  We could use a naming convention like foo.ORIG.o and foo.PATCHED.o.  In this case, the test framework can automatically detect the .o files and pass them to create-diff-object, ignoring the output but checking for errors.
- I guess we will also need to have "dummy" vmlinux and other needed .ko files in the directory, which create-diff-object takes as its `$KOBJFILE` input.  That may be a pain point, maybe there's a way we can restructure create-diff-object a bit, such that its inputs are a little more test-friendly.  This needs more thought.
- For some tests, we might want to run some script on the output .o file to test that it looks ok.  We could use a naming convention for the script name, like foo.test, which is a script file test which runs on the foo.OUTPUT.o file after create-diff-object runs.
- For some tests, a create-diff-object error might be the success condition.  I presume we could come up with another naming convention for that, like foo.ORIG.FAIL.o and foo.PATCHED.FAIL.o
- Note these are all just examples, I don't really care what the naming convention is.  But having a convention makes maintenance of the tests easier.
- We will need arch-specific .o files.  Currently we don't support cross-compilation, so x86 tests have to run on an x86 host and ppc64le tests have to run on a ppc64le host.  We can just put them in their own x86 and ppc64le directories for now.  Eventually we probably want to make create-diff-object smart enough to read arches other than the host.
- Every time somebody fixes a bug, they should also add .o file(s) to the unit tests which would have caught the bug.
- Try to figure out the purpose of most of our integration tests, and convert them to unit tests.
[...]

PR #818, opened 2018-04-05
Unittests

sm00th:
Here's a simple create-diff-obj unit test framework as discussed in #798.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,10 @@ INSTALL_DIRS = $(SUBDIRS:%=install-%)
 UNINSTALL_DIRS = $(SUBDIRS:%=uninstall-%)
 CLEAN_DIRS   = $(SUBDIRS:%=clean-%)
 
-.PHONY: all install uninstall clean check
+UNITTEST_DIR = test/unit
+CLEAN_DIRS  += clean-$(UNITTEST_DIR)
+
+.PHONY: all install uninstall clean check unit
 .PHONY: $(SUBDIRS) $(BUILD_DIRS) $(INSTALL_DIRS) $(CLEAN_DIRS)
 
 
@@ -25,6 +28,9 @@ $(UNINSTALL_DIRS):
 clean: $(CLEAN_DIRS)
 $(CLEAN_DIRS):
 	$(MAKE) -C $(@:clean-%=%) clean
+
+unit: $(UNITTEST_DIR)/Makefile build-kpatch-build
+	$(MAKE) -C $(UNITTEST_DIR)
 
 check:
 	shellcheck kpatch/kpatch kpatch-build/kpatch-build kpatch-build/kpatch-gcc
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
-*.o filter=lfs diff=lfs merge=lfs -text
+*.ORIG.o filter=lfs diff=lfs merge=lfs -text
+*.PATCHED.o filter=lfs diff=lfs merge=lfs -text
+*.PATCHED.FAIL.o filter=lfs diff=lfs merge=lfs -text
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,9 @@
 *.d
 *.so
 *.test.out
+!*.ORIG.o
+!*.PATCHED.o
+!*.PATCHED.FAIL.o
 .tmp_versions
 Module.symvers
 kmod/core/.cache.mk
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 *.swp
 *.d
 *.so
+*.test.out
 .tmp_versions
 Module.symvers
 kmod/core/.cache.mk
--- a/test.o
+++ b/test.o
@@ -1,3 +1,0 @@
-version https://git-lfs.github.com/spec/v1
-oid sha256:851a621d0ab0dd17184d2c6b8ae920cf0e28b676973bb961dea02bc392c79669
-size 1192
--- a/test/unit/Makefile
+++ b/test/unit/Makefile
@@ -1,0 +1,7 @@
+ARCH=$(shell uname -m)
+
+all: Makefile.include $(ARCH)/Makefile
+	$(MAKE) -C $(ARCH)
+
+clean: Makefile.include $(ARCH)/Makefile
+	$(MAKE) -C $(ARCH) clean
--- a/test/unit/Makefile.include
+++ b/test/unit/Makefile.include
@@ -1,0 +1,40 @@
+EXT_ORIG        ?= ORIG.o
+EXT_PATCHED     ?= PATCHED.o
+EXT_FAIL        ?= PATCHED.FAIL.o
+EXT_TEST        ?= test
+EXT_OUTPUT      ?= OUTPUT.o
+EXT_TEST_OUTPUT ?= test.out
+TNAME            = $(@:.$(EXT_OUTPUT)=)
+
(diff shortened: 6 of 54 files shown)


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "test/unit/kpatch-unit-objs"]
+	path = test/unit/kpatch-unit-objs
+	url = https://github.com/sm00th/kpatch-unit-objs.git
```
sm00th: This is just for reference, should probably be switched to ad dynup repo before merging.

jpoimboe:
Ok, here's yet another idea for object storage...

Github supports git LFS:

https://git-lfs.github.com/
https://help.github.com/articles/about-storage-and-bandwidth-usage/

Which looks really nice to me.  The only downsides are that you have to install git-lfs to get the "lfs" subcommand; and that there are bandwidth/storage limits of 1GB/mo (but I seriously doubt we would ever hit those).

What do you think?

sm00th:
git lfs sounds like a uniform lookaside solution which is much easier than custom scripts, although it does require some set up on contributor's side. There seems to be a catch. I tried using it and got shut down on push:
```
$ git push mine unittests-lfs
batch response: @sm00th can not upload new objects to public fork sm00th/kpatch
```
According to https://github.com/git-lfs/git-lfs/issues/1906 object pushes to public forks only works with 'Github Enterprise'. There is also an older issue that suggests that it should work once parent repo has some object files uploaded, but this might not be true anymore: https://github.com/git-lfs/git-lfs/issues/1449

jpoimboe:
I pushed an LFS test file to try to enable it.  Can you try again?

jpoimboe:
Yeah, go for it.

sm00th:
Not sure if github sends out notifications on push, but I did update this to lfs version.

jpoimboe:
Github doesn't send notifications for force pushes.

Looks like this needs a rebase onto master, since I changed .gitattributes.  You can just overwrite my changes, as your version looks better.

sm00th:
> You can just overwrite my changes, as your version looks better.

Did that. I also deleted 'test.o' as we won't need it.

joe-lawrence:
This is pretty cool!

One question: after running the testes, git status notices an untracked file: `test/unit/x86_64/gcc-static-local-var-4.test.out`   was that intentional?  A subsequent `make clean` takes care it of anyway.

Also, would it make sense, maybe in a future commit, to create subdirs for various kernel versions?  Right now it looks like the collection of object files come from the integration tests... in the future we may get bug reports against a specific kernel/config that we'd like to add.  Just wondering what your thoughts were on that.

:+1: :+1:

jpoimboe:
In theory the unit tests are supposed to be independent of the kernel version, so we should be able to mix/match .o files from different kernel versions in the same x86_64 dir.  (Though we discussed in #798 that this isn't completely possible yet if the special section environment variables change.)

BTW, a few minor improvements which can be done with this PR or another one:
- Fail if any .o files aren't properly stripped
- Fail if git-lfs isn't installed (or the .o files are missing)
- Top-level 'make unit' target
- Hook it up to Travis CI for every pull request

sm00th:
v3:
 - .test.out in .gitignore
 - strip/lfs checks
 - top-level unit target

Makefile
```diff
@@ -26,5 +29,8 @@ clean: $(CLEAN_DIRS)
 $(CLEAN_DIRS):
 	$(MAKE) -C $(@:clean-%=%) clean
 
+unit: $(UNITTEST_DIR)/Makefile
```
jpoimboe: I don't think it should be dependent on the unit Makefile, because a .o could be added without touching the makefile.  (and the same comment for a similar dependency in test/unit/Makefile).  I'm not sure if there's an easy way to track the .o changes/additions?  Maybe you would know, your Makefile-fu seems pretty strong :-)

I think it's probably fine for now to just remove any dependencies and run the unit tests every time.

jpoimboe: It also might make sense to add a dependency on create-diff-object here.

jpoimboe: (but only if we also figured out how to make a dependency on the .o files, I think)

sm00th: That dependency does not affect when it runs because `unit` is not an automatic target(and I would argue against making it one because of git-lfs dependency), so it will always run when invoked by `make unit`. $(UNITTEST_DIR)/Makefile is just a dependency that make doesn't know how to create so it will fail early if it is not there.

It is, however a good idea to make `unit` a PHONY target and add $(BUILD_DIRS) dependency so it will automatically build everything if it is not built yet.

It would also be a good idea to add create-diff-object dependency to test/unit/Makefile.include to the rules where it is used, that way it will retest everything on next `make unit` invocation if create-diff-object was rebuilt.

jpoimboe: That all sounds good.  But if I run 'make unit' twice, the second run doesn't do anything.  Is that what we want?  I presume that means that if we add a new .o file, it won't detect it and rerun the tests, right?

sm00th:
v4:
 - make unit a PHONY target
 - add $(BUILD_DIRS) as unit dependency
 - add create-diff-object as *.OUTPUT.o dependency so every test will rerun if create-diff-object is updated.

Makefile
jpoimboe: (And if so, maybe that's ok, we'll just have to remember to `make clean` before running the tests again, after adding a .o file)

sm00th: > That all sounds good. But if I run 'make unit' twice, the second run doesn't do anything. Is that what we want?

If you touch any of the .ORIG.o/.PATCHED.o it would retest the touched testcase and it would detect new ones as well. When you do `make unit` it unconditionally ends up in test/unit/$(uname -m)/Makefile that has all of the automated targets and the only dependency logic that matters is there.

jpoimboe:
A few more minor issues:
- I'm not sure it makes sense to make `$(BUILD_DIRS)` a dependency of `make unit` because then it rebuilds the kpatch.ko core module everytime, which is unnecessary.  Instead maybe just make it depend on create-diff-object?  (But then I tried adding such a dependency and apparently our crappy makefiles don't allow rebuilding create-diff-object from the root directory, maybe it's an easy fix or there's a better way.)
- Is there a way we can tweak the .gitignore rules, such that .o files are ignored, but .ORIG.o and .PATCHED.o files are *not* ignored?  Otherwise a new uncommited test won't show up in "git status".
- It would be nice if it were less verbose by default (but this can also be a future improvement)

sm00th:
> How silent should the output of "make -s unit" be?

Not very, it is a test "description" + create-diff-object output like so:
```
- Testing 'gcc-mangled-3'. Expected pass. 
gcc-mangled-3.ORIG.o: changed function: get_slabinfo                                 
```

Because of the number of cases it does add up to a wall of text (currently 119 lines). I don't mind getting rid of create-diff-object output or even all of it if you'd prefer that.

sm00th:
> make it depend on create-diff-object?

The best we can do at this point is depend on just the kpatch-build dir which is still a bit more than we need but is much less of a pain than the module (no KPATCH_BUILD needed)

> Is there a way we can tweak the .gitignore rules, such that .o files are ignored, but .ORIG.o and .PATCHED.o files are not ignored?

Yes, not sure when it got added but git 2.13 supports "!exclude" syntax in .gitignore.

> It would be nice if it were less verbose by default (but this can also be a future improvement)

I'll get rid of test text and will make create-diff-object output optional (off by default). When I started out with just 2 testcases the output was much nicer :)

PR #819, opened 2018-04-06
create-diff-object: Refactor rela_equal() using toc_rela()

koding5:
Heavy lifting of reading .toc rela entries for rela symbols
referring to .toc + offset, can be simplified using toc_rela() in
rela_equal() and remove the #ifdery guarding PowerPC code.

This patch also trims the commentary related to PowerPC.


koding5:
oops, wrong github username. Will redo the request.

PR #820, opened 2018-04-06
create-diff-object: Refactor rela_equal() using toc_rela()

kamalesh-babulal:
Heavy lifting of reading .toc rela entries for rela symbols
referring to .toc + offset, can be simplified using toc_rela() in
rela_equal() and remove the #ifdery guarding PowerPC code.

This patch also trims the commentary related to PowerPC.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -252,129 +252,80 @@ static int kpatch_mangled_strcmp(char *s1, char *s2)
 
 static int rela_equal(struct rela *rela1, struct rela *rela2)
 {
-#ifdef __powerpc__
-	struct section *toc_relasec1, *toc_relasec2;
-	struct rela *r_toc_relasec1, *r_toc_relasec2;
+	struct rela *rela_toc1, *rela_toc2;
 	unsigned long toc_data1, toc_data2;
-#endif
 
 	if (rela1->type != rela2->type ||
 	    rela1->offset != rela2->offset)
 		return 0;
 
-	if (rela1->string)
-		return rela2->string && !strcmp(rela1->string, rela2->string);
-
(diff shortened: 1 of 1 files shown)


kpatch-build/create-diff-object.c
```diff
@@ -252,129 +252,80 @@ static int kpatch_mangled_strcmp(char *s1, char *s2)
 
 static int rela_equal(struct rela *rela1, struct rela *rela2)
 {
-#ifdef __powerpc__
-	struct section *toc_relasec1, *toc_relasec2;
-	struct rela *r_toc_relasec1, *r_toc_relasec2;
+	struct rela *r_rela1, *r_rela2;
```
jpoimboe: Let's call them rela_toc1 and rela_toc1 for consistency with the rest of the code.

kpatch-build/create-diff-object.c
```diff
@@ -252,129 +252,80 @@ static int kpatch_mangled_strcmp(char *s1, char *s2)
...
 
-		if ((r_toc_relasec1->addend != r_toc_relasec2->addend))
-			return 0;
+	if (!r_rela1 && !r_rela2)
+		return toc_data1 == toc_data2;
 
-		if (is_special_static(r_toc_relasec1->sym))
-			return !kpatch_mangled_strcmp(r_toc_relasec1->sym->name,
-					r_toc_relasec2->sym->name);
+	 if (r_rela1->string)
```
jpoimboe: Extra whitespace here

kamalesh-babulal:
@jpoimboe Thanks for the review.

**V2 Changes:**
- Rename `r_rela` to `rela_toc` as suggested by Josh to maintain consistency with the code base.
- Fix the whitespace issue.

PR #821, opened 2018-04-06
Testcase: Clean up gcc-static-local-var-2

kamalesh-babulal:
The test case gcc-static-local-var-2 doesn't uses any macros
from kpatch-macros.h, so remove the hunk including it.

Cc: Joe Lawrence <jdl1291@gmail.com>


joe-lawrence:
:+1:

PR #822, opened 2018-04-06
patch-author-guide: sleeping locks in callbacks

sm00th:
Add a note about dealing with sleeping locks in callbacks.

### DIFF
--- a/doc/patch-author-guide.md
+++ b/doc/patch-author-guide.md
@@ -193,7 +193,10 @@ static void kpatch_post_unpatch_tcp_send_challenge_ack(patch_object *obj)
 +KPATCH_POST_UNPATCH_CALLBACK(kpatch_post_unpatch_tcp_send_challenge_ack);
 ```
 
-Don't forget to protect access to the data as needed.
+Don't forget to protect access to the data as needed. Please note that mutexes
+and other sleeping locks can't be used from stop_machine context, so you will
+have to check if any locks protecting the data are already held and if so
+return an error from your callback function.
 
 Also be careful when upgrading.  If patch A has a pre/post-patch callback which
 writes to X, and then you load patch B which is a superset of A, in some cases


joe-lawrence:
:+1: 

Thanks for adding.  BTW, do you think the recently used technique of adding a `nop()` to all the functions that take/release a mutex should get its own mention somewhere?

doc/patch-author-guide.md
```diff
@@ -193,7 +193,10 @@ static void kpatch_post_unpatch_tcp_send_challenge_ack(patch_object *obj)
 +KPATCH_POST_UNPATCH_CALLBACK(kpatch_post_unpatch_tcp_send_challenge_ack);
 ```
 
-Don't forget to protect access to the data as needed.
+Don't forget to protect access to the data as needed. Please note that mutexes
+and other sleeping locks can't be used from stop_machine context, so you will
+have to check if any of locks protecting the data are already held and if so
```
jpoimboe: "any of locks" -> "any locks"

joe-lawrence:
>> do you think the recently used technique of adding a nop() to all the functions that take/release a mutex should get its own mention somewhere?

> Yes, however it is only useful for older kpatch versions where hooks had no return value, so not sure if it will be of any help in current documentation.

Ah right.

I need to think more about the following scenario:  a kpatch modifies code in a module that is loaded later (meaning we'll trigger our module notifier and run from there, not `stop_machine()`).  See **Test Y** in [tests.txt](https://github.com/dynup/kpatch/files/1785500/tests.txt) from PR #780.

Differences from a vmlinux or already loaded callback:
* stop_machine() vs. modprobe context
  * we can probably just spin or sleep from the callback
* kpatch_verify_activeness_safety() is only called from stop_machine()
  * can't use the nop-mutex-holders trick for this case
* pre-patch callback failure does *not* prevent loading module from staying loaded
  * we can protect against running our patched code, but not the target code

Similar weirdness exists on the removal side.  Removing the target module first will invoke the module notifier, but unloading the patch module first would cause a `stop_machine()`.

Maybe it's just Friday and my brain is worrying too much about what hurdles an implementation of this advice might need to clear. 

:+1:  for the current text as it's accurate despite all of the above.

jpoimboe:
Ugh.  Good find.

This is yet another complication from patching modules in the module notifier.  I've floated the idea of removing this "feature" in livepatch upstream, but nobody else seems interested.  Some options:

- move the callbacks out of stop_machine (but this would make them less useful)
- call the callbacks from stop_machine, even when they're called from the module notifier (not sure this really makes sense either)
- force all to-be-patched modules to be loaded before loading the patch module
- document this limitation in the (kpatch.ko) callbacks documentation

I don't see any good options other than documenting this quirk.

joe-lawrence:
I can't seem to distill this down into something more approachable.  I think my brain would rather move onto the livepatch consistency model where there are fewer corner cases :disappointed:

-->8-- -->8-- -->8-- -->8-- -->8-- -->8-- -->8-- -->8-- -->8-- -->8--

Don't forget to protect access to the data as needed. Please note that
spinlocks and mutexes / sleeping locks can't be used from stop_machine
context. Also note the pre-patch callback return code will be ignored by the
kernel's module notifier, so it does not effect the target module or livepatch
module status. This means:

* Pre-patch callbacks to loaded objects (vmlinux, loaded kernel modules) are
  run from stop_machine(), so they may only inspect lock state (i.e.
  spin_is_locked(), mutex_is_locked()) and optionally return -EBUSY to prevent
  patching.

* Post-patch, pre-unpatch, and post-unpatch callbacks to loaded objects are
  also run from stop_machine(), so the same locking context applies.  No
  return status is supported.

* Deferred pre-patch callbacks to newly loading objects do not run from
  stop_machine(), so they may spin or schedule, i.e. spin_lock(),
  mutex_lock()).  Return status is ignored.

* Post-patch, pre-unpatch, and post-unpatch callbacks to unloading objects are
  also *not* run from stop_machine(), so they may spin or sleep.  No return
  status is supported.

Unfortunately there is no simple, all-case-inclusive kpatch callback
implementation that handles data structures and mutual exclusion.

A few workarounds:

1. If a given lock/mutex is held and released by the same set of functions
(that is, functions that take a lock/mutex always release it before
returning), a trivial change to those functions can re-purpose kpatch's
activeness safety check to avoid patching when the lock/mutex may be held.
This assumes that all lock/mutex holders can be patched.

2. If it can be assured that all patch targets will be loaded before the
kpatch patch module, pre-patch callbacks may return -EBUSY if the lock/mutex
is held to block the patching.

3. If all patch targets will be loaded after the kpatch patch module, then
pre-patch callbacks will execute in process context via the module notifier
and may be able to directly participate in mutual exclusion primitives.

    Similarly, if all patch targets will be unloaded before the kpatch patch
module is removed or disabled, all unpatch callbacks will execute in process
context.

4. Finally, if a kpatch is disabled or removed and while all patch targets are
still loaded, then all unpatch callbacks will run from stop_machine() -- the
unpatching cannot be stopped at this point and the callbacks cannot spin or
sleep.

    It may be easiest to omit any unpatching code at this point.

    If unpatching and mutual exclusion is required, a separate mechanism, such as
a sysfs file, should be employed to provide process context to safely
spin/sleep to take a lock/mutex and unpatch.

-->8-- -->8-- -->8-- -->8-- -->8-- -->8-- -->8-- -->8-- -->8-- -->8--

I think this points to the fact that kpatch has mainly focused on applying function patches.  We don't normally need to worry about mutual exclusion by manipulating existing data.

jpoimboe:
Thanks, want to close this PR and open up another one to make it easier to review?  I'm not sure about the third workaround, is it realistic?  But otherwise it all seems to make sense to me.

Issue #618, opened 2016-11-10
Detect out-of-range relocations


jpoimboe:
It would be good to add a check at the end of create-diff-object to find any out-of-range relocations so we can find issues like #615 earlier.

stettberger:
My gcc (8.3.0-6) produces such bad relocations in the first place. I got to a situation, where this relocation check failed on the produced patch file. However, when I called the check on the input ELFs there where already bad relocations in place. Has anyone encountered this Problem before? (GCC Producing relocations that point beyond a section boundary).

jpoimboe:
@stettberger I'm confused, from reading the code in kpatch_check_relocations() it seems that it would already support such jump table relocations.  Do you have a more specific example?

stettberger:
With the Follwing sections
```
  [70] .text.ap_send_error_response PROGBITS        0000000000000000 002e20 00094b 00  AX  0   0 16
  [72] .rodata.ap_send_error_response PROGBITS        0000000000000000 00376c 000350 00   A  0   0  4
```
the relocations

```
Relocation section '.rela.rodata.ap_send_error_response' at offset 0x182a0 contains 212 entries:
    Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
...
00000000000002d8  0000003800000002 R_X86_64_PC32          0000000000000000 .text.ap_send_error_response + b13
00000000000002dc  0000003800000002 R_X86_64_PC32          0000000000000000 .text.ap_send_error_response + b17
```

fails with the current version of `kpatch_check_relocations()` fails as 0xb13 > 0x94b.

jpoimboe:
So why do the rodata relocations (presumably from a GCC switch statement jump table) reference code which doesn't exist?  There seems to be a mismatch, is it pulling .rela.rodata.ap_send_error_response from the original unpatched .o file somehow?  I don't see how that can happen.

stettberger:
The compiler produces the following trampoline (L466, L463 are labels within the code section):

```asm
.section .text.ap_send_error_response
        leaq    .L430(%rip), %rdx
        movslq  (%rdx,%r12,4), %rax
        addq    %rdx, %rax
        jmp     *%rax

.section        .rodata.ap_send_error_response,"a",@progbits
.align 4
.align 4
.L430:
        .long   .L466-.L430
        .long   .L463-.L430
        .long   .L463-.L430
        .long   .L465-.L430
```

[...]

PR #823, opened 2018-04-10
create-diff-object: check for out-of-range relocations

sm00th:
Check that none of the relocations are out-of-range of their
corresponding sections before writing the output elf.

Fixes: #618

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1923,6 +1923,27 @@ static void kpatch_regenerate_special_section(struct kpatch_elf *kelf,
 	sec->base->data->d_size = dest_offset;
 }
 
+static void kpatch_check_relocations(struct kpatch_elf *kelf)
+{
+	struct rela *rela;
+	struct section *sec;
+	Elf_Data *sdata;
+
+	list_for_each_entry(sec, &kelf->sections, list) {
+		if (!is_rela_section(sec))
+			continue;
+		list_for_each_entry(rela, &sec->relas, list) {
+			if (rela->sym->sec) {
+				sdata = rela->sym->sec->data;
+				if (rela->addend > sdata->d_size) {
+					ERROR("out-of-range relocation %s+%x in %s", rela->sym->sec->name,
+							rela->addend, sec->name);
+				}
+			}
+		}
+	}
+}
+
 static void kpatch_include_debug_sections(struct kpatch_elf *kelf)
 {
 	struct section *sec;
@@ -3082,6 +3103,7 @@ int main(int argc, char *argv[])
 		sec->sh.sh_info = sec->base->index;
 		kpatch_rebuild_rela_section_data(sec);
 	}
+	kpatch_check_relocations(kelf_out);
 
 	kpatch_create_shstrtab(kelf_out);
 	kpatch_create_strtab(kelf_out);


kpatch-build/create-diff-object.c
```diff
@@ -1923,6 +1923,28 @@ static void kpatch_regenerate_special_section(struct kpatch_elf *kelf,
...
+	struct section *sec;
+	Elf_Data *sdata;
+
+	list_for_each_entry(sec, &kelf->sections, list) {
+		if (!is_rela_section(sec))
+			continue;
+		list_for_each_entry_safe(rela, safe, &sec->relas, list) {
+			if (rela->sym->sec) {
+				sdata = rela->sym->sec->data;
+				if (rela->addend < sdata->d_off ||
```
jpoimboe: On x86, rela->addend is often negative in relation to the symbol, e.g.:
```
Relocation section [20] '.rela.text.semctl_stat' for section [19] '.text.semctl_stat' at offset 0x720 contains 6 entries:
  0x0000000000000001  X86_64_NONE     000000000000000000      -4 __fentry__
  0x000000000000002d  X86_64_PC32     000000000000000000      -4 ipc_obtain_object_check
  0x000000000000004f  X86_64_PC32     000000000000000000      -4 ipcperms
  0x0000000000000062  X86_64_PC32     000000000000000000      -4 security_sem_semctl
  0x0000000000000074  X86_64_PC32     000000000000000000      -4 ipc_obtain_object_idr
  0x0000000000000090  X86_64_PC32     000000000000000000      -4 kernel_to_ipc64_perm
```
That's because on x86, the relative address is adjusted by the length of the instruction which uses it.

So for X86_64_PC32 relocations, the check may need to take into account the length of the instruction which uses the rela's address.  Though that's probably overkill.  Instead I would recommend just removing the `<` check, since I don't think we don't change the addend anyway and I don't see how it could go beyond the beginning of the section.

The `>` check is much more important IMO.

kpatch-build/create-diff-object.c
```diff
@@ -1923,6 +1923,28 @@ static void kpatch_regenerate_special_section(struct kpatch_elf *kelf,
...
+						rela->addend > (sdata->d_off + sdata->d_size)) {
```
jpoimboe: Realistically, I believe there's always only *one* data descriptor per section and `d_off` will always be zero, so it should be fine to do `if (rela->addend >= sdata->d_size)`.

sm00th:
v2 changes:
  - list_for_each_entry_safe -> list_for_each_entry since the list is never modified
  - removed lower bound check
  - removed data offset from checks since it is expected to always be 0
  - added addend to the error message

kpatch-build/create-diff-object.c
sm00th: Should this be >= or >? I've been testing with the patch from #615 on top of current Linus' tree and am getting a relocation that would fail >= check. The resulting module does load alright though:

```
  [12] .text.compat_SyS_getdents PROGBITS        0000000000000000 0002d0 00012f 00  AX  0   0 16
                                                                                   
Relocation section '.rela.debug_ranges' at offset 0x39850 contains 34 entries:  
...                                                                                
00000000000004d8  0000000a00000001 R_X86_64_64            0000000000000000 .text.compat_SyS_getdents + 12f 
...                                                                                

```

jpoimboe: Ah, that's one of the DWARF debug sections.  I think '>' is probably fine then.

PR #824, opened 2018-04-11
patch-author-guide: more on locks in callbacks

joe-lawrence:
Elaborate on the difficulties in using locks/mutexes from the kpatch
callback routines and suggest a few workarounds.

### DIFF
--- a/doc/patch-author-guide.md
+++ b/doc/patch-author-guide.md
@@ -193,10 +193,51 @@ static void kpatch_post_unpatch_tcp_send_challenge_ack(patch_object *obj)
 +KPATCH_POST_UNPATCH_CALLBACK(kpatch_post_unpatch_tcp_send_challenge_ack);
 ```
 
-Don't forget to protect access to the data as needed. Please note that mutexes
-and other sleeping locks can't be used from stop_machine context, so you will
-have to check if any locks protecting the data are already held and if so
-return an error from your callback function.
+Don't forget to protect access to the data as needed. Please note that
+spinlocks and mutexes / sleeping locks can't be used from stop_machine
+context. Also note the pre-patch callback return code will be ignored by the
+kernel's module notifier, so it does not affect the target module or livepatch
+module status. This means:
+
+* Pre-patch callbacks to loaded objects (vmlinux, loaded kernel modules) are
+  run from stop_machine(), so they may only inspect lock state (i.e.
+  spin_is_locked(), mutex_is_locked()) and optionally return -EBUSY to prevent
+  patching.
+
+* Post-patch, pre-unpatch, and post-unpatch callbacks to loaded objects are
+  also run from stop_machine(), so the same locking context applies.  No
+  return status is supported.
+
+* Deferred pre-patch callbacks to newly loading objects do not run from
+  stop_machine(), so they may spin or schedule, i.e. spin_lock(),
+  mutex_lock()).  Return status is ignored.
+
+* Post-patch, pre-unpatch, and post-unpatch callbacks to unloading objects are
+  also *not* run from stop_machine(), so they may spin or sleep.  No return
+  status is supported.
+
+Unfortunately there is no simple, all-case-inclusive kpatch callback
+implementation that handles data structures and mutual exclusion.
+
+A few workarounds:
+
+1. If a given lock/mutex is held and released by the same set of functions
+(that is, functions that take a lock/mutex always release it before
+returning), a trivial change to those functions can re-purpose kpatch's
+activeness safety check to avoid patching when the lock/mutex may be held.
+This assumes that all lock/mutex holders can be patched.
+
+2. If it can be assured that all patch targets will be loaded before the
+kpatch patch module, pre-patch callbacks may return -EBUSY if the lock/mutex
+is held to block the patching.
+
+3. Finally, if a kpatch is disabled or removed and while all patch targets are
+still loaded, then all unpatch callbacks will run from stop_machine() -- the
+unpatching cannot be stopped at this point and the callbacks cannot spin or
+sleep.
+
+    With that in mind, it is probably easiest to omit unpatching callbacks
+at this point.
 
 Also be careful when upgrading.  If patch A has a pre/post-patch callback which
 writes to X, and then you load patch B which is a superset of A, in some cases


doc/patch-author-guide.md
```diff
@@ -193,10 +193,62 @@ static void kpatch_post_unpatch_tcp_send_challenge_ack(patch_object *obj)
...
 ```
 
-Don't forget to protect access to the data as needed. Please note that mutexes
-and other sleeping locks can't be used from stop_machine context, so you will
-have to check if any locks protecting the data are already held and if so
-return an error from your callback function.
+Don't forget to protect access to the data as needed. Please note that
+spinlocks and mutexes / sleeping locks can't be used from stop_machine
+context. Also note the pre-patch callback return code will be ignored by the
+kernel's module notifier, so it does not effect the target module or livepatch
```
jpoimboe: s/effect/affect/

doc/patch-author-guide.md
```diff
@@ -193,10 +193,62 @@ static void kpatch_post_unpatch_tcp_send_challenge_ack(patch_object *obj)
...
+kpatch patch module, pre-patch callbacks may return -EBUSY if the lock/mutex
+is held to block the patching.
+
+3. If all patch targets will be loaded after the kpatch patch module, then
+pre-patch callbacks will execute in process context via the module notifier
+and may be able to directly participate in mutual exclusion primitives.
+
+    Similarly, if all patch targets will be unloaded before the kpatch patch
+module is removed or disabled, all unpatch callbacks will execute in process
+context.
```
jpoimboe: Unlike 2, I'm not sure 3 is very feasible.  It seems unlikely and difficult to enforce with tooling.

doc/patch-author-guide.md
```diff
@@ -193,10 +193,62 @@ static void kpatch_post_unpatch_tcp_send_challenge_ack(patch_object *obj)
...
+
+4. Finally, if a kpatch is disabled or removed and while all patch targets are
+still loaded, then all unpatch callbacks will run from stop_machine() -- the
+unpatching cannot be stopped at this point and the callbacks cannot spin or
+sleep.
+
+    It may be easiest to omit any unpatching code at this point.
```
jpoimboe: I'm not sure what this means.  Does it mean to leave the unpatch callbacks empty?

doc/patch-author-guide.md
```diff
@@ -193,10 +193,62 @@ static void kpatch_post_unpatch_tcp_send_challenge_ack(patch_object *obj)
...
+
+    If unpatching and mutual exclusion is required, a separate mechanism, such as
+a sysfs file, should be employed to provide process context to safely
+spin/sleep to take a lock/mutex and unpatch.
```
jpoimboe: Interesting idea, but I wonder if this is really feasible (and worth documenting)?  It seems like a tricky feature to implement and it would require tooling to support it.

doc/patch-author-guide.md
```diff
@@ -193,10 +193,62 @@ static void kpatch_post_unpatch_tcp_send_challenge_ack(patch_object *obj)
...
+
+1. If a given lock/mutex is held and released by the same set of functions
+(that is, functions that take a lock/mutex always release it before
+returning), a trivial change to those functions can re-purpose kpatch's
+activeness safety check to avoid patching when the lock/mutex may be held.
+This assumes that all lock/mutex holders can be patched.
+
+2. If it can be assured that all patch targets will be loaded before the
+kpatch patch module, pre-patch callbacks may return -EBUSY if the lock/mutex
+is held to block the patching.
```
jpoimboe: side note: I think I'll post a livepatch RFC to get rid of patching modules before they're loaded, by creating patch module depencies on the target objects.  It will make so many of these things easier... If it were accepted upstream (probably unlikely) then we could propose something similar for kpatch.ko.

doc/patch-author-guide.md
joe-lawrence: This section was only a listing of possibilities,  I never said this one was reasonable :)  I agree that it is probably not a useful suggestion, so we could drop it if you like.

doc/patch-author-guide.md
joe-lawrence: Yes.  My take on it is that undo-ing the patch is probably going to be more trouble than its worth.

doc/patch-author-guide.md
joe-lawrence: I think @sm00th was tinkering with this idea for one of our kpatches, so I thought I'd mention it.  I agree that kpatch tooling doesn't help us here.

doc/patch-author-guide.md
jpoimboe: The document is meant to guide, I'm not sure we want to guide people in that direction :-)

doc/patch-author-guide.md
jpoimboe: Agreed, but that wasn't clear to me so this statement should be clarified.

doc/patch-author-guide.md
jpoimboe: I'd say let's leave such ideas out of the guide unless it's something we'd be willing to do ourselves (and recommend).

doc/patch-author-guide.md
joe-lawrence: It would cut down the support/testing matrix... maybe saleable if it really simplifies the code.

One thing to consider -- modules can to fail their init (see #816 for example) if hardware is not in an expected state.  This might be a very unlikely scenario, but the module dependencies would require all of those modules to successfully load before the kpatch module.  Maybe not worrying about.

The weirdest thing I found about this feature (in livepatch context) was that a failing pre-patch callback had two different outcomes based on how it was called:  If the kpatch module is loading, its failure would prevent the kpatch module from loading.  If kpatch was already loaded and a module is loading, it would prevent the module from loading.   Not completely crazy, but on the other hand, forces the patch writer to understand from which any particular callback will being made.

joe-lawrence:
Do we want to use this opportunity to brainstorm better solutions and/or kpatch enhancements to support such ideas?  I'm just trying to better frame this PR and the recommended text.

doc/patch-author-guide.md
jpoimboe: Good point about modules failing their init due to missing HW.  We should see how common that is, and figure out what to do about it.  I get the sense that the vast majority of modules don't do that.

jpoimboe:
Sure, what ideas did you have?  I thought the text was headed in the right direction, except for the bits that were more brainstorming and less practical.

joe-lawrence:
I'll have to think on it a little bit, though I'm kinda hesitant to expend a lot of time and energy when livepatch is going to be the future.  We went down a lot of roads with last week's kpatch, but ended up concluding that most ideas solved 90% of the problem, "except for this one thing."   Maybe if kpatch can be enhanced a little, some of those ideas would have been closer to 100%.

joe-lawrence:
v2:
- s/effect/affect
- remove suggestion #3 (load all module before patching, unload all before unpatching)
- reword the advice to omit unpatch callbacks in final case

PR #826, opened 2018-04-12
Fix false-positives in kpatch_check_relocations

sm00th:
Because of signdness difference kpatch_check_relocations() would trigger
an error on any negative addend.
Fix by casting Elf_Data->d_size from size_t to long.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1935,7 +1935,7 @@ static void kpatch_check_relocations(struct kpatch_elf *kelf)
 		list_for_each_entry(rela, &sec->relas, list) {
 			if (rela->sym->sec) {
 				sdata = rela->sym->sec->data;
-				if (rela->addend > sdata->d_size) {
+				if (rela->addend > (int)sdata->d_size) {
 					ERROR("out-of-range relocation %s+%x in %s", rela->sym->sec->name,
 							rela->addend, sec->name);
 				}


jpoimboe:
No idea... or maybe `int` would be better since that's what `rela->addend` is?

sm00th:
> pedantic C question: would casting to ssize_t be more accurate?

I think it would.

> No idea... or maybe int would be better since that's what rela->addend is?

I was a bit worried about int size on 32bit systems (do we even support those?) Can we be sure d_size would never overflow into negative values?

joe-lawrence:
I didn't think we supported 32-bit, but I notice that the 32-bit alt-arch CentOS 7 has a kpatch rpm in its repo:

http://mirror.centos.org/altarch/7/os/i386/Packages/kpatch-0.4.0-1.el7.noarch.rpm

so I'll defer to @jpoimboe on that question.

jpoimboe:
No, we don't support 32-bit, that RPM is just a bash script.

sm00th:
v2:
 - fixed typo in comment
 - casting to int since it is the same as rela->addend and should be enough on 64bit systems

joe-lawrence:
Looks good, but I should have suggested earlier:
```
diff --git a/kpatch-build/Makefile b/kpatch-build/Makefile
index 32e1995..9587d5e 100644
--- a/kpatch-build/Makefile
+++ b/kpatch-build/Makefile
@@ -1,6 +1,6 @@
 include ../Makefile.inc
 
-CFLAGS += -MMD -MP -I../kmod/patch -Iinsn -Wall -g -Werror
+CFLAGS += -MMD -MP -I../kmod/patch -Iinsn -Wall -Wsign-compare -g -Werror
 LDLIBS = -lelf
 
 TARGETS = create-diff-object create-klp-module create-kpatch-module
```
to see the other seven similar instances in create-diff-object.c.

I can merge this PR if you prefer to look at those in other commit.  Let me know.

PR #827, opened 2018-04-12
kpatch-build: fix Ubuntu kernel detection on successive retries

vincentbernat:
If a patch failed a first time, kpatch-build is using the previous
cache directory on subsequent builds. The UBUNTU_KERNEL=1 variable is
not set in this case. Therefore, utsrelease.h is not updated correctly
and the appropriate structures are not used. Just check if distro is
Ubuntu and we didn't request our own specific directory.

Signed-off-by: Vincent Bernat <vincent@bernat.im>


joe-lawrence:
@vincentbernat  Thanks for fixing

PR #828, opened 2018-04-12
livepatch-patch-hook: make callbacks addition depend on the kernel ve…

rppt:
…rsion

Since commit 926e4e0c7d14b43896f9814da236b4d6408c8ff0 ("kmod: add support
for in-kernel livepatch hooks") livepatch-patch-hook.c unconditionally
creates callbacks. This causes build error for kernels older than 4.15:

make -C /home/rppt/git/linux M=/home/rppt/.kpatch/tmp/patch livepatch-proc-vmalloc-live.ko
make[1]: Entering directory '/home/rppt/git/linux'
  CC [M]  /home/rppt/.kpatch/tmp/patch/patch-hook.o
In file included from /home/rppt/.kpatch/tmp/patch/patch-hook.c:21:0:
/home/rppt/.kpatch/tmp/patch/livepatch-patch-hook.c:82:23: error: field ‘callbacks’ has incomplete type
  struct klp_callbacks callbacks;
                       ^
/home/rppt/.kpatch/tmp/patch/livepatch-patch-hook.c: In function ‘patch_init’:
/home/rppt/.kpatch/tmp/patch/livepatch-patch-hook.c:395:10: error: ‘struct klp_object’ has no member named ‘callbacks’
   lobject->callbacks = object->callbacks;
          ^
scripts/Makefile.build:302: recipe for target '/home/rppt/.kpatch/tmp/patch/patch-hook.o' failed
make[2]: *** [/home/rppt/.kpatch/tmp/patch/patch-hook.o] Error 1
Makefile:1687: recipe for target 'livepatch-proc-vmalloc-live.ko' failed
make[1]: *** [livepatch-proc-vmalloc-live.ko] Error 2
make[1]: Leaving directory '/home/rppt/git/linux'
Makefile:20: recipe for target 'livepatch-proc-vmalloc-live.ko' failed
make: *** [livepatch-proc-vmalloc-live.ko] Error 2

Introduce HAVE_LIVEPATCH_CALLBACKS to allow conditional compilation of the
callbacks addition.

Signed-off-by: Mike Rapoport <rppt@linux.vnet.ibm.com>

### DIFF
--- a/kmod/patch/livepatch-patch-hook.c
+++ b/kmod/patch/livepatch-patch-hook.c
@@ -53,6 +53,14 @@
 #define HAVE_IMMEDIATE
 #endif
 
+#ifdef RHEL_RELEASE_CODE
+# if RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(7, 5)
+#  define HAVE_CALLBACKS
+# endif
+#elif LINUX_VERSION_CODE >= KERNEL_VERSION(4, 15, 0)
+# define HAVE_CALLBACKS
+#endif
+
 /*
  * There are quite a few similar structures at play in this file:
  * - livepatch.h structs prefixed with klp_*
@@ -79,7 +87,9 @@ struct patch_object {
 	struct list_head list;
 	struct list_head funcs;
 	struct list_head relocs;
+#ifdef HAVE_CALLBACKS
 	struct klp_callbacks callbacks;
+#endif
 	const char *name;
 	int funcs_nr, relocs_nr;
 };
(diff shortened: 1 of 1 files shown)


jpoimboe:
Thanks!

Slight nit, I would call it HAVE_CALLBACKS, since the "LIVEPATCH" is already a given.

Also for the non-HAVE_CALLBACKS case, I wonder if we should return an error in add_callbacks_to_patch_objects() if the user specified any callbacks but the kernel doesn't support them?  Otherwise it would be a silent failure.

rppt:
> Slight nit, I would call it HAVE_CALLBACKS, since the "LIVEPATCH" is already a given.

agree

> Also for the non-HAVE_CALLBACKS case, I wonder if we should return an error in add_callbacks_to_patch_objects() if the user specified any callbacks but the kernel doesn't support them? Otherwise it would be a silent failure.

Maybe it'd be better to fail the build, e.g.:

```
static inline int add_callbacks_to_patch_objects
{
        BUILD_BUG_ON(__kpatch_callbacks_pre_patch != __kpatch_callbacks_pre_patch_end);
        /* etc */
}
```
?

jpoimboe:
Agreed, the BUILD_BUG_ON is even better.

rppt:
Apparently I was too optimistic with `BUILD_BUG_ON()` because somehow I've missed the fact that compilation of `livepatch-pacth-hook.c` does not know what would be the values of `__kpatch_callbacks_*`
So, in the end I've made `add_callbacks_to_patch_objects()` to return an error if user defined some callbacks but `HAVE_CALLBACKS=n`

PR #830, opened 2018-04-16
Enable TravisCI

sm00th:
Add .travis.yml with build/unit/shellcheck tests.


sm00th:
I've also enabled travisci for dynup/kpatch repo so we should have check results on this pull-request and at  least people in dynup organisation should have access to travis dashboard at https://travis-ci.org/dynup/kpatch

PR #831, opened 2018-04-17
unit: don't suppress stderr

jpoimboe:
Don't suppress stderr in the unit tests, so that create-diff-object
error messages will still be printed.


sm00th:
I originally removed those because we have cases when create-diff-object is supposed to fail, but I do understand this might be unwanted for the normal case. Maybe instead of removing it we can move it to [failure case](https://github.com/dynup/kpatch/pull/831/files#diff-c409ac299319bd1a72dc474515116500R58)?

jpoimboe:
Ah, ok.  Yeah, that would be better.

jpoimboe:
v2:
- Continue to suppress stderr for FAIL tests

PR #833, opened 2018-04-19
ppc64le: fix gcc-plugin installation

sm00th:
Because $PLUGIN variable is a part of $TARGETS we are trying to install
gcc-plugins twice which is treated as an error by install command:

```
/usr/bin/install create-diff-object create-klp-module create-kpatch-module gcc-plugins/ppc64le-plugin.so kpatch-gcc gcc-plugins/ppc64le-plugin.so /usr/local/libexec/kpatch
/usr/bin/install: will not overwrite just-created '/usr/local/libexec/kpatch/ppc64le-plugin.so' with 'gcc-plugins/ppc64le-plugin.so'
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,4 @@ script:
   - make
   - make unit
   - make check
+  - sudo make install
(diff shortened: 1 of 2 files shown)


jpoimboe:
Should we add 'make install' to the Travis tests?

sm00th:
> Should we add 'make install' to the Travis tests?

Might be a good idea. I don't expect there to be many issues like this, but it might catch an occasional slip-up in the build system.

jpoimboe:
Yeah, this is the first such 'make install' bug I can remember, but it doesn't hurt to add a test for it.

Issue #764, opened 2017-11-27
testing fedora 27: macro-printk: build failed


joe-lawrence:
Integration test results from PR #762 
- Fedora release 27 (Twenty Seven)
- 4.13.13-300.fc27.x86_64
- gcc version 7.2.1 20170915 (Red Hat 7.2.1-2) (GCC) 
```
ERROR: macro-printk: build failed
fib_frontend.o: changed function: inet_rtm_newroute
ERROR: fib_semantics.o: .rodata.__msg.61343 section header details differ
/root/kpatch.joe/kpatch-build/create-diff-object: unreconcilable difference
fib_trie.o: changed function: fib_table_insert
meminfo.o: changed function: meminfo_proc_show
```

joe-lawrence:
Reran the build with a little more debugging text:
```
ERROR: fib_semantics.o: .rodata.__msg.61364 vs .rodata.__msg.61341 section header details differ
```

Let's get more information about these two sections.  `readelf` confirms that the alignment is different.  The two sizes are different as well:
```
ORIG=~/.kpatch/tmp/orig/net/ipv4/fib_semantics.o
PATCHED=~/.kpatch/tmp/patched/net/ipv4/fib_semantics.o

% (readelf --wide -S $PATCHED | head -n4; \
   echo "(orig)"; \
   readelf --wide -S $ORIG | grep '61341'; \
   echo "(patched)"; \
   readelf --wide -S $PATCHED | grep '61364') | column -t

There      are                  109       section           headers,  starting  at  offset  0xa3c88:       
Section    Headers:                                                                                        
[Nr]       Name                 Type      Address           Off       Size      ES  Flg     Lk        Inf  Al
(orig)                                                                                                     
[68]       .rodata.__msg.61341  PROGBITS  0000000000000000  002e78    00000e    00  A       0         0    8
(patched)                                                                                                  
[66]       .rodata.__msg.61364  PROGBITS  0000000000000000  003240    000018    00  A       0         0    16
```

`objdump` the contents of the sections and we can see that they don't match:
```
% objdump -s -j '.rodata.__msg.61341' $ORIG 
...
Contents of section .rodata.__msg.61341:
 0000 496e7661 6c696420 73636f70 6500      Invalid scope.  


% objdump -s -j '.rodata.__msg.61364' $PATCHED
...
Contents of section .rodata.__msg.61364:
 0000 496e7661 6c696420 70726566 73726320  Invalid prefsrc 
[...]

Issue #763, opened 2017-11-27
testing fedora 27: gcc-static-local-var-5: build failed


joe-lawrence:
Integration test results from PR #762 
- Fedora release 27 (Twenty Seven)
- 4.13.13-300.fc27.x86_64
- gcc version 7.2.1 20170915 (Red Hat 7.2.1-2) (GCC) 
```
ERROR: gcc-static-local-var-5: build failed
ERROR: audit.o: reference to static local variable lock.61272 in audit_log_end was removed
/root/kpatch.joe/kpatch-build/create-diff-object: unreconcilable difference
meminfo.o: changed function: meminfo_proc_show
```

joe-lawrence:
Fedora 27 is way EOL and gcc-static-local-var-5.patch does not cause any problems on later RHEL versions.  Closing.

PR #838, opened 2018-04-19
Test cleanups

jpoimboe:
Some cleanups for the unit and integration tests.


jpoimboe:
Agreed, I was thinking F27 could be the base for all future patch rebases.  Ubuntu 16.04 and Centos 7 can be cleaned up later, though I suppose it wouldn't hurt anything to leave the old tests there.

sm00th:
Great, the output looks much nicer now :+1:

joe-lawrence:
Nice cleanups.  A few review questions/notes:

- On removing macro-printk.patch:  this means that there no more tests for our ```KPATCH_PRINTK``` macro.  Also, this patch currently fails on F27 (issue #764).
- On removing gcc-static-local-var-5.patch:  this currently fails on F27 (see issue #763).
- I'm on board with removing a bunch of these from the integration tests directory, but it leaves some of the unit test objects files (like bug-table-section, data-read-mostly, etc.) without source patch files.
- `assert_num_funcs()` - would verifying the function names be too much?

joe-lawrence:
> Maybe there's some other way to flush out such random issues.

If we could ever figure out a clean way to import all those historical CVE patches that we worked on, it would offer more than enough input for kpatch to chew on once a release cycle or so.

> And we can continue to brainstorm about how to flush out these kinds of issues, but I don't really like our current method of constantly rebasing the integration tests, as they eventually end up not testing what they were originally intended to test. So I want to keep the integration tests as lean as possible.

I think there are too many and the result is that nobody ends up running them with any regularity.  So trimming them down and splitting off the unit tests moves us to a much better position to do so.

PR #842, opened 2018-04-26
Unittests submodule

sm00th:
Switching from git-lfs to a separate repo for objectfiles.

I am planning to change the [personal repo](https://github.com/sm00th/kpatch-unit-objs.git) to dynup/kpatch-unit-objs before we merge this. Just wanted to make sure everyone is on board with the contents before creating an organization repo.


joe-lawrence:
git submodules are new to me, so perhaps a short README in test/unit to explain how to set it up would helpful.  Googling around, I found something like `git submodule init && git submodule update` would populate the test/unit/kpatch-unit-objs directory.  Do projects ever add something like that into their Makefiles to do this automatically?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "test/unit/kpatch-unit-objs"]
+	path = test/unit/kpatch-unit-objs
```
jpoimboe: Can we make the path `test/unit/objs`?  Easier to type and less redundant.

joe-lawrence:
> I'd propose a check_submodule make command ... 'make unit' would be self documenting and we wouldn't necessarily need a README.

:+1:

.gitmodules
jpoimboe: Yeah, sounds fine to me.  Or even "kpatch-unit-test-objs" if that's not too verbose.

sm00th:
> Similar to the check_lfs make command we had before, I'd propose a check_submodule make command as part of the make unit target to make sure the unit test objects are up to date. It should either update the submodule automatically, or at least print the commands needed to do so.

We can run "git submodule update --init" before every test run. The only problem is - it will reset the submodule to the recorded commit on every run unless you specifically tell it not to (by means of `git add test/unit/objs`), it doesn't seem to affect newly added files though so it is only a problem if you are trying to test a different commit from objs repo, which should be an obscure enough case to ignore this.

jpoimboe:
Maybe do something like "git submodule update --init --rebase" so any existing local changes will be rebased onto the latest?

sm00th:
v2:
  - submodule path changed to test/unit/objs
  - updating submodule on every run with `git submodule update --init --rebase`

jpoimboe:
Looks good to me.  Adding new tests is going to be a pain (two pull requests instead of one), but this seems to be our least worst option.

sm00th:
v3:
 - switched to dynup repo

Issue #771, opened 2017-12-01
(enhancement) testing: improve build times


joe-lawrence:
In #752 we talked about a few possible ways to avoid extraneous builds and general integration test time.

While rebasing the tests for Fedora 27, I continued tinkering with two of these ideas to see what their workflow would look like and generate ballpark test times.  I don't have review-able code at the moment, but wanted to talk a bit more the general ideas:

1. Reuse an existing kpatch-build cache from `~/.kpatch`.  To do this, I added a kpatch-build command line option `--cache` which accepted two actions: `build` and `copy`.  Before building a series of kpatches, the `--cache build` option downloads, builds the specified targets, and then stops.  (It is not compatible with any input patches on the command line.)  The cachedir can now be renamed `.kpatch.copy`, backed up, whatever.  Then when building kpatch modules with `--cache copy` option, kpatch-build will `rsync --archive --delete .kpatch.copy .kpatch` instead of (re)downloading and (re)building the original kernel.

   Rsync is nice here as it will only restore what changed between builds.   Therefore, reusing the kpatch cache works best when few objects are updated.  However, if a patch causes an entire tree rebuild, then a subsequent `--cache copy` rsync could take more than a few moments.

2. Add object target specification(s) to the integration patch files.  Obviously we need to know what targets will be affected by the patch, so this needs to be used with care.  To test the idea, I added the target name(s) to the top of the patch file description and fed those as kpatch-build `--target` options. 

Here are some very preliminary numbers.  Not all of the Fedora 27 test patches build yet, so the final numbers might be a little different.  However I think the data is still interesting to see:

```
VM
==
qemu with kvm enabled
Memory 4096 MB
CPUs 6

==> out.VM <==
real    1770m34.007s
==> out.VM.cache <==
real    450m31.223s
==> out.VM.cache.target <==
real    307m18.664s

Beaker machine
==============
Model Name      Intel(R) Xeon(R) CPU X3470 @ 2.93GHz
Memory  32768 MB
CPUs 8

==> out.beaker <==
real    295m58.502s
==> out.beaker.cache <==
real    115m2.115s
==> out.beaker.cache.target <==
real    36m30.839s
```

These two optimizations led to full integration tests of only 13-17% of their original wall times.  (The numbers may change a bit once all of the tests build successfully for Fedora 27.)

[...]

PR #845, opened 2018-04-28
added function for archiving builded kernels or using it to kpatch-bu…

M1cRO:
Added function for archiving builded kernels or using it to kpatch-build as kpatch-build -b option.
You can set -b options to directory when you want to have builded kernel with ccache cache, if you already have builded kernel it will be used instead full building, in my case xz pack workin 2.30 min, unpack 5min (one thread for unpack, can't be multithreaded), building full kernel takes around 20 minutes.

Actually need add some options for set archive threads, archiver

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -722,7 +722,7 @@ if [ -n "${BUILDEDKERNELDIR}" ]; then
 		CROSS_COMPILE="$TOOLSDIR/kpatch-gcc " make "-j$CPUS" $TARGETS 2>&1 | logger || die
 
 		echo "Archive builded kernel"
-		XZ_OPT="-1v -T8" tar -C /tmp/$KVER-$KREL.x86_64 -cJf ${BUILDEDKERNELDIR}/builded-${KVER}-${KREL}.x86_64.tar.xz .
+		XZ_OPT="-1v -T8" tar -C /tmp/${KVER}-${KREL}.x86_64 -cJf ${BUILDEDKERNELDIR}/builded-${KVER}-${KREL}.x86_64.tar.xz .
 	fi
 else
 	echo "Building original kernel"
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -564,7 +564,7 @@ if [ -n "${BUILDEDKERNELDIR}" ]; then
 		echo "${BUILDEDKERNELDIR}/builded-${KVER}-${KREL}.x86_64.tar.xz will be created"
 	fi
 fi
-if [ -n "${BUILDEDKERNELDIR}" && ! -f "${BUILDEDKERNELDIR}/builded-${KVER}-${KREL}.x86_64.tar.xz" || ! -n "${BUILDEDKERNELDIR}" ]; then
+if [ -n "${BUILDEDKERNELDIR}" ] && [ ! -f "${BUILDEDKERNELDIR}/builded-${KVER}-${KREL}.x86_64.tar.xz" ] || [ ! -n "${BUILDEDKERNELDIR}" ]; then
 	if [[ -n "$USERSRCDIR" ]]; then
 		echo "Using source directory at $USERSRCDIR"
 
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -371,26 +371,27 @@ module_name_string() {
 
 usage() {
 	echo "usage: $(basename "$0") [options] <patch1 ... patchN>" >&2
-	echo "		patchN             Input patchfile(s)" >&2
-	echo "		-h, --help         Show this help message" >&2
-	echo "		-a, --archversion  Specify the kernel arch version" >&2
-	echo "		-r, --sourcerpm    Specify kernel source RPM" >&2
-	echo "		-s, --sourcedir    Specify kernel source directory" >&2
-	echo "		-c, --config       Specify kernel config file" >&2
-	echo "		-v, --vmlinux      Specify original vmlinux" >&2
-	echo "		-j, --jobs         Specify the number of make jobs" >&2
-	echo "		-t, --target       Specify custom kernel build targets" >&2
-	echo "		-n, --name         Specify the name of the kpatch module" >&2
-	echo "		-o, --output       Specify output folder" >&2
-	echo "		-d, --debug        Enable 'xtrace' and keep scratch files" >&2
-	echo "		                   in <CACHEDIR>/tmp" >&2
-	echo "		                   (can be specified multiple times)" >&2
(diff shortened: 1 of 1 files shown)


jpoimboe:
I tried it out, but it didn't seem to work:
```
$ kpatch-build/kpatch-build -b /tmp/abc ~/meminfo-string.patch -t vmlinux
/tmp/abc/builded-4.16.4-200.fc27.x86_64.tar.xz will be created
Using cache at /home/jpoimboe/.kpatch/src
Testing patch file(s)
Reading special section data
Building original kernel
Archive builded kernel
tar: /tmp/4.16.4-200.fc27.x86_64: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
Building patched kernel
  100 %                  32 B / 0 B > 9.999                                    
Patched objects: vmlinux
Building patch module: kpatch-meminfo-string.ko
SUCCESS
$ ls -l /tmp/abc
total 4
-rw-rw-r--. 1 jpoimboe jpoimboe 32 Apr 30 13:25 builded-4.16.4-200.fc27.x86_64.tar.xz
```

joe-lawrence:
Thanks for taking a look at this, @M1cRO.  The integration test build times can definitely take a considerably amount of time to complete.  Those timings of course, depend on the setup:

* Kernel config (Fedora can take much longer than RHEL for example)
* ccache (configuration, size, previous builds, etc.)
* processor, disk, etc.

We've floated various ideas in the past toward addressing this.  I think https://github.com/dynup/kpatch/issues/771 and its [WIP test-branch](https://github.com/dynup/kpatch/compare/master...joe-lawrence:kpatch_readonly_cache?expand=1) was the last round.  In that case, we proposed a "read-only" cache directory to stash the build artifacts.  

Would something like that meet the needs behind this enhancement?  If disk space is a concern, perhaps the read-only cache would be compressed externally to the kpatch-build tool?  I like the idea of archiving a reference kernel build, but I think it would be more flexible if kpatch-build itself didn't have to worry about the tar and compression options.

M1cRO:
@jpoimboe I thought that the directory with the kernel (like /tmp/4.16.4-200.fc27.x86_64) is the same for all distributions.
Sorry I tested it on CentOS7,  what the directory with kernel in Fedora?

M1cRO:
@joe-lawrence 
- Actually yes, it looks like the same idea with difference realisation. (I building many different kernels for different CentOS with different patches)
- If you have problem with 3gb space I think you will have problem with building ~15gb too :) (I did it in mock and put builded kernel by mock and building it on tmpfs ^_^)
- Regarding "if kpatch-build itself didn't have to worry about the tar and compression options" then who should doing it?

joe-lawrence:
@M1cRO 
> (joe) The integration test build times can definitely take a considerably amount of time to complete.

Oops, I thought this PR had referred to the integration tests, but after re-reading, it's doesn't.  My assumption (ie, running multiple kpatch builds in succession) lead me to think the following ...

> (M1cRO) Regarding "if kpatch-build itself didn't have to worry about the tar and compression options" then who should doing it?

If the compression is baked into the script, then every use case would incur the packing and unpacking time penalty.  I may want to re-use the kernel build if I'm building a bunch of patches against the same kernel ... for that, the archive packing and unpacking adds extra overhead that I don't need.

My first reaction was to wonder if one could do something like (fixing the tar options accordingly):
```
kpatch-build --cache build
tar -Jcvf ~/kpatch.copy-$KVER ~/.kpatch.copy       # archive the cache copy
rm ~/.kpatch.copy                                  # remove if desired
```
and then to use it:
```
tar -Jxvf ~/kpatch.copy-$KVER ~/.kpatch.copy       # extract the cache copy
kpatch-build --cache copy <my build options here>
rm ~/.kpatch.copy                                  # remove if desired
```

Another idea that may be do-able would be to specify the cache creation and unpacking commands themselves on the kpatch-build line.  That way we could pull the rsync commands out of [WIP test-branch](https://github.com/dynup/kpatch/compare/master...joe-lawrence:kpatch_readonly_cache?expand=1) to make the kpatch-build even more archive/rsync/etc agnostic.

M1cRO:
@joe-lawrence 
> If the compression is baked into the script, then every use case would incur the packing and unpacking time penalty.
Nope, just one pack all other times just unpack.
But if you do many separeted patches (modules) for one kernel it will be many unpacks and it will be a problem (still not a fast, 5 minutes for one unpack ;(), but if we will be able to use zfs snapshots (sorry i'm freebsd user ^_^) or somethink like that it can improve this operation. Also if we have enough place we can just move it if we are using tmpfs (but it not for all).

When we are using rsync we still needed many spaces (same 15gb?). In my case, we pack builded kernel (with ccache) onсe and it take around ~2.5gb (for earlier version 1.6gb), and when I building new modules I just unpack it one time (yes I do it like one module for all pacthes)

joe-lawrence:
Ah, you're right about the unpack penalty.  But anyway, I think the point stands that something like a pre-build command (to restore a copy, cache, or snapshot) and a post-build command (to create a copy, cache, or snapshot) would be the most flexible implementation.  I had thought about tinkering with dm-snapshots at one point, but though it was too specific to the integration test case and would impose too much on the end user.

joe-lawrence:
@M1cRO 
> (joe) ... something like a pre-build command (to restore a copy, cache, or snapshot) and a post-build command (to create a copy, cache, or snapshot) would be the most flexible implementation ...

I threw together a [prototype](https://github.com/dynup/kpatch/compare/master...joe-lawrence:archived-build) of this idea.  Tested once, sharp around the edges, etc.

The idea is that you could pass in commands to the kpatch-build script.  First, to construct a cache and archive it:
```
  ./kpatch-build/kpatch-build \
      --archive-cmd "tar -C $HOME -Jcvf /tmp/kpatch.tar.xz .kpatch/" \
      -t vmlinux
```
And then when building a kpatch, a command to unpack and restore the archive:
```
  ./kpatch-build/kpatch-build \
      --unpack-cmd "tar -C $HOME -Jxf /tmp/kpatch.tar.xz" \
      -t vmlinux \
      test/integration/centos-7/meminfo-string.patch
```
I didn't bother removing $HOME/.kpatch in these examples, but it wouldn't be difficult to run more elaborate commands or scripts that could cp, rsync, FS-snapshot, wget, etc. the archive.

Do you think this would meet your needs?

Also, I couldn't think of better terms than "archive-cmd" and "unpack-cmd", so suggestions on terminology welcome :)

M1cRO:
@joe-lawrence I've commented it there https://github.com/joe-lawrence/kpatch/commit/7437e9ef1fe984da30091e83b2e61fb22c734aaf, all other looks fine for me ;) and this idea/realization better that mine 👍

joe-lawrence:
Closing and handling in #848

PR #846, opened 2018-05-03
Update kpatch.spec, keep it updated

M1cRO:


jpoimboe:
@M1cRO Can you rebase them into a single commit and force push to the same branch?

M1cRO:
@jpoimboe  somethink like that?

joe-lawrence:
Thanks, @M1cRO

Issue #843, opened 2018-04-27
ppc64le: WARN detection


jpoimboe:
ppc64le needs to support the kpatch_line_macro_change_only() function so that the WARN macros dont't cause functions to be detected as changed unnecessarily, due to their use of the `__LINE__` macro.

PR #847, opened 2018-05-04
ppc64le: add WARN detection

sm00th:
ppc64le lacked kpatch_line_macro_change_only() implementation. Add one
based on x86_64 version and available unit tests.

Fixes: #843 

---

Tested on objectfiles available in our unit tests and this works fine, it might not cover other cases, but I guess it is still better than nothing.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -509,6 +509,72 @@ static int kpatch_line_macro_change_only(struct section *sec)
 
 	return 1;
 }
+#elif __powerpc__
+#define PPC_INSTR_LEN 4
+#define PPC_RA_OFFSET 16
+
+static int kpatch_line_macro_change_only(struct section *sec)
+{
+	unsigned long start1, start2, size, offset;
+	unsigned int instr1, instr2;
+	struct rela *rela;
+	int lineonly = 0, found;
+
+	if (sec->status != CHANGED ||
+	    is_rela_section(sec) ||
+	    !is_text_section(sec) ||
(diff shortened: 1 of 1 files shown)


kpatch-build/create-diff-object.c
```diff
@@ -509,6 +509,74 @@ static int kpatch_line_macro_change_only(struct section *sec)
...
+		found = 0;
+		list_for_each_entry(rela, &sec->rela->relas, list) {
+			if (rela->offset < offset + PPC_INSTR_LEN)
+				continue;
+			if (rela->string)
+				continue;
+			if (!strncmp(rela->sym->name, "__warned.", 9))
+				continue;
+			if (!strcmp(rela->sym->name, ".toc"))
+				continue;
```
jpoimboe: Instead of checking for a .toc rela, I think it would be good to be more specific here.  When you saw this, what was the "second level" rela, i.e. what did toc_rela(rela) refer to?

sm00th: For all relas toc_rela was refering to a rela with sym->name ".rodata.netdev_is_rx_handler_busy.str1.8". Will "second level" rela have toc_rela() as non-null and we can just check for that?

jpoimboe: Ah, that's a string, so I think you can just change the `if (rela->string)` to `if (toc_rela(rela)->string)`.

jpoimboe:
Looks great, assuming the unit tests need to be updated.

sm00th:
v3:
 - uni-test-obj submodule update

PR #848, opened 2018-05-04
kpatch-build: add save/restore-cache archive command options

joe-lawrence:
Add kpatch-build command arguments that the script will execute after
the reference kernel build (--archive-cmd) or in lieu of building the
reference kernel (--unpack-cmd).

* If an --archive-cmd is provided, then no input patch files should be
   specified.  kpatch-build will exit immediately after running the
   --archive-cmd.

* If --unpack-cmd is provided, input patches will be required.  These
  will be applied immediately after running the --unpack-cmd and
  kpatch-build will continue with building a patched kernel.

Possible use cases include creating reference kernel build tarball
archives:

  ./kpatch-build/kpatch-build \
      --archive-cmd "tar -C $HOME -Jcf /tmp/kpatch.tar.xz .kpatch/" \
      -t vmlinux

  [ ... $HOME/.kpatch/ may be removed ... ]

  ./kpatch-build/kpatch-build \
      --unpack-cmd "tar -C $HOME -Jxf /tmp/kpatch.tar.xz" \
      -t vmlinux \
      test/integration/centos-7/meminfo-string.patch

or rsyncing the reference build to another location:

  ./kpatch-build/kpatch-build \
    --archive-cmd "rsync --archive --delete $HOME/.kpatch/ /tmp/kpatch-archive/" \
    -t vmlinux

  [ ... $HOME/.kpatch/ may be removed ... ]

  ./kpatch-build/kpatch-build \
    --unpack-cmd "rsync --archive --delete /tmp/kpatch-archive/$HOME/.kpatch" \
    -t vmlinux \
    test/integration/centos-7/meminfo-string.patch

Suggested-by: Ilya A. Arkhipov ( M1cRO )

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -49,6 +49,7 @@ LOGFILE="$CACHEDIR/build.log"
 DEBUG=0
 SKIPCLEANUP=0
 SKIPGCCCHECK=0
+SAVEARCHIVE=0
 ARCH_KCFLAGS=""
 declare -a PATCH_LIST
 APPLIED_PATCHES=0
@@ -392,12 +393,14 @@ usage() {
 	echo "		-d, --debug        Enable 'xtrace' and keep scratch files" >&2
 	echo "		                   in <CACHEDIR>/tmp" >&2
 	echo "		                   (can be specified multiple times)" >&2
+	echo "		--save-cache       save archive of kpatch cache" >&2
+	echo "		--restore-cache    restore kpatch cache archive before building" >&2
 	echo "		--skip-cleanup     Skip post-build cleanup" >&2
 	echo "		--skip-gcc-check   Skip gcc version matching check" >&2
 	echo "		                   (not recommended)" >&2
 }
 
-options="$(getopt -o ha:r:s:c:v:j:t:n:o:d -l "help,archversion:,sourcerpm:,sourcedir:,config:,vmlinux:,jobs:,target:,name:,output:,debug,skip-gcc-check,skip-cleanup" -- "$@")" || die "getopt failed"
+options="$(getopt -o ha:r:s:c:v:j:t:n:o:d -l "help,archversion:,sourcerpm:,sourcedir:,config:,vmlinux:,jobs:,target:,name:,output:,debug,save-cache,restore-cache:,skip-gcc-check,skip-cleanup" -- "$@")" || die "getopt failed"
 
 eval set -- "$options"
 
@@ -455,6 +458,14 @@ while [[ $# -gt 0 ]]; do
 			echo "DEBUG mode enabled"
 		fi
 		;;
+	--save-cache)
+		SAVEARCHIVE=1
+		;;
+	--restore-cache)
+		[[ -z "$2" ]] && die "--restore-cache requires archive filename"
+		RESTOREARCHIVE="$2"
+		shift
+		;;
 	--skip-cleanup)
 		echo "Skipping cleanup"
 		SKIPCLEANUP=1
@@ -472,10 +483,23 @@ while [[ $# -gt 0 ]]; do
 	shift
 done
 
-if [[ ${#PATCH_LIST[@]} -eq 0 ]]; then
-	warn "no patch file(s) specified"
-	usage
-	exit 1
+if [[ $SAVEARCHIVE -ne 0 ]] ; then
+	if [[ -n "$RESTOREARCHIVE" ]] ; then
+		warn "--save-cache is incompatible with --restore-cache"
+		exit 1
+	fi
+
+	if [[ ${#PATCH_LIST[@]} -ne 0 ]]; then
+		warn "--save-cache is incompatible with input .patch files"
+		exit 1
+	fi
+else
+	# patch(es) required unless building a kpatch cache for archiving
+	if [[ ${#PATCH_LIST[@]} -eq 0 ]]; then
+		warn "no patch file(s) specified"
+		usage
+		exit 1
+	fi
 fi
 
 if [[ $DEBUG -eq 1 ]] || [[ $DEBUG -ge 3 ]]; then
@@ -564,6 +588,10 @@ if [[ -n "$USERSRCDIR" ]]; then
 
 	# save original vmlinux before it gets overwritten by sourcedir build
 	[[ "$VMLINUX" -ef "$SRCDIR"/vmlinux ]] && cp -f "$VMLINUX" "$TEMPDIR/vmlinux" && VMLINUX="$TEMPDIR/vmlinux"
+
+elif [[ -n "$RESTOREARCHIVE" ]]; then
+	echo "Restoring kpatch archive $RESTOREARCHIVE ..."
+	tar -C "$CACHEDIR" -zxf "$RESTOREARCHIVE"
 
 elif [[ -e "$SRCDIR"/.config ]] && [[ -e "$VERSIONFILE" ]] && [[ "$(cat "$VERSIONFILE")" = "$ARCHVERSION" ]]; then
 	echo "Using cache at $SRCDIR"
@@ -690,10 +718,12 @@ fi
 # unsupported kernel option checking: CONFIG_DEBUG_INFO_SPLIT
 grep -q "CONFIG_DEBUG_INFO_SPLIT=y" "$CONFIGFILE" && die "kernel option 'CONFIG_DEBUG_INFO_SPLIT' not supported"
 
-echo "Testing patch file(s)"
 cd "$SRCDIR" || die
-apply_patches
-remove_patches
+if [[ $SAVEARCHIVE -eq 0 ]]; then
+	echo "Testing patch file(s)"
+	apply_patches
+	remove_patches
+fi
 
 cp -LR "$DATADIR/patch" "$TEMPDIR" || die
 
@@ -710,8 +740,21 @@ if [[ $DEBUG -ge 4 ]]; then
 	export KPATCH_GCC_DEBUG=1
 fi
 
-echo "Building original kernel"
-./scripts/setlocalversion --save-scmversion || die
+if [[ -z "$RESTOREARCHIVE" ]]; then
+	echo "Building original kernel"
+	./scripts/setlocalversion --save-scmversion || die
+	unset KPATCH_GCC_TEMPDIR
+	# $TARGETS used as list, no quotes.
+	# shellcheck disable=SC2086
+	CROSS_COMPILE="$TOOLSDIR/kpatch-gcc " make "-j$CPUS" $TARGETS 2>&1 | logger || die
+fi
+
+if [[ $SAVEARCHIVE -ne 0 ]]; then
+	echo "Original kernel build in $CACHEDIR is complete, creating archive ..."
+	tar -C "$CACHEDIR" -zcf "${BASE}/kpatch-archive-${KVER}-${KREL}.tar.gz" .
+	echo "kpatch-archive-${KVER}-${KREL}.tar.gz saved."
+	exit 0
+fi
 unset KPATCH_GCC_TEMPDIR
 # $TARGETS used as list, no quotes.
 # shellcheck disable=SC2086


jpoimboe:
The interface is pretty difficult to use.  Does it really need to be that flexible?  What if we just have a `--save-cache` which spits out a tarball and `--restore-cache=<file>` which restores it?

joe-lawrence:
> The interface is pretty difficult to use. Does it really need to be that flexible? What if we just have a --save-cache which spits out a tarball and --restore-cache=<file> which restores it?

If the interface is difficult, what if we make these undocumented options and then add documented ones  (--save-cache, --restore-cache) that setup the script variables to do tarballs?  Or alternatively make these environment variables that can be overridden by the user (default is tarball).

The flexibility was added since I don't really know what the best time or space saving archive/unpack combination should be.  Some use cases might be interested in compression, some not.  Ditto for tarring up.  Maybe making an archive of ~/.ccache would be more productive than backing up the kpatch cache.  Etc.  I think it depends on what the kpatch-build sequence is going to be.  The open-ended versions make alternatives possible.

joe-lawrence:
> If we make the interfaces undocumented, they would still presumably be used, and we'd still have to support them. Does anybody actually need them?

Fair point.  What about the second alternative?  Override-able variables in the script, ie, `ARCHIVECMD="{ARCHIVECMD:-tar -Jcvf /tmp/kpatch-archive.tar.xz $HOME/.kpatch}`  I guess you don't get to specify a location for the tar file with this though.

> I'd rather not have to support hypothetical scenarios. Can we just assume sane defaults? Presumably, users of the interface are trying to save compile time/resources, so wouldn't a basic tar.gz or bz2 file be generally sufficient?

Situation 1:  I want to build kpatches 1 through 10 against the same kernel and I don't care about disk space.  Uncompressing the full cache archive ten times is a waste of time.  I just need to restore the delta back to the reference original.  I'm willing to sacrifice storage for time.

Situation 2: I want to build kpatch 1 against 10 kernels.  In this case, a tar archive makes sense.  The time penalty doesn't matter since I'm only using each cache once.

The latter situation is the most common, so I get it -- why make this more complicated than it needs to be.  But the hypothetical scenarios didn't seem too far fetched IMHO.  /two-cents

jpoimboe:
> Fair point. What about the second alternative? Override-able variables in the script, ie, ARCHIVECMD="{ARCHIVECMD:-tar -Jcvf /tmp/kpatch-archive.tar.xz $HOME/.kpatch} I guess you don't get to specify a location for the tar file with this though.

Will you actually use that interface though?  If not then I don't think we should expose it.

> Situation 1: I want to build kpatches 1 through 10 against the same kernel and I don't care about disk space. Uncompressing the full cache archive ten times is a waste of time. I just need to restore the delta back to the reference original. I'm willing to sacrifice storage for time.

(Are you thinking about the integration tests?  If so, I'm not sure we'll need to build the tests individually anymore; in the future I'm thinking the single combined test would be good enough?)

But anyway, the vast majority of (well written) kpatch patches only touch a few files, and don't touch any common header files, so there wouldn't be much of a rebuild between patches.  So ~/.kpatch would be largely the same after the first build.  So in this scenario, wouldn't it be good enough to just use what's already in ~/.kpatch after the previous build, instead of restoring the cache every time?

Even if that's not enough, somebody could just use rsync manually, outside of kpatch-build, to keep ~/.kpatch in sync with the original cache.  That would be pretty easy to script up.  But this seems like an esoteric case anyway and is way outside the scope of kpatch-build, IMO.

M1cRO:
I've using kpatch-build with koji and I need doing many building for same kernel and it always rebuilding kernel ;( For new CVE you should againg rebuild all kernels... It take so much time.
I've building kpatch modules by spec file and unpack looks very friendly for me.

M1cRO:
Anyway kpatch-build should be ready for using already builded kernel.

M1cRO:
Any updates/ideas?

joe-lawrence:
Hi @M1cRO , sorry for getting away from this, I was distracted by other work.

> (jpoimboe) Will you actually use that interface though? If not then I don't think we should expose it.
> (Are you thinking about the integration tests? If so, I'm not sure we'll need to build the tests individually anymore; in the future I'm thinking the single combined test would be good enough?)

That was the first thought that popped into my head, however I will admit to having to run kpatch-build for the same patch a few times (typos, missing header files, etc.) in which I would be running against the same kernel more than once.  Also handy sometimes when doing kpatch development.

> But anyway, the vast majority of (well written) kpatch patches only touch a few files, and don't touch any common header files, so there wouldn't be much of a rebuild between patches. So ~/.kpatch would be largely the same after the first build. So in this scenario, wouldn't it be good enough to just use what's already in ~/.kpatch after the previous build, instead of restoring the cache every time?

This is probably right for the majority of cases.  For patches that do touch the same file(s), or repeat builds, those files need to be put back into place or else the patch verification fails (ie, already patched).

> Even if that's not enough, somebody could just use rsync manually, outside of kpatch-build, to keep ~/.kpatch in sync with the original cache. That would be pretty easy to script up. But this seems like an esoteric case anyway and is way outside the scope of kpatch-build, IMO.

This is fact what I do sometimes, though I have to manually edit the kpatch-build script so it stops after the reference kernel build and I can copy the cache myself.

If I'm the only one who would be using the open-ended interface, then I'd rather not complicate matters for something I'd use only sporadically.

aliceinwire:
looks like an interesting feature to have

jpoimboe:
> > (jpoimboe) Will you actually use that interface though? If not then I don't think we should expose it.
> (Are you thinking about the integration tests? If so, I'm not sure we'll need to build the tests individually anymore; in the future I'm thinking the single combined test would be good enough?)
> 
> That was the first thought that popped into my head, however I will admit to having to run kpatch-build for the same patch a few times (typos, missing header files, etc.) in which I would be running against the same kernel more than once. Also handy sometimes when doing kpatch development.

In that case, isn't the pre-existing ~/.kpatch cache sufficient?  (assuming you're not patching include/linux/kernel.h or something)

> > But anyway, the vast majority of (well written) kpatch patches only touch a few files, and don't touch any common header files, so there wouldn't be much of a rebuild between patches. So ~/.kpatch would be largely the same after the first build. So in this scenario, wouldn't it be good enough to just use what's already in ~/.kpatch after the previous build, instead of restoring the cache every time?
> 
> This is probably right for the majority of cases. For patches that do touch the same file(s), or repeat builds, those files need to be put back into place or else the patch verification fails (ie, already patched).

Hm?  kpatch-build un-applies patches automatically at the end.

> > Even if that's not enough, somebody could just use rsync manually, outside of kpatch-build, to keep ~/.kpatch in sync with the original cache. That would be pretty easy to script up. But this seems like an esoteric case anyway and is way outside the scope of kpatch-build, IMO.
> 
> This is fact what I do sometimes, though I have to manually edit the kpatch-build script so it stops after the reference kernel build and I can copy the cache myself.
> 
> If I'm the only one who would be using the open-ended interface, then I'd rather not complicate matters for something I'd use only sporadically.

But with my proposed interface, you wouldn't have to manually edit kpatch-build.  You could
- kpatch-build --save-cache (you could ignore the outputed tarball as this would leave ~/.kpatch build in its "original" built state)
- rsync ~/.kpatch <otherdir>
- kpatch-build foo.patch
- rsync <otherdir> ~/.kpatch
- kpatch-build bar.patch

sm00th:
> -   kpatch-build --save-cache (you could ignore the outputed tarball as this would leave ~/.kpatch in its "original" built state)
> -  rsync ~/.kpatch <otherdir>
> -  kpatch-build foo.patch
> -  rsync <otherdir> ~/.kpatch
> -  kpatch-build bar.patch

Actual patch-build steps need an option to skip `make mrproper` though. I think that an option like this alone (in conjunction with existing `--skip-cleanup`) covers a lot of usecases because as you said average patch touches small amount of files, so building unpatched kernel on top of pre-built patched objectfiles should give us substantial boost anyway. This is a bare minimum that we need, adding --save-cache can save some trouble, especially when kernel sources are taken from rpm/deb. The rest should be outside of kpatch-build imo, it can be done as shell-scripts in contrib dir for example.

jpoimboe:
> Actual patch-build steps need an option to skip make mrproper though. I think that an option like this alone (in conjunction with existing --skip-cleanup) covers a lot of usecases because as you said average patch touches small amount of files, so building unpatched kernel on top of pre-built patched objectfiles should give us substantial boost anyway. This is a bare minimum that we need, adding --save-cache can save some trouble, especially when kernel sources are taken from rpm/deb. The rest should be outside of kpatch-build imo, it can be done as shell-scripts in contrib dir for example.

Hm, I completely forgot about mrproper.  Do we even need to do that between builds?  Removing that could really speed things up (especially for people who don't have ccache).

Also, why do you think `--skip-cleanup` would help?  All it does is unapply the patch and remove the temporary patch objects.

sm00th:
> Also, why do you think --skip-cleanup would help? All it does is unapply the patch and remove the temporary patch objects.

Ohhh... for some reason I thought that kpatch-builds keeps all objfiles in $TEMPDIR which it does not (yet), but it looks like with `--save-cache` it will.

joe-lawrence:
In reviewing this, I think a bunch of my complaints stem from using `--skip-cleanup` so I could inspect the temporary object files.  While I still prefer an open-ended `eval` on a user provided archive/restore command for flexibility, I can see how the general use case is straightforward.  So to keep it simple for the user, where are we with this feature?  Do we want `--save-cache` and `--restore-cache file` options?  And what sane defaults do we want to use (xz / gzip / bzip2, compression levels, etc?)

jpoimboe:
> In reviewing this, I think a bunch of my complaints stem from using --skip-cleanup so I could inspect the temporary object files.

BTW, I always use '-d' because it's more user friendly than '--skip-cleanup'.  It leaves the objects in ~/.kpatch/tmp and also unpatches the source files.

> While I still prefer an open-ended eval on a user provided archive/restore command for flexibility, I can see how the general use case is straightforward. So to keep it simple for the user, where are we with this feature? Do we want --save-cache and --restore-cache file options?

The simpler --save-cache/--restore-cache interface sounds good to me, if there are no objections.

> And what sane defaults do we want to use (xz / gzip / bzip2, compression levels, etc?)

I would say just use gzip (or .tar.gz), as it's nice and fast, and I think speed is the primary goal.

joe-lawrence:
v2:
* Convert the `archive-cmd` code into `save-cache`
  * no pathname, tarball will automatically be named `kpatch-archive-${KVER}-${KREL}.tar.gz` in current directory.
  * not compatible with input patch files since we only build a reference kernel for caching
* Convert the `unpack-cmd` code into `restore-cache`
  * accepts pathname argument

I could add a pathname argument to `save-cache` if that is preferred.

Also, would it be worth filtering out some files in the archive (maybe vmlinux.o, .tmp_*, etc?)

kpatch-build/kpatch-build
```diff
@@ -693,14 +723,23 @@ if [[ $DEBUG -ge 4 ]]; then
...
-CROSS_COMPILE="$TOOLSDIR/kpatch-gcc " make "-j$CPUS" $TARGETS 2>&1 | logger || die
+if [[ -z "$RESTOREARCHIVE" ]]; then
+	echo "Building original kernel"
+	./scripts/setlocalversion --save-scmversion || die
+	make mrproper 2>&1 | logger || die
+	cp -f "$CONFIGFILE" "$SRCDIR/.config"
+	unset KPATCH_GCC_TEMPDIR
+	# $TARGETS used as list, no quotes.
+	# shellcheck disable=SC2086
+	CROSS_COMPILE="$TOOLSDIR/kpatch-gcc " make "-j$CPUS" $TARGETS 2>&1 | logger || die
```
sm00th: I think we'll either have to use `O=` here or tar $USERSRCDIR for this to work properly with `-s`. If `O=` approach is taken (which I think is a more universal one) actual make command should probably be executed even if we are restoring just to make sure source haven't changed. I have a specific usecase in mind: I build a lot against linus' tree which is constantly updated and with described approach it will be possible to partially use the cache even though the kernel version is different. Because of this case I also don't think that `--restore-cache` and `--save-cache` should be mutually exclusive. I can take on that in a later pullrequest though if you prefer it this way.

jpoimboe: @sm00th Would removing the "make mrproper" solve your problem? I don't see why the mrproper is needed, and it would be very useful if somebody figured that out, as that seems to be one of the main motivations for this feature.

Also have you tried ccache, or does that not help for some reason?

jpoimboe: BTW, the use of `O=` is problematic, see 2e99d6b7a44876a4368e7cf7ef6e75f203728723.

sm00th:
> Would removing the "make mrproper" solve your problem?

Yes, that should solve it. With this approach `-s` would be incompatible with `--save/restore-cache`, right?

>  I don't see why the mrproper is needed, and it would be very useful if somebody figured that out, as that seems to be one of the main motivations for this feature.

To me it looks like mrproper only does something in `-s` mode which is kind of an "on your own risk" mode, so we probably can replace mrproper with a note in README or at the start of kpatch-build when this option is used.

> Also have you tried ccache, or does that not help for some reason?

I am using ccache and it helps a lot. However the i/o time a kernel build takes to recreate objectfiles (especially on bigger configs) is non-negligible. An SSD would probably be another solution to the problem.

> BTW, the use of O= is problematic, see 2e99d6b.

I see. I suspected that the project might have ran into issues with it before.

sm00th:
> I don't see why the mrproper is needed

Ah, I just realised mrproper makes sure we rebuild the kernel with proper CFLAGS, i.e. with -ffunction-sections.

jpoimboe:
> Ah, I just realised mrproper makes sure we rebuild the kernel with proper CFLAGS, i.e. with -ffunction-sections.

That shouldn't be necessary because the kernel makefiles detect and rebuild on CFLAG changes.  And also the original and patched kernels are both built with -ffunction-sections, so there's no need to do it between rebuilds.

mrproper _might_ be needed after first setting up a new ~/.kpatch/src dir, but I still don't see why it's needed every time.

joe-lawrence:
What's the status for v2?  I didn't really follow the "make mrproper" subthread -- could that could be handled in a follow-up?

jpoimboe:
I would rather fix the mrproper issue first, because fixing that (if possible) would fix at least some (or even all?) of the motivations[*] for this feature.  And I hope it will help provide more clarity about what exactly the real world use cases are, so we can know what it is we're trying to fix or improve.

[*] The two motivations I remember seeing are
1. building multiple patches against the same kernel -- removing mrproper should make this faster, so would this feature still be used here?
2. building the same patch against multiple kernels -- removing mrproper might allow the ~/.kpatch dir to be manually saved/restored/rsynced from outside the kpatch-build script.  would that be sufficient?

M1cRO:
Can't anybody give answer about mrproper?
We can do one more option for enable/disable mrproper or open one more issue for it?

sm00th:
> I just tested back-to-back kernel builds, modifying CFLAGS in between, and can confirm that the second build does indeed rebuild all the objects.

Yes it does. Forgot to reply to Josh's message but I looked into it and it is a neat feature of kbuild I didn't know about: it saves the full build command for every generated file in a `.${FILE}.cmd` and if anything changes, including any flags, rebuilds it.

joe-lawrence:
> Yes it does. Forgot to reply to Josh's message but I looked into it and it is a neat feature of kbuild I didn't know about: it saves the full build command for every generated file in a .${FILE}.cmd and if anything changes, including any flags, rebuilds it.

Does that leave us with any requirement to run `make mrproper` then?

jpoimboe:
> When you say "new" ~/.kpatch/src dir, do you mean a completely fresh one, or if I were to build for kernel X, then build for kernel Y?

Yeah.

> Does that leave us with any requirement to run make mrproper then?

Not that I know of...

M1cRO:
Have you any plan to merge it?
Or we still have some not resolved issues?

joe-lawrence:
v3: rebased to incorporate f32363e634a6e81a82b52d049f77a3b85ce1e104 ("kpatch-build: don't do "make mrproper" on every build"), but not tested

joe-lawrence:
This PR has sat for nearly a year without further comment.  I don't have a strong need for this feature any more, so I'm inclined to close/cancel it... Does any one still care?

M1cRO:
Ok, close it. Anyway I using patch with different logic.

Issue #836, opened 2018-04-19
function pointers to global functions don't get converted to dynrelas


jpoimboe:
As discovered in #834.

jpoimboe:
@sm00th After looking at this bug, I'm starting to think that our unit test hacks aren't sufficient.  Instead of having fake vmlinux and Module.symvers files, I think we will need the real things, to properly test the dynrela code.  Can you update the unit test framework accordingly?

I guess we'll need to target our unit tests against 4.16 (or similar kernels) for now.  Maybe the tests could be moved to a 4.16 subdir of the x86_64 dir, or something similar, so we can eventually add new tests for later kernels.  The 4.16 subdir can have a stripped vmlinux and Module.symvers in it, along with a file containing the values of PARA_STRUCT_SIZE and friends.

git-lfs space may become an issue, so it would also be good to add a check to ensure vmlinux is stripped. 

Also I think we can stay on 4.16 for a while.  Even test binaries built against later kernels will probably work with the old vmlinux for quite a while.

sm00th:
I assume config used to build vmlinux/Module.symvers does matter? Should it be f27 config + olddefconfig or something else?

jpoimboe:
BTW, one possible issue is that some of the tests might be against modules instead of vmlinux.  If so, I'm not sure the best way to deal with those.

sm00th:
> BTW, one possible issue is that some of the tests might be against modules instead of vmlinux. If so, I'm not sure the best way to deal with those.

I think it won't be much of a problem to come up with a solution for modules, but I think we need to draw a line somewhere.

Stripped vmlinux I got is 53Mb + 5Mb of objectfiles and we get ~17 downloads a month. And thats just x86_64. Modules, while much smaller, still may add up to a lot.
Another thing: I don't think there is a way to purge old binaries from lfs, so updating vmlinux even during pullrequest review would mean using twice the storage and once we reach 1Gb there will be no way of cleaning old stuff up.

I propose that the tests that would need to be built against modules (or even against real vmlinux if you don't expect there to be many of those) be moved to integration tests rather than unit tests.

jpoimboe:
Hm, I forgot about the download quota.  Maybe a 53M vmlinux file isn't such a good idea then.  Especially looking forward where the unit tests support multiple kernels and arches.

I think all we _really_ need is the symbol table, so maybe we need to change create-diff-object somehow so that it can take a symbol table as input.  Will think about it a little more.

[...]

PR #849, opened 2018-05-15
create-diff-object symtab

sm00th:
As discussed in #836 this pull request is a merge of mine and @jpoimboe's attempts to change create-diff-object to take symtable (`eu-readelf -s` output) as an input instead of an object-file.  This will make unit-tests more flexible and will allow us to cover more cases with them.

There is something to discuss though: when I tried to get a symtab that would work with preexisting test objectfiles I couldn't as I lost the config I used. Even if I rebuild all of those now so that they all work with a single vmlinux.symtab - it would be unreasonable to expect other contributors to rebuild their objectfiles using a specific config/kernel version. Should we just always use per-test symtab(and possibly symvers) files then? As I understand in most cases they can be generated from .ORIG.o (we can even do it on the fly) and only in rare cases we'll need to use vmlinux/parent module symtab.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -773,16 +773,26 @@ for i in $FILES; do
 	mkdir -p "output/$(dirname "$i")"
 	cd "$SRCDIR" || die
 	find_kobj "$i"
-	if [[ "$KOBJFILE" = vmlinux ]]; then
-		KOBJFILE="$VMLINUX"
-	else
-		KOBJFILE="$TEMPDIR/module/$KOBJFILE"
-	fi
 	cd "$TEMPDIR" || die
 	if [[ -e "orig/$i" ]]; then
-		# create-diff-object orig.o patched.o kernel-object output.o Module.symvers patch-mod-name
-		"$TOOLSDIR"/create-diff-object "orig/$i" "patched/$i" "$KOBJFILE" \
-			"output/$i" "$SRCDIR/Module.symvers" "${MODNAME//-/_}" 2>&1 | logger 1
+		if [[ "$KOBJFILE" = vmlinux ]]; then
+			KOBJFILE_NAME=vmlinux
+			KOBJFILE_PATH="$VMLINUX"
+			SYMTAB="${TEMPDIR}/${KOBJFILE_NAME}.symtab"
+		else
+			KOBJFILE_NAME=$(basename "${KOBJFILE%.ko}")
+			KOBJFILE_NAME="${KOBJFILE_NAME/-/_}"
+			KOBJFILE_PATH="${TEMPDIR}/module/$KOBJFILE"
+			SYMTAB="${KOBJFILE_PATH}.symtab"
+		fi
+
+		eu-readelf -s "$KOBJFILE_PATH" > "$SYMTAB"
+
+		# create-diff-object orig.o patched.o parent-name parent-symtab
+		#		     Module.symvers patch-mod-name output.o
+		"$TOOLSDIR"/create-diff-object "orig/$i" "patched/$i" "$KOBJFILE_NAME" \
+			"$SYMTAB" "$SRCDIR/Module.symvers" "${MODNAME//-/_}" \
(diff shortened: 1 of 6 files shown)


jpoimboe:
Yeah, let's use a per-test symtab/symvers.  If the symtab doesn't exist, it can be auto-generated and if the symvers doesn't exist, it can be /dev/null.

kpatch-build/kpatch-build
```diff
@@ -773,16 +773,25 @@ for i in $FILES; do
...
-		# create-diff-object orig.o patched.o kernel-object output.o Module.symvers patch-mod-name
-		"$TOOLSDIR"/create-diff-object "orig/$i" "patched/$i" "$KOBJFILE" \
-			"output/$i" "$SRCDIR/Module.symvers" "${MODNAME//-/_}" 2>&1 | logger 1
+		if [[ "$KOBJFILE" = vmlinux ]]; then
+			KOBJFILE_NAME=vmlinux
+			KOBJFILE_PATH="$VMLINUX"
+		else
+			KOBJFILE_NAME="${KOBJFILE%.ko}"
+			KOBJFILE_NAME="${KOBJFILE_NAME/-/_}"
+			KOBJFILE_PATH="module/$KOBJFILE"
```
jpoimboe: I think I wrote some of this code but now it looks buggy.  Shouldn't this be `$TEMPDIR/module/$KOBJFILE`?

kpatch-build/kpatch-build
```diff
@@ -773,16 +773,25 @@ for i in $FILES; do
...
+		fi
+
+		SYMTAB="$KOBJFILE_PATH.symtab"
```
jpoimboe: In the vmlinux case, without -s, $KOBJFILE_PATH is going to point out to a non-writeable path.  Instead it should be written somewhere in $TEMPDIR.

kpatch-build/kpatch-build
```diff
@@ -773,16 +773,25 @@ for i in $FILES; do
...
+		eu-readelf -s "$KOBJFILE" > "$SYMTAB"
```
jpoimboe: I think this should be $KOBJFILE_PATH instead of $KOBJFILE.

kpatch-build/kpatch-build
sm00th: Will $TEMPDIR/module/$KOBFILE work in non `-s` case?

sm00th:
v2:
  - switched to per-test symtab/symvers
  - added now-required elfutils to travis
  - KOBJFILE_NAME passed through basename to strip path
  - KOBJFILE_PATH explicitly in $TEMPDIR
  - SYMTAB explicitly in $TEMPDIR

kpatch-build/kpatch-build
```diff
@@ -773,16 +773,25 @@ for i in $FILES; do
...
+		if [[ "$KOBJFILE" = vmlinux ]]; then
+			KOBJFILE_NAME=vmlinux
+			KOBJFILE_PATH="$VMLINUX"
+		else
+			KOBJFILE_NAME=$(basename "${KOBJFILE%.ko}")
+			KOBJFILE_NAME="${KOBJFILE_NAME/-/_}"
+			KOBJFILE_PATH="${TEMPDIR}/module/$KOBJFILE"
+		fi
+
+		SYMTAB="${TEMPDIR}/${KOBJFILE_NAME}.symtab"
```
jpoimboe: This could overwrite another symtab if there are two modules being patched which have the same name (but are in different directories).  Wouldn't break anything, but could make debugging confusing.

Maybe set SYMTAB conditionally.  In the vmlinux case it could be `$TEMPDIR/vmlinux.symtab`, and in the module case it could be `$KOBJFILE.symtab`.

sm00th:
v3:
 - use different symtab paths for vmlinux/module cases: `${TEMPDIR}/${KOBJFILE_NAME}.symtab` in vmlinux case; `${KOBJFILE_PATH}.symtab` in module case where `${KOBJFILE_PATH}` is explicitly in `${TEMPDIR}`

PR #859, opened 2018-05-29
create-diff-object: don't convert .TOC. references to dynrelas

jpoimboe:
When I made a patch to the nfsd module on a ppc64le system with a RHEL 7
based kernel, livepatch prevented the target module from loading with:

  livepatch: symbol '.TOC.' not found in symbol table

References to this symbol are treated specially by the kernel module
loader, so references to it should never be converted to dynrelas.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2467,8 +2467,17 @@ static int function_ptr_rela(const struct rela *rela)
 
 static int may_need_dynrela(const struct rela *rela)
 {
+	/*
+	 * References to .TOC. are treated specially by the module loader and
+	 * should never be converted to dynrelas.
+	 */
+	if (rela->type == R_PPC64_REL16_HA || rela->type == R_PPC64_REL16_LO ||
+	    rela->type == R_PPC64_REL64)
+		return 0;
+
 	if (!rela->sym->sec)
 		return 1;
+
 	/*
 	 * Nested functions used as callbacks are a special case.
 	 * They are not supposed to be visible outside of the


kamalesh-babulal:
Hi @jpoimboe What do you think of the below patch. `.TOC.` symbol is generally accompanied by an empty symbol of relocation type `R_PPC64_ENTRY`as part of the function prologue sequence generated by GCC >= 6. Move the check for `.TOC.` symbol of relocation type `R_PPC64_REL64` along with empty symbol relocation type check and mark them normal rela.
```
index 451a21c..747b3d9 100644
@@ -2604,9 +2604,10 @@ static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
                         *      .localentry my_func, .-my_func
                         *
                         * The R_PPC64_ENTRY is optional and its symbol might
-                        * have an empty name.  Leave it as a normal rela.
+                        * have an empty name.  .TOC. is of R_PPC64_REL64
+                        * relocation type, leave them as a normal rela.
                         */
-                       if (rela->type == R_PPC64_ENTRY)
+                       if (rela->type == R_PPC64_ENTRY || rela->type == R_PPC64_REL64)
                                continue;
 
                        if (rela->sym->bind == STB_LOCAL) {
```

jpoimboe:
> Are those relocation types used for anything else in the kernel?

Answering my own question, it doesn't look like they're used for anything else.

jpoimboe:
v2: change check to PPC64_REL16_HA or PPC64_REL16_LO or R_PPC64_REL64.

kpatch-build/create-diff-object.c
```diff
@@ -2467,8 +2467,17 @@ static int function_ptr_rela(const struct rela *rela)
 
 static int may_need_dynrela(const struct rela *rela)
 {
+	/*
+	 * References to .TOC. are treated specially by the module loader and
+	 * should never be converted to dynrelas.
+	 */
+	if (rela->type == PPC64_REL16_HA || rela->type PPC64_REL16_LO ||
```
kamalesh-babulal: > Yes, they are not used for other relocations. The v2 looks good, '==' missing in above check.

jpoimboe: Doh, guess I should have at least compiled it.

jpoimboe:
v3: fix compile issues

PR #863, opened 2018-05-30
test/unit: run submodule command from base git directory

jpoimboe:
This fixes the following error with an older version of git
(1.8.3.1):

  make -C test/unit
  make[1]: Entering directory `/root/kpatch/test/unit'
  git submodule update --init --rebase
  You need to run this command from the toplevel of the working tree.

### DIFF
--- a/test/unit/Makefile
+++ b/test/unit/Makefile
@@ -4,7 +4,7 @@ OBJDIR ?= objs/$(ARCH)
 .PHONY: all clean
 
 all: Makefile.include
-	git submodule update --init --rebase
+	cd $(shell git rev-parse --show-toplevel) && git submodule update --init --rebase
 	$(MAKE) -C $(OBJDIR)
 
 clean: Makefile.include


test/unit/Makefile
```diff
@@ -4,7 +4,7 @@ OBJDIR ?= objs/$(ARCH)
 .PHONY: all clean
 
 all: Makefile.include
-	git submodule update --init --rebase
+	(cd ../.. && git submodule update --init --rebase)
 	$(MAKE) -C $(OBJDIR)
```
joe-lawrence: `git rev-parse --show-toplevel` will also give you the top level directory if you want to get fancy.  Also is the extra '(' shell necessary?

jpoimboe: Good points

jpoimboe:
v2:
- use `git rev-parse --show-toplevel`
- remove extra subshell

joe-lawrence:
:+1:

PR #864, opened 2018-05-31
livepatch-patch-hook: fix non-rhel builds

sm00th:
Preprocessor doesn't stop unwinding macroses in #if clauses if one of
the conditions is false resulting in the following error when trying to
build on non-rhel system:

patch/livepatch-patch-hook.c:53:48: error: missing binary operator before token "("
       RHEL_RELEASE_CODE <= RHEL_RELEASE_VERSION(7, 5))

Fix by using 2 levels of #ifs instead of &&. The same way HAVE_CALLBACKS
does.


sm00th:
I've only tested this on non-rhel, so if anyone have a rhel setup ready it would be nice if you could check if it successfully builds there as well.

jpoimboe:
BTW, it builds fine on RHEL.

joe-lawrence:
> BTW, it builds fine on RHEL.

Was waiting for a RHEL-build to finish, but you beat me to it. :)

PR #868, opened 2018-06-01
testing: kpatch-build should set all module names

### LINKED ISSUE #867: test: kpatch-test assumes "kpatch-" module prefix
c0105ea46774 ("kpatch-build: set default module prefix accordingly") modified kpatch-build to set a "kpatch-" or "livepatch-" module filename prefix depending on its type.  The `test/integration/kpatch-test` script still assumes that the modules that it generates are prefixed by "kpatch-".

joe-lawrence:
Since c0105ea46774 ("kpatch-build: set default module prefix
accordingly") kpatch-build has automatically named resulting patch
modules with a "kpatch-" or "livepatch-" prefix (depending on its type).

To avoid naming confusion, have kpatch-test name all of its patch
modules with a "test-" prefix.

Fixes #867.

### DIFF
--- a/test/integration/kpatch-test
+++ b/test/integration/kpatch-test
@@ -133,7 +133,8 @@ unload_all() {
 build_module() {
 	file=$1
 	prefix=$(basename ${file%%.patch})
-	module=kpatch-$prefix.ko
+	modname="test-$prefix"
+	module="${modname}.ko"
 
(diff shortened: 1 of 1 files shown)


joe-lawrence:
Tested on RHEL7.5 with a few of the integration test patches (three or four just to test the individual module and combined module logic).

PR #870, opened 2018-06-02
kpatch-build: RHEL kernel-alt support

jpoimboe:
Deal with a few RHEL kernel-alt quirks for ppc64le:

- The RPM and spec names are "kernel-alt".

- 7.6 ALT is based on 4.14 but it doesn't have the 'immediate' flag.


jpoimboe:
v2:
- removed quotes around regex

PR #873, opened 2018-06-02
kpatch-build: detect truncated Module.symvers

### LINKED ISSUE #589: kpatch-build with -t argument only builds patch modules with partial Module.symvers
Documenting the bug found in issue #587, since the `-t` target argument only compiles the relevant portion of the kernel, only a partial Module.symvers gets generated, which is then compiled with the patch module. The full Module.symvers needs to be used instead.

jpoimboe:
"make mrproper" combined with the '-t' flag is dangerous, as it results
in the Module.symvers file getting truncated, which causes
create-diff-object to create some funky dynrelas.  Detect this condition
in kpatch-build and error out.

We will hopefully also be removing "make mrproper" soon, which will make
'-t' even more useful.

Fixes #589.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -724,6 +724,8 @@ if [[ ! -e "$TEMPDIR/changed_objs" ]]; then
 	die "no changed objects found"
 fi
 
+grep -q vmlinux "$SRCDIR/Module.symvers" || die "truncated $SRCDIR/Module.symvers file"
+
 # Read as words, no quotes.
 # shellcheck disable=SC2013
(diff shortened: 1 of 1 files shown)


jpoimboe:
> Is vmlinux is supposed to be the last symbol in the file?

I don't think so, the Module.symvers order seems pretty random.  Why?

joe-lawrence:
>>  Is vmlinux is supposed to be the last symbol in the file?

> I don't think so, the Module.symvers order seems pretty random. Why?

Just curious.  If it were at the end of the file, it's absence would imply truncation ... the test right now just verifies that vmlinux is present somewhere in the symvers file.  (Good enough for our purposes.)

Issue #866, opened 2018-06-01
gcc8: parainstructions-section: .text -> .text.unlikely movements cause build failure


sm00th:
parainstructions-section.patch from integration tests fails to build with gcc8 and 4.17-rc6 with the following error:
```
ERROR: generic.o: symbol changed sections: proc_register                             
/home/asavkov/dev/kpatch/kpatch-build/create-diff-object: unreconcilable difference  
```
Due to optimizations patched function's (`proc_alloc_inum()`) section got renamed to `.text.unlikely.*`. Same happened to a bunch of functions calling `proc_allic_inum()`. Most of them were just renames, while some retained original `.text.` sections separating unlikely code into a `.cold.*` symbol in `.text.unlikely.` sections.

Issue #865, opened 2018-05-31
gcc8: special-static-2: function has no fentry/mcount call, unable to patch


sm00th:
special-static-2.patch from integration tests fails to build with gcc8 and 4.17-rc6 with the following error:

```
x86.o: function kvm_set_msr_common.cold.140 has no fentry/mcount call, unable to patch 
ERROR: x86.o: 1 function(s) can not be patched                                         
/home/asavkov/dev/kpatch/kpatch-build/create-diff-object: unreconcilable difference
```

This is something @jpoimboe mentioned in #856, quoting here for convenience:

> I ran into these GCC 8 cold subfunctions with objtool. I suspect there are more kpatch-build issues to fix related to GCC 8, because the cold subfunctions aren't real patchable functions; they're instead an extension of the non-cold function. So if the cold text changes, I bet kpatch-build will fail because it won't recognize that it's a part of the parent function.

jpoimboe:
Yeah, we'll need to figure out a way to treat the subfunction as an extension of the function.  So if the subfunction changes, the function itself should be patched, with the new subfunction also included.

PR #881, opened 2018-06-13
gcc8 fixes

sm00th:
This is my attempt at fixing #865 and #866. I tried to keep the commits small so that it would be easier to follow and fix if some of the approaches are sub-optimal.

Build/load/unload/run tested with [unlikely-likely.patch.txt](https://github.com/dynup/kpatch/files/2097337/unlikely-likely.patch.txt) on top of 4.17 which covers both cases. Also build-tested with all of the integration-tests we have and it now builds all of the patches gcc7 does successfully.

Note: if the patch causes a function to change subsections (.e.g. .text -> .text.unlikely) it will be marked as CHANGED and included in the output even if the actual contents of the function were not changed. This makes sense to me but can be a source of confusion I guess.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2296,6 +2296,10 @@ static void kpatch_mark_ignored_functions_same(struct kpatch_elf *kelf)
 			log_normal("NOTICE: no change detected in function %s, unnecessary KPATCH_IGNORE_FUNCTION()?\n", rela->sym->name);
 		rela->sym->status = SAME;
 		rela->sym->sec->status = SAME;
+
+		if (rela->sym->child)
+			rela->sym->child->status = SAME;
+
 		if (rela->sym->sec->secsym)
 			rela->sym->sec->secsym->status = SAME;
 		if (rela->sym->sec->rela)
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -644,8 +644,14 @@ static void kpatch_compare_sections(struct list_head *seclist)
 			if (sec->base->sym && sec->base->sym->status != CHANGED)
 				sec->base->sym->status = sec->status;
 		} else {
-			if (sec->sym && sec->sym->status != CHANGED)
-				sec->sym->status = sec->status;
+			struct symbol *sym = sec->sym;
+
+			if (sym && sym->status != CHANGED)
+				sym->status = sec->status;
+
+			if (sym && sym->child && sym->status == SAME &&
+			    sym->child->sec->status == CHANGED)
+				sym->status = CHANGED;
 		}
 	}
 }
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -184,6 +184,38 @@ static void kpatch_bundle_symbols(struct kpatch_elf *kelf)
 }
 
 /*
+ * During optimization gcc may move unlikely execution branches into *.cold
+ * subfunctions. kpatch_detect_child_functions detects such subfunctions and
+ * crossreferences them with their parent functions through parent/child
+ * pointers.
+ */
+static void kpatch_detect_child_functions(struct kpatch_elf *kelf)
+{
+	struct symbol *sym;
+
+	list_for_each_entry(sym, &kelf->symbols, list) {
+		char *coldstr;
+
+		coldstr = strstr(sym->name, ".cold.");
+		if (coldstr != NULL) {
+			char *pname;
+
+			pname = strndup(sym->name, coldstr - sym->name);
+			if (!pname)
+				ERROR("strndup");
+
+			sym->parent = find_symbol_by_name(&kelf->symbols, pname);
+			free(pname);
+
+			if (!sym->parent)
+				ERROR("failed to find parent function for %s", sym->name);
+
+			sym->parent->child = sym;
+		}
+	}
+}
+
+/*
  * This function detects whether the given symbol is a "special" static local
  * variable (for lack of a better term).
  *
@@ -3162,6 +3194,9 @@ int main(int argc, char *argv[])
 
 	kpatch_bundle_symbols(kelf_base);
 	kpatch_bundle_symbols(kelf_patched);
+
+	kpatch_detect_child_functions(kelf_base);
+	kpatch_detect_child_functions(kelf_patched);
 
 	kpatch_compare_elf_headers(kelf_base->elf, kelf_patched->elf);
 	kpatch_check_program_headers(kelf_base->elf);
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -102,6 +102,11 @@ static int is_bundleable(struct symbol *sym)
 			  !strncmp(sym->sec->name + 15, sym->name, strlen(sym->sec->name) - 15))))
 		return 1;
 
+	if (sym->type == STT_FUNC &&
+	    !strncmp(sym->sec->name, ".text.hot.",10) &&
+	    !strcmp(sym->sec->name + 10, sym->name))
+		return 1;
+
 	if (sym->type == STT_OBJECT &&
 	   !strncmp(sym->sec->name, ".data.",6) &&
 	   !strcmp(sym->sec->name + 6, sym->name))
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -68,6 +68,12 @@
 
 char *childobj;
 
+enum subsection {
+	SUBSECTION_NORMAL,
+	SUBSECTION_HOT,
+	SUBSECTION_UNLIKELY
+};
+
 enum loglevel loglevel = NORMAL;
 
 /*******************
@@ -656,6 +662,22 @@ static void kpatch_compare_sections(struct list_head *seclist)
 	}
 }
 
+static enum subsection kpatch_subsection_type(struct section *sec)
+{
+	if (!strncmp(sec->name, ".text.unlikely.", 15))
+		return SUBSECTION_UNLIKELY;
+
+	if (!strncmp(sec->name, ".text.hot.", 10))
+		return SUBSECTION_HOT;
+
+	return SUBSECTION_NORMAL;
+}
+
+static int kpatch_subsection_changed(struct section *sec1, struct section *sec2)
+{
+	return kpatch_subsection_type(sec1) != kpatch_subsection_type(sec2);
+}
+
 static void kpatch_compare_correlated_symbol(struct symbol *sym)
 {
 	struct symbol *sym1 = sym, *sym2 = sym->twin;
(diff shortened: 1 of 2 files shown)


sm00th:
> Do you expect many more parent/child types than just ".cold." ?

I am yet to meet it in the wild and failing to navigate gcc's sources properly but I suspect there might be a '.hot.' counterpart, or at least it is possible it will be added in the future.

jpoimboe:
> I am yet to meet it in the wild and failing to navigate gcc's sources properly but I suspect there might be a '.hot.' counterpart, or at least it is possible it will be added in the future.

/me cries.  As objtool maintainer I've never seen '.hot.' (and I'm pretty certain I would have heard about it already if it's out there).  Maybe that's something we can look forward to next year for GCC 9.

kpatch-build/create-diff-object.c
```diff
@@ -608,15 +637,41 @@ static void kpatch_compare_sections(struct list_head *seclist)
...
+{
+	if (strncmp(sec->name, ".text.", 6)) {
+		DIFF_FATAL("unsupported section type: %s", sec->name);
+	}
+
+	if (!strncmp(sec->name, ".text.unlikely.", 15)) {
+		return SUBSECTION_UNLIKELY;
+	} else if (!strncmp(sec->name, ".text.hot.", 10)) {
+		return SUBSECTION_HOT;
+	}
```
jpoimboe: else's aren't necessary after returns

kpatch-build/create-diff-object.c
```diff
@@ -608,15 +637,41 @@ static void kpatch_compare_sections(struct list_head *seclist)
...
+			if (sec->sym && sec->sym->status != CHANGED) {
 				sec->sym->status = sec->status;
+				if (sec->sym->parent)
+					sec->sym->parent->status = sec->status;
+			}
 		}
 	}
 }
 
+static sstype_t kpatch_section_type(struct section *sec)
```
jpoimboe: Section type already has another meaning.  Maybe 'kpatch_subsection_type'?

kpatch-build/create-diff-object.c
```diff
@@ -608,15 +637,41 @@ static void kpatch_compare_sections(struct list_head *seclist)
...
+{
+	if (strncmp(sec->name, ".text.", 6)) {
+		DIFF_FATAL("unsupported section type: %s", sec->name);
+	}
```
jpoimboe: I think this DIFF_FATAL could be confusing, based on how this function is called.  For example, if were a data section, this error would hit instead of the "symbol changed sections" error.  So maybe just remove this check altogether and assume that it's just always SUBSECTION_NORMAL if it doesn't match the other string comparisons?

jpoimboe:
Some unit tests would be good for this.

sm00th:
> - Does it handle the case where a child function appears or disappears as a result of the patch? That would mean the parent should be marked changed (and the child, if it exists, included in the patch).

It does, however all of the patches I tested with resulted in parent function being changed as well. It won't mark the parent changed if just the .cold. subfunction appears/disappears, but I am under an impression that it can't happen. Might this assumption be wrong?

> - If only the parent changes, but not the child, does the child still get pulled into the patch?

It does not. Won't the patched function be able to use the already loaded .cold. subfunction?

> - Some unit tests would be good for this.

Yes, I was planning to add them once all of the major missing parts are included in the pull-request.

jpoimboe:
> It does, however all of the patches I tested with resulted in parent function being changed as well. It won't mark the parent changed if just the .cold. subfunction appears/disappears, but I am under an impression that it can't happen. Might this assumption be wrong?

Yes, the subfunction can appear/disappear.  The subfunction is basically a subset of the function which the compiler deems unlikely to execute, so it moves part of the function to the unlikely text section, thus creating the "subfunction".  If the patch file removes that piece of unlikely code from the function, the subfunction will disappear.  Similarly, if the patch file adds a piece of unlikely to a function, it may be placed in a (new) subfunction.

> If only the parent changes, but not the child, does the child still get pulled into the patch?
> It does not. Won't the patched function be able to use the already loaded .cold. subfunction?

Yes, I suppose that would be fine.  Either way works.

sm00th:
> If the patch file removes that piece of unlikely code from the function, the subfunction will disappear. Similarly, if the patch file adds a piece of unlikely to a function, it may be placed in a (new) subfunction.

Yes, but the thing is every time I tried to just add a cold subfunction (if (!jiffies) printk) the parent function got changed as well. Do you happen to know a way to do this without affecting parent function so that we can add it to unittests?

jpoimboe:
> Yes, but the thing is every time I tried to just add a cold subfunction (if (!jiffies) printk) the parent function got changed as well. Do you happen to know a way to do this without affecting parent function so that we can add it to unittests?

Yeah, find a function which already has a subfunction and figure out which part of the function's source corresponds to the subfunction.  It will presumably be in an "unlikely" clause.  Then just add an asm("nop") to that clause.  That should in theory modify the subfunction without modifying the parent function, assuming there aren't any other unlikely clauses after the changed one (as that would affect some other relocations in the parent function).

sm00th:
> Yeah, find a function which already has a subfunction and figure out which part of the function's source corresponds to the subfunction. It will presumably be in an "unlikely" clause. Then just add an asm("nop") to that clause. That should in theory modify the subfunction without modifying the parent function, assuming there aren't any other unlikely clauses after the changed one (as that would affect some other relocations in the parent function).

But that will only change the subfunction, not add or remove it. Anyway I changed it so that should now cover add/delete cases even if parent function is not change. Shouldn't hurt even if it never triggers.

sm00th:
v2:
  - Suggested codestyle and function names changes
  - added children skips to kpatch_reate_patches_sections() and function_ptr_rela()
  - addition/deletion of subfunction will now mark parent as changed
  - kpatch_mark_ignored_functions_same() will now mark children of ignored functions same

jpoimboe:
Sorry, I neglected to read and I completely missed your point about adding/removing subfunctions.  Adding/removing a subfunction will always change the parent, because the parent will always jump to the subfunction.  So adding/removing the subfunction will add/remove a jump to the parent function.

sm00th:
Okay, we can drop the last patch (create-diff-object: make deleted subfunctions mark parents as changed ) then.

kpatch-build/create-diff-object.c
```diff
@@ -608,16 +665,40 @@ static void kpatch_compare_sections(struct list_head *seclist)
...
-			if (sec->base->sym && sec->base->sym->status != CHANGED)
-				sec->base->sym->status = sec->status;
-		} else {
-			if (sec->sym && sec->sym->status != CHANGED)
-				sec->sym->status = sec->status;
+		struct symbol *sym;
+		if (is_rela_section(sec))
+			sym = sec->base->sym;
+		else
+			sym = sec->sym;
```
jpoimboe: optional: could probably do `sym = is_rela_section(sec) ? sec->base->sym : sec->sym;` which I think would be more succinct and readable

kpatch-build/create-diff-object.c
```diff
@@ -608,16 +665,40 @@ static void kpatch_compare_sections(struct list_head *seclist)
...
+
+		if (sym && sym->status != CHANGED) {
+			sym->status = sec->status;
+			if (sym->parent) {
+				if (sec->status == NEW && sym->parent->sec->twin)
+					sym->parent->status = CHANGED;
```
jpoimboe: Since we decided that a new subfunction will always change the parent, I think this isn't needed?

kpatch-build/create-diff-object.c
```diff
@@ -585,6 +622,26 @@ static int kpatch_line_macro_change_only(struct section *sec)
 }
 #endif
 
+static void kpatch_find_deleted_subsections(struct kpatch_elf *kelf)
```
jpoimboe: I agree we can remove this

kpatch-build/create-diff-object.c
sm00th: I think we better leave it here, otherwise we are dependent on the section ordering in the list. I.e. if the patch results in a NEW subfunction and parent section is already marked as CHANGED without this if we will be marking parent as NEW. Or am I missing something?

jpoimboe: It seems a little convoluted, what do you think about changing it to something like this?
```
	/* sync symbol status */
	list_for_each_entry(sec, seclist, list) {
		if (is_rela_section(sec)) {
			if (sec->base->sym && sec->base->sym->status != CHANGED)
				sec->base->sym->status = sec->status;
		} else {
			struct symbol *sym = sec->sym;

			if (sym && sym->status != CHANGED)
				sym->status = sec->status;

			if (sym && sym->child && sym->status == SAME &&
			    sym->child->sec->status == CHANGED)
				sym->status = CHANGED;
		}
	}
```

sm00th: That should work and is easier to read. Will use that, thanks.

sm00th:
v3:
  - codestyle fixes
  - dropped "create-diff-object: make deleted subfunctions mark parents as changed" patch
  - kpatch_compare_sections() patch reworked to be less confusing
  - added unittests

kpatch-build/create-diff-object.c
```diff
@@ -612,12 +655,34 @@ static void kpatch_compare_sections(struct list_head *seclist)
...
 		} else {
-			if (sec->sym && sec->sym->status != CHANGED)
-				sec->sym->status = sec->status;
+			struct symbol *sym = sec->sym;
+
+			if (sym && sym->status != CHANGED)
+				sym->status = sec->status;
+
+			if (sym && sym->child && sym->status == SAME &&
+					sym->child->sec->status == CHANGED)
```
jpoimboe: The alignment is off here, it should look like:
```
			if (sym && sym->child && sym->status == SAME &&
			    sym->child->sec->status == CHANGED)
```

kpatch-build/create-diff-object.c
```diff
@@ -630,10 +695,12 @@ static void kpatch_compare_correlated_symbol(struct symbol *sym)
...
 	 * If two symbols are correlated but their sections are not, then the
 	 * symbol has changed sections.  This is only allowed if the symbol is
-	 * moving out of an ignored section.
+	 * moving out of an ignored section, or moving between normal/hot/unlikely
+	 * subsections.
 	 */
 	if (sym1->sec && sym2->sec && sym1->sec->twin != sym2->sec) {
-		if (sym2->sec->twin && sym2->sec->twin->ignore)
+		if ((sym2->sec->twin && sym2->sec->twin->ignore) ||
+				kpatch_subsection_changed(sym1->sec, sym2->sec))
```
jpoimboe: another alignment issue here

kpatch-build/create-diff-object.c
sm00th: Ah, my vim's tabstop/shiftwidth combination made it look like it was aligned.

sm00th:
v4:
  - indent fixes

Issue #876, opened 2018-06-05
testing: rhel-7.6 meminfo-cmdline-rebuild-SLOW.patch build fails


joe-lawrence:
When testing against 3.10.0-862.el7.x86_64, a couple errors, split by group:
```
create-diff-object: ERROR: aes-x86_64-asm_64.o: main: 3086: FILE symbol not found in base. Stripped?
create-diff-object: ERROR: ia32entry.o: main: 3086: FILE symbol not found in base. Stripped?
create-diff-object: ERROR: entry_64.o: main: 3086: FILE symbol not found in base. Stripped?
create-diff-object: ERROR: wakeup_64.o: main: 3086: FILE symbol not found in base. Stripped?
create-diff-object: ERROR: hibernate_asm_64.o: main: 3086: FILE symbol not found in base. Stripped?
create-diff-object: ERROR: head_64.o: main: 3086: FILE symbol not found in base. Stripped?
create-diff-object: ERROR: xen-asm.o: main: 3086: FILE symbol not found in base. Stripped?
create-diff-object: ERROR: copy_user_nocache_64.o: main: 3086: FILE symbol not found in base. Stripped?
create-diff-object: ERROR: getuser.o: main: 3086: FILE symbol not found in base. Stripped?
create-diff-object: ERROR: putuser.o: main: 3086: FILE symbol not found in base. Stripped?
```
For all of these "FILE symbol not found in base" errors, that is because they are assembly files and not .c files.  I'm guessing that somehow `kernel.h` changes percolate down to those files and we need to rethink this patch.

```
create-diff-object: ERROR: dma-noop.o: find_local_syms: 174: find_local_syms for dma-noop.c: found_none
create-diff-object: ERROR: proportions.o: find_local_syms: 174: find_local_syms for proportions.c: found_none
```
Both .c files are built into `lib.a`, so they don't have `fentry` calls for us to hook.  We could ignore them, or figure out why they are being modified by this patch.  (Again `kernel.h` I'm looking at you.)

```
cmdline.o: changed function: cmdline_proc_show
meminfo.o: changed function: meminfo_proc_show
ERROR: cx2341x.o: reference to static local variable cx2341x_video_spatial_filter_mode_menu.28486 in CSWTCH.36 was removed
/root/kpatch/kpatch-build/create-diff-object: unreconcilable difference
```
Need to do more digging on this one.

jpoimboe:
[...]

PR #882, opened 2018-06-13
kpatch-build misc fixups

joe-lawrence:
Collection of small changes to cover some issues that popped up as part of #876.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -52,11 +52,12 @@
 #include "kpatch-patch.h"
 #include "kpatch-elf.h"
 #include "kpatch-intermediate.h"
+#include "kpatch.h"
 
 #define DIFF_FATAL(format, ...) \
 ({ \
 	fprintf(stderr, "ERROR: %s: " format "\n", childobj, ##__VA_ARGS__); \
-	error(2, 0, "unreconcilable difference"); \
+	error(EXIT_STATUS_DIFF_FATAL, 0, "unreconcilable difference"); \
 })
 
 #ifdef __powerpc__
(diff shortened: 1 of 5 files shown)


kpatch-build/create-diff-object.c
```diff
@@ -3082,8 +3082,10 @@ int main(int argc, char *argv[])
 			break;
 		}
 	}
-	if (!hint)
-		ERROR("FILE symbol not found in base. Stripped?\n");
+	if (!hint) {
+		log_normal("WARNING: FILE symbol not found in base. Stripped object file or assembly source?\n");
+		return 3; /* NO_CHANGE */
```
sm00th: nit: not about this particular change, but we might want to add an enum for returncodes, it is inconvenient to search for those in comments.

joe-lawrence: I thought the same while modifying the code, but for some reason assumed it was a stylistic choice that @jpoimboe made when he introduced the numbers.  If I remember correctly, the return codes are also used by one of the fatal error message macros, so the enum would need to go into a common header file.  There's also stdlib.h's notion of EXIT_SUCCESS and EXIT_FAILURE ... is there a safe way to use those and add a EXIT_NO_CHANGE?

sm00th: It is used in DIFF_FATAL macros from create-diff-object.c (retval 2) and in ERROR macros from log.h (retval1).
I think using stdlib's defines alongside our own might be confusing.

joe-lawrence:
v2
- Added a commit to implement Artem's enum suggestion, compile tested only.

joe-lawrence:
v3
- git add the sources for v2 :)

PR #883, opened 2018-06-15
kpatch-build: fix find_special_section_data with blank AWK_OPTIONS

sm00th:
If $AWK_OPTIONS are blank gawk would treat "" as a blank script
resulting in none of the special struct being detected.

Fixes: 1330dcc "create-diff-object: add ORC section support"


joe-lawrence:
I'd be ok with this work around, but does stuffing the awk script into a variable help us any?  (ie, it could bring the entire `SPECIAL_VARS` assignment into a single line... which I wonder if make check is getting a little confused with.)

sm00th:
> I'd be ok with this work around, but does stuffing the awk script into a variable help us any?

It doesn't do much but what is the alternative? Just eval it right away without `$SPECIAL_VARS`? That doesn't seem to remove any quoting requirements.

joe-lawrence:
My thought was that the linter was getting confused by the linebreak, as there is already a quote at the beginning of `SPECIAL_VARS="$(` but perhaps there need to be more once we get inside the command substitution `$(`.

PR #886, opened 2018-06-18
kpatch-build: less agressive clean_cache()

sm00th:
Some of the provisioned machines I sometimes use don't have enought
diskspace for a full kpatch-patch build in a home partitions, I usually
solve this by symlinking .kpatch(and .ccache) dirs to a different
partition, however this only works with -s option because of
clean_cache().

clean_cache() currently removes .kpatch directory completely, recreating
it from scratch, change it to only remove the contents of the directory
instead.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -125,7 +125,7 @@ cleanup() {
 }
 
 clean_cache() {
-	rm -rf "$CACHEDIR"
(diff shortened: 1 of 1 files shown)


sm00th:
v2: `$CACHEDIR` -> `${CACHEDIR:?}` as suggested by shellcheck

sm00th:
v3: commit message spelling

PR #890, opened 2018-06-21
readme: Update gcc-plugin-devel dependency

kamalesh-babulal:
Update gcc-plugin-devel package dependency for livepatch
modules on ppc64le.

Suggested-by: Joe Lawrence <jdl1291@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,9 @@ sudo dnf install pesign yum-utils openssl wget numactl-devel
 sudo dnf builddep kernel-${UNAME%.*}
 sudo dnf debuginfo-install kernel-${UNAME%.*}
 
+# required on ppc64le
+sudo dnf install gcc-plugin-devel
+
 # optional, but highly recommended
 sudo dnf install ccache
 ccache --max-size=5G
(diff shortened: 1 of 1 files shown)


joe-lawrence:
You can add `gcc-plugin-devel` to the RHEL section as well.

Also, we can close it #890 manually, but if you add "Fixes #890." to your commit msg, github will close the corresponding issue automatically when we merge it.

Thanks, @kamalesh-babulal !

kamalesh-babulal:
v2: Update `gcc-plugin-devel` dependency for RHEL too.

README.md
```diff
@@ -44,6 +44,9 @@ sudo dnf install pesign yum-utils openssl wget numactl-devel
 sudo dnf builddep kernel-${UNAME%.*}
 sudo dnf debuginfo-install kernel-${UNAME%.*}
 
+# required, on ppc64le for building livepatch modules
```
jpoimboe: The "for building livepatch modules" is already implicit here.  Also the comma kind of makes it sound like it's always required.  So maybe change it to something like "required on ppc64le"?

kamalesh-babulal:
v3: Reword the dependency description

PR #894, opened 2018-07-17
kpatch-build: fix clean_cache

jpoimboe:
Commit d86c1113cc25 ("kpatch-build: less aggressive clean_cache()")
broke clean_cache().  Instead of expanding the wildcard, it tries to
delete a file named '*'.


joe-lawrence:
Too bad checkpatch doesn't flag that one.

Issue #895, opened 2018-07-24
kpatch-build fails with -mrecord-mcount


sm00th:
4.18 rc's contain 96f60dfa5819 "trace: Use -mcount-record for dynamic ftrace" that adds -mrecord-mcount to KBUILD_CFLAGS when supported by compiler.

This option moves all of the profiling call pointers to a __mcount_loc section. From gcc's man page:

```
       -mrecord-mcount
       -mno-record-mcount
           If profiling is active (-pg), generate a __mcount_loc section that contains pointers
           to each profiling call. This is useful for automatically patching and out calls.
```

create-diff-object fails unable to find fentry/mcount in the function section:

```
ERROR: meminfo.o: 1 function(s) can not be patched
meminfo.o: function meminfo_proc_show has no fentry/mcount call, unable to patch
/home/asavkov/dev/kpatch/kpatch-build/create-diff-object: unreconcilable difference
```

jpoimboe:
I think the difference is that now the `__fentry__` relocation type is `R_X86_64_PC32` instead of `R_X86_64_NONE`.  Should be an easy fix in kpatch_find_func_profiling_calls().

sm00th:
Sorry, should have mentioned that I tried that and it is a bit more complicated. With `R_X86_64_PC32` added as an option in `kpatch_find_func_profiling_calls()` it fails in `kpatch_create_mcount_sections()` with `unexpected instruction at the start of a function`. The first instruction in every function looks like this:

```
   0:   e8 00 00 00 00          callq  5 <meminfo_proc_show+0x5>
```

So it is already what `kpatch_create_mcount_sections()` is trying to change it to and first rela is already of a proper type. Can we just skip create_mcount step in this case or the last part of it?

jpoimboe:
> So it is already what kpatch_create_mcount_sections() is trying to change it to and first rela is already of a proper type. Can we just skip create_mcount step in this case or the last part of it?

Yeah, sounds like it.

PR #897, opened 2018-07-26
create-diff-object: -mcount-record support

sm00th:
4.18 adds -mcount-record to KBUILD_FLAGS when supported by the compiler.
This results in most of kpatch_create_mcount_sections()'s work being
already done, so we can at least skip the last part of it that updates
the first instruction in patched functions.

--

Resulting patch loads and functions ok, but I don't completely understand the __mcount_loc magic so maybe this is not the best solution for #895.


kpatch-build/create-diff-object.c
```diff
@@ -3056,18 +3056,23 @@ static void kpatch_create_mcount_sections(struct kpatch_elf *kelf)
...
-		insn[2] = 0;
-		insn[3] = 0;
-		insn[4] = 0;
 
 		rela = list_first_entry(&sym->sec->rela->relas, struct rela,
 					list);
-		rela->type = R_X86_64_PC32;
+		if (!(insn[0] == 0xe8 && insn[1] == 0 && insn[2] == 0 &&
+		    insn[3] == 0 && insn[4] == 0) ||
+		    rela->type != R_X86_64_PC32) {
```
kamalesh-babulal: Can the above check for opcode/rela type replaced with `if (rela->type == R_X86_64_NONE)`

sm00th: Hm, yes. I think right now this should work as well and definitely does look better. It will be a bit less strict though, but I guess it is worth it. If there won't be any objections to this I'll change it to this, thank you.

sm00th:
v2:
  - updated if kpatch_create_mcount_sections() + comment
  - included unittest for this case

sm00th:
v3:
  - proper comment style

PR #899, opened 2018-08-01
kpatch-build: allow setting skipcleanup/skipgcccheck through env vars

sm00th:
kpatch-build currently forcefully sets SKIPCLEANUP/SKIPGCCCHECK to 0 on
startup so the only way to set them is through command line arguments.
Setting this through env vars might be useful for automation, for
example this will allow to skip gcc check during integration tests
without modifying kpatch-test script.


jpoimboe:
Environment vars are a bit sneaky...  Why exactly is this needed?  Command-line args are preferable.

If we _do_ expose environment vars, they should probably be namespaced with a "KPATCH_" prefix.  I realize we're not doing that with CACHEDIR, but we probably should be.

sm00th:
No longer needed.

PR #901, opened 2018-08-15
Integration test improvements

sm00th:
Couple of small improvements:
 - saving build logs for failed tests
 - include -FAIL tests in "quick" target

### DIFF
--- a/test/integration/kpatch-test
+++ b/test/integration/kpatch-test
@@ -28,8 +28,6 @@
 #
 # - foo.patch: patch that should build successfully
 #
-# - foo-SLOW.patch: patch that should be skipped in the quick test
-#
 # - bar-FAIL.patch: patch that should fail to build
 #
 # - foo-LOADED.test: executable which tests whether the foo.patch module is
@@ -106,7 +104,6 @@ else
 		prefix=${file%%.patch}
 		[[ -e "$prefix-FAIL.test" ]]   && TEST_LIST+=("$prefix-FAIL.test")
 		[[ -e "$prefix-LOADED.test" ]] && TEST_LIST+=("$prefix-LOADED.test")
-		[[ -e "$prefix-SLOW.test" ]]   && TEST_LIST+=("$prefix-SLOW.test")
 	done
 fi
 
@@ -230,7 +227,6 @@ build_combined_module() {
 	declare -a COMBINED_LIST
 	for file in "${PATCH_LIST[@]}"; do
 		[[ $file =~ -FAIL ]] && log "combine: skipping $file" && continue
-		[[ $file =~ -SLOW ]] && log "combine: skipping $file" && continue
 		COMBINED_LIST+=($file)
 	done
 	if [[ ${#COMBINED_LIST[@]} -le 1 ]]; then
@@ -258,7 +254,6 @@ run_combined_test() {
 
 	for testprog in "${TEST_LIST[@]}"; do
 		[[ $testprog != *-LOADED.test ]] && continue
-		[[ $testprog =~ -SLOW ]] && continue
 		if $testprog >> $LOG 2>&1; then
 			error "combined: $testprog succeeded before kpatch load"
 			return
@@ -272,7 +267,6 @@ run_combined_test() {
 
 	for testprog in "${TEST_LIST[@]}"; do
 		[[ $testprog != *-LOADED.test ]] && continue
-		[[ $testprog =~ -SLOW ]] && continue
 		if ! $testprog >> $LOG 2>&1; then
 			error "combined: $testprog failed after kpatch load"
 		fi
@@ -285,7 +279,6 @@ run_combined_test() {
 
 	for testprog in "${TEST_LIST[@]}"; do
 		[[ $testprog != *-LOADED.test ]] && continue
-		[[ $testprog =~ -SLOW ]] && continue
 		if $testprog >> $LOG 2>&1; then
 			error "combined: $testprog succeeded after kpatch unload"
 			return
--- a/test/integration/kpatch-test
+++ b/test/integration/kpatch-test
@@ -258,6 +258,7 @@ run_combined_test() {
 
 	for testprog in "${TEST_LIST[@]}"; do
 		[[ $testprog != *-LOADED.test ]] && continue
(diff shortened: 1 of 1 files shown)


test/integration/kpatch-test
```diff
@@ -152,7 +152,10 @@ build_module() {
 	log "build: $prefix"
 
 	if ! $KPATCHBUILD -n $modname $file >> $LOG 2>&1; then
-		[[ $shouldfail -eq 0 ]] && error "$prefix: build failed"
+		if [[ $shouldfail -eq 0 ]]; then
+			error "$prefix: build failed"
+			cp ${HOME}/.kpatch/build.log $(dirname ${LOG})/${prefix}.$(basename ${LOG})
```
kamalesh-babulal: I am not familiar with the test framework code, I might be missing something. Should the copied over build failure log also removed during the invocation of the script again or the idea is to retain it for future reference.

sm00th: That's a good point. I think they should be treated the same as main log which [is removed on every start](https://github.com/dynup/kpatch/blob/master/test/integration/kpatch-test#L54). I'll post a v2, thank you.

sm00th:
v2: remove copied over build logs on script start

joe-lawrence:
Changes look like they would work (I did have a comment about `$LOG`), a few high level questions that wouldn't hold up the review, but came to mind:

Just to be sure here, adding the FAIL tests will add three additional kpatch builds to the "quick" target... I haven't run the numbers but wouldn't this mean pushing a "quick" test from ~15 minutes to an hour?  (Maybe it doesn't scale that poorly, but just pulling numbers off the top of my head).

How large are the build.log files and should we gzip, tail or otherwise try to summarize them?

sm00th:
> Just to be sure here, adding the FAIL tests will add three additional kpatch builds to the "quick" target... I haven't run the numbers but wouldn't this mean pushing a "quick" test from ~15 minutes to an hour? 

Because we are now reusing objectfiles from previous builds it doesn't add that much. I don't have any solid data because data I have comes from different machines and measurements include the time it takes to spin up virtual machines. So far it looks like it bumps the overall time from 30-35 mins to 35-40.

> How large are the build.log files and should we gzip, tail or otherwise try to summarize them?

Thanks to not rebuilding objectfiles all the time they generally shouldn't be that big. Worst case would be 2x full kernel buildlog which is not small. gzipping might be ok, but it adds some inconvenience, so maybe we should gzip only after some threshold, say 1Mb? I also would expect these tests to mostly pass so we shouldn't see those logs too often.

test/integration/kpatch-test
```diff
@@ -51,7 +51,7 @@ unset CCACHE_HASHDIR
 KPATCHBUILD="$ROOTDIR"/kpatch-build/kpatch-build
 ERROR=0
 LOG=test.log
-rm -f $LOG
+rm -f $(dirname ${LOG})/*$(basename ${LOG})
```
jpoimboe: Using dirname here seems a bit overkill, since the test log is always in the current directory.  If we ever decided to put the logs in another directory I think it would be simpler to just add a LOGDIR variable at that point.

Also maybe we can just use a .log convention so we don't need all the basenames everywhere.  Then this can just be `rm -f *.log` which is much easier to read.

test/integration/kpatch-test
jpoimboe: Similar to my other comments, this could just be `cp $HOME/.kpatch/build.log $prefix.log` which is more readable.

test/integration/kpatch-test
jpoimboe: The convention in this file is that we don't use the `{` brackets, so in general e.g. use `$LOG` instead of `${LOG}`.

test/integration/kpatch-test
```diff
@@ -290,21 +294,21 @@ run_combined_test() {
...
 
 build_combined_module
 
 unload_all
 
-if [[ $QUICK != 1 ]]; then
-	for file in "${PATCH_LIST[@]}"; do
+for file in "${PATCH_LIST[@]}"; do
+	if [[ $QUICK != 1 || $file =~ -FAIL ]]; then
 		run_load_test $file
```
jpoimboe: Why run a load test for a `-FAIL` module which presumably failed to build?

sm00th:
v3:
  - simpler log names/handling
  - skip -FAIL test runs on top level instead of leaving this decision to run_load_test

test/integration/kpatch-test
```diff
@@ -152,7 +152,10 @@ build_module() {
 	log "build: $prefix"
 
 	if ! $KPATCHBUILD -n $modname $file >> $LOG 2>&1; then
-		[[ $shouldfail -eq 0 ]] && error "$prefix: build failed"
+		if [[ $shouldfail -eq 0 ]]; then
+			error "$prefix: build failed"
+			cp $HOME/.kpatch/build.log $prefix.log
```
joe-lawrence: The commit msg says, "copy the build log to \<prefix\>.test.log file", so should this have been `cp $HOME/.kpatch/build.log $prefix.test.log` ?

sm00th:
v4: commit message fixups

test/integration/kpatch-test
```diff
@@ -254,6 +258,7 @@ run_combined_test() {
 
 	for testprog in "${TEST_LIST[@]}"; do
 		[[ $testprog != *-LOADED.test ]] && continue
+		[[ $testprog =~ -SLOW ]] && continue
```
jpoimboe: I realize this is consistent with existing behavior, but why don't we include -SLOW patches in the combined test?  Should we revisit that decision?

joe-lawrence: I think this was part of the original setup that Seth created.  But yeah, if we're going to build all the patches (and most kernel objects) anyway, I don't think the -SLOW patches will add much to the build time.

jpoimboe: Looking back, it was actually me :-) about four years ago.  The original intent was that 'combined' was supposed to be a quick sanity test, so it avoided the slow rebuilds.  But if we're going to be doing combined patches for CI, I think the -SLOW patches should be included in that.  So maybe we can just get rid of the -SLOW concept altogether.

Issue #906, opened 2018-09-13
readelf: Warning: Section has an out of range sh_link value


joe-lawrence:
I noticed while debugging on an alternate architecture. 

The intermediate tmp.ko ELF file has a bunch of sections with a sh_link pointing to the .symtab section (in this case section 62):
```
% readelf --wide --sections tmp.ko

There are 65 section headers, starting at offset 0x7bab8:

Section Headers:
  [Nr] Name              Type            Address          Off    Size   ES Flg Lk Inf Al
  [ 0]                   NULL            0000000000000000 000000 000000 00      0   0  0
  [ 1] .group            GROUP           0000000000000000 000040 000008 04     62 5470  4
  [ 2] .group            GROUP           0000000000000000 000048 000008 04     62 5483  4
  [ 3] .group            GROUP           0000000000000000 000050 000008 04     62 5479  4
  [ 4] .note.gnu.build-id NOTE           0000000000000000 000058 000024 00   A  0   0  4
  [ 5] .text             PROGBITS        0000000000000000 00007c 000000 00  AX  0   0  4
  [ 6] .text.unlikely    PROGBITS        0000000000000000 000080 000270 00  AX  0   0  8
  [ 7] .rela.text.unlikely RELA          0000000000000000 053550 0001c8 18   I 62   6  8
  [ 8] .init.text        PROGBITS        0000000000000000 0002f0 000480 00  AX  0   0  8
  [ 9] .rela.init.text   RELA            0000000000000000 053718 0004b0 18   I 62   8  8
  [10] .exit.text        PROGBITS        0000000000000000 000770 000048 00  AX  0   0  8
  [11] .rela.exit.text   RELA            0000000000000000 053bc8 000048 18   I 62  10  8
                                     [ ... snip ... ]
  [62] .symtab           SYMTAB          0000000000000000 02ddd8 020250 18     63 5457  8
                                     [ ... snip ... ]
```
The final ELF object doesn't appear to have fully reset all the sh_links to .symtab's new index (section 57):
```
% readelf --wide --sections livepatch-cmdline-string.ko
There are 61 section headers, starting at offset 0x82e98:
readelf: livepatch-cmdline-string.ko: Warning: Section 1 has an out of range sh_link value of 62
readelf: livepatch-cmdline-string.ko: Warning: Section 2 has an out of range sh_link value of 62
readelf: livepatch-cmdline-string.ko: Warning: Section 3 has an out of range sh_link value of 62

Section Headers:
[...]

PR #907, opened 2018-09-13
create-klp-module: fixup GROUP section sh_links

joe-lawrence:
remove_intermediate_sections() deletes a few sections from the ELF file,
potentially moving the .strtab's section index.  Be sure to adjust not
only RELA sections that sh_link to the .strtab but also any GROUP
sections.

Fixes: #906

### DIFF
--- a/kpatch-build/create-klp-module.c
+++ b/kpatch-build/create-klp-module.c
@@ -435,7 +435,6 @@ static struct argp argp = { options, parse_opt, args_doc, 0 };
 int main(int argc, char *argv[])
 {
 	struct kpatch_elf *kelf;
-	struct section *symtab, *sec;
 	struct section *ksymsec, *krelasec, *strsec;
 	struct arguments arguments;
 	char *strings;
@@ -493,16 +492,6 @@ int main(int argc, char *argv[])
 
 	remove_intermediate_sections(kelf);
 	kpatch_reindex_elements(kelf);
-
-	/* Rebuild rela sections, new klp rela sections will be rebuilt too. */
-	symtab = find_section_by_name(&kelf->sections, ".symtab");
-	list_for_each_entry(sec, &kelf->sections, list) {
-		if (!is_rela_section(sec))
-			continue;
-		sec->sh.sh_link = symtab->index;
-		sec->sh.sh_info = sec->base->index;
-		kpatch_rebuild_rela_section_data(sec);
-	}
 
 	kpatch_create_shstrtab(kelf);
 	kpatch_create_strtab(kelf);
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -3225,7 +3225,6 @@ int main(int argc, char *argv[])
 	struct arguments arguments;
 	int num_changed, callbacks_exist, new_globals_exist;
 	struct lookup_table *lookup;
-	struct section *sec, *symtab;
 	struct symbol *sym;
 	char *hint = NULL, *orig_obj, *patched_obj, *parent_name;
 	char *parent_symtab, *mod_symvers, *patch_name, *output_obj;
@@ -3353,18 +3352,6 @@ int main(int argc, char *argv[])
 	kpatch_strip_unneeded_syms(kelf_out, lookup);
 	kpatch_reindex_elements(kelf_out);
 
-	/*
-	 * Update rela section headers and rebuild the rela section data
-	 * buffers from the relas lists.
-	 */
-	symtab = find_section_by_name(&kelf_out->sections, ".symtab");
-	list_for_each_entry(sec, &kelf_out->sections, list) {
-		if (!is_rela_section(sec))
-			continue;
-		sec->sh.sh_link = symtab->index;
-		sec->sh.sh_info = sec->base->index;
-		kpatch_rebuild_rela_section_data(sec);
-	}
 	kpatch_check_relocations(kelf_out);
 
 	kpatch_create_shstrtab(kelf_out);
--- a/kpatch-build/create-kpatch-module.c
+++ b/kpatch-build/create-kpatch-module.c
@@ -185,7 +185,6 @@ static struct argp argp = { options, parse_opt, args_doc, 0 };
 int main(int argc, char *argv[])
 {
 	struct kpatch_elf *kelf;
-	struct section *symtab, *sec;
 	struct section *ksymsec, *krelasec, *strsec;
 	struct arguments arguments;
 	int ksyms_nr, krelas_nr;
@@ -229,15 +228,6 @@ int main(int argc, char *argv[])
 	remove_intermediate_sections(kelf);
 
 	kpatch_reindex_elements(kelf);
-
-	symtab = find_section_by_name(&kelf->sections, ".symtab");
-	list_for_each_entry(sec, &kelf->sections, list) {
-		if (!is_rela_section(sec))
-			continue;
-		sec->sh.sh_link = symtab->index;
-		sec->sh.sh_info = sec->base->index;
-		kpatch_rebuild_rela_section_data(sec);
-	}
 
 	kpatch_create_shstrtab(kelf);
 	kpatch_create_strtab(kelf);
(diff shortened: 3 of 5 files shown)


joe-lawrence:
Should we be more generic about fixing up `sh_link` before and after `remove_intermediate_sections()`?  I feel like the current method (and this follow up fix) rely on the fact that the only `sh_link` values we ever need to update are those pointing to .symtab.

joe-lawrence:
Thanks for the link @kamalesh-babulal !  I think you're right about updating the `sh_info` field as well as the `sh_link`... once I get my testing box up and running again, I'll look into v2.

joe-lawrence:
v2: I couldn't think of any clever way to retain `sh_info` symbol index, so I just stuffed a copy of the corresponding symbol pointer in the `struct section` so I could apply its new index value after re-indexing.

Some testing notes showing both the `sh_link` and `sh_info` values updated:

```
Temporary kernel-object file:

  % readelf --wide --sections tmp.ko | grep 'GROUP'
    [Nr] Name              Type            Address          Off    Size   ES Flg Lk Inf Al
    [ 1] .group            GROUP           0000000000000000 000040 000008 04     62 5468  4
    [ 2] .group            GROUP           0000000000000000 000048 000008 04     62 5481  4
    [ 3] .group            GROUP           0000000000000000 000050 000008 04     62 5477  4

  % readelf --wide --syms tmp.ko | grep -e '5468:' -e '5481:' -e '5477:'
     Num:    Value          Size Type    Bind   Vis      Ndx Name
    5468: 0000000000000000     0 FUNC    GLOBAL HIDDEN    12 __s390_indirect_jump_r4
    5477: 0000000000000000     0 FUNC    GLOBAL HIDDEN    14 __s390_indirect_jump_r14
    5481: 0000000000000000     0 FUNC    GLOBAL HIDDEN    13 __s390_indirect_jump_r5

Final livepatch object file, after reindexing:

  % readelf --wide --sections livepatch-cmdline-string.ko | grep 'GROUP'
    [ 1] .group            GROUP           0000000000000000 000040 000008 04     57 5465  4
    [ 2] .group            GROUP           0000000000000000 000048 000008 04     57 5478  4
    [ 3] .group            GROUP           0000000000000000 000050 000008 04     57 5474  4

  % readelf --wide --syms livepatch-cmdline-string.ko | grep -e '5465:' -e '5478:' -e '5474:'
    5465: 0000000000000000     0 FUNC    GLOBAL HIDDEN    12 __s390_indirect_jump_r4
    5474: 0000000000000000     0 FUNC    GLOBAL HIDDEN    14 __s390_indirect_jump_r14
    5478: 0000000000000000     0 FUNC    GLOBAL HIDDEN    13 __s390_indirect_jump_r5
```

kpatch-build/create-klp-module.c
```diff
@@ -491,17 +491,32 @@ int main(int argc, char *argv[])
...
+	 * new symbol table indexes, so stash their 'identifying entry'
+	 * symbol pointer for use when rebuilding the section.
+	 */
+	list_for_each_entry(sec, &kelf->sections, list) {
+		if (is_group_section(sec)) {
+			sec->sh_link_info = find_symbol_by_index(&kelf->symbols, sec->sh.sh_info);
+		}
+	}
+
 	remove_intermediate_sections(kelf);
```
kamalesh-babulal: `kpatch_reindex_elements()` does a reindex of the symbols, after removing some sections from ELF, will it be better to move the loop to update `sec->sh_link_info` after `kpatch_reindex_elements()`

kamalesh-babulal: ah !! right. I can not think of better name, probably drop prefix `sh_` from `sh_link_info`, as `sh` is redundant for section struct member.

kpatch-build/create-klp-module.c
```diff
@@ -491,17 +491,32 @@ int main(int argc, char *argv[])
...
-		sec->sh.sh_info = sec->base->index;
-		kpatch_rebuild_rela_section_data(sec);
+		if (is_rela_section(sec)) {
+			sec->sh.sh_link = symtab->index;
+			sec->sh.sh_info = sec->base->index;
+			kpatch_rebuild_rela_section_data(sec);
+		}
+		else if (is_group_section(sec)) {
+			sec->sh.sh_link = symtab->index;
+			sec->sh.sh_info = sec->sh_link_info ? sec->sh_link_info->index : 0;
```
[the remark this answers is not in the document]
kamalesh-babulal: Interesting, were there any stacktrace when reading '/proc/cmdline' after loading the livepatch module. I will vote for bailing out.

joe-lawrence:
v3: per @kamalesh-babulal 's suggestions:
- s/sec->sh_link_info/sec->link_info/g
- ERROR out if we can't find the 'identifying entry' symbol (by its index) and dump the associated ternary operator assignment

jpoimboe:
- I don't remember seeing these GROUP sections before, is it only s390 which is using them?  What are they used for?

- How did these sections end up in the intermediate object?  Is create-diff-object putting them there somehow?  I didn't see where that could happen.

- create-diff-object also does section shuffling, maybe it would make sense to put all this in common code, in kpatch-elf.c.  sec->link_info could be initialized at the beginning with all the other data.  And then sh_link and sh_info could be updated inside kpatch_reindex_elements().  I think the rela sh_link/sh_info reindexing also should be moved there.

I think `link_sym` would be a clearer name for the field instead of `link_info`.

joe-lawrence:
> I don't remember seeing these GROUP sections before, is it only s390 which is using them? What are they used for?
>
> How did these sections end up in the intermediate object? Is create-diff-object putting them there somehow? I didn't see where that could happen.

I have only started seeing them on s390x very recently.  It might something new out of gcc?  I'll compare versions and try to dig into their origin today.  (s390x machine finally reserved.)

> create-diff-object also does section shuffling, maybe it would make sense to put all this in common code, in kpatch-elf.c. sec->link_info could be initialized at the beginning with all the other data. And then sh_link and sh_info could be updated inside kpatch_reindex_elements(). I think the rela sh_link/sh_info reindexing also should be moved there.

OK, I'll take a look at adding a refactoring commit to the end of this as well.

> I think link_sym would be a clearer name for the field instead of link_info.

OK, will rename.

joe-lawrence:
The group sections appear as part of the kernel build:
```
% readelf --sections ~/.kpatch/src/fs/proc/cmdline.o                               
There are 40 section headers, starting at offset 0x1b220:

Section Headers:
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
  [ 0]                   NULL             0000000000000000  00000000
       0000000000000000  0000000000000000           0     0     0
  [ 1] .group            GROUP            0000000000000000  00000040
       0000000000000008  0000000000000004          37   1183     4
  [ 2] .group            GROUP            0000000000000000  00000048
       0000000000000008  0000000000000004          37   1181     4

% readelf --section-groups ~/.kpatch/src/fs/proc/cmdline.o 

COMDAT group section [    1] `.group' [__s390_indirect_jump_r4] contains 1 sections:
   [Index]    Name
   [   17]   .text.__s390_indirect_jump_r4

COMDAT group section [    2] `.group' [__s390_indirect_jump_r14] contains 1 sections:
   [Index]    Name
   [   18]   .text.__s390_indirect_jump_r14
```

And I think these particular sections may have been introduced by gcc as part of Spectre mitigation:

Feb 8 14:45:53 2018 - S/390: Disable prediction of indirect branches
https://gcc.gnu.org/viewcvs/gcc?view=revision&revision=257489

joe-lawrence:
@jpoimboe : was this your intention for `kpatch_reindex_elements()` ... I wasn't clear where you thought `link_sym` should be initialized.

v4:
- s/link_info/link_sym/g
- pushed common code into kpatch_reindex_elements()
- tested on s390x (livepatch prototype) and x86_64 (both livepatch and kpatch modules)

joe-lawrence:
Ping ... as far as I remember, these GROUP sections were s390x only.  Shall I close this PR until / if kpatch gains more s390x support?

jpoimboe:
@joe-lawrence Sorry for dropping the ball on this one.  We can close it until if/when kpatch-build supports s390.

PR #914, opened 2018-10-15
Added distros to man page

sanskriti-s:
The man page earlier said that kpatch-build only works for Fedora, so I
added all the other distros it works for based on the README.

Signed-off-by: Sanskriti Sharma <sansharm@redhat.com>

### DIFF
--- a/man/kpatch-build.1
+++ b/man/kpatch-build.1
@@ -11,9 +11,6 @@ currently running and creates a kernel module that will replace
 modified functions in the kernel such that the patched code takes 
 effect.
 
-This script currently only works on Fedora and will need to be adapted 
-to work on other distros.  
-
 .SH OPTIONS
 
 -h|--help


man/kpatch-build.1
```diff
@@ -11,8 +11,9 @@ currently running and creates a kernel module that will replace
 modified functions in the kernel such that the patched code takes 
 effect.
 
-This script currently only works on Fedora and will need to be adapted 
-to work on other distros.  
+This script currently works on Fedora, RHEL, CentOS, Oracle Linux,
+Ubuntu, Debian and Gentoo, and can be adapted to work on other distros.
+See README for specific releases.
```
jpoimboe: Maybe we should just remove this paragraph altogether?  It now works on most distros, and even for the distros it doesn't support officially, the --sourcedir option can be used.  So the original intent was to warn about limitations, but those limitations are pretty much gone now.

joe-lawrence: That works, too.  I rarely look at these files, so I don't have a strong opinion :)

PR #915, opened 2018-10-15
Added option to kpatch man page

sanskriti-s:
Added option version to man page, under commands, based on usage function in
kpatch/kpatch usage()

Signed-off-by: Sanskriti sharma <sansharm@redhat.com>

### DIFF
--- a/man/kpatch.1
+++ b/man/kpatch.1
@@ -35,6 +35,9 @@ info <module>
 list
        list installed patch modules
 
+signal
+       signal/poke any process stalling the current patch transition
+
 version
        display the kpatch version
 


man/kpatch.1
```diff
@@ -35,6 +35,9 @@ info <module>
 list
        list installed patch modules
 
+signal
+	signal/poke any process stalling the current patch transition
```
joe-lawrence: nitpick ... `% man -l man/kpatch.1` looks like this:
```
       list
              list installed patch modules

       signal      signal/poke any process stalling the current patch transition

       version
              display the kpatch version
```
My guess is that is man wants spaces and not tabs here.

PR #916, opened 2018-10-16
Move lookup_open() call to a later stage.

### LINKED ISSUE #910: find_local_syms() failures on optimized-out objectfiles
While testing f27 with meminfo-cmdline-rebuild-SLOW.patch I stumbled upon a couple examples when the whole objectfile is getting discarded during final vmlinux linking.

The two examples I have are:

arch/x86/lib/error-inject.c that is built if CONFIG_FUNCTION_ERROR_INJECTION is enabled but only used if CONFIG_BPF_KPROBE_OVERRIDE or CONFIG_FAIL_FUNCTION is enabled.

drivers/net/dsa/dsa_loop_bdinfo.c - not exactly sure what happens with this one, but it is missing from vmlinux (or any modules) as well.

It looks like config-dependency issues, but can we somehow detect cases like this?

sm00th:
Sometimes due to config-dependency issues or other reasons whole
object-files would get optimized out from final vmlinux/module, in cases
like this create-diff-object would fail during symbol lookup table
creation in lookup_open(). Because lookup_open() call is situated before
we established that objectfile has changed this triggers not only on
real problems, but also during mass-rebulds caused by changes to
header-files. While it usually indicates a real issue with config this
should not prevent kpatch from building.

Move lookup_open() call so that it is called only for changed
object-files.

Fixes #910

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2482,7 +2482,7 @@ static struct sym_compare_type *kpatch_elf_locals(struct kpatch_elf *kelf)
 			continue;
 
 		sym_array[i].type = sym->type;
-		sym_array[i++].name = sym->name;
+		sym_array[i++].name = strdup(sym->name);
 	}
 	sym_array[i].type = 0;
 	sym_array[i].name = NULL;
@@ -3229,7 +3229,7 @@ int main(int argc, char *argv[])
 	struct symbol *sym;
 	char *hint = NULL, *orig_obj, *patched_obj, *parent_name;
 	char *parent_symtab, *mod_symvers, *patch_name, *output_obj;
-	struct sym_compare_type *base_locals;
+	struct sym_compare_type *base_locals, *sym_comp;
 
 	arguments.debug = 0;
 	argp_parse (&argp, argc, argv, 0, NULL, &arguments);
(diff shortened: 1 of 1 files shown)


kpatch-build/create-diff-object.c
```diff
@@ -3332,6 +3325,13 @@ int main(int argc, char *argv[])
 	 */
 	kpatch_elf_teardown(kelf_patched);
 
+	/* create symbol lookup table */
+	base_locals = kpatch_elf_locals(kelf_base);
+	lookup = lookup_open(parent_symtab, mod_symvers, hint, base_locals);
+	free(base_locals);
+	kpatch_elf_teardown(kelf_base);
+	kpatch_elf_free(kelf_base);
```
jpoimboe: We should try to leave the kpatch_elf_teardown/free calls where they are, because they provide protection against accessing obsolete data later on, which was an occasional source of bugs in the past.

Would it work if we leave the teardown/free calls where they are, and put the kpatch_elf_locals() call immediately before them?  Then lookup_open() can still be done here, without having to also move the teardown/free calls.

sm00th: > Would it work if we leave the teardown/free calls where they are, and put the kpatch_elf_locals() call immediately before them? Then lookup_open() can still be done here, without having to also move the teardown/free calls.

We can leave the teardown where it is now, but I get segfaults in find_local_syms() unless kpatch_elf_free(kelf_base) is located after that call.

sm00th: Its this and hint as well. Updated the request.

PR #917, opened 2018-10-16
Add "__addressable_" to maybe_discarded_sym().

sm00th:
Starting with 1b1eeca7e4c1 "init: allow initcall tables to be emitted using
relative references" [1] __init functions are generating an "__addressable_"
symbol in a ".discarded.addressable" section so it does not show up in final
vmlinux triggering find_local_syms failures. Add "_addressable_" to the list
in maybe_discarded_sym().

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=1b1eeca7e4c19fa76d409d4c7b338dba21f2df45

### DIFF
--- a/kpatch-build/lookup.c
+++ b/kpatch-build/lookup.c
@@ -80,7 +80,8 @@ static int maybe_discarded_sym(const char *name)
 	 */
 	if (!strncmp(name, "__exitcall_", 11) ||
 	    !strncmp(name, "__brk_reservation_fn_", 21) ||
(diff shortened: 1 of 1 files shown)


jpoimboe:
It's too bad our lookup code doesn't have access to the name of the symbol's section, as it's a common pattern for discarded symbols to be placed into a ".discard.*" section.  Then we could just ignore all symbols belonging to such sections, which would be a lot less fragile than the current approach.

This fix looks good though.

PR #933, opened 2019-01-17
Fix NULL pointer deref in main due to base_locals

bsingharora:
For fun I tried to create a livepatch of upstream patch
ad211f3e94b314a910d4af03178a0b52a7d1ee0a for my kernel. This
caused kpatch-build to fail with a NULL pointer derefence because
base_locals was NULL (returned via kpatch_elf_locals(), which
can return a NULL pointer). This patch fixes the SIGSEGV
via a NULL check. The end result is a live patch is created
and loaded.

Signed-off-by: Balbir singh <bsingharora@gmail.com>


joe-lawrence:
Thanks, Balbir.

PR #934, opened 2019-01-28
integration/centos-7: fix shadow-newpid.patch warnings

sm00th:
When quickly loading/unloading this patch multiple times it is possible
to hit "Duplicate shadow variable" warnings since the patch doesn't have
any cleanup hooks on unload.
Switch to klp_shadow_get_or_alloc to ignore these.


sm00th:
> There's also a fedora-27 version of this test, does it not also need to be fixed?

fedora-27 uses kpatch api (they don't have livepatch enabled), which doesn't throw this warning.

PR #936, opened 2019-01-29
livepatch-patch-hook: skip klp_(un)register_patch() for v5.0+ kernels

joe-lawrence:
In v5.0, upstream kernel commit 958ef1e39d24 ("livepatch: Simplify API
by removing registration step") removed klp_(un)register_patch().  We
only need to call klp_enable_patch() now.

### DIFF
--- a/kmod/patch/livepatch-patch-hook.c
+++ b/kmod/patch/livepatch-patch-hook.c
@@ -64,6 +64,10 @@
 # define HAVE_CALLBACKS
 #endif
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 1, 0)
+# define HAVE_SIMPLE_ENABLE
+#endif
+
 /*
  * There are quite a few similar structures at play in this file:
  * - livepatch.h structs prefixed with klp_*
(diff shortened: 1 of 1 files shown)


joe-lawrence:
> Also, we may want to wait until this change is actually pulled into the main kernel tree (though I don't expect it to miss 5.0).

For some reason I thought this was going to make 5.0, but I just noticed Jiri's [pull request](https://lore.kernel.org/lkml/nycvar.YFH.7.76.1903051601280.19912@cbobk.fhfr.pm/) for 5.1.  I'll update the `KERNEL_VERSION(5, 0, 0)` line to 5.1 to account for that.

joe-lawrence:
- Updated version check and commit msg for 5.1

kamalesh-babulal:
@joe-lawrence #954 also touchs the function `patch_exit()` based on which gets merged first, the other patch needs to be rebased.

joe-lawrence:
@kamalesh-babulal thanks for the heads up... I was going to merge this one soon since v5.1 will be tagged soon, but I'll make it simpler for #954 and handle the rebase here.

PR #937, opened 2019-02-06
Add jump label support

### LINKED ISSUE #931: Livepatch does not handle static keys
TL;DR:
If the patched function has a condition with [static keys](https://www.kernel.org/doc/Documentation/static-keys.txt), the condition will always remain in its initial state there. It will not be switched by `static_key_slow_inc()/static_key_slow_dec()`. A workaround is to check the condition with `static_key_enabled()` in the patched code.

kpatch-build should at least detect such conditions and fail the build of the patch or build the patch so that the sure static keys are handled automatically.

---

Here is the long story.

While we were testing the live patch with [the fix for CVE-2018-9568](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/net/core/sock.c?id=9d538fa60bad4f7b23193c89e843797a1cf71ef3) for Virtuozzo 7, some of the tests caught kernel crashes due to incorrect reference counts of cgroups.

It took a while to debug but finally the root cause was found.
I reproduced the problem in CentOS 7.6 with the kernel 3.10.0-957.1.3.el7.

The source patch looked OK, it just added `newsk->sk_prot_creator = sk->sk_prot;` in `sk_clone_lock()`. It was far from obvious that `sk_update_clone()` was inlined into that function and made it into the patched `sk_clone_lock()`. It was even less obvious that sk_update_clone() used a static key condition:

net/core/sock.c:
```
static void sk_update_clone(const struct sock *sk, struct sock *newsk)
{
	if (mem_cgroup_sockets_enabled && sk->sk_cgrp) // here
		sock_update_memcg(newsk);
```
`mem_cgroup_sockets_enabled` is actually not a variable:
include/net/sock.h:
```
#define mem_cgroup_sockets_enabled static_key_false(&memcg_socket_limit_enabled)
```
The default state of the condition `if (mem_cgroup_sockets_enabled)` translates to a nop in the binary code:
```
   0xffffffffa82201fb:  0f 1f 44 00 00  nopl   0x0(%rax,%rax,1)  // here
   0xffffffffa8220200:  48 8b 43 28     mov    0x28(%rbx),%rax
<...>
   // the rest of the conditional, this part is not used by default
   0xffffffffa8220248:  49 83 bc 24 88 02 00 00 00      cmpq   $0x0,0x288(%r12)
   0xffffffffa8220251:  74 ad   je     0xffffffffa8220200
   0xffffffffa8220253:  48 89 df        mov    %rbx,%rdi
   0xffffffffa8220256:  e8 c5 06 c1 ff  callq  0xffffffffa7e30920  // sock_update_memcg
   0xffffffffa822025b:  eb a3   jmp    0xffffffffa8220200
```
When the static key is enabled by setting a limit via `memory.kmem.tcp.limit_in_bytes` in the memory cgroup, nop becomes a jump to the rest of the conditional:
```
   0xffffffffa82201fb:  e9 48 00 00 00  jmpq   0xffffffffa8220248  // here
   0xffffffffa8220200:  48 8b 43 28     mov    0x28(%rbx),%rax
<...>
   // the rest of the conditional is executed in this case
   0xffffffffa8220248:  49 83 bc 24 88 02 00 00 00      cmpq   $0x0,0x288(%r12)
   0xffffffffa8220251:  74 ad   je     0xffffffffa8220200
   0xffffffffa8220253:  48 89 df        mov    %rbx,%rdi
   0xffffffffa8220256:  e8 c5 06 c1 ff  callq  0xffffffffa7e30920  // sock_update_memcg
   0xffffffffa822025b:  eb a3   jmp    0xffffffffa8220200
```
This is how everything works for the original (not patched) binary code. So far so good.

The condition in `sk_clone_lock()` from the livepatch, however, remains in its default, disabled, state forever (checked that in the binary code of the loaded patch).

As a result, `sock_update_memcg()` will not be called for the given sockets, so `mem_cgroup_get(memcg)` will not be called either (see mm/memcontrol.c). 
```
void sock_update_memcg(struct sock *sk)
{
	if (mem_cgroup_sockets_enabled) {
<...>
		if (!mem_cgroup_is_root(memcg) && memcg_proto_active(cg_proto)) {
			mem_cgroup_get(memcg);
			sk->sk_cgrp = cg_proto;
		}
<...>
```
However, `sock_release_memcg()` is not patched, so `mem_cgroup_put(memcg)` will be called in time. 
```
void sock_release_memcg(struct sock *sk)
{
	if (mem_cgroup_sockets_enabled && sk->sk_cgrp) {
		struct mem_cgroup *memcg;
		WARN_ON(!sk->sk_cgrp->memcg);
		memcg = sk->sk_cgrp->memcg;
		mem_cgroup_put(memcg); // here
	}
}
```
Put() without get() => broken reference count and potential memory corruption.

---

To demonstrate the problem in CentOS 7.6, I have built a livepatch for the upstream fix for CVE-2018-9568 using kpatch rev. b29227547d6e.
A modified reproducer program from that commit can be used to trigger `sk_clone_lock()` ([test-clone.c.txt](https://github.com/dynup/kpatch/files/2691314/test-clone.c.txt))

    # kpatch load ./livepatch-CVE-2018-9568.ko 
    loading patch module: ./livepatch-CVE-2018-9568.ko
    waiting (up to 5 seconds) for patch transition to complete...
    transition complete (1 seconds)

    # mkdir /sys/fs/cgroup/memory/my_test/
    # echo 655360 > /sys/fs/cgroup/memory/my_test/memory.kmem.tcp.limit_in_bytes
    # sh
    # echo $$ > /sys/fs/cgroup/memory/my_test/tasks
    # /home/user/temp/test-clone

    (kernel crashed at this point)

`WARNING: CPU: 1 PID: 4004 at mm/page_counter.c:29 page_counter_cancel+0x42/0x50`, followed by a `BUG: unable to handle kernel paging request`, probably due to memory corruption.

mm/page_counter.c:29:
```
int page_counter_cancel(struct page_counter *counter, unsigned long nr_pages)
{
	long new;

	new = atomic_long_sub_return(nr_pages, &counter->count);

	/* More uncharges than charges? */
	WARN_ON_ONCE(new < 0);   // here

	return new > 0;
}
```
The full dmesg is attached: [vmcore-dmesg.txt](https://github.com/dynup/kpatch/files/2691312/vmcore-dmesg.txt)

---

A workaround for that particular patch is simple: just patch `sk_update_clone()` too, so that it checked the value of the static key explicitly:

```
--- a/net/core/sock.c
+++ b/net/core/sock.c
@@ -1470,7 +1470,7 @@ EXPORT_SYMBOL(sk_release_kernel);
 
 static void sk_update_clone(const struct sock *sk, struct sock *newsk)
 {
-	if (mem_cgroup_sockets_enabled && sk->sk_cgrp)
+	if (static_key_enabled(&memcg_socket_limit_enabled) && sk->sk_cgrp)
 		sock_update_memcg(newsk);
 }
```

However, I think, such things in the source patches are almost impossible to find during the code review. It would be better if kpatch-build could at least detect such conditions and fail the build of the patch or build the patch so that the sure static keys were handled automatically.

---

Files:
[test-clone.c.txt](https://github.com/dynup/kpatch/files/2691314/test-clone.c.txt)
[vmcore-dmesg.txt](https://github.com/dynup/kpatch/files/2691312/vmcore-dmesg.txt)

jpoimboe:
Add support for jump labels, also known as static jumps, static keys, static branches, and jump tables.  Luckily, kpatch_process_special_sections() is already generic enough to make this an easy fix.

Fixes #931.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -227,33 +227,40 @@ gcc_version_check() {
 }
 
 find_special_section_data_ppc64le() {
+
+	[[ "$CONFIG_JUMP_LABEL" -eq 0 ]] && AWK_OPTIONS="-vskip_j=1"
+
 	SPECIAL_VARS="$(readelf -wi "$VMLINUX" |
 		gawk --non-decimal-data '
-		BEGIN { f = b = e = 0 }
+		BEGIN { f = b = e = j = 0 }
 
 		# Set state if name matches
 		f == 0 && /DW_AT_name.* fixup_entry[[:space:]]*$/ {f = 1; next}
 		b == 0 && /DW_AT_name.* bug_entry[[:space:]]*$/ {b = 1; next}
 		e == 0 && /DW_AT_name.* exception_table_entry[[:space:]]*$/ {e = 1; next}
+		j == 0 && /DW_AT_name.* jump_entry[[:space:]]*$/ {j = 1; next}
 
 		# Reset state unless this abbrev describes the struct size
 		f == 1 && !/DW_AT_byte_size/ { f = 0; next }
 		b == 1 && !/DW_AT_byte_size/ { b = 0; next }
 		e == 1 && !/DW_AT_byte_size/ { e = 0; next }
+		j == 1 && !/DW_AT_byte_size/ { j = 0; next }
 
 		# Now that we know the size, stop parsing for it
 		f == 1 {printf("export FIXUP_STRUCT_SIZE=%d\n", $4); f = 2}
 		b == 1 {printf("export BUG_STRUCT_SIZE=%d\n", $4); b = 2}
 		e == 1 {printf("export EX_STRUCT_SIZE=%d\n", $4); e = 2}
+		j == 1 {printf("export JUMP_STRUCT_SIZE=%d\n", $4); j = 2}
 
 		# Bail out once we have everything
-		f == 2 && b == 2 && e == 2 {exit}')"
+		f == 2 && b == 2 && e == 2 && (j == 2 || skip_j) {exit}')"
 
 	[[ -n "$SPECIAL_VARS" ]] && eval "$SPECIAL_VARS"
 
 	[[ -z "$FIXUP_STRUCT_SIZE" ]] && die "can't find special struct fixup_entry size"
 	[[ -z "$BUG_STRUCT_SIZE" ]] && die "can't find special struct bug_entry size"
 	[[ -z "$EX_STRUCT_SIZE" ]]  && die "can't find special struct exception_table_entry size"
+	[[ -z "$JUMP_STRUCT_SIZE" && "$CONFIG_JUMP_LABEL" -ne 0 ]]  && die "can't find special struct jump_entry size"
 
 	return
 }
@@ -266,12 +273,13 @@ find_special_section_data() {
 
 	[[ "$CONFIG_PARAVIRT" -eq 0 ]] && AWK_OPTIONS="-vskip_p=1"
 	[[ "$CONFIG_UNWINDER_ORC" -eq 0 ]] && AWK_OPTIONS="$AWK_OPTIONS -vskip_o=1"
+	[[ "$CONFIG_JUMP_LABEL" -eq 0 ]] && AWK_OPTIONS="$AWK_OPTIONS -vskip_j=1"
 
 	# If $AWK_OPTIONS are blank gawk would treat "" as a blank script
 	# shellcheck disable=SC2086
 	SPECIAL_VARS="$(readelf -wi "$VMLINUX" |
 		gawk --non-decimal-data $AWK_OPTIONS '
-		BEGIN { a = b = p = e = o = 0 }
+		BEGIN { a = b = p = e = o = j = 0 }
 
 		# Set state if name matches
 		a == 0 && /DW_AT_name.* alt_instr[[:space:]]*$/ {a = 1; next}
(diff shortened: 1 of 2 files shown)


sm00th:
Looks like we'll need to update `CDO_ENV` in https://github.com/dynup/kpatch-unit-test-objs/blob/master/x86_64/Makefile

jpoimboe:
"make unit" works now

joe-lawrence:
I never got around to researching this, but if we kpatch func A which has a static key, toggle the key, and then disable/unload the kpatch, will func A reflect the new key value?  A simpler way to ask: static keys can be shared across kernel and modules and all key-sites are updated when they are toggled?

jpoimboe:
> I never got around to researching this, but if we kpatch func A which has a static key, toggle the key, and then disable/unload the kpatch, will func A reflect the new key value?

Yes, the original function will see the new key value.

> A simpler way to ask: static keys can be shared across kernel and modules and all key-sites are updated when they are toggled?

Yes, changing a static key will patch all the users of the key across all functions, in both kernel and modules.

kpatch-build/kpatch-build
```diff
@@ -304,6 +311,7 @@ find_special_section_data() {
 	[[ -z "$EX_STRUCT_SIZE" ]]  && die "can't find special struct paravirt_patch_site size"
 	[[ -z "$PARA_STRUCT_SIZE" && "$CONFIG_PARAVIRT" -ne 0 ]] && die "can't find special struct paravirt_patch_site size"
 	[[ -z "$ORC_STRUCT_SIZE" && "$CONFIG_UNWINDER_ORC" -ne 0 ]] && die "can't find special struct orc_entry size"
+	[[ -z "$JUMP_STRUCT_SIZE" ]] && die "can't find special struct jump_entry size"
```
joe-lawrence: Not a big deal since we generally target distro kernels, but I believe this introduces a new configuration requirement of `CONFIG_JUMP_LABEL=y`.  The commit looks fine, I just wanted to highlight this in case you might want to make it optional.

jpoimboe: Good point

joe-lawrence: Happy to approve/merge if you like it as is... alternatively we can make it optional or document it somewhere (though it's not like we've collected up and doc'd all the other CONFIG_ requirements to date.)

jpoimboe:
Updated the code to check for CONFIG_JUMP_LABEL

PR #938, opened 2019-02-08
test/unit: add support for .env files

sm00th:
Add support for optional <test-name>.env files that contain additional
environment variables. This can also be used to override per-arch env
vars we have because last value assigned will be used.


kamalesh-babulal:
> ... was scratching head wondering why we were running the "file" unix command here until I realized this was a Makefile function. It looks like a good fit for this usage and will gracefully give an empty string if `$(TNAME).$(EXT_ENV)` doesn't exist.

Thanks for adding this comment, I was also wondering the same. It all makes sense now.

PR #942, opened 2019-02-14
Out of tree module fixes

joe-lawrence:
Two small changes to out-of-tree module support.  

@pcd1193182 : were we building out-of-tree module kpatches differently when working on #923?  I just want to make sure I'm not breaking another use case with this change

### DIFF
--- a/README.md
+++ b/README.md
@@ -649,7 +649,7 @@ sys_nanosleep(), etc?**
 
 Yes! There's a few requirements, and the feature is still in its infancy.
 
-1. You need to use the `--out-of-tree` flag to specify the version of the
+1. You need to use the `--oot-module` flag to specify the version of the
 module that's currently running on the machine.
 2. `--sourcedir` has to be passed with a directory containing the same
 version of code as the running module, all set up and ready to build with a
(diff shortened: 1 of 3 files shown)


joe-lawrence:
> We can improve this paragraph by using the matching `flag options` in `Can you patch out-of-tree modules` steps and the `Sample invocation`. Having matching version of options would make it more relatable, instead of checking for the flags in the help.

Done.  I used the long format options (IMHO they are better for documentation purposes as they usually better describe the option than something like `-z`.)

README.md
```diff
@@ -659,15 +659,15 @@ currently-running module.
...
+4. Usually you'll need to pass the `--target` flag as well, to specify the
+proper `make` target names.
 5. This has only been tested for a single out-of-tree module per patch, and
 not for out-of-tree modules with dependencies on other out-of-tree modules
 built separately.
 
 ***Sample invocation***
 
-`kpatch-build -s ~/test/ -t default -e /lib/modules/$(uname -r)/extra/test.ko test.patch`
+`kpatch-build --sourcedir ~/test/ --target default -oot-module /lib/modules/$(uname -r)/extra/test.ko test.patch`
```
kamalesh-babulal: Minor nitpick , `--oot-module` is missing a hyphen.

joe-lawrence:
One more change since this PR is still open:  add --oot-module to the kpatch-build manpage file.

pcd1193182:
@joe-lawrence I don't think any of this should break the stuff I was working with. Changes look good in general, thanks for making this stuff more clear.

joe-lawrence:
No more PR comments, so merging this one.

PR #944, opened 2019-02-19
Revert "create-diff-object: add jump label support"

jpoimboe:
This reverts commit 87c64519fce02424e00725ddaeff8b619d40d160.

The jump label support doesn't work with upstream livepatch.  Joe
Lawrence found the following ordering issue:

```
load_module

  apply_relocations

    /* Livepatch relocation sections are applied by livepatch */
    if (info->sechdrs[i].sh_flags & SHF_RELA_LIVEPATCH)
            continue;

  post_relocation
    module_finalize
      jump_label_apply_nops        << crash

  ...

  do_init_module
    do_one_initcall(mod->init)
      __init patch_init [kpatch-patch]
        klp_register_patch
          klp_init_patch
            klp_init_object
              klp_init_object_loaded
                klp_write_object_relocations
```

So jump_label_apply_nops() is called *before*
klp_write_object_relocations() has had a chance to write the klp
relocations (.klp.rela.kvm_intel.__jump_table, for example).

We need to resolve this upstream first.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1771,21 +1771,6 @@ static int ex_table_group_size(struct kpatch_elf *kelf, int offset)
 	return size;
 }
 
-static int jump_table_group_size(struct kpatch_elf *kelf, int offset)
-{
-	static int size = 0;
-	char *str;
-
-	if (!size) {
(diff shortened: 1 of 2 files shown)


jpoimboe:
This reverts #937.

PR #945, opened 2019-02-21
README: use correct command line option for out-of-tree modules

rudis:


joe-lawrence:
Hi @rudis , thanks for opening a PR to fix the readme.  We have already opened PR #942 that corrects the flag as well as a small relative path fixup in kpatch-gcc wrapper script.  If you are currently building out of tree modules, can you comment on that PR to whether it breaks your current workflow?   Thanks!

rudis:
@joe-lawrence I missed that PR. I'm not building out of tree modules, I only noticed that typo and fixed it a few months ago (and forgot to create a PR then). So this PR can be closed.

PR #951, opened 2019-04-16
create-diff-objject: fix memleak of the struct lookup_table

wwheart:
reason: Firstly, in the function lookup_open use the malloc to
              allocate some memory, but call the function lookup_close
               to free the memory.
               Secondly, table->obj_sym->name, table->exp_sym->name and
               table->exp_sym->objname used the strdup, so them should
               free also.
               Thirdly, adjust the order of make_nodname, if not, it
               will cause an exception when free(exp_sym->objname) in
                lookup_close.

### DIFF
--- a/kpatch-build/lookup.c
+++ b/kpatch-build/lookup.c
@@ -178,19 +178,23 @@ static void find_local_syms(struct lookup_table *table, char *hint,
 /* Strip the path and replace '-' with '_' */
 static char *make_modname(char *modname)
 {
-	char *cur;
+	char *cur, *name;
 
 	if (!modname)
 		return NULL;
 
-	cur = modname;
+	name = strdup(basename(modname));
+	if (!name)
+		perror("strdup");
+
+	cur = name; /* use cur as tmp */
 	while (*cur != '\0') {
 		if (*cur == '-')
 			*cur = '_';
 		cur++;
 	}
 
-	return basename(modname);
+	return name;
 }
 
 static void symtab_read(struct lookup_table *table, char *path)
@@ -305,11 +309,7 @@ static void symvers_read(struct lookup_table *table, char *path)
 		if (!symname)
 			perror("strdup");
 
-		objname = strdup(mod);
-		if (!objname)
-			perror("strdup");
-		/* Modifies objname in-place */
-		objname = make_modname(objname);
+		objname = make_modname(mod);
 
 		table->exp_syms[i].name = symname;
 		table->exp_syms[i].objname = objname;
@@ -338,7 +338,18 @@ struct lookup_table *lookup_open(char *symtab_path, char *symvers_path,
 
 void lookup_close(struct lookup_table *table)
 {
+	int i;
+	struct object_symbol *obj_sym;
+	struct export_symbol *exp_sym;
+
+	for_each_obj_symbol(i, obj_sym, table)
+		free(obj_sym->name);
 	free(table->obj_syms);
+
+	for_each_exp_symbol(i, exp_sym, table) {
+		free(exp_sym->name);
+		free(exp_sym->objname);
+	}
 	free(table->exp_syms);
 	free(table);
 }
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -3378,6 +3378,7 @@ int main(int argc, char *argv[])
 	kpatch_dump_kelf(kelf_out);
 	kpatch_write_output_elf(kelf_out, kelf_patched->elf, output_obj);
 
+	lookup_close(lookup);
 	kpatch_elf_free(kelf_patched);
 	kpatch_elf_teardown(kelf_out);
 	kpatch_elf_free(kelf_out);


kpatch-build/lookup.c
```diff
@@ -305,11 +305,11 @@ static void symvers_read(struct lookup_table *table, char *path)
 		if (!symname)
 			perror("strdup");
 
-		objname = strdup(mod);
+		/* Modifies objname in-place */
+		objname = make_modname(mod);
```
sm00th: I don't think this is the right way to fix the issue you are seeing. make_modname() modifies the variable in-place so this way it will modify "mod" and we don't want that.

The pointer returned by basename() in make_modname() is not supposed to be freed, here's an excerpt from the man page:
```
RETURN VALUE
       Both dirname() and basename() return pointers to null-terminated strings.  (Do not pass these pointers to free(3).)
```

Looks like make_modname() needs to be a bit smarter and do something like:
```
tmp = strdup(modname);
<fixup tmp>
ret = strdup(basename(tmp));
free(tmp)
return ret;
```

wwheart: yet, I have changed, please reviev again, thanks!

kpatch-build/lookup.c
```diff
@@ -338,7 +341,22 @@ struct lookup_table *lookup_open(char *symtab_path, char *symvers_path,
 
 void lookup_close(struct lookup_table *table)
 {
+	int i;
+	struct object_symbol *obj_sym;
+	struct export_symbol *exp_sym;
+
+	for_each_obj_symbol(i, obj_sym, table) {
+		if (obj_sym->name != NULL)
```
kamalesh-babulal: The code can be simplified by removing the checks for `NULL`. Calling `free(NULL)` is no-op.

wwheart: yet, I have removed the checks for NULL, please review again, thanks !

kpatch-build/lookup.c
```diff
@@ -178,19 +178,25 @@ static void find_local_syms(struct lookup_table *table, char *hint,
...
 	while (*cur != '\0') {
 		if (*cur == '-')
 			*cur = '_';
 		cur++;
 	}
 
-	return basename(modname);
+	name = basename(cur);
+	name = strdup(name);
+	free(cur);
```
kamalesh-babulal: Sorry, I should have done a careful review before, the code segfaults at `free (cur)`.  Segfault happens because the string start position (length) of `cur` is different while allocation vs free'ing,  that also mean the basename is wrong.  I was curious and checking the memory saving, with and without the patch:
```
Before:
==12851== LEAK SUMMARY:
==12851==    definitely lost: 2,257,092 bytes in 94,142 blocks
==12851==    indirectly lost: 6,044,367 bytes in 114,647 blocks

After: (fixing the segfault by fixing cur using tmp variable)
==13365==    definitely lost: 2,257,060 bytes in 94,141 blocks
==13365==    indirectly lost: 454,869 bytes in 74 blocks
```
This patch does fixes huge amount memory leak. I guess the memory leak seen after the patch due to the un-freed allocation from other paths.   One way to fix is, the segfault:
```
@@ -187,12 +187,14 @@ static char *make_modname(char *modname)
        if (!cur)
                perror("strdup");
 
+       name = cur; /* use name as tmp */
        while (*cur != '\0') {
                if (*cur == '-')
                        *cur = '_';
                cur++;
        }
 
+       cur = name;
        name = basename(cur);
        name = strdup(name);
        free(cur);
```

kpatch-build/lookup.c
```diff
@@ -338,7 +341,19 @@ struct lookup_table *lookup_open(char *symtab_path, char *symvers_path,
 
 void lookup_close(struct lookup_table *table)
 {
+	int i;
+	struct object_symbol *obj_sym;
+	struct export_symbol *exp_sym;
+
+	for_each_obj_symbol(i, obj_sym, table) {
+			free(obj_sym->name);
+	}
```
kamalesh-babulal: Please fix the indentation here and in the following block. We do not require the braces for single statement in the for loop.

kpatch-build/lookup.c
wwheart: Sorry， I am careless, thank you for your careful review.

kpatch-build/lookup.c
wwheart: Have fixed your review, thanks !

kpatch-build/lookup.c
```diff
@@ -305,11 +313,8 @@ static void symvers_read(struct lookup_table *table, char *path)
 		if (!symname)
 			perror("strdup");
 
-		objname = strdup(mod);
-		if (!objname)
-			perror("strdup");
 		/* Modifies objname in-place */
```
jpoimboe: This comment can be removed

wwheart: OK，The comment is removed now. Thanks!

kpatch-build/lookup.c
jpoimboe: Sorry, I thought I left this comment earlier, but it got lost somehow -- maybe I forgot to submit it.

I think this code could be simplified a bit, like:
```
	name = strdup(basename(modname));
	if (!name)
		perror("strdup");

	cur = name;
	while (*cur != '\0') {
		if (*cur == '-')
			*cur = '_';
		cur++;
	}

	return name;
```

wwheart: OK, it's good. Thanks！

PR #952, opened 2019-04-20
fix memleak in the create-klp-module.c

wwheart:
reason: The strdup() function returns a pointer to a new string
	     which is a duplicate of the string s.  Memory for the
	     new string is obtained with malloc, and can be freed
	     with free.

	     here, the name and objname do not use anymore, so it
	     should be free.

### DIFF
--- a/kpatch-build/create-klp-module.c
+++ b/kpatch-build/create-klp-module.c
@@ -57,9 +57,7 @@ static struct symbol *find_or_add_ksym_to_symbols(struct kpatch_elf *kelf,
 	if (!rela)
 		ERROR("name of ksym not found?");
 
-	name = strdup(strings + rela->addend);
-	if (!name)
-		ERROR("strdup");
+	name = strings + rela->addend;
 
 	/* Get objname of ksym */
 	rela = find_rela_by_offset(ksymsec->rela,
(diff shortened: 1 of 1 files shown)


kpatch-build/create-klp-module.c
```diff
@@ -74,6 +74,8 @@ static struct symbol *find_or_add_ksym_to_symbols(struct kpatch_elf *kelf,
 	snprintf(pos, 32, "%lu", ksym->pos);
 	/* .klp.sym.objname.name,pos */
 	snprintf(buf, 256, KLP_SYM_PREFIX "%s.%s,%s", objname, name, pos);
+	free(name);
+	free(objname);
```
jpoimboe: There's no need for the original strdup's in the first place.  e.g.
```
objname = strdup(strings + rela->addend);
```
can just be simplified to
```
objname = strings + rela->addend;
```
And the same comment applies for the other changes.

wwheart: Yet, it's good ideas, it make the code more simple. Thank you for your review.

wwheart:
> You missed similar changes in other functions, which were changed in the previous version of this patch.

Yet, I have seen the commit merged. This maybe have other memleak.

jpoimboe:
@wwheart Even if they were merged, if you could clean them up like this commit was by removing the unnecessary strdups, that would be great.

wwheart:
> @wwheart Even if they were merged, if you could clean them up like this commit was by removing the unnecessary strdups, that would be great.

Yet, you will continue review the code, if I find any issues, I will push the commit agian.

wwheart:
> @wwheart Even if they were merged, if you could clean them up like this commit was by removing the unnecessary strdups, that would be great.

I have found more two unnecessary strdups, I have updated the commit, yet.

PR #954, opened 2019-04-30
kmod: fix memleak in the function patch_exit

wwheart:
reason: after calling the function klp_unregister_patch in patch_exit,
             the lpatch must be freed, otherwise, it would cause memory leak.

### DIFF
--- a/kmod/patch/livepatch-patch-hook.c
+++ b/kmod/patch/livepatch-patch-hook.c
@@ -460,6 +460,7 @@ out:
 static void __exit patch_exit(void)
 {
 	WARN_ON(klp_unregister_patch(lpatch));
+	patch_free_livepatch(lpatch);
 }
 
 module_init(patch_init);


wwheart:
> I would prefer the commit title to be `livepatch-patch-hook: fix memleak in the function patch_exit`. Which indicates the file, being modified over a directory name.

OK！

joe-lawrence:
Looks good to me... two quick questions: 1) this fix looks fairly obvious, but just wondering if you tested with slub_debug=FZPU just to ensure no use-after-frees and 2) was this found via code inspection, or did you use any other debug build options to discover the leaked data structures.   Thanks!

kamalesh-babulal:
> 1) this fix looks fairly obvious, but just wondering if you tested with slub_debug=FZPU just to ensure no use-after-frees.

I tested it with `slub_debug=FZPU`, found no use-after-frees issue.

wwheart:
I am in the holidays, when I insmod and rmmod the livepatch module for 10000 times, I found the system memory inceasing. I can test by slub debug later.

joe-lawrence:
Thanks for the fix and testing, guys.  I wonder if `CONFIG_DEBUG_KMEMLEAK` would find any additional leaks that might be lurking in the code?  I've never used it, so that's why I was curious to know how these past few were discovered.

PR #957, opened 2019-05-09
create-diff-object: remove unneeded condition

wwheart:
The create-diff-object don't create .kpatch.dynrelas and
.rela.kpatch.dynrelas sections which are removed to
create-kpatch-module, so the condition is unneeded yet.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2748,7 +2748,8 @@ static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
 		if (!is_rela_section(sec))
 			continue;
 		if (!strcmp(sec->name, ".rela.kpatch.funcs") ||
-		    !strcmp(sec->name, ".rela.kpatch.dynrelas"))
+		    !strcmp(sec->name, ".rela.kpatch.relocations") ||
+		    !strcmp(sec->name, ".rela.kpatch.symbols"))
 			continue;
 		list_for_each_entry_safe(rela, safe, &sec->relas, list) {
 			if (!rela->need_dynrela)


joe-lawrence:
87643703a757 ("create-diff-object: create .kpatch.relocations and .kpatch.symbols sections") modified create-diff-object.c to create intermediate ".kpatch.relocations" sections instead of ".kpatch.dynrelas" sections.

Instead of removing the check, should the loop have been updated to skip the newly created ".rela.kpatch.relocations" and ".rela.kpatch.symbols" sections?

wwheart:
> [8764370](https://github.com/dynup/kpatch/commit/87643703a75746e03c6942f7ecc722113fa21783) ("create-diff-object: create .kpatch.relocations and .kpatch.symbols sections") modified create-diff-object.c to create intermediate ".kpatch.relocations" sections instead of ".kpatch.dynrelas" sections.
> 
> Instead of removing the check, should the loop have been updated to skip the newly created ".rela.kpatch.relocations" and ".rela.kpatch.symbols" sections?

I had think this, the ".rela.kpatch.relocations" and ".rela.kpatch.symbols" sections are created after the need_dynrela is set, so these sections's need_dynrela are false,  which will be skip later.  The code is follow:

2717                          * in (.rela.toc)->need_dynrela is set, as they represent the function sym.
2718                          */
2719                         if (may_need_dynrela(rela))
2720                                 toc_rela(rela)->need_dynrela = 1;--------------------need_dynrela  is set here
2721                 }
2722         }
2723
2724         /* create .kpatch.relocations text/rela section pair */
2725         krela_sec = create_section_pair(kelf, ".kpatch.relocations", sizeof(*krelas), nr);--------section is created here.
2726         krelas = krela_sec->data->d_buf;
2727
2728         /* create .kpatch.symbols text/rela section pair */
2729         ksym_sec = create_section_pair(kelf, ".kpatch.symbols", sizeof(*ksyms), nr);
2730         ksyms = ksym_sec->data->d_buf;
2731
2732         /* create .kpatch.symbols section symbol (to set rela->sym later) */
2733         ALLOC_LINK(ksym_sec_sym, &kelf->symbols);
2734         ksym_sec_sym->sec = ksym_sec;
2735         ksym_sec_sym->sym.st_info = GELF_ST_INFO(STB_LOCAL, STT_SECTION);
2736         ksym_sec_sym->type = STT_SECTION;
2737         ksym_sec_sym->bind = STB_LOCAL;
2738         ksym_sec_sym->name = ".kpatch.symbols";
2739
2740         /* lookup strings symbol */
2741         strsym = find_symbol_by_name(&kelf->symbols, ".kpatch.strings");
2742         if (!strsym)
2743                 ERROR("can't find .kpatch.strings symbol");
2744
2745         /* populate sections */
2746         index = 0;
2747         list_for_each_entry(sec, &kelf->sections, list) {
2748                 if (!is_rela_section(sec))
2749                         continue;
2750                 if (!strcmp(sec->name, ".rela.kpatch.funcs"))
2751                         continue;
2752                 list_for_each_entry_safe(rela, safe, &sec->relas, list) {
2753                         if (!rela->need_dynrela)
2754                                 continue;                            -----------------------------skip here.
2755
2756                         /*
2757                          * Allow references to core module symbols to remain as

kamalesh-babulal:
> 8764370 ("create-diff-object: create .kpatch.relocations and .kpatch.symbols sections") modified create-diff-object.c to create intermediate ".kpatch.relocations" sections instead of ".kpatch.dynrelas" sections.

> Instead of removing the check, should the loop have been updated to skip the newly created ".rela.kpatch.relocations" and ".rela.kpatch.symbols" sections?

AFAIK, with the current approach we will only loop through `.rela.kpatch.funcs` section and relocation entries from that section, are the onces which would have `rela->need_dynrela` set.  If we consider skipping newly created `.rela.kpatch.relocations` and `.rela.kpatch.symbols` sections alone, we would loop through all the relocation sections for which `need_dynrela` is never set.  Am I missing something in the code path.

joe-lawrence:
The original code with added comments:
```C
        /* Loop through all sections */
        list_for_each_entry(sec, &kelf->sections, list) {

                /* Skip non-rela sections */
                if (!is_rela_section(sec))
                        continue;

                /* Skip .rela.kpatch funcs and rela.kpatch.dynrelas */
                if (!strcmp(sec->name, ".rela.kpatch.funcs") ||
                    !strcmp(sec->name, ".rela.kpatch.dynrelas"))
                        continue;

                /* Loop through all of this section's relocations */

                        /* Skip relas that don't need dynrela */
                        if (!rela->need_dynrela)
                                continue;
```
@wwheart : you're right that we could rely on the need_dynrela check, but if we're already specifically skipping over ".rela.kpatch.funcs", why not add those new sections to the same check?  It would avoid traversing through their relocations in the first place (as was done prior to 8764370 ("create-diff-object: create .kpatch.relocations and .kpatch.symbols sections")).

@kamalesh-babulal : "Am I missing something in the code path."  Aren't those `!strcmp`'s going to skip said sections?  ie, am I interpreting those loop short-circuits backwards?

Note: I'm not super familiar with this part of the code, so I'm only asking this question having looked at this PR and then the previous one noted to see what changed.

jpoimboe:
Good catch @joe-lawrence.  I agree with that we should be skipping .rela.kpatch.relocations.  And similarly we should also be skipping .rela.kpatch.symbols.

This reminds me that I have an old branch which cleans up this code a bit, which I need to dust off.  Funnily enough, it includes this exact patch.

kamalesh-babulal:
> Aren't those !strcmp's going to skip said sections? ie, am I interpreting those loop short-circuits backwards?

@joe-lawrence /me hides in a brown paper bag, I shouldn't be reviewing in a hurry while travelling. Yes, its the other way around. Agree that we should be skipping ` .rela.kpatch.relocations` and `.rela.kpatch.symbols` sections.

joe-lawrence:
No worries @kamalesh-babulal, those strcmp's get me every time and I have to reread to make sure I've got the match vs. no-match cases straight.

PR #962, opened 2019-05-15
Update fixup-section patches

sm00th:
These patches previously would try to dereference a userspace pointer
directly which can cause issue on some systems. Since the aim of these
patches is to check fixup section changes we can don't need to do any
meaningful changes to the function, so change the patches to just add
asm("nop") instead.


sm00th:
> The fixup-section objects in kpatch-unit-test-objs, also need to updated using this patch.

Hm. I think we can skip that since that code never executes and we only use those to check if create-diff-object still functions properly. So if there is no real issue we are hitting with those I'd rather refrain from updating them because it is another ~100Kb in git repo.

kamalesh-babulal:
> Hm. I think we can skip that since that code never executes and we only use those to check if create-diff-object still functions properly.

Sorry, for not being clear. I meant to say replace the old objects with new objects. It was just to have the object files updated when the relvant testcase changed. Agree, old objects serves the purpose. We may very well skip the update.

PR #968, opened 2019-05-28
kpatch-build: use readelf instead of eu-readelf

wwheart:
readelf is more standard, using readelf insteaded we should solve there
issues:

First, using "readelf -s", the symbol name would truncated by 25 chars,
to solve this issue, add option "--wide".

Second, the size may be mixed of decimal and hex, we get the size by "%s",
and use strtoul(size, NULL, 0) to convert the size.

Third, the symbol type is SHN_UNDE, the Ndx display "UND", so changed to
compare with "UND".

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -845,7 +845,7 @@ for i in $FILES; do
 			SYMVERS_FILE="$SRCDIR/Module.symvers"
 		fi
 
-		eu-readelf -s "$KOBJFILE_PATH" > "$SYMTAB"
+		readelf -s --wide "$KOBJFILE_PATH" > "$SYMTAB"
 
 		# create-diff-object orig.o patched.o parent-name parent-symtab
 		#		     Module.symvers patch-mod-name output.o
--- a/kpatch-build/lookup.c
+++ b/kpatch-build/lookup.c
@@ -200,17 +200,17 @@ static char *make_modname(char *modname)
 static void symtab_read(struct lookup_table *table, char *path)
 {
 	FILE *file;
-	long unsigned int value, size;
+	long unsigned int value;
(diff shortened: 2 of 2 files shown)


sm00th:
Looks good and seems to pass integration tests. Can we also "fix" the other two instances of eu-readelf in this pull-request?

```
$ git grep eu-readelf
test/difftree.sh:       [[ $(eu-readelf -s $i | grep FILE | wc -l) -ne 1 ]] && continue
test/unit/Makefile.include:     eu-readelf -s $(patsubst %.$(EXT_SYMTAB),%.$(EXT_ORIG),$(@)) >$@
```

Looks like both can be substituted with `readelf -s --wide` as well.

wwheart:
> Looks good and seems to pass integration tests. Can we also "fix" the other two instances of eu-readelf in this pull-request?
> 
> ```
> $ git grep eu-readelf
> test/difftree.sh:       [[ $(eu-readelf -s $i | grep FILE | wc -l) -ne 1 ]] && continue
> test/unit/Makefile.include:     eu-readelf -s $(patsubst %.$(EXT_SYMTAB),%.$(EXT_ORIG),$(@)) >$@
> ```
> 
> Looks like both can be substituted with `readelf -s --wide` as well.

In my view, it's better to modity these two in another new pull-request.
and  "[[ $(eu-readelf -s $i | grep FILE | wc -l) -ne 1 ]] && continue" may be changed to  
"[[ $(readelf -s sub_test.ko | awk -F " " '{print $4}' | grep FILE | wc -l) -ne 1 ]]" && continue"
because the "$i" maybe contain a symbol name with "FILE".

wwheart:
> If it is more convenient to do it in a separate pull-request that is fine with me.

yet, I should push another pull-request after this commit is merged.

PR #969, opened 2019-05-29
test: use readelf instead of eu-readelf

wwheart:
readelf is more standard, using readelf insteaded

test/difftree.sh: the symbol name may be with "FILE", which may be
	get a incorrect count, here add "awk".

test/unit/Makefile.include: use "readelf -s --wide" instead of
	"eu-readelf -s".

### DIFF
--- a/test/difftree.sh
+++ b/test/difftree.sh
@@ -46,7 +46,7 @@ do
 		;;
 	esac
 	# skip objects that are the linked product of more than one object file
-	[[ $(eu-readelf -s $i | grep FILE | wc -l) -ne 1 ]] && continue
+	[[ $(readelf -s $i | awk '$4=="FILE" {n++} END {print n}') -ne 1 ]] && continue
 	$SCRIPTDIR/../kpatch-build/create-diff-object $i $i /usr/lib/debug/lib/modules/$(uname -r)/vmlinux "$TEMPDIR/output.o" > "$TEMPDIR/log.txt" 2>&1
 	RETCODE=$?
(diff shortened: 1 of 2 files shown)


test/difftree.sh
```diff
@@ -46,7 +46,7 @@ do
 		;;
 	esac
 	# skip objects that are the linked product of more than one object file
-	[[ $(eu-readelf -s $i | grep FILE | wc -l) -ne 1 ]] && continue
+	[[ $(readelf -s $i | awk -F " " '{print $4}' | grep FILE | wc -l) -ne 1 ]] && continue
```
joe-lawrence: This could be simplified a little bit by incorporating the filtering and counting into the awk script:
```
readelf -s $(find . -name vmlinux) | awk '$4=="FILE" {n++} END {print n}'
```

wwheart: > This could be simplified a little bit by incorporating the filtering and counting into the awk script:
> 
> ```
> readelf -s $(find . -name vmlinux) | awk '$4=="FILE" {n++} END {print n}'
> ```

yet, but I don't think we can use "$(find . -name vmlinux)" replace the "$i".

joe-lawrence:
@wwheart : yea, you are right about using `$i` ... I had just copied/pasted from my terminal using vmlinux as an example.

Btw, I wonder if using awk and an exact column match on the Type field ($4) instead of a simple grep for "FILE" fixes a possible bug where a symbol name could have been "FILE".  Probably unlikely, but using the field match tightens things up a bit.  Thanks for updating.

PR #971, opened 2019-06-04
lookup: Fix format string for symtab_read() on PPC64LE

kamalesh-babulal:
commit 767d9669bd13 ("kpatch-build: use readelf instead of eu-readelf")
replaced eu-readelf with readelf for constructing symbol table.  The
format of symbol table entries differs a little on Power when the symbol
is a function with binding type LOCAL.  For example, consider:

23: 0000000000000008    96 FUNC    LOCAL  DEFAULT [<localentry>: 8]    4 cmdline_proc_show

An extra column preceding index of the symbol denoting symbol value to
be local entry point offset of the function is printed, with the
current sscanf format string in lookup::symtab_read the values will
mismatch ending with in-accurate lookup table getting constructed. This
patch fixes it by introducing a Power specific format string for
function symbols with bind type LOCAL.

Fixes: 767d9669 ("kpatch-build: use readelf instead of eu-readelf")

### DIFF
--- a/kpatch-build/lookup.c
+++ b/kpatch-build/lookup.c
@@ -209,8 +209,18 @@ static void symtab_read(struct lookup_table *table, char *path)
 		ERROR("fopen");
 
 	while (fgets(line, 256, file)) {
-		matched = sscanf(line, "%*s %lx %s %s %s %*s %s %s\n",
-				 &value, size, type, bind, ndx, name);
+		if (strstr(line, "[<localentry>: 8]")) {
+			/*
+			 * 23: 0000000000000008    96 FUNC    LOCAL  DEFAULT [<localentry>: 8]     4 cmdline_proc_show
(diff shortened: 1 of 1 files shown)


joe-lawrence:
Looks reasonable to me, will the 8 offset in "[<localentry>: 8]" ever change?

kamalesh-babulal:
Offset 8 means that the localentry point is 2 instructions past the global entry point of the function and it remains consistent across kernel and shall not change, as the function prolouge sequence is:
```
my_func:
    addis r2, r12, (.TOC.-my_func)@ha
    addi r2, r2, (.TOC.-my_func)@l
    .localentry my_func, .-my_func
    ... ; function definition
    blr
```
The offset is encoded in the three most significant bits of `sym->st_other` field and hence the output
from `readelf -s` decodes the information about the localentry.

Issue #967, opened 2019-05-28
documentation: i miss an information about the lifetime of a livepatch module


disaster123:
I'm not sure whether the topic is clear most probably not.

What i'm missing is when should a module be loaded, when unloaded? Something like a best practice.

For example if i build a newer livepatch for my kernel which also contains the changes of the old livepatch. Should i just load the newer one or should i unload the old one before?

This one https://github.com/dynup/kpatch/blob/master/contrib/kpatch.service also contains an unload on stop which means, when the server reboots or shuts down it will unload all patches. Is this really a good idea on shutdown? It may contain fixes for unmounting filesystems, ...

jpoimboe:
If you have a cumulative livepatch which is a superset of the old one, the old one can just be left loaded.  That way there's no window of time where the system is unpatched.  And also you don't have to unload the patch, which may introduce additional risk.  It can just be left inert, in memory.

Upstream livepatch now has a better way to deal with this.  If you set klp_patch.replace, the new patch will automatically replace and disable the old patch(es).  This is the best way to do it because a) any compiler optimization changes between the patches could potentially change which functions are affected; and b) it allows you to unload the old patch afterwards, if you want.

Please feel free to suggest any changes to the patch author guide to make it clearer.

Also, that kpatch.service file looks like a bug - I don't think it's a good idea to unload the patch when shutting down, as it introduces unnecessary risk.

disaster123:
Thanks a lot for clarification. I think your text might just go directly to the doc ;-)

regarding klp_patch.replace - i found:
https://lore.kernel.org/patchwork/patch/921683/

but this is not in our 4.19 kernel so i need to stick to the default behaviour.

PR #972, opened 2019-06-10
contrib/service: don't unload modules on stop

joe-lawrence:
The kpatch.service file shouldn't unload patch modules on service stop
(this is also executed by systemd on reboot).  Patch modules may not be
designed to be safely unloaded and/or may patch kernel routines that
need to continue to run throughout system bring down.

Suggested-by: disaster123

### DIFF
--- a/contrib/kpatch.service
+++ b/contrib/kpatch.service
@@ -6,7 +6,6 @@ ConditionKernelCommandLine=!kpatch.enable=0
 Type=oneshot
 RemainAfterExit=yes
 ExecStart=PREFIX/sbin/kpatch load --all
-ExecStop=PREFIX/sbin/kpatch unload --all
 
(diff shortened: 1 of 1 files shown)


joe-lawrence:
@disaster123: let me know if you prefer a full name / email Suggested-by attribution.

jpoimboe:
I think we can add "Fixes #967" to the commit message.

joe-lawrence:
> I think we can add "Fixes #967" to the commit message.

I read #967 as mostly a documentation request.  We can add something similar to your text from https://github.com/dynup/kpatch/issues/967#issuecomment-496596042 to this PR if you like.

jpoimboe:
Oh right, I forgot that part of the issue.  Never mind :-)

PR #973, opened 2019-06-11
kpatch-build: remove localentry data from ppc64le symtab

joe-lawrence:
```
commit f8213c87f64a ("lookup: Fix format string for symtab_read() on
PPC64LE") fixed the symbol table lookup when readelf adds ppc64le
"[<localentry>: 8]" info for functions like so:

  23: 0000000000000008    96 FUNC    LOCAL  DEFAULT [<localentry>: 8]    4 cmdline_proc_show

however, it seems that readelf 2.30-57.el8 displays this in a slightly
different format:

  24493: c000000000587970    96 FUNC    LOCAL  DEFAULT    2 cmdline_proc_show    [<localentry>: 8]

Instead of adding more cases to kpatch-build's lookup.c scanf format,
let's just delete this information from the symtab file with a quick and
dirty sed regex.  This allows us to handle both observed cases (and
perhaps others) while removing the arch-specific scanf formatting in
lookup.c

Fixes: f8213c87f64a ("lookup: Fix format string for symtab_read() on PPC64LE")
```

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -846,6 +846,9 @@ for i in $FILES; do
 		fi
 
 		readelf -s --wide "$KOBJFILE_PATH" > "$SYMTAB"
+		if [[ "$ARCH" = "ppc64le" ]]; then
+			sed -ri 's/\s+\[<localentry>: 8\]//' "$SYMTAB"
+		fi
 
 		# create-diff-object orig.o patched.o parent-name parent-symtab
 		#		     Module.symvers patch-mod-name output.o
--- a/kpatch-build/lookup.c
+++ b/kpatch-build/lookup.c
@@ -209,18 +209,8 @@ static void symtab_read(struct lookup_table *table, char *path)
 		ERROR("fopen");
 
 	while (fgets(line, 256, file)) {
-		if (strstr(line, "[<localentry>: 8]")) {
-			/*
-			 * 23: 0000000000000008    96 FUNC    LOCAL  DEFAULT [<localentry>: 8]     4 cmdline_proc_show
-			 * On Power, local func syms Ndx is preceded with "[<localentry>: 8]"
-			 * denoting local entry point offset of the function.
-			 */
-			matched = sscanf(line, "%*s %lx %s %s %s %*s [<localentry>: 8] %s %s\n",
-					 &value, size, type, bind, ndx, name);
-		} else {
-			matched = sscanf(line, "%*s %lx %s %s %s %*s %s %s\n",
-					 &value, size, type, bind, ndx, name);
-		}
+		matched = sscanf(line, "%*s %lx %s %s %s %*s %s %s\n",
+				 &value, size, type, bind, ndx, name);
 
 		if (matched == 5) {
 			name[0] = '\0';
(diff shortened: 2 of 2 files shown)


joe-lawrence:
Hey @kamalesh-babulal ... I stumbled upon this when testing out a simple [demo.patch.txt](https://github.com/dynup/kpatch/files/3278384/demo.patch.txt) on RHEL-8.1.  The error I got was:
```
/root/kpatch/kpatch-build/create-diff-object: ERROR: cmdline.o: find_local_syms: 175: find_local_syms for cmdline.c: couldn't find in vmlinux symbol table
```
and I noticed that the localentry data was in a different column:
```
% grep '\<cmdline_proc_show\>' ~/.kpatch/tmp/vmlinux.symtab 
 24493: c000000000587970    96 FUNC    LOCAL  DEFAULT    2 cmdline_proc_show    [<localentry>: 8]
```

This PR is very lightly tested, but let me know what you think of the change.

kpatch-build/kpatch-build
```diff
@@ -846,6 +846,9 @@ for i in $FILES; do
 		fi
 
 		readelf -s --wide "$KOBJFILE_PATH" > "$SYMTAB"
+		if [[ "$ARCH" = "ppc64le" ]]; then
+			sed -i 's/\[<localentry>: 8\]//' "$SYMTAB"
```
sm00th:
```suggestion
			sed -ri 's/\s+\[<localentry>: 8\]//' "$SYMTAB"
```

I think we might want to eat up preceding whitespaces as well.

kamalesh-babulal:
I tired reading the symtab using the readelf shipped with RHEL-8.1 and upstream readelf
```
# readelf --version
GNU readelf version 2.30-54.el8

# readelf -s --wide ./fs/proc/cmdline.o
...
14: 0000000000000000    96 FUNC    LOCAL  DEFAULT    1 cmdline_proc_show    [<localentry>: 8]

# ./binutils/readelf -v
GNU readelf (GNU Binutils) 2.32.51.20190612

# ./binutils/readelf -s --wide /home/kamalesh/linux/fs/proc/cmdline.o
...
23: 0000000000000008    96 FUNC    LOCAL  DEFAULT [<localentry>: 8]     4 cmdline_proc_show
```
Looks like some custom patch is breaking the order on RHEL. I liked the idea of the patch.

kpatch-build/kpatch-build
joe-lawrence: Yeah, the extra whitespace tab... I wasn't sure how to handle both RHEL-8 and @kamalesh-babulal 's originally reported format cases.  One uses `\t[<localentry>: 8\]` and the other (I think) `[<localentry>: 8\]\t`.  Can you think of a regex that handles both cases without concatenating columns?

sm00th: I think only handling preceding tab case should be enough since we will never have `[<localentry>: 8]` at the beginning of the line.

joe-lawrence:
v2
-- Updated with @sm00th's suggestion to remove leading whitespace

Also, @kamalesh-babulal was curious about the origins of the RHEL-specific behavior... as far as I can tell it was included with RHEL8's version of binutils (and will be carried forward) to move the "other" symbol information to the *end* of the line.  The intent was to not break old scripts that expect to see readelf symbol format in particular columns..  *ahem*  :)

Issue #974, opened 2019-06-12
powerpc: support alternatives (aka feature fixups)


jpoimboe:
As mentioned in #963, I think we have two gaps in powerpc support related to alternatives ("feature fixups"):

1) kpatch-build doesn't create .klp.arch sections
2) the kernel powerpc arch livepatch code doesn't read .klp.arch sections

At least in theory this means that if any of the feature fixup sections refers to a patched function, and the alternative original code has klp relas associated with it, code corruption may occur when the klp relas clobber the copied alternative replacement code, thinking that it's the alternative original code.

Phase 1 fix:
- For kernel versions which don't have livepatch-specific powerpc code (currently all kernel versions), kpatch-build needs to assert an error if it detects that one of the following sections refers to a patched function: __ftr_fixup, __mmu_ftr_fixup, __fw_ftr_fixup.

Phase 2 fix:
- The upstream kernel needs a powerpc implementation of arch_klp_init_object_loaded().
- kpatch-build needs to create klp.arch sections for powerpc for those kernel versions, instead of the phase 1 error

Phase 3 fix:
- Convince upstream that loading patches for modules which haven't been loaded yet is still a bad idea.  The easiest solution is just to modprobe all target modules before loading the patch.
- Remove all these .klp section hacks

joe-lawrence:
The .klp arch sections are an interesting idea, but I wonder if they will be a never-ending source of weird bugs since they kind of violate a bunch of assumptions of ordinary module loading / symbol resolution.  Removing support for future modules would also simplify a bunch of callback and other core livepatch code as well.

jpoimboe:
To see if this is more than a hypothetical issue, I looked at all C object files which use feature fixups, and there are only a handful of them for a RHEL8-based config on an upstream kernel:
```
# for i in `find . -name "*.o"`; do if eu-readelf -S $i |grep -q ftr_fixup; then if [[ -e ${i/\.o/.c} ]]; then echo $i; eu-readelf -S $i | grep ftr_fixup; fi; fi; done 
./arch/powerpc/kernel/dbell.o
[ 4] __ftr_fixup          PROGBITS     0000000000000000 00000398 00000090  0 A      0   0  8
[ 5] .rela__ftr_fixup     RELA         0000000000000000 00023738 00000120 24 I     30   4  8
./arch/powerpc/kernel/process.o
[45] __ftr_fixup          PROGBITS     0000000000000000 00003450 00000060  0 A      0   0  8
[...]

PR #980, opened 2019-06-19
ppc64le: Feature fixup

kamalesh-babulal:
This patch series implements the phase 1 and other powerpc related quirks discussed at #974.
1. First patch fixes the group size of lwsync fixup special section
2. Second patch add support for __barrier_nospec_fixup special section
3. The last patch adds checks to assert if the function being patched, calls any of the *_fixup section and error out for such patches.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1825,7 +1825,7 @@ static int fixup_entry_group_size(struct kpatch_elf *kelf, int offset)
 
 static int fixup_lwsync_group_size(struct kpatch_elf *kelf, int offset)
 {
-	return 4;
+	return 8;
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


kamalesh-babulal:
I will raise a separate pull request for fixup sections.

jpoimboe:
Thanks @kamalesh-babulal .  Instead of the 3rd commit, could we instead just add an "unsupported" field to the special_section struct?  Then we could do something like
```
@@ -1943,6 +1948,9 @@ static int should_keep_rela_group(struct section *sec, unsigned int start,
                }
        }
 
+       if (special->unsupported)
+               DIFF_FATAL("unsupported reference to special section %s", sec->base->name);
+
        return found;
 }
 
```

kamalesh-babulal:
>  Instead of the 3rd commit, could we instead just add an "unsupported" field to the special_section struct? Then we could do something like

Yes, this approach looks neat :)

Issue #979, opened 2019-06-18
kpatch.service and subsequent starts


joe-lawrence:
PR #972 modified the kpatch systemd service file to _not_ unload kpatch modules on service stop.  We did this so that kpatch modules would remain in place during system shutdown.

The systemd service may also stopped manually or in the case of a kpatch package upgrade.  A subsequent `systemctl start kpatch` would then fail as `kpatch --load-all` will trip over the still-loaded modules.

A few ways to work around this:

1. Enhance the kpatch.service script to detect this scenario
2. Enhance the kpatch script with a new option for this scenario
3. (My preferred) Enhance `kpatch --load-all` so that it doesn't "die" with failure over already-loaded modules.  Ideally it would simply skip over those and then continue trying to load any that aren't loaded yet.

Option 3 would change behavior of the kpatch script that other scripts may rely upon, so I'd like to make sure that won't break anything else before pursuing it.

joe-lawrence:
#3 could be as simple as this (untested):
```diff
diff --git a/kpatch/kpatch b/kpatch/kpatch
index 328f1197259d..1343168e06a6 100755
--- a/kpatch/kpatch
+++ b/kpatch/kpatch
@@ -314,7 +314,8 @@ load_module () {
                                die "error: cannot re-enable patch module $modname, cannot verify checksum match"
                        fi
                else
-                       die "error: module named $modname already loaded and enabled"
+                       warn "error: module named $modname already loaded and enabled"
+                       return 0
                fi
        fi
```
that would let load_module() complain about the already present module, but not fail on it.  The modprobe command already does something similar for the same scenario, sans msg:
```
% ls -d /sys/module/e1000e/
/sys/module/e1000e/
% modprobe e1000e
% echo $?
0
```

sm00th:
Looks good. I *think* that if there are scripts relying on that they are likely to be used for testing.

PR #981, opened 2019-06-19
kpatch script: don't fail if module already loaded+enabled

joe-lawrence:
For "kpatch load" invocations, don't set failing return status if the
kpatch module is already loaded and enabled.  This allows the user to
more gracefully restart the kpatch.service, or re-run "kpatch load
--all" commands to pick up new kpatch modules.

Fixes: #979

### DIFF
--- a/kpatch/kpatch
+++ b/kpatch/kpatch
@@ -314,30 +314,30 @@ load_module () {
 				die "error: cannot re-enable patch module $modname, cannot verify checksum match"
 			fi
 		else
-			die "error: module named $modname already loaded and enabled"
+			echo "module named $modname already loaded and enabled"
 		fi
+	else
+		echo "loading patch module: $module"
+		local i=0
+		while true; do
+			out="$(LC_ALL=C insmod "$module" 2>&1)"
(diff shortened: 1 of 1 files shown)


joe-lawrence:
Q: do we want to elaborate on this to continue executing `load_module()` such that it runs the `wait_for_patch_transition` logic at the bottom for said module?  This would handle a corner case where the service is restarted (or kpatch re-invoked) while a livepatch module is currently transitioning.  Dunno if it's worth pushing that change right now?

sm00th:
> Q: do we want to elaborate on this to continue executing load_module() such that it runs the wait_for_patch_transition logic at the bottom for said module?

I think we do, otherwise we are risking having an untransitioned livepatch module just sitting there forever. Unless a user spots it and unloads it (which I wouldn't count on).

joe-lawrence:
v2: move the "loading patch module" logic underneath an `[[ -d "$moddir" ]]` else clause.  This allows code to fall-through down to the `wait_for_patch_transition` logic.  (Note that the "module already loaded, re-enabling" transition logic is slightly different, so I left that one in place.)

PR #985, opened 2019-06-29
create-diff-object: fix failed in kpatch_check_relocations for compre…

wwheart:
…ss section

If using the CFLAGS of "-gz=zlib", the section would be compressed,
and the section's sh_flags is set with "SHF_COMPRESSED", so the
rela->addend may be bigger than sdata->d_size, which cause
ERROR in kpatch_check_relocations of "out-of-range relocation".

The section infomation as follow:

Relocation section '.rela.debug_info' at offset 0xac8 contains 76 entries:
Offset          Info           Type           Sym. Value    Sym. Name + Addend
000000000006  00080000000a R_X86_64_32       0000000000000000 .debug_abbrev + 0
00000000000c  000c0000000a R_X86_64_32       0000000000000000 .debug_str + 326
000000000011  000c0000000a R_X86_64_32       0000000000000000 .debug_str + 657
000000000015  000c0000000a R_X86_64_32       0000000000000000 .debug_str + 748

[16] .debug_str        PROGBITS         0000000000000000  00000424
00000000000003f7  0000000000000001 MSC       0     0     1

Therefore, we may check whether the "SHF_COMPRESSED" is set, if so
we can get the section uncompressed size by "ch_size", and compare
with "ch_size".

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -441,8 +441,7 @@ static void kpatch_compare_correlated_section(struct section *sec)
 		goto out;
 	}
 
-	if (sec1->sh.sh_size != sec2->sh.sh_size ||
-	    sec1->data->d_size != sec2->data->d_size) {
+	if (sec1->size != sec2->size) {
 		sec->status = CHANGED;
 		goto out;
 	}
@@ -498,14 +497,14 @@ static int kpatch_line_macro_change_only(struct section *sec)
 	if (sec->status != CHANGED ||
 	    is_rela_section(sec) ||
 	    !is_text_section(sec) ||
-	    sec->sh.sh_size != sec->twin->sh.sh_size ||
+	    sec->size != sec->twin->size ||
 	    !sec->rela ||
 	    sec->rela->status != SAME)
 		return 0;
 
 	start1 = (unsigned long)sec->twin->data->d_buf;
 	start2 = (unsigned long)sec->data->d_buf;
-	size = sec->sh.sh_size;
+	size = sec->size;
 	for (offset = 0; offset < size; offset += length) {
 		insn_init(&insn1, (void *)(start1 + offset), 1);
 		insn_init(&insn2, (void *)(start2 + offset), 1);
@@ -581,14 +580,14 @@ static int kpatch_line_macro_change_only(struct section *sec)
 	if (sec->status != CHANGED ||
 	    is_rela_section(sec) ||
 	    !is_text_section(sec) ||
-	    sec->sh.sh_size != sec->twin->sh.sh_size ||
+	    sec->size != sec->twin->size ||
 	    !sec->rela ||
 	    sec->rela->status != SAME)
 		return 0;
 
 	start1 = (unsigned long)sec->twin->data->d_buf;
 	start2 = (unsigned long)sec->data->d_buf;
-	size = sec->sh.sh_size;
+	size = sec->size;
 	for (offset = 0; offset < size; offset += PPC_INSTR_LEN) {
 		if (!memcmp((void *)start1 + offset, (void *)start2 + offset,
 			    PPC_INSTR_LEN))
@@ -765,10 +764,10 @@ static void kpatch_correlate_sections(struct list_head *seclist1, struct list_he
 			 * Changed group sections are currently not supported.
 			 */
 			if (sec1->sh.sh_type == SHT_GROUP) {
-				if (sec1->data->d_size != sec2->data->d_size)
+				if (sec1->size != sec2->size)
 					continue;
 				if (memcmp(sec1->data->d_buf, sec2->data->d_buf,
-				           sec1->data->d_size))
+				           sec1->size))
 					continue;
 			}
 			sec1->twin = sec2;
@@ -869,7 +868,7 @@ static void kpatch_mark_grouped_sections(struct kpatch_elf *kelf)
 		if (groupsec->sh.sh_type != SHT_GROUP)
 			continue;
 		data = groupsec->data->d_buf;
-		end = groupsec->data->d_buf + groupsec->data->d_size;
+		end = groupsec->data->d_buf + groupsec->size;
 		data++; /* skip first flag word (e.g. GRP_COMDAT) */
 		while (data < end) {
 			sec = find_section_by_index(&kelf->sections, *data);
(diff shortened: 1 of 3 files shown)


wwheart:
The reason compile failed is the libelf version is old, I compile success with libelf-0.176.so.

wwheart:
this compile options as follow:  
-flto-compression-level=<number> Use zlib compression level <number> for IL.
  -gz                         Generate compressed debug sections.
  -gz=<format>                Generate compressed debug sections in format <format>.

jpoimboe:
Can you please make sure it compiles with older libelf by adding something like
```
#ifndef SHF_COMPRESSED
#define SHF_COMPRESSED (1 << 11)
#endif
```

kpatch-build/create-diff-object.c
```diff
@@ -2168,15 +2168,15 @@ static void kpatch_check_relocations(struct kpatch_elf *kelf)
...
+	int sec_size;
 
 	list_for_each_entry(sec, &kelf->sections, list) {
 		if (!is_rela_section(sec))
 			continue;
 		list_for_each_entry(rela, &sec->relas, list) {
 			if (rela->sym->sec) {
-				sdata = rela->sym->sec->data;
-				if (rela->addend > (int)sdata->d_size) {
+				sec_size = rela->sym->sec->uncompressed_size;
```
jpoimboe: No need for an extra variable here

jpoimboe: Also, please just call it `sec->size`, and replace all uses of sec->data_d_size and sec->sh.sh_size with it.

kpatch-build/kpatch-elf.c
```diff
@@ -238,6 +239,16 @@ void kpatch_create_section_list(struct kpatch_elf *kelf)
 
 		if (!gelf_getshdr(scn, &sec->sh))
 			ERROR("gelf_getshdr");
+		else
+			sec->uncompressed_size = sec->sh.sh_size;
+
+		if ((sec->sh.sh_flags & SHF_COMPRESSED)) {
+			if (!gelf_getchdr(scn, &ch))
+				ERROR("gelf_getchdr");
+			else
```
jpoimboe: These else conditions aren't necessary because ERROR never returns.

wwheart:
> Can you please make sure it compiles with older libelf by adding something like
> 
> ```
> #ifndef SHF_COMPRESSED
> #define SHF_COMPRESSED (1 << 11)
> #endif
> ```

yet, the elfutils version I used is elfutils-0.176.tar.bz2

wwheart@linux-41:~/libelf 0 > grep SHF_COMPRESSED -rn *.h
elf.h:481:#define SHF_COMPRESSED             (1 << 11)  /* Section with compressed data. */

jpoimboe:
I guess you'll also need to define the other compile failures in the same ifdef.  Maybe a wrapper function kpatch_get_compressed_size() which is empty for older libelfs (that don't have SHF_COMPRESSED defined).

wwheart:
> I guess you'll also need to define the other compile failures in the same ifdef. Maybe a wrapper function kpatch_get_compressed_size() which is empty for older libelfs (that don't have SHF_COMPRESSED defined).

yet, it's a good suggestion.

kpatch-build/create-diff-object.c
wwheart: > No need for an extra variable here

yet, remove the unneeded variable of sec_size

kpatch-build/kpatch-elf.c
```diff
@@ -209,6 +209,33 @@ void kpatch_create_rela_list(struct kpatch_elf *kelf, struct section *sec)
 	}
 }
 
+#ifdef SHF_COMPRESSED
+void kpatch_get_uncompressed_size(Elf_Scn *scn, struct section *sec)
```
jpoimboe: Please call it kpatch_get_section_size

kpatch-build/create-diff-object.c
wwheart: > No need for an extra variable here

yet, the value sdata had deleted.

kpatch-build/kpatch-elf.c
wwheart: > Please call it kpatch_get_section_size

OK, the function name had changed

wwheart:
> There are still a lot of read references to sec->sh.sh_size and sec->data->d_size in the code. Any place where those variables are read, I think it should instead be reading sec->size.
> 
> Similarly, everywhere those variables are set, sec->size should also be set.

I don't know it's need to instead be reading sec->size, because the sec->size is decompressed size, which the elf is loading and decompressing, not the section size is the elf file, so if we use sec->size to read or write may cause out of ranges.

jpoimboe:
> I don't know it's need to instead be reading sec->size, because the sec->size is decompressed size, which the elf is loading and decompressing, not the section size is the elf file, so if we use sec->size to read or write may cause out of ranges.

As far as I know, compressed sections are only used for debug sections, and create-diff-object just treats those as opaque data.  sec->size is just an internal variable, we don't (and shouldn't) use it for writing the section back AFAIK.  So it should be ok to read sec->size everywhere in create-diff-object.c.  It makes the code more consistent and readable.

If we ever do need to access the data of uncompressed sections, I assume the decompressing and recompressing will happen in kpatch-elf.c, and create-diff-object.c can still ignorantly use sec->size, since as far as it knows the sections will always be uncompressed.

wwheart:
> > I don't know it's need to instead be reading sec->size, because the sec->size is decompressed size, which the elf is loading and decompressing, not the section size is the elf file, so if we use sec->size to read or write may cause out of ranges.
> 
> As far as I know, compressed sections are only used for debug sections, and create-diff-object just treats those as opaque data. sec->size is just an internal variable, we don't (and shouldn't) use it for writing the section back AFAIK. So it should be ok to read sec->size everywhere in create-diff-object.c. It makes the code more consistent and readable.
> 
> If we ever do need to access the data of uncompressed sections, I assume the decompressing and recompressing will happen in kpatch-elf.c, and create-diff-object.c can still ignorantly use sec->size, since as far as it knows the sections will always be uncompressed.

static void kpatch_compare_correlated_nonrela_section(struct section *sec)
{
        struct section *sec1 = sec, *sec2 = sec->twin;

        if (sec1->sh.sh_type != SHT_NOBITS &&
            memcmp(sec1->data->d_buf, sec2->data->d_buf, sec1->data->d_size)) 
                sec->status = CHANGED;
        else
                sec->status = SAME;
}

for example the code above, the "sec1->data->d_size" can't be instead of sec1->size, otherwise will cause access the data out of range for the debug section.

kpatch-build/kpatch-elf.c
wwheart: > These else conditions aren't necessary because ERROR never returns.

yet, the else condition had been remove.

PR #986, opened 2019-07-02
Patch replacement fixes for the old KPatch core

euspectre:
Virtuozzo uses the old KPatch core (kpatch.ko) for binary patches for the older kernels and relies on atomic replacement of patches.

When I was trying to replace a loaded binary patch (P1) with a new one (P2) that changed a smaller set of kernel functions (a faulty part was removed in P2), kernel crashed in certain cases.

Assume that patch P1 changes functions F1(), F2() and F3() while patch P2 - F1() only. I saw in the crash reports that the kernel was executing its original variant of F2() and the variant of F3() from P1 at that moment. This inconsistency caused problems because F2() and F3() were dependent on one another.

kpatch_apply_patch() does check if the functions from the patch to be replaced are currently running. However, the functions are removed from
'kpatch_func_hash' in kpatch_register() only after stop_machine() and kpatch_apply_patch() have finished. So, they can be running at that point, which is bad.

While investigating this, I also found that kpatch_backtrace_address_verify() searched the whole
set of functions for the ones being replaced (func->op == KPATCH_OP_UNPATCH). This is a waste of time when the patch is loaded or unloaded rather than replaced, because the search is done for each stack entry being checked.

This patchset fixes both issues.

I doubt that Livepatch has the same issues, its implementation of atomic replacement is completely different. I haven't checked though.

### DIFF
--- a/kmod/core/core.c
+++ b/kmod/core/core.c
@@ -358,8 +358,10 @@ static int kpatch_apply_patch(void *data)
 	struct kpatch_apply_patch_args *args = data;
 	struct kpatch_module *kpmod;
 	struct kpatch_func *func;
+	struct hlist_node *tmp;
 	struct kpatch_object *object;
 	int ret;
+	int i;
 
 	kpmod = args->kpmod;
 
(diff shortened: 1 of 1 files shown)


jpoimboe:
Looks good to me.  FWIW, I consider you the maintainer of the core module code now.  Let us know when you're no longer using it ;-)

euspectre:
> FWIW, I consider you the maintainer of the core module code now. Let us know when you're no longer using it ;-)

:-)
Yes, will do.
Looks like we are going to use the old-style patches and the core module for a while.

PR #989, opened 2019-07-05
Add support for R_X86_64_PLT32

### LINKED ISSUE #975: Support R_X86_64_PLT32 relocations
Starting with binutils 2.31, the Linux kernel may have R_X86_64_PLT32 relocations.  Make sure we support them.  This should be as simple as treating R_X86_64_PLT32 exactly like R_X86_64_PC32 everywhere.  For more details see upstream commit torvalds/linux@b21ebf2fb4cde1618915a97cc773e287ff49173e.

jpoimboe:
Starting with binutils 2.31, the Linux kernel may have R_X86_64_PLT32
relocations. Make sure we support them. This should be as simple as
treating R_X86_64_PLT32 exactly like R_X86_64_PC32 everywhere. For more
details see upstream commit torvalds/linux@b21ebf2.

This also fixes the following issue seen on Fedora 29:

```
$ kpatch-build/kpatch-build -t vmlinux ./test/integration/fedora-27/convert-global-local.patch
Using cache at /home/jpoimboe/.kpatch/src
Testing patch file(s)
Reading special section data
Building original source
Building patched source
ERROR: slub.o: 1 function(s) can not be patched
slub.o: function __kmalloc has no fentry/mcount call, unable to patch
/home/jpoimboe/git/kpatch/kpatch-build/create-diff-object: unreconcilable difference
ERROR: 1 error(s) encountered. Check /home/jpoimboe/.kpatch/build.log for more details.
```

Fixes #975.

### DIFF
--- a/kmod/core/core.c
+++ b/kmod/core/core.c
@@ -700,6 +700,7 @@ static int kpatch_write_relocations(struct kpatch_module *kpmod,
 		case R_X86_64_NONE:
 			continue;
 		case R_X86_64_PC32:
+		case R_X86_64_PLT32:
 			loc = dynrela->dest;
 			val = (u32)(dynrela->src + dynrela->addend -
 				    dynrela->dest);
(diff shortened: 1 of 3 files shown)


jpoimboe:
True, I will add one.  I'm also working on the unit tests to get my jump label warning change to pass.

PR #993, opened 2019-07-08
test/integration: Rebase to RHEL 8.0

kamalesh-babulal:
Rebase the tests to the closet upstream kernel to RHEL 8.0, as per
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/8.0_release_notes/index#overview
it is kernel v4.18. So rebase it on top of upstream v4.18.


kamalesh-babulal:
I have only compile tested v4.18 and RHEL 8.0, with all of the patch(s) applied.

joe-lawrence:
Hi @kamalesh-babulal,

If I understand correctly, this PR rebases the integration tests on top of upstream v4.18.  That's going to be _very_ close to RHEL-8.0 GA kernel, but not exactly the same.  For example, the tracepoints-sections.patch modifies run_timer_softirq() and expects to see `base->must_forward_clk = false` in the code context leading up to the added code.  When I look at kernel-4.18.0-80.el8.src.rpm however, its run_timer_softirq() doesn't touch `must_forward_clk`.  So I think RHEL-8.0 must have included a backport of 363e934d8811 ("timers: Clear timer_base::must_forward_clk with timer_base::lock held"), which is upstream v4.19+.

FWIW, I think that is the only patch that had a v4.18 vs. RHEL-8.0 discrepancy.  If you can update the PR with rebase against kernel-4.18.0-80.el8.src.rpm, I can verify that they build / run on one of our internal test boxes.

kamalesh-babulal:
@joe-lawrence Thanks for the review and for help with a test run.  Agree, there is the difference between the upstream v4.18 and RHEL 8.0 kernel source.  When I tried, it applied with minor fuzz on a couple of test cases over RHEL 8.0 sources. Yeah, it makes sense to have rebased against RHEL 8.0 sources, instead of closet upstream. I will redo it and push it.

kamalesh-babulal:
**v2 changes**:
- Rebased to RHEL 8.0 kernel version 4.18.0-80.el8
- Added README file, with the kernel version the integration tests were rebased.

joe-lawrence:
Thanks for rebasing @kamalesh-babulal.  Here are results from an x86_64 test run:
[combined.log](https://github.com/dynup/kpatch/files/3388819/combined.log)
[gcc-static-local-var-5.log](https://github.com/dynup/kpatch/files/3388820/gcc-static-local-var-5.log)
[make.log](https://github.com/dynup/kpatch/files/3388821/make.log)
[test.log](https://github.com/dynup/kpatch/files/3388822/test.log)

One build failure, perhaps a result of the objtool warnings?
```
build: gcc-static-local-var-5
Using cache at /root/.kpatch/src
Testing patch file(s)
Reading special section data
Building original source
Building patched source
ERROR: audit.o: reference to static local variable lock.65414 in audit_log_end was removed
ERROR: gcc-static-local-var-5: build failed
```
and one test script failure, due to print formatting changes (so something like 
[data-new-LOADED.patch.txt](https://github.com/dynup/kpatch/files/3388829/data-new-LOADED.patch.txt) to fix:
```
loading patch module: test-data-new.ko
ERROR: data-new: rhel-8.0/data-new-LOADED.test failed after kpatch load
disabling patch module: test_data_new
unloading patch module: test_data_new
```
I'll try to do a similar ppc64le test run when I get a chance.

joe-lawrence:
And the same for ppc64le

[make.out.txt](https://github.com/dynup/kpatch/files/3392496/make.out.txt)
[test.log](https://github.com/dynup/kpatch/files/3392497/test.log)
[macro-printk.1.log](https://github.com/dynup/kpatch/files/3392501/macro-printk.1.log)
[module-call-external.1.log](https://github.com/dynup/kpatch/files/3392502/module-call-external.1.log)

Build errors:
```
build: macro-printk
Using cache at /root/.kpatch/src
Testing patch file(s)
Reading special section data
Building original source
Building patched source
fib_frontend.o: changed function: inet_rtm_newroute
ERROR: fib_semantics.o: object size mismatch: __msg.66237
fib_trie.o: changed function: fib_table_insert
ERROR: macro-printk: build failed
```

```
build: module-call-external
Using cache at /root/.kpatch/src
Testing patch file(s)
Reading special section data
Building original source
Building patched source
export.o: changed function: e_show
af_netlink.o: new function: kpatch_string
Patched objects: fs/nfsd/nfsd.ko vmlinux
Building patch module: test-module-call-external.ko
ERROR: Undefined symbols: kpatch_string . Check /root/.kpatch/build.log for more details.
ERROR: module-call-external: build failed
```
Test errors:

data-new (same as x86_64 above)

shadow-newpid crashes the machine:
```
livepatch: enabling patch 'test_shadow_newpid'
livepatch: 'test_shadow_newpid': starting patching transition
livepatch: 'test_shadow_newpid': patching complete
Unable to handle kernel paging request for data at address 0x00000000
Faulting instruction address: 0xc000000000cc6b38
Oops: Kernel access of bad area, sig: 11 [#1]
LE SMP NR_CPUS=2048 NUMA pSeries
Modules linked in: test_shadow_newpid(OEK) sg pseries_rng xts vmx_crypto xfs libcrc32c sd_mod ibmvscsi ibmveth scsi_transport_srp dm_mirror dm_region_hash dm_log dm_mod [last unloaded: test_parainstructions_section]
CPU: 7 PID: 579 Comm: systemd-journal Kdump: loaded Tainted: G           OE K  --------- -  - 4.18.0-80.el8.ppc64le #1
NIP:  c000000000cc6b38 LR: c0000000004e84d0 CTR: c0000000004e8450
REGS: c0000007f8fa7850 TRAP: 0300   Tainted: G           OE K  --------- -  -  (4.18.0-80.el8.ppc64le)
MSR:  8000000000009033 <SF,EE,ME,IR,DR,RI,LE>  CR: 88000228  XER: 20040000
CFAR: c000000000008934 DAR: 0000000000000000 DSISR: 40000000 IRQMASK: 0 
GPR00: d00000000b4a3030 c0000007f8fa7ad0 c0000000015bd100 c0000007f2020361 
GPR04: 000000000000fc9f 0000000000000000 c0000007f8fa7c20 0000000000000000 
GPR08: 0000000000000030 000000007fffffff 0000000000000000 d00000000b4a3660 
GPR12: c0000000004e8450 c000000007fa6200 0000000000000000 0000000000000000 
GPR16: 0000000000000000 0000000000000000 c0000007f8fa7c20 0000000000000000 
GPR20: 0000000000000000 0000000000000000 000000000000003f 00000000ffffffff 
GPR24: 000000000000003f 00000000ffffffff 000000000000003f 00000000ffffffff 
GPR28: 0000000000000000 000000000000fc9f c0000007f2020361 c0000007f2030000 
NIP [c000000000cc6b38] vsnprintf+0x58/0x5c0
LR [c0000000004e84d0] seq_printf+0x80/0xd0
Call Trace:
[c0000007f8fa7ad0] [c0000007f8fa7b10] 0xc0000007f8fa7b10 (unreliable)
[c0000007f8fa7bc0] [c0000000004e84d0] seq_printf+0x80/0xd0
[c0000007f8fa7bf0] [d00000000b4a3030] proc_pid_status+0xcf8/0x2598 [test_shadow_newpid]
[c0000007f8fa7ce0] [c00000000007610c] livepatch_handler+0x30/0x64
[c0000007f8fa7d30] [c0000000004e6b6c] seq_read+0x1cc/0x650
[c0000007f8fa7dd0] [c0000000004a09d8] sys_read+0x108/0x310
[c0000007f8fa7e30] [c00000000000b388] system_call+0x5c/0x70
Instruction dump:
7fe32214 fa210078 fa410080 fba100d8 fbc100e0 fa810090 7c9d2378 7fa3f840 
7c7e1b78 7cb12b78 7cd23378 419d043c <89310000> 7fd4f378 2f890000 419e0134 
---[ end trace eb35e9a9bf3754e2 ]---
```
I'll look into this crash later today...

kamalesh-babulal:
@joe-lawrence Thanks a lots for testing, I was trying to `gcc-static-local-var-5` on ppc64le and hitting a segfault. I will debug more and update.

joe-lawrence:
Hmm, shadow-newpid.patch crashes even when I sub out all the code for a nop().  Some more files:  
[crash.tar.gz](https://github.com/dynup/kpatch/files/3393904/crash.tar.gz)

joe-lawrence:
FYI, shadow-newpid.patch no longer builds with kpatch v0.7.0, specifically 4f4870dd058f3 ("create-diff-object: Don't allow jump labels"):
```
fork.o: WARNING: unable to correlate static local variable ctr.65763 used by .toc, assuming variable is new
fork.o: changed function: _do_fork.constprop.7
fork.o: changed function: sys_fork
fork.o: changed function: sys_vfork
fork.o: changed function: _do_fork
array.o: changed function: proc_pid_status
/root/kpatch/kpatch-build/create-diff-object: ERROR: exit.o: kpatch_regenerate_special_section: 2091: Found a jump label at do_exit()+0x87c, using key devmap_managed_key.  Jump labels aren't currently supported.  Use static_key_enabled() instead.
```
but as the crash was in the patched proc_pid_status() and not do_exit(), I suspect that jump labels may not be the real culprit.

jpoimboe:
I can confirm the proc_pid_status() panic is fixed by #1007.

joe-lawrence:
Nice, thanks for verifying @jpoimboe!

julien-thierry:
For context:
- The first 3 patches aim to fix the issue on x86_64 pointed by Joe on gcc-static-local-var-5.
However with those changes (especially the fact of considering .part. as child function), this introduces a sibling call warning on ppc64el on the same test. This is due to the parent of the .part. now being considered for inclusion since one of its child's section has changed.
If we agree to include those changes, I can add the no-optimize attribute to the test patch.

- Fourth patch fixes the error on macro-printk, although this feels like making the issue more unlikely to happen, I do not have an absolute solution for that.

The rest of the patches is just the actual rebase and silencing issues that won't be solved right now.

kpatch-build/create-diff-object.c
```diff
@@ -205,6 +205,15 @@ static void kpatch_bundle_symbols(struct kpatch_elf *kelf)
 	}
 }
 
+static void kpatch_add_child_symbol(struct symbol *parent, struct symbol *child)
+{
+	parent->nb_children++;
+	parent->children = realloc(parent->children,
+				   sizeof(*parent->children) * parent->nb_children);
```
joe-lawrence: A nit, but realloc() could fail so we should catch that like other allocation failures.

kpatch-build/create-diff-object.c
```diff
@@ -226,13 +236,12 @@ static void kpatch_detect_child_functions(struct kpatch_elf *kelf)
 			if (!pname)
 				ERROR("strndup");
 
-			sym->parent = find_symbol_by_name(&kelf->symbols, pname);
+			parent = find_symbol_by_name(&kelf->symbols, pname);
 			free(pname);
 
-			if (!sym->parent)
-				ERROR("failed to find parent function for %s", sym->name);
+			if (!parent)
```
joe-lawrence: Are we missing ERROR("failed to find parent function for %s", parent) here?

test/integration/rhel-8.0/data-read-mostly.patch.disabled
```diff
@@ -1,3 +1,5 @@
+Disabled due to https://github.com/dynup/kpatch/issues/940
```
joe-lawrence: Perhaps a follow up commit could rework these tests to avoid functions that use jump labels?  I realize that's ignoring the problem, but would let us continue to test read-mostly and shadow variables in the meantime.

test/integration/rhel-8.0/new-function.patch
```diff
@@ -15,8 +15,8 @@ index 3ad4602..f3cda7e 100644
  	return (b - buf) ? b - buf : retval;
```
joe-lawrence: Note for future work: we should probably explain this in the patch author guide.

test/integration/rhel-8.0/multiple.test
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+
+SCRIPTDIR="$(readlink -f $(dirname $(type -p $0)))"
+ROOTDIR="$(readlink -f $SCRIPTDIR/../..)"
```
joe-lawrence: As @sm00th mentioned in PR #1040, this needs to be ROOTDIR="$(readlink -f $SCRIPTDIR/../../..)"

kpatch-build/create-diff-object.c
julien-thierry: Yes, will update.

kpatch-build/create-diff-object.c
julien-thierry: Yes, seems I messed my patch splitting, thanks for spotting this!

I'll update as well.

test/integration/rhel-8.0/data-read-mostly.patch.disabled
julien-thierry: I can have a look at it, but what might end up happening is that the test becomes completely different (patching other functions or maybe files) than the similar test in fedora/rhel folders. Which might be confusing if we keep the same name.

Perhaps we can keep this one disabled and add a second test "data-read-mostly-2.patch" that would (hopefully) work on RHEL-8 with the current state of kpatch, by avoiding to modify functions with jump labels.

test/integration/rhel-8.0/multiple.test
julien-thierry: Thanks, will update.

kpatch-build/create-diff-object.c
sm00th: It also now needs to be freed at some point.

test/integration/rhel-8.0/data-read-mostly.patch.disabled
joe-lawrence: You're right, creating a newly named test file would be less-confusing.  Anyway, that was just an idea for a follow up PR if you wanted to explore that, not needed for this one.

test/integration/rhel-8.0/new-function.patch
joe-lawrence: > Note for future work: we should probably explain this in the patch author guide.

I'm working on it now, I'll post up a PR with author guide changes soon.

kpatch-build/create-diff-object.c
```diff
@@ -205,9 +205,20 @@ static void kpatch_bundle_symbols(struct kpatch_elf *kelf)
...
+	child->parent = parent;
+	parent->children[parent->nb_children - 1] = child;
+}
+
 /*
  * During optimization gcc may move unlikely execution branches into *.cold
- * subfunctions. kpatch_detect_child_functions detects such subfunctions and
+ * subfunctions. Some functions can also be split into multiple *.part
+ * functions.
+ * kpatch_detect_child_functions detects such subfunctions and
```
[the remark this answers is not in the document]
julien-thierry: So the initial motivation for this change was a failure on the test gcc-static-local-var-5. The failure was due to a static local variable "lock" in function audit_log_end() being moved in the patched object into audit_log_end.part.17 .

One particular thing in the patched object is that both the audit_log_end and audit_log_end.part.* symbols exists, while the original only has audit_log_end. So the original audit_log_end is twined with the patched audit_log_end and static local references are only looked in the patched section related to audit_log_end and kpatch misses the fact that a reference was moved to the .part. section.

So, maybe the "child" concideration is not correct. But there needs to be a link between .part. functions and their original symbol (when it is also present on the patched object, as most of the time I've noticed it isn't) and the static variable correlation needs to be able to follow that link when looking for relocations.

Should I create a separate list of "parts" of a symbol? Other suggestions?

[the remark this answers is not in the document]
jpoimboe: Hm.  Does the patched audit_log_end() also reference the "lock" static local variable?  IIRC, "part" functions are clones of the originals.  So it seems like audit_log_end() should also reference the static local, unless I'm missing something.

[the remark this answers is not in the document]
julien-thierry: No, it doesn't have a reference to the lock. audit_log_end() almost only has a jump to audit_log_end.part in this case.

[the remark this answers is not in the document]
jpoimboe: Ah... I think I had a fundamental misunderstanding of this optimization.  Looking at gcc/ipa-split.c in the GCC source confirms that.  I had been thinking this was more like the constprop and isra optimizations.

So I take back my original comment!  It's very similar to the .cold optimization and I think the parent/child relationship makes sense after all.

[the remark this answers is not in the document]
julien-thierry: Thanks for having a look!

So how do you suggest I proceed with this PR? Split it in two, one with the test addition and patch modifications and the other with the create-diff-object changes?

Any suggestion for which one should be base on the other?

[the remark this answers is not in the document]
jpoimboe: Since we don't have a CI running these yet, I think I'd suggest doing the integration tests first, in one PR.  That way the other fixes can refer to them.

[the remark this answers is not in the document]
julien-thierry: Makes sense, I'll try sending that later today after rebasing on master and checking there aren't new issues!

julien-thierry:
After the last update the following errors are withstanding:
PPCLE64:
* build: macro-printk
ERROR: fib_semantics.o: object size mismatch: __msg.66237

This can be fixed with the following PR:
https://github.com/dynup/kpatch/pull/1053

x86:
* build: gcc-static-local-var-5
ERROR: audit.o: reference to static local variable lock.65414 in audit_log_end was removed

This can be fixed with the following PR:
https://github.com/dynup/kpatch/pull/1054

julien-thierry:
> I see `gcc-static-local-var-5` failing on x86_64 with:
> `ERROR: audit.o: reference to static local variable lock.65414 in audit_log_end was removed`
> 
> And `macro-printk` on ppc64le with:
> `ERROR: fib_semantics.o: object size mismatch: __msg.66237`
> (which I guess is caused by one of recent merges)
> 
> But I think it will be easier to deal with those in a separate PR since this one is already hard to review, but it has improved drastically from where we started. Thank you.

Both of these issues are expected (in the current state) and are fixed by the pending PRs #1054 and #1053

PR #997, opened 2019-07-17
create-diff-object: Don't allow jump labels

### LINKED ISSUE #946: kpatch-build should warn about static keys
kpatch-build drops `__jump_table` sections from the special section handling, so static keys do not operate as expected.  (#931 "Livepatch does not handle static keys".)

We attempted including them (#937 "Add jump label support"), but it turned out that wasn't the full story as we ran into issues (#940 "Jump label support still cannot work".)

The main problem is that "special" livepatch symbols are not fully relocated before the jump label core code attempts to initialize them.  (See the call chain in the commit message for #944 Revert "create-diff-object: add jump label support".)

While we are waiting for a kernel fix (work in progress), we should:

1. Update the patch author's guide explaining these short comings
2. Warn the user (if we can) that functions referencing static keys have changed and that the patched code may not execute as intended.  In some cases simply loading the patch module may crash the host.

cc: @euspectre, @ufo19890607

jpoimboe:
Create an error if a patched function uses a jump label.  We need this
until upstream livepatch supports them with a .klp.arch section.

Fixes #946.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1773,6 +1773,21 @@ static int ex_table_group_size(struct kpatch_elf *kelf, int offset)
 	return size;
 }
 
+static int jump_table_group_size(struct kpatch_elf *kelf, int offset)
+{
+	static int size = 0;
+	char *str;
+
+	if (!size) {
+		str = getenv("JUMP_STRUCT_SIZE");
+		if (!str)
+			ERROR("JUMP_STRUCT_SIZE not set");
+		size = atoi(str);
+	}
+
+	return size;
+}
+
 #ifdef __x86_64__
 static int parainstructions_group_size(struct kpatch_elf *kelf, int offset)
 {
@@ -1897,6 +1912,10 @@ static struct special_section special_sections[] = {
 		.name		= "__ex_table", /* must come after .fixup */
 		.group_size	= ex_table_group_size,
 	},
+	{
+		.name		= "__jump_table",
+		.group_size	= jump_table_group_size,
+	},
 #ifdef __x86_64__
 	{
 		.name		= ".smp_locks",
@@ -1994,6 +2013,7 @@ static void kpatch_regenerate_special_section(struct kpatch_elf *kelf,
 	struct rela *rela, *safe;
 	char *src, *dest;
 	unsigned int group_size, src_offset, dest_offset, include;
+	int jump_table = !strcmp(special->name, "__jump_table");
 
 	LIST_HEAD(newrelas);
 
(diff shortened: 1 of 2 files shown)


jpoimboe:
@joe-lawrence @sm00th are the unit tests working for you for this PR?  Not sure why this is failing on Travis on the new dynamic debug test, it works fine for me locally.

jpoimboe:
@sm00th I added a debugging commit and the problem seems to be in the test/unit/Makefile.include support for .env files which override the default.  I have a dynamic-debug-special-static-issue-990.env file which overrides the default value of JUMP_STRUCT_SIZE from 24 to 16.  When I run the test locally, it works.  With "make unit VERBOSE=1", notice the 2nd JUMP_STRUCT_SIZE comes after the 1st:
```
PARA_STRUCT_SIZE=16 EX_STRUCT_SIZE=12 BUG_STRUCT_SIZE=12 ALT_STRUCT_SIZE=13 JUMP_STRUCT_SIZE=24 ORC_STRUCT_SIZE=6 JUMP_STRUCT_SIZE=16 /home/jpoimboe/git/kpatch/kpatch-build/create-diff-object dynamic-debug-special-static-issue-990.ORIG.o dynamic-debug-special-static-issue-990.PATCHED.o \
	vmlinux dynamic-debug-special-static-issue-990.symtab /dev/null \
	test_dynamic-debug-special-static-issue-990 dynamic-debug-special-static-issue-990.OUTPUT.o 
```
But on Travis, there is no 2nd version.
```
PARA_STRUCT_SIZE=16 EX_STRUCT_SIZE=12 BUG_STRUCT_SIZE=12 ALT_STRUCT_SIZE=13 JUMP_STRUCT_SIZE=24 ORC_STRUCT_SIZE=6  /home/travis/build/dynup/kpatch/kpatch-build/create-diff-object dynamic-debug-special-static-issue-990.ORIG.o dynamic-debug-special-static-issue-990.PATCHED.o \
		vmlinux dynamic-debug-special-static-issue-990.symtab /dev/null \
		test_dynamic-debug-special-static-issue-990 dynamic-debug-special-static-issue-990.OUTPUT.o 
```
I verified that Travis at least sees the env file.
```
$ cat test/unit/objs/x86_64/dynamic-debug-special-static-issue-990.env
JUMP_STRUCT_SIZE=16
```
Weird...

jpoimboe:
Ah, travis uses Make 3.81 whereas I have Make 4.2.1.  I'm guess the older version doesn't support one of the features in the makefile.

jpoimboe:
Yeah, Make 3.x doesn't support the "file" function.  I'll use "shell" instead.

jpoimboe:
Makefile fix in #998.

jpoimboe:
Yay, travis is happy now.

PR #1000, opened 2019-07-18
Bump version to v0.7.0

sm00th:
Increment version to v0.7.0 due to changed behavior of kpatch.service and
kpatch load subcommand. Full list of changes:

- Multiple memory leak fixes in kpatch-build
- livepatch-patch-hook compatability fixes for kernels 5.1+
- Making kpatch-build compatible with custom gcc names
- Added rhel-rebased integration tests
- kpatch.service will no longer unload modules on stop
- kpatch load will no longer fail if a module is already loaded and enabled
- kpatch-build will now check for *_fixup section changes on ppc64le and will
  fail on such changes
- Add support for R_X86_64_PLT32
- don't allow jump labels
- ppc64le-specific kpatch-build fixes

### DIFF
--- a/contrib/kpatch.spec
+++ b/contrib/kpatch.spec
@@ -1,6 +1,6 @@
 Name: kpatch
 Summary: Dynamic kernel patching
-Version: 0.6.3
+Version: 0.7.0
 License: GPLv2
 Group: System Environment/Kernel
 URL: http://github.com/dynup/kpatch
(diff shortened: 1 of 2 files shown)


contrib/kpatch.spec
```diff
@@ -93,6 +93,19 @@ rm -rf %{buildroot}
...
 %changelog
+* Thu Jul 18 2019 Artem Savkov <asavkov@redhat.com> - 0.7.0
+- Multiple memory leak fixes in kpatch-build
+- livepatch-patch-hook compatability fixes for kernels 5.1+
+- Making kpatch-build compatible with custom gcc names
+- Added rhel-rebased integration tests
+- kpatch.service will no longer unload modules on stop
+- kpatch load will no longer fail if a module is already loaded and enabled
+- kpatch-build will now check for *_fixup section changes on ppc64le and will
+  fail on such changes
```
jpoimboe: Does RPM allow multi-line changelog entries?

jpoimboe:
PR #1000 !!!!

contrib/kpatch.spec
jpoimboe: I see other packages doing it, so it should be ok.

PR #1001, opened 2019-07-19
create-diff-object: Add ppc64le mcount support

jpoimboe:
Add the __mcount_loc section on ppc64le.  It has pointers to all the
mcount calls.  This will enable the ftrace hook to be used for patched
functions.

We have also been seeing a mysterious ppc64le livepatch bug where r2
(TOC pointer) gets intermittently corrupted.  My only theory is that the
call to _mcount is doing that somehow.  Now that call will be replaced
with a NOP when the patch module is loaded.


jpoimboe:
I rescind my comment about the r2 corruption thing, this didn't fix it.

jpoimboe:
Please don't merge this yet, I want to do some more testing.

PR #1002, opened 2019-07-22
create-diff-object/ppc64le: Don't allow sibling calls

jpoimboe:
Before doing a sibling call, the patched function restores the stack to
its caller's stack.  The kernel-generated call stub then writes the
patch module's r2 (toc) value to the caller's stack, corrupting it,
eventually causing a panic after it returns to the caller.

Fix it by disallowing sibling calls from patched functions on ppc64le.

In theory we could instead a) generate a custom stub, or b) modify the
kernel livepatch_handler code to save/restore the stack r2 value, but
this is easier.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -3253,6 +3253,60 @@ static void kpatch_build_strings_section_data(struct kpatch_elf *kelf)
 	}
 }
 
+/*
+ * Don't allow sibling calls from patched functions on ppc64le.  Before doing a
+ * sibling call, the patched function restores the stack to its caller's stack.
+ * The kernel-generated stub then writes the patch module's r2 (toc) value to
+ * the caller's stack, corrupting it, eventually causing a panic after it
+ * returns to the caller and the caller tries to use the livepatch module's toc
+ * value.
+ *
+ * In theory we could instead a) generate a custom stub, or b) modify the
+ * kernel livepatch_handler code to save/restore the stack r2 value, but this
+ * is easier for now.
+ */
+static void kpatch_no_sibling_calls_ppc64le(struct kpatch_elf *kelf)
+{
+#ifdef __powerpc64__
+	struct symbol *sym;
+	unsigned int insn;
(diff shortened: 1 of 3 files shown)


jpoimboe:
I plan to add a unit test for this soon, once I figure out why I can't get the ppc64le unit tests to work.

jpoimboe:
The unit tests for this PR have been merged, so it should be ready now.

joe-lawrence:
(@jpoimboe) perhaps copy/paste explanation you provided in private BZ?  I just re-read that and it explains the situation well.  It might be a bit long for the commit message, but could be helpful for other reviewers.

jpoimboe:
@joe-lawrence I fleshed out the commit message, hopefully it's much better now.  Let me know if that clears up all your questions.

kpatch-build/create-diff-object.c
```diff
@@ -3253,6 +3253,42 @@ static void kpatch_build_strings_section_data(struct kpatch_elf *kelf)
...
+	void *insn;
+
+	list_for_each_entry(sym, &kelf->symbols, list) {
+		if (sym->type != STT_FUNC || sym->status != CHANGED)
+			continue;
+
+		for (insn = sym->sec->data->d_buf;
+		     insn < sym->sec->data->d_buf + sym->sec->data->d_size;
+		     insn += 4) {
+			if (*(unsigned int *)insn != 0x48000000)
```
joe-lawrence: Endianness shouldn't matter, so treating it as an unsigned int is okay, I think, rather than a series of bytes.  However 0x48000000 looks like a magic number.. can we make this a preprocessor constant or at least add a comment to what this pattern is?

jpoimboe:
@joe-lawrence I added a comment and also tweaked the code a bit.

jpoimboe:
Just made one more trivial change, put the for loop on a single line :-)

joe-lawrence:
> Just made one more trivial change, put the for loop on a single line :-)

Looks good to me, thanks.

PR #1004, opened 2019-07-22
kpatch-build: Use rpmbuild --nodeps

jpoimboe:
The list of prerequsite RPMs which are needed to build the kernel RPM is
constantly growing.  But at least some of those RPMs aren't strictly
necessary for building the kernel, at least for kpatch-build's purposes.
Requiring them all to be installed is a bit overkill, and sometimes
causes kpatch-build to fail when it doesn't need to.

If the build does fail, we can always check the kpatch.log file and
update the dependencies listed in the README as needed.


joe-lawrence:
This bites me all. the. time.  Even worse is that yum-builddep doesn't seem to always accurately install the build requirements.  Thanks for this, hopefully it turns out better than it has been.

Issue #877, opened 2018-06-07
ERROR: meminfo.o: kpatch_check_relocations: 1941: out-of-range relocation .debug_str+9668 in .rela.debug_info


aliceinwire:
Building modules, stage 2.
MODPOST 259 modules
Extracting new and modified ELF sections
meminfo.o: changed function: meminfo_proc_show
/usr/libexec/kpatch/create-diff-object: ERROR: meminfo.o: kpatch_check_relocations: 1941: out-of-range relocation .debug_str+9668 in .rela.debug_info

create-diff-object debug output
[create-diff-patch.log](https://github.com/gentoo/elivepatch-server/files/2074874/create-diff-patch.log)

this is the patch tested:
```
diff -Nupr src.orig/fs/proc/meminfo.c src/fs/proc/meminfo.c
--- src.orig/fs/proc/meminfo.c  2017-11-17 15:58:41.113211972 -0500
+++ src/fs/proc/meminfo.c       2017-11-17 15:58:58.554211972 -0500
@@ -42,6 +42,8 @@ static void show_val_kb(struct seq_file
		seq_write(m, " kB\n", 4);
 }

+static int foo = 5;
+
 static int meminfo_proc_show(struct seq_file *m, void *v)
 {
		struct sysinfo i;
@@ -153,6 +155,7 @@ static int meminfo_proc_show(struct seq_
		show_val_kb(m, "CmaFree:        ",
					global_page_state(NR_FREE_CMA_PAGES));
 #endif
+       seq_printf(m, "kpatch: %d\n", foo);

		hugetlb_report_meminfo(m);


kpatch-build
	create-diff-object.c
```

aliceinwire:
I tried to use also different patch but the error is similar.

sm00th:
What are you building against? I couldn't reproduce it with `v4.17-1306-g716a685fdb89`.

aliceinwire:
thanks i wlll try with a more new kernel i was using 4.9.52 if i remeber correctly.i dont have my pc with me now

sm00th:
It is still a bug worth investigating even if it is on an older kernel, I'll try to reproduce it with something older

jpoimboe:
If you can run it with '-d' and then attach the following data, it will probably help us debug it:
- ~/.kpatch/tmp/orig/fs/proc/meminfo.o
- ~/.kpatch/tmp/patched/fs/proc/meminfo.o
- ~/.kpatch/tmp/vmlinux.symtab
[...]

PR #1011, opened 2019-07-29
create-diff-object: disable DWARF compression explicitly

steils:
On some systems the linker produces compressed debug sections by
default. It is not supported by create-diff-object for now.

Fixes: #877

Signed-off-by: Stefan Strogin <steils@gentoo.org>


jpoimboe:
Thanks @steils !  #985 also tried to fix this issue by adding support for compressed sections, but it's still under review.  So maybe we should go with this approach in the meantime.

Issue #984, opened 2019-06-26
Multiple issues found by Coverity scan.


sm00th:
I've ran the latest kpatch sources (8e3daab) through coverity scan and it found a bunch of issues. There are some false-positives but most of these are real issues we better take care of.

Here is the full log: [scan-results.log](https://github.com/dynup/kpatch/files/3330824/scan-results.log)

PR #1014, opened 2019-08-01
Covscan fixes

sm00th:
A bunch of small fixes of issues found by covscan. Please note that some of these are harmless or very unlikely so we might want to drop some of them, I just opted to clean-up the scan log as much as possible.

There are still some outstanding issues it still finds with these changes but I think all of those (except maybe list ones) are false positives. Here is the full covscal log: [scan-results.log](https://github.com/dynup/kpatch/files/3456376/scan-results.log)

Fixes: #984

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1238,6 +1238,10 @@ static void rela_insn(struct section *sec, struct rela *rela, struct insn *insn)
 
 	start = (unsigned long)sec->base->data->d_buf;
 	end = start + sec->base->sh.sh_size;
+
+	if (end <= start)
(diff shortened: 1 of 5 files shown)


kpatch-build/create-diff-object.c
```diff
@@ -1319,6 +1326,7 @@ static void kpatch_replace_sections_syms(struct kpatch_elf *kelf)
 				int start, end;
 
 				if (sym->type == STT_SECTION ||
+				    sym->sec == NULL ||
```
jpoimboe: change to '!sym->sec'

sm00th: fixed

PR #1020, opened 2019-08-12
livepatch-patch-hook: skip klp_(un)register_patch() for RHEL-7.8+ ker…

joe-lawrence:
…nels

RHEL-7.8 backported upstream kernel commit 958ef1e39d24 ("livepatch:
Simplify API by removing registration step"), so add it to the kernels
using the simple registration API.


joe-lawrence:
RHEL-7.8 is still under development, but I figured I'd queue this up before I forget to do it :)

Issue #1022, opened 2019-08-13
Stalled process signalling in v5.1+ kernels


joe-lawrence:
Ever since upstream v5.1 (specifically commits [0b3d52790e1c](https://github.com/torvalds/linux/commit/0b3d52790e1c) ("livepatch: Remove signal sysfs attribute") and [cba82dea3061](https://github.com/torvalds/linux/commit/cba82dea3061) ("livepatch: Send a fake signal periodically") the kpatch utility script's signal_stalled_processes() functionality has been redundant.

Currently it is coded as:
- If `/sys/kernel/livepatch/$module/signal` is present, use the sysfs interface
- Otherwise iterate through all "stalled" tasks and manually signal them

I tested this against a RHEL-7.8 development release (which includes the mentioned upstream backports) and the old behavior is benign:  the kernel automatically sends "fake" signals while kpatch utility tries to do the same manually.

Going forward we would need to introduce kernel and distro-specific checks into kpatch utility if we wanted to eliminate the redundant signalling.  This should be fairly easy to add, but I wanted to gauge other's opinions as this would be the beginning of such conditional code to kpatch utility.

jpoimboe:
The easiest solution would be to drop support for the almost-two-year-old kernels which have neither the sysfs file nor the automatic signalling, and remove the manual signalling logic altogether.  That would be my vote.

joe-lawrence:
(Thinking out load)  I think ripping removing it from the kpatch utility would okay for older distro kernels without the automatic signalling mechanism -- they may use newer kpatch releases to _build_ kpatch .ko's, but leave the userspace utility alone to retain its old, manual signalling logic.

PR #1024, opened 2019-08-15
kpatch: remove manual signaling logic

joe-lawrence:
Ever since upstream v5.1 (specifically commits 0b3d52790e1c ("livepatch:
Remove signal sysfs attribute") and cba82dea3061 ("livepatch: Send a
fake signal periodically") the kpatch utility script's
signal_stalled_processes() functionality has been redundant.

Remove this code from the script to avoid user confusion or potentially
weird user vs. kernel signalling race conditions.

Fixes: #1022

### DIFF
--- a/kpatch/kpatch
+++ b/kpatch/kpatch
@@ -230,21 +230,8 @@ signal_stalled_processes() {
 	[[ -z "$module" ]] && return
 
 	if [[ -e "/sys/kernel/livepatch/$module/signal" ]] ; then
+		echo "signaling stalled process(es):"
 		echo 1 > "/sys/kernel/livepatch/$module/signal"
-	else
-		for proc_task in /proc/[0-9]*/task/[0-9]*; do
-			tid=${proc_task#*/task/}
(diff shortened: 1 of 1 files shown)


jpoimboe:
I was actually thinking we should keep the /sys/kernel/livepatch/$module/signal logic, since there are still relatively recent kernels which have it.

jpoimboe:
And also because there's an easy way to detect whether the sysfs file exists, like we did before, so there's little harm in continuing to support it.

joe-lawrence:
> And also because there's an easy way to detect whether the sysfs file exists, like we did before, so there's little harm in continuing to support it.

Ah yes, this is a better compromise than completely removing the code.  Here is a v2 (untested) where it modifies the user-visible notices a bit.  The only strangeness would be if one ran 'kpatch signal' on ancient kernel -- it would silently do nothing.  Without checking kernel versions, the script doesn't know if the sysfs interface absence indicates a new or older kernel version.  Given how infrequently 'signal' is used, it seems reasonable not to introduce version checking just for that case.

Issue #902, opened 2018-09-05
ERROR: invalid ancestor arch/x86/realmode/rm/pasyms.h for arch/x86/realmode/rm/reboot.o. Check /root/.kpatch/build.log for more details.


disaster123:
Hello,

while trying to build a kpatch i get:
ERROR: invalid ancestor arch/x86/realmode/rm/pasyms.h for arch/x86/realmode/rm/reboot.o. Check /root/.kpatch/build.log for more details.

.kpatch/build.log says nothing special it look fine to me.

Greets,
Stefan

kamalesh-babulal:
Can you share more information about the environment:
- patch your passing to `kpatch-build`
- Kernel version 
- gcc version

disaster123:
gcc:
```
gcc version 6.3.0 20170516 (Debian 6.3.0-18+deb9u1) 
```

Patch
```
diff --git a/arch/x86/include/asm/msr-index.h b/arch/x86/include/asm/msr-index.h
index e9b90b0e8..493dfe385 100644
--- a/arch/x86/include/asm/msr-index.h
+++ b/arch/x86/include/asm/msr-index.h
@@ -69,6 +69,7 @@
 #define MSR_IA32_ARCH_CAPABILITIES     0x0000010a
 #define ARCH_CAP_RDCL_NO               (1 << 0)   /* Not susceptible to Meltdown */
 #define ARCH_CAP_IBRS_ALL              (1 << 1)   /* Enhanced IBRS support */
+#define ARCH_CAP_SKIP_VMENTRY_L1DFLUSH (1 << 3)   /* Skip L1D flush on vmentry */
 #define ARCH_CAP_SSB_NO                        (1 << 4)   /*
                                                    * Not susceptible to Speculative Store Bypass
                                                    * attack, so no Speculative Store Bypass
diff --git a/arch/x86/include/asm/vmx.h b/arch/x86/include/asm/vmx.h
index fa2359410..d41efb49b 100644
--- a/arch/x86/include/asm/vmx.h
+++ b/arch/x86/include/asm/vmx.h
@@ -577,6 +577,9 @@ enum vmx_l1d_flush_state {
        VMENTER_L1D_FLUSH_COND,
        VMENTER_L1D_FLUSH_ALWAYS,
        VMENTER_L1D_FLUSH_EPT_DISABLED,
+#ifndef __GENKSYMS__
+       VMENTER_L1D_FLUSH_NOT_REQUIRED,
+#endif
 };
 
[...]

PR #1030, opened 2019-08-21
kpatch-build: sanity check patched filenames

joe-lawrence:
Run the input patch(es) through lsdiff and then verify that no obviously
unsupported files are directly modified (e.g. assembly .S files).

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ sudo dnf install gcc kernel-devel-${UNAME%.*} elfutils elfutils-devel
 Install the dependencies for the "kpatch-build" command:
 
 ```bash
-sudo dnf install pesign yum-utils openssl wget numactl-devel
+sudo dnf install pesign yum-utils openssl wget numactl-devel patchutils
 sudo dnf builddep kernel-${UNAME%.*}
 sudo dnf debuginfo-install kernel-${UNAME%.*}
 
@@ -50,9 +50,6 @@ sudo dnf install gcc-plugin-devel
 # optional, but highly recommended
 sudo dnf install ccache
 ccache --max-size=5G
-
-# optional, for kpatch-test
-sudo dnf install patchutils
 ```
 
 #### RHEL 7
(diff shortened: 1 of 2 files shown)


joe-lawrence:
I was revisiting issue #902 to see if we could improve the user experience a bit.  Here's an enhancement that could head off some low-hanging fruit if the user tries to patch file paths that are obviously unsupported.  Since this operates on the patch file itself, feedback is provided fairly early in the process and not after long kernel builds.

Here's what the output looks like with test patches:
```
% ./kpatch-build/kpatch-build asm.patch 
Using cache at /home/cloud-user/.kpatch/src
Testing patch file(s)
ERROR: /home/cloud-user/github-kpatch/asm.patch: unsupported patch to assembly: arch/x86_64/kernel/entry_64.S
ERROR: Unsupported changes detected.

% ./kpatch-build/kpatch-build lib.patch
Using cache at /home/cloud-user/.kpatch/src
Testing patch file(s)
ERROR: /home/cloud-user/github-kpatch/lib.patch: unsupported patch to lib/: lib/syscall.c
ERROR: /home/cloud-user/github-kpatch/lib.patch: unsupported patch to lib/: lib/A/B/syscall.c
ERROR: Unsupported changes detected.
```

I can add additional pathname patterns if desired.  I'd be interested in other sanity checks that could be applied through simple patch file inspection as well.

jpoimboe:
Looks good, though I guess our (horribly out-of-date) README should also be updated to make patchutils a real dependency instead of only being needed for unit tests.

joe-lawrence:
v2: updated the README.md file

joe-lawrence:
While we're at it, is there anything else that would make sense to check?  Other un-patchable files?  Directory depth (ie, one that applies with `patch -p1`) etc?

jpoimboe:
Not that I can think of, but you might want to look at all the files that are ignored by kpatch-gcc.  For example, it ignores `lib/` files.  That check can probably be removed now?  There might be others as well.

joe-lawrence:
Merging this for now, we can update paths later if we need to.

Issue #1031, opened 2019-08-21
kpatch-build creates incompatible livepatches for RHEL-8 kernels


joe-lawrence:
When loading a simple cmdline string kpatch, Yulia is seeing this error when it's built with kpatch-build v0.8.0:
```
loading patch module: /var/lib/kpatch/4.18.0-135.el8.x86_64/kpatch-cmdline-string.ko
insmod: ERROR: could not insert module /var/lib/kpatch/4.18.0-135.el8.x86_64/kpatch-cmdline-string.ko: Invalid parameters
kpatch: failed to load module /var/lib/kpatch/4.18.0-135.el8.x86_64/kpatch-cmdline-string.ko
```
the same input patch loads fine when built with v0.7.1

In looking at the commits between those two tags, I don't think considered that RHEL-8 kernels would always fall into the "simple enable API" check in b913b4bae:
``` diff
diff --git a/kmod/patch/livepatch-patch-hook.c b/kmod/patch/livepatch-patch-hook.c
index ccc6ee8aba7c..2ddd153ceb7d 100644
--- a/kmod/patch/livepatch-patch-hook.c
+++ b/kmod/patch/livepatch-patch-hook.c
@@ -64,7 +64,11 @@
 # define HAVE_CALLBACKS
 #endif
 
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 1, 0)
+#ifdef RHEL_RELEASE_CODE
+# if RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(7, 8)      << oops
+#  define HAVE_SIMPLE_ENABLE                              <<
+# endif
+#elif LINUX_VERSION_CODE >= KERNEL_VERSION(5, 1, 0)
 # define HAVE_SIMPLE_ENABLE
 #endif
```

Those conditions should be re-written to protect RHEL-8 kernel versions that do not contain the simple enable API.

PR #1032, opened 2019-08-22
kmod/patch: check for RHEL8 when setting HAVE_SIMPLE_ENABLE

sm00th:
Make sure we don't set HAVE_SIMPLE_ENABLE on rhel8. This condition will
need to be updated once 958ef1e39d24 ("livepatch: Simplify API by removing
registration step") is backported to rhel8.

Fixes: b913b4b ("livepatch-patch-hook: skip klp_(un)register_patch() for RHEL-7.8+ kernels")
Fixes: #1031


joe-lawrence:
Thanks for fixing this @sm00th.  Since we will most likely be backporting 958ef1e39d24 to RHEL-8.2, my vote would be to just set that up in this fix so we don't have to revisit it in a few weeks.

PR #1033, opened 2019-08-22
test/integration: add an ability to check specific repo/commit

sm00th:
Add an ability to run vagrant integration tests on specified
repos/commits.

### DIFF
--- a/test/integration/vm-integration-run
+++ b/test/integration/vm-integration-run
@@ -1,16 +1,20 @@
 #!/bin/bash
 
 KPATCH_SLOW=0
+KPATCH_GIT=${KPATCH_GIT:-https://github.com/dynup/kpatch.git}
+KPATCH_REV=${KPATCH_REV:-HEAD}
 LOGDIR="/vagrant/logs"
 
 usage()
 {
   echo "usage: $(basename "${0}") [options]" >&2
-  echo "-h, --help This message" >&2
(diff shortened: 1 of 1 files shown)


test/integration/vm-integration-run
```diff
@@ -1,36 +1,54 @@
...
   esac
-  shift
 done
 
-git clone https://github.com/dynup/kpatch.git || exit 1
+git clone "${KPATCH_GIT}" || exit 1
 
 cd kpatch || exit 1
 
+git reset --hard "${KPATCH_REV}" || exit 1
```
jpoimboe: Maybe 'git checkout ${KPATCH_REV}' instead?  'git reset --hard' is a bit severe as it unilaterally wipes out any local changes and it also changes the current branch head.

sm00th: That is intentional. This code is ran in a one-off vm so we don't really care about the branch head or local changes and we want this command to bring us as close to KPATCH_REV as possible.

jpoimboe: Ah, right.  Though since it's run in a VM and it's a fresh clone I guess there shouldn't be any local changes anyway?  I guess either way works.

joe-lawrence: Too bad git doesn't support a shallow, depth=1 clone of a specific commit hash.

PR #1038, opened 2019-09-17
kpatch-build: fix two parent matches error

wwheart:
when I make a patch for a module, cause the error as follow:
ERROR: two parent matches for hello.o. Check /tmp/kpatch-build.log for more details.

The Makefile is as follow:
	obj-m += m_hello.o
	m_hello-y = hello.o
	default:
        	$(MAKE) -C /lib/modules/4.4.21-69-default/build M=$(shell pwd) modules
	clean:
        	$(MAKE) -C /lib/modules/4.4.21-69-default/build M=$(shell pwd) clean

and than cmd "grep -l hello.o ./.*.cmd | grep -Fv hello.o" would get two parant
	.m_hello.ko.cmd
	.m_hello.o.cmd
but the ".m_hello.ko.cmd" is the parant for "m_hello.o", and ".m_hello.o.cmd" is the
parant for "hello.o".

because the "hello.o" is a substring of "m_hello.o", which will cause the "m_hello.o"
matched for the "hello.o".

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -373,17 +373,17 @@ find_parent_obj() {
 	if [[ "$DEEP_FIND" -eq 1 ]]; then
 		num=0
 		if [[ -n "$last_deep_find" ]]; then
-			parent="$(grep -l "$grepname" "$last_deep_find"/.*.cmd | filter_parent_obj "${pdir}" "${file}" | head -n1)"
-			num="$(grep -l "$grepname" "$last_deep_find"/.*.cmd | filter_parent_obj "${pdir}" "${file}" | wc -l)"
+			parent="$(grep -lw "$grepname" "$last_deep_find"/.*.cmd | filter_parent_obj "${pdir}" "${file}" | head -n1)"
+			num="$(grep -lw "$grepname" "$last_deep_find"/.*.cmd | filter_parent_obj "${pdir}" "${file}" | wc -l)"
 		fi
 		if [[ "$num" -eq 0 ]]; then
-			parent="$(find . -name ".*.cmd" -print0 | xargs -0 grep -l "$grepname" | filter_parent_obj "${pdir}" "${file}" | cut -c3- | head -n1)"
-			num="$(find . -name ".*.cmd" -print0 | xargs -0 grep -l "$grepname" | filter_parent_obj "${pdir}" "${file}" | wc -l)"
+			parent="$(find . -name ".*.cmd" -print0 | xargs -0 grep -lw "$grepname" | filter_parent_obj "${pdir}" "${file}" | cut -c3- | head -n1)"
+			num="$(find . -name ".*.cmd" -print0 | xargs -0 grep -lw "$grepname" | filter_parent_obj "${pdir}" "${file}" | wc -l)"
 			[[ "$num" -eq 1 ]] && last_deep_find="$(dirname "$parent")"
 		fi
 	else
-		parent="$(grep -l "$grepname" "$dir"/.*.cmd | filter_parent_obj "${dir}" "${file}" | head -n1)"
-		num="$(grep -l "$grepname" "$dir"/.*.cmd | filter_parent_obj "${dir}" "${file}" | wc -l)"
+		parent="$(grep -lw "$grepname" "$dir"/.*.cmd | filter_parent_obj "${dir}" "${file}" | head -n1)"
+		num="$(grep -lw "$grepname" "$dir"/.*.cmd | filter_parent_obj "${dir}" "${file}" | wc -l)"
 	fi
 
 	[[ "$num" -eq 0 ]] && PARENT="" && return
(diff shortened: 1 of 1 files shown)


kpatch-build/kpatch-build
```diff
@@ -385,7 +385,7 @@ find_parent_obj() {
 	PARENT="$(basename "$parent")"
 	PARENT="${PARENT#.}"
 	PARENT="${PARENT%.cmd}"
-	PARENT="$dir/$PARENT"
+	[[ $dir != "." ]] && PARENT="$dir/$PARENT"
```
sm00th: Why don't we want './' here? It shouldn't hurt so I don't see any reason for this change.

wwheart: > Why don't we want './' here? It shouldn't hurt so I don't see any reason for this change.

For the "dir" is get from dir="$(dirname "$parent")", it would be either "." or "xxxx/yyy/".

sm00th: But why can't we just leave `PARENT="$dir/$PARENT"` so that if $dir is "." we get `PARENT="./$PARENT"`? This is a completely valid path so we can have a unified solution for both cases without doing this check.

wwheart: > But why can't we just leave `PARENT="$dir/$PARENT"` so that if $dir is "." we get `PARENT="./$PARENT"`? This is a completely valid path so we can have a unified solution for both cases without doing this check.

The original input parameter of "find_parent_obj" is got from the file "changed_objs", which maybe either "x1.o" or "dir1/x2.o".
for the "x1.o",  the result of the cmd "dirname x1.o" is "."，and than we would get the parant with "./", such as "./y1.o", therefore when we add "-w" for grep, the result of the cmd  "grep -lw "./y1.o" ./.*.cmd" is none. This is why we don't want "./".
for the "dir1/x2.o", the result of the cmd "dirname dir1/x2.o" is "dir1", and than we would get
the parant with "dir1/", such as "dir1/y2.o".

sm00th: Oh, I see now. Thank you for the explanation.

wwheart:
> The '-w' addition looks ok, but I am not sure '.' dir special case though.
> 
> This will also conflict with #1036 and will need to be rebased.

I have rebased the code yet.

joe-lawrence:
> > The '-w' addition looks ok, but I am not sure '.' dir special case though.
> > This will also conflict with #1036 and will need to be rebased.
> 
> I have rebased the code yet.

The changes look good, however I have two comments:

1 - Here's a revised commit message for this change:
```
kpatch-build: use whole word filename matching in find_parent_obj()

Building a kpatch for a module with this Makefile:

  obj-m += m_hello.o
  m_hello-y = hello.o
  default:
          $(MAKE) -C /lib/modules/4.4.21-69-default/build M=$(shell pwd) modules
  clean:
          $(MAKE) -C /lib/modules/4.4.21-69-default/build M=$(shell pwd) clean

results in kpatch-build "ERROR: two parent matches for hello.o".

The problem is that find_parent_obj() looks for filenames like so:

  % grep -l hello.o ./.*.cmd | grep -Fv hello.o
  .m_hello.ko.cmd
  .m_hello.o.cmd

where .m_hello.ko.cmd is the parent for m_hello.o and .m_hello.o.cmd is the
parent for hello.o, but because "hello.o" is a substring of "m_hello.o",
it will cause m_hello.o to be matched for hello.o as well.

Fix this by using grep's -w|--word-regexp option to force it to match
whole words instead of substrings.
```

2 - Your (master) branch has an additional merge commit on top.  Github thinks it has no conflicts with the base branch so I guess it will work, but ordinarily we rebase the PR branch against the dynup/master latest and don't see that extra commit.  I wonder if this is a side effect of doing work on master and then pulling down the latest dynup/master and that forced the merge.

wwheart:
> > > The '-w' addition looks ok, but I am not sure '.' dir special case though.
> > > This will also conflict with #1036 and will need to be rebased.
> > 
> > 
> > I have rebased the code yet.
> 
> The changes look good, however I have two comments:
> 
> 1 - Here's a revised commit message for this change:
> 
> ```
> kpatch-build: use whole word filename matching in find_parent_obj()
> 
> Building a kpatch for a module with this Makefile:
> 
>   obj-m += m_hello.o
>   m_hello-y = hello.o
>   default:
>           $(MAKE) -C /lib/modules/4.4.21-69-default/build M=$(shell pwd) modules
>   clean:
>           $(MAKE) -C /lib/modules/4.4.21-69-default/build M=$(shell pwd) clean
> 
> results in kpatch-build "ERROR: two parent matches for hello.o".
> 
> The problem is that find_parent_obj() looks for filenames like so:
> 
>   % grep -l hello.o ./.*.cmd | grep -Fv hello.o
>   .m_hello.ko.cmd
>   .m_hello.o.cmd
> 
> where .m_hello.ko.cmd is the parent for m_hello.o and .m_hello.o.cmd is the
> parent for hello.o, but because "hello.o" is a substring of "m_hello.o",
> it will cause m_hello.o to be matched for hello.o as well.
> 
> Fix this by using grep's -w|--word-regexp option to force it to match
> whole words instead of substrings.
> ```
> 
> 2 - Your (master) branch has an additional merge commit on top. Github thinks it has no conflicts with the base branch so I guess it will work, but ordinarily we rebase the PR branch against the dynup/master latest and don't see that extra commit. I wonder if this is a side effect of doing work on master and then pulling down the latest dynup/master and that forced the merge.

OK， it's ok now.

PR #1040, opened 2019-09-18
integration/rhel-*: fix multiple.test

sm00th:
When these from internal depths of Red Hat upstream paths changed and
now we are one level deeper in directory tree.

The issue probably also exist in rhel8.0 rebase pr #993.


joe-lawrence:
Weird, how did these ever get the "../.." when the multiple.test for other distros already had the three directory version.  /shrugs

PR #1042, opened 2019-10-01
kmod: don't build module by default

### LINKED ISSUE #966: kmod core module: v5.2-rc1 stack trace updates break build
tglx renovated the stack trace code for 5.2 (see patchset in [lkml archive](https://lore.kernel.org/lkml/20190425094453.875139013@linutronix.de/)), changing `struct stack_trace` and its related APIs.

Sooner or later we're going to deprecate the kpatch core module since its functionality can be provided by the kernel livepatch API.  But in case anyone wants to continue to using it with newer kernels, consider this a heads up that work will be required to build it for 5.2.

cc: @euspectre

yhcote:
The stacktrace internal interface has been updated for v5.2. On x86,
among others, this means that stack_trace_save_tsk* is to be used
instead of instantiating a stack_trace struct. Those routines are
currently not exported for module use.

See linux kernel: <25e39e32b0a3> (livepatch: Simplify stack trace retrieval)
for more info about the stacktrace interface changes.

The kernel/livepatch subsystem is the way forward for using kpatch and
there are plans to phase out 'kmod' altogether in the near future.

It is still possible to build and use kmod by doing something like:
$ env BUILDMOD="yes" make

Closes #966

Signed-off-by: Yannick Cote <ycote@redhat.com>

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -17,10 +17,5 @@ MANDIR     = $(DESTDIR)$(PREFIX)/share/man/man1
 SYSTEMDDIR = $(DESTDIR)$(PREFIX)/lib/systemd/system
 UPSTARTDIR = $(DESTDIR)/etc/init
 
-# The core module is only supported on x86_64
-ifeq ($(ARCH),x86_64)
-BUILDMOD  ?= yes
-endif
-
 .PHONY: all install clean
 .DEFAULT: all


kmod/core/Makefile
```diff
@@ -1,24 +1,29 @@
...
 KPATCH_BUILD ?= /lib/modules/$(shell uname -r)/build
 KERNELRELEASE := $(lastword $(subst /, , $(dir $(patsubst %/,%,$(KPATCH_BUILD)))))
+VERCODE := $(shell awk '/LINUX_VERSION_CODE/{print $$NF}' $(KPATCH_BUILD)/include/generated/uapi/linux/version.h)
 THISDIR := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
 
 ifeq ($(wildcard $(KPATCH_BUILD)),)
 $(error $(KPATCH_BUILD) doesn\'t exist.  Try installing the kernel-devel-$(KERNELRELEASE) RPM or linux-headers-$(KERNELRELEASE) DEB.)
 endif
 
+ifeq ($(shell if [ $(VERCODE) -lt 328192 ];then echo "y";fi),y)
```
julien-thierry: Not feeling strongly about this, but you could define a:
`VERCODE_5_2 := 328192`
To clarify what that value is. (Arguably you can guess from the 'else' message, but I feel it reduces ambiguity).

joe-lawrence: Geez, `make` makes it difficult to do something simple like compare two integers.  I'd almost be tempted to tack this comparison into the awk command, let it print "kpatch.ko" and then save the output into a $TARGETS variable... but I'm not sure that's any more or less elegant than the ifeq shell trick.  Either way, setting up a `VERCODE_5_2` variable might save somebody time if they aren't paying close attention to the comment that Julien mentioned.

kmod/core/Makefile
```diff
@@ -1,24 +1,29 @@
 # make rules
 KPATCH_BUILD ?= /lib/modules/$(shell uname -r)/build
 KERNELRELEASE := $(lastword $(subst /, , $(dir $(patsubst %/,%,$(KPATCH_BUILD)))))
+VERCODE := $(shell awk '/LINUX_VERSION_CODE/{print $$NF}' $(KPATCH_BUILD)/include/generated/uapi/linux/version.h)
```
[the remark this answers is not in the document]
yhcote: Yes it needs to be there.

kmod/core/Makefile
yhcote: I think it's documented by the info line. I had commented it, but removed it would clutter more than anything.

yhcote:
> commit msg nits:
> 
> * Not sure if "This fixes..." will be detected by github.  I'd copy/paste from https://help.github.com/en/articles/closing-issues-using-keywords to be sure
> * Usually we write upstream commit references as `<commid-id-12chars> ("Subject line")`
> 
> the commit:
> 
> * Makefile.inc already includes a `BUILDMOD` decision, maybe this logic could go there as well?
> 
> meta:
> 
> * I wonder if it would be better to just remove the support module entirely?  I would think the various kpatch-build kernel `CONFIG_` checks would indirectly enforce a minimum kernel version (and not worry about distro releases complicating the version code logic).  It's code that we're bringing along, but not  testing any longer... @jpoimboe thoughts?

* Github "does" pickup the "this fixes", but I will change this and the commit reference to comply to the format you propose.

* I thought about going into Makefile.inc before, but this seems to clutter that file. And seems better to confine that temporary fix to the leaf Makefile.

* For more background context, this change is temporary where like you mentioned, we may remove the module entirely in the future, or we may be able to have a better fix if the new functions get exported in later kernel version. Finally, I have proposed to implement 'makeit' in the tree to allow for more flexibility, simplicity and packaging abilities.

kmod/core/Makefile
yhcote: > Geez, `make` makes it difficult to do something simple like compare two integers. I'd almost be tempted to tack this comparison into the awk command, let it print "kpatch.ko" and then save the output into a $TARGETS variable... but I'm not sure that's any more or less elegant than the ifeq shell trick. Either way, setting up a `VERCODE_5_2` variable might save somebody time if they aren't paying close attention to the comment that Julien mentioned.

Oh, and for that. I could not agree more. The right way to do it in my opinion is NOT by using GNU make extensions but my having an external shell script that does the work and simply call that from the Makefile. At the moment it seems out of place. But if we use 'makeit', then this is certainly how that kind of logic is isolated from target definition.

jpoimboe:
>     * I wonder if it would be better to just remove the support module entirely?  I would think the various kpatch-build kernel `CONFIG_` checks would indirectly enforce a minimum kernel version (and not worry about distro releases complicating the version code logic).  It's code that we're bringing along, but not  testing any longer... @jpoimboe thoughts?

I'd love to rip kpatch.ko out entirely, but I think it still has some users for older kernels, like @euspectre if I'm not mistaken.

kmod/core/Makefile
```diff
@@ -1,24 +1,29 @@
...
+ifeq ($(shell if [ $(VERCODE) -lt 328192 ];then echo "y";fi),y)
 KPATCH_MAKE = $(MAKE) -C $(KPATCH_BUILD) M=$(THISDIR)
 
 kpatch.ko: core.c
 	$(KPATCH_MAKE) kpatch.ko
 
 all: kpatch.ko
+else
+all:
+$(info !!! Starting at v5.2, needed stacktrace routines are no longer available; not building kmod)
```
jpoimboe: When running make from the top-level dir, I see this message twice because it does a clean first.

euspectre:
> 
> I'd love to rip kpatch.ko out entirely, but I think it still has some users for older kernels, like @euspectre if I'm not mistaken.

Yes, we still use it for the kernels 3.x and 4.x.
I think it is OK to stop building kpatch.ko for 5.2 and newer.

joe-lawrence:
> > I'd love to rip kpatch.ko out entirely, but I think it still has some users for older kernels, like @euspectre if I'm not mistaken.
> 
> Yes, we still use it for the kernels 3.x and 4.x.
> I think it is OK to stop building kpatch.ko for 5.2 and newer.

Good know that it is still being used. But if the integration tests no longer target the stop_machine model and PR #1041 assumes the livepatch model in the author guide, does it make sense to build/use it automatically for those kernel releases?  Would it be safer to turn it off by default and let you guys manually flip it on with `BUILD_MOD=true make` or something like that?  That seems safer and simpler than @yhcote having to play version games.  /2 cents

joe-lawrence:
> 1. I have another patch that "fixes" that issue, so that we can build the module with v5.2+. It is blocked by the fact that some routines needed are not exported to modules and only accessible from built-in code. In the case where those get exported in the future, the patch would be submitted.

If you were to add kallsym lookups to eliminate the need for those symbol exports, @euspectre might be interested in that patch. 😄

euspectre:
> But if the integration tests no longer target the stop_machine model and PR #1041 assumes the livepatch model in the author guide, does it make sense to build/use it automatically for those kernel releases? Would it be safer to turn it off by default and let you guys manually flip it on with `BUILD_MOD=true make` or something like that?

From my point of view, it is OK not to build kpatch.ko by default.

euspectre:
> If you were to add kallsym lookups to eliminate the need for those symbol exports, @euspectre might be interested in that patch.

:-) Well, it is not decided yet, but it is very likely that we'll finally switch to upstream livepatch for the kernels 5.x. If noone else uses kpatch.ko, then, perhaps, there is no point in fixing it for 5.2+.

yhcote:
> > > I'd love to rip kpatch.ko out entirely, but I think it still has some users for older kernels, like @euspectre if I'm not mistaken.
> > 
> > 
> > Yes, we still use it for the kernels 3.x and 4.x.
> > I think it is OK to stop building kpatch.ko for 5.2 and newer.
> 
> Good know that it is still being used. But if the integration tests no longer target the stop_machine model and PR #1041 assumes the livepatch model in the author guide, does it make sense to build/use it automatically for those kernel releases? Would it be safer to turn it off by default and let you guys manually flip it on with `BUILD_MOD=true make` or something like that? That seems safer and simpler than @yhcote having to play version games. /2 cents

I'd be happy with that. If we decide to go that route and not build the module by default, I'll change this accordingly.

sm00th:
> I'd be happy with that. If we decide to go that route and not build the module by default, I'll change this accordingly.

If we are voting I'd vote for this easier approach in favor of make gymnastics.

PR #1043, opened 2019-10-02
create-diff-object: simplify mangled function correlation

### LINKED ISSUE #935: "duplicate" mangled fuctions
While running integration tests on ppc64le I found out that gcc-constprop.patch is failing with 
```
ERROR: timekeeping.o: symbol changed sections: .text.timekeeping_update.constprop.9
```

It turns out that there are 2 .constprop.* subfunctions in each objectfile:

```
# eu-readelf -s orig/kernel/time/timekeeping.o  | grep timekeeping_update
  137: 0000000000000000    552 FUNC    LOCAL  DEFAULT       48 timekeeping_update.constprop.9
  138: 0000000000000000    568 FUNC    LOCAL  DEFAULT       50 timekeeping_update.constprop.8
# eu-readelf -s patched/kernel/time/timekeeping.o  | grep timekeeping_update
  137: 0000000000000000    552 FUNC    LOCAL  DEFAULT       48 timekeeping_update.constprop.10
  138: 0000000000000000    568 FUNC    LOCAL  DEFAULT       50 timekeeping_update.constprop.9
```

So `kpatch_rename_mangled_functions()` gets very confused and we end up with two .9 functions in patched objfile.

Objfiles can be found at http://dump.trac3r.net/doubleconstprop/

jpoimboe:
The RHEL powerpc kernel is compiled with -O3, which triggers some
"interesting" new optimizations.  One of them, which seems to be
relatively common, is the replacing of a function with two separate
"constprop" functions.

Previously we only ever saw a single constprop clone, so we just renamed
the patched version of the function to match the original version.  Now
that we can have two such clones, that no longer makes sense.

Instead of renaming functions, just improve the correlation logic such
that they can be correlated despite having slightly different symbol
names.  The first clone in the original object is correlated with the
first clone in the patched object; the second clone is correlated with
the second clone; and so on.

This assumes that the order of the symbols and sections doesn't change,
which seems to be a reasonable assumption based on past experience with
the compiler.  Otherwise it will just unnecessarily mark the cloned
constprop functions as changed, which is annoying but harmless, and
noticeable by a human anyway.

Fixes #935.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -297,6 +297,13 @@ static int is_special_static(struct symbol *sym)
  */
 static int kpatch_mangled_strcmp(char *s1, char *s2)
 {
+	/*
+	 * ELF string sections aren't mangled, though they look that way.  Just
+	 * compare them normally.
+	 */
+	if (strstr(s1, ".str1."))
+		return strcmp(s1, s2);
+
 	while (*s1 == *s2) {
 		if (!*s1)
 			return 0;
@@ -389,11 +396,7 @@ static int rela_equal(struct rela *rela1, struct rela *rela2)
 	if (rela_toc1->addend != rela_toc2->addend)
 		return 0;
 
-	if (is_special_static(rela_toc1->sym))
-		return !kpatch_mangled_strcmp(rela_toc1->sym->name,
-					      rela_toc2->sym->name);
-
-	return !strcmp(rela_toc1->sym->name, rela_toc2->sym->name);
+	return !kpatch_mangled_strcmp(rela_toc1->sym->name, rela_toc2->sym->name);
 }
 
 static void kpatch_compare_correlated_rela_section(struct section *sec)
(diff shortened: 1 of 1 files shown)


jpoimboe:
I will add a powerpc unit test for this hopefully tomorrow.

jpoimboe:
I think it's a bit of a chicken-egg problem, but we can merge the test PR immediately before rebasing this commit with the updated test-objs pointer and then merging this PR.

sm00th:
I think it is a normal submodule workflow. If we merge test-obj PR right now it won't affect the main repo since the submodule pointer will still use old commit-id, we can then add a commit updating the pointer to this PR so once it is merged we'll have both the test and accompanying changes. No races anywhere.

jpoimboe:
> I think it is a normal submodule workflow. If we merge test-obj PR right now it won't affect the main repo since the submodule pointer will still use old commit-id, we can then add a commit updating the pointer to this PR so once it is merged we'll have both the test and accompanying changes. No races anywhere.

Right, of course.... Not sure how I managed to forget that.  I'll go ahead and merge the test PR now.

jpoimboe:
This PR is now in shape for merging, once reviews are done.

julien-thierry:
While running the yet unmerged test for rhel-8.0, I run into errors such as:

ERROR: aio.o: kpatch_create_patches_sections: 2708: lookup_local_symbol put_aio_ring_file.isra.19

(this is on gcc-static-local-var-4.patch)

The reason is that when creating the patch object, the lookups for existing symbols are done on the original object which has the original names. Since we no longer rename the mangled functions, lookups done for changed mangled symbols no longer work.

My suggestion would be to do the renaming of the symbol after the correlation has been done (i.e. once we have set the twin pointer + compared the correlated objects).

If that sounds like the right approach I can send a patch for this.

PR #1047, opened 2019-10-10
[RFC] per-object patches

sm00th:
A stab at making kpatch produce per-object modules instead of a single combined one. Throwing it out there in case someone else is working (or was planning to) on this as well.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -785,8 +785,6 @@ verify_patch_files
 apply_patches
 remove_patches
 
-cp -LR "$DATADIR/patch" "$TEMPDIR" || die
-
 if [[ "$ARCH" = "ppc64le" ]]; then
 	ARCH_KCFLAGS="-mcmodel=large -fplugin=$PLUGINDIR/ppc64le-plugin.so"
 fi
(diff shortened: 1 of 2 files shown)


joe-lawrence:
Just curious, what kinds of input patches have you thrown at this?

Might be interesting to see how kpatch-build blows up if you modify some heavily used header file that rebuilds all the modules :)

sm00th:
> Just curious, what kinds of input patches have you thrown at this?

I've been trying out small patches that affect couple of modules + vmlinux. But I do plan to ran through integration tests that will compile on a recent kernel. 

> Might be interesting to see how kpatch-build blows up if you modify some heavily used header file that rebuilds all the modules :)

Thats a great testcase. I tried a meminfo-SLOW patch today and it worked as expected to the point where I hit a CSWITCH issue (and thats why we don't run SLOW patches in integration tests anymore).

The only reason for it to blow up I currently see is a module name collision, which I am not sure can happen.

joe-lawrence:
> The only reason for it to blow up I currently see is a module name collision, which I am not sure can happen.

Was that two similarly named modules from different parts of the source tree?

sm00th:
> > The only reason for it to blow up I currently see is a module name collision, which I am not sure can happen.
> Was that two similarly named modules from different parts of the source tree?

Yes, I've never seen this in a wild but if there are two modules with the exact same name we might have a problem because for now it keeps object files under 'modname.ko/path/to/objfile.o'. I wonder how modprobe handles this or if this is explicitly prohibited somehow.

yhcote:
Work brought into another dev branch.

PR #1048, opened 2019-10-10
Correlation rename fixes

julien-thierry:
As mentioned on https://github.com/dynup/kpatch/pull/1043.

This change actually breaks some tests I was rebasing for rhel-8.0 with the following kind of error:
(this is on gcc-static-local-var-4.patch)
ERROR: aio.o: kpatch_create_patches_sections: 2708: lookup_local_symbol put_aio_ring_file.isra.19

The issue is due to the fact that we dropped symbol renaming when correlating. For symbols that end with status "SAME" there is no problem as we don't include those in the final object. For symbols that have status "CHANGED", we need to lookup the original symbols to know their address/position in the kernel/module getting patched, but at the point of the lookup, we only have the name from the patched object.
In the above example, put_aio_ring_file.isra.19 from the patched object gets correlated to put_aio_ring_file.isra.20 from the original object. But when creating the patch module itself, we no longer have the information of the original name.

My suggestion is to always rename mangled elements with the name of the object they get correlated to (last patch).

I also took the opportunity to do some cleanup related to correlation.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -776,6 +776,13 @@ do {								\
 	}							\
 } while (0)
 
+#define UNCORRELATE_ELEMENT(_elem_)				\
+do {								\
+	typeof(_elem_) elem = (_elem_);				\
+	elem->twin->twin = NULL;				\
+	elem->twin = NULL;					\
+} while (0)
+
 static void __kpatch_correlate_section(struct section *sec1, struct section *sec2)
 {
 	CORRELATE_ELEMENT(sec1, sec2, "section");
@@ -784,6 +791,11 @@ static void __kpatch_correlate_section(struct section *sec1, struct section *sec
 static void kpatch_correlate_symbol(struct symbol *sym1, struct symbol *sym2)
 {
 	CORRELATE_ELEMENT(sym1, sym2, "symbol");
+}
+
+static void kpatch_correlate_static_local(struct symbol *sym1, struct symbol *sym2)
+{
+	CORRELATE_ELEMENT(sym1, sym2, "static local");
 }
 
 static void kpatch_correlate_section(struct section *sec1, struct section *sec2)
@@ -1036,23 +1048,17 @@ static void kpatch_correlate_static_local_variables(struct kpatch_elf *base,
 		if (!kpatch_is_normal_static_local(sym))
 			continue;
 
-		if (sym->twin) {
-			sym->twin->twin = NULL;
-			sym->twin = NULL;
-		}
+		if (sym->twin)
+			UNCORRELATE_ELEMENT(sym);
 
 		bundled = sym == sym->sec->sym;
 		if (bundled && sym->sec->twin) {
-			sym->sec->twin->twin = NULL;
-			sym->sec->twin = NULL;
+			UNCORRELATE_ELEMENT(sym->sec);
 
-			sym->sec->secsym->twin->twin = NULL;
-			sym->sec->secsym->twin = NULL;
+			UNCORRELATE_ELEMENT(sym->sec->secsym);
 
-			if (sym->sec->rela) {
-				sym->sec->rela->twin->twin = NULL;
-				sym->sec->rela->twin = NULL;
-			}
+			if (sym->sec->rela)
+				UNCORRELATE_ELEMENT(sym->sec->rela);
 		}
 	}
 
@@ -1106,28 +1112,10 @@ static void kpatch_correlate_static_local_variables(struct kpatch_elf *base,
 				ERROR("sections %s and %s aren't correlated for symbol %s",
 				      sym->sec->name, patched_sym->sec->name, sym->name);
 
-			log_debug("renaming and correlating static local %s to %s\n",
-				  patched_sym->name, sym->name);
+			kpatch_correlate_static_local(sym, patched_sym);
 
-			patched_sym->name = strdup(sym->name);
-			sym->twin = patched_sym;
-			patched_sym->twin = sym;
-
-			/* set initial status, might change */
-			sym->status = patched_sym->status = SAME;
-
-			if (bundled) {
-				sym->sec->twin = patched_sym->sec;
-				patched_sym->sec->twin = sym->sec;
-
-				sym->sec->secsym->twin = patched_sym->sec->secsym;
-				patched_sym->sec->secsym->twin = sym->sec->secsym;
-
-				if (sym->sec->rela && patched_sym->sec->rela) {
-					sym->sec->rela->twin = patched_sym->sec->rela;
-					patched_sym->sec->rela->twin = sym->sec->rela;
-				}
-			}
+			if (bundled)
+				kpatch_correlate_section(sym->sec, patched_sym->sec);
 		}
 	}
 
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -760,11 +760,51 @@ static void kpatch_compare_symbols(struct list_head *symlist)
 	}
 }
 
+#define CORRELATE_ELEMENT(_e1_, _e2_)				\
+do {								\
+	typeof(_e1_) e1 = (_e1_);				\
+	typeof(_e2_) e2 = (_e2_);				\
+	e1->twin = e2;						\
+	e2->twin = e1;						\
+	/* set initial status, might change */			\
+	e1->status = e2->status = SAME;				\
+} while (0)
+
+static void __kpatch_correlate_section(struct section *sec1, struct section *sec2)
+{
+	CORRELATE_ELEMENT(sec1, sec2);
+}
+
+static void kpatch_correlate_symbol(struct symbol *sym1, struct symbol *sym2)
+{
+	CORRELATE_ELEMENT(sym1, sym2);
+}
+
+static void kpatch_correlate_section(struct section *sec1, struct section *sec2)
+{
+	__kpatch_correlate_section(sec1, sec2);
+
+	if (is_rela_section(sec1)) {
+		__kpatch_correlate_section(sec1->base, sec2->base);
+		sec1 = sec1->base;
+		sec2 = sec2->base;
+	} else if (sec1->rela) {
+		__kpatch_correlate_section(sec1->rela, sec2->rela);
+	}
+
+	if (sec1->secsym)
+		kpatch_correlate_symbol(sec1->secsym, sec2->secsym);
+	if (sec1->sym)
+		kpatch_correlate_symbol(sec1->sym, sec2->sym);
+}
+
 static void kpatch_correlate_sections(struct list_head *seclist1, struct list_head *seclist2)
 {
 	struct section *sec1, *sec2;
 
 	list_for_each_entry(sec1, seclist1, list) {
+		if (sec1->twin)
+			continue;
 		list_for_each_entry(sec2, seclist2, list) {
 			if (kpatch_mangled_strcmp(sec1->name, sec2->name) ||
 			    sec2->twin)
@@ -786,10 +826,8 @@ static void kpatch_correlate_sections(struct list_head *seclist1, struct list_he
 				           sec1->data->d_size))
 					continue;
 			}
-			sec1->twin = sec2;
-			sec2->twin = sec1;
-			/* set initial status, might change */
-			sec1->status = sec2->status = SAME;
+
+			kpatch_correlate_section(sec1, sec2);
 			break;
 		}
 	}
@@ -800,6 +838,8 @@ static void kpatch_correlate_symbols(struct list_head *symlist1, struct list_hea
 	struct symbol *sym1, *sym2;
 
 	list_for_each_entry(sym1, symlist1, list) {
+		if (sym1->twin)
+			continue;
 		list_for_each_entry(sym2, symlist2, list) {
 			if (kpatch_mangled_strcmp(sym1->name, sym2->name) ||
 			    sym1->type != sym2->type || sym2->twin)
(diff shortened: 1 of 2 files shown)


julien-thierry:
I added a unit test highlighting the issue:
https://github.com/dynup/kpatch-unit-test-objs/pull/19

kpatch-build/kpatch-elf.c
```diff
@@ -858,6 +860,8 @@ void kpatch_elf_teardown(struct kpatch_elf *kelf)
 	}
 
 	list_for_each_entry_safe(sym, safesym, &kelf->symbols, list) {
+		if (sym->twin)
+			sym->twin = NULL;
```
joe-lawrence: Just curious, where do the stale references occur?  
In this instance, wouldn't the subsequent memset be clearing sym->twin value anyway?

kpatch-build/create-diff-object.c
```diff
@@ -760,11 +760,39 @@ static void kpatch_compare_symbols(struct list_head *symlist)
 	}
 }
 
+#define CORRELATE_ELEMENTS(e1, e2)				\
```
joe-lawrence: nit: not a big deal, but wouldn't the compiler figure out if inlining is a win here?

kpatch-build/kpatch-elf.c
julien-thierry: Whoops, typo from my end, this should be like for the sections:
`sym->twin->twin = NULL;`

I don't think there is currently a bug due to that. It is just to make sure no-one is tempted to check the `twin` pointers of the patched object when building the final object as they reference freed object.

kpatch-build/create-diff-object.c
julien-thierry: I didn't go for a macro for inlining here but because I can use the macro for both `struct symbol *` and `struct section *` expressions.

If that's not considered good practice, I can have two static functions, but their code will be pretty much identical.

kpatch-build/create-diff-object.c
```diff
@@ -760,11 +760,45 @@ static void kpatch_compare_symbols(struct list_head *symlist)
...
+do {								\
+	e1->twin = e2;						\
+	e2->twin = e1;						\
+	/* set initial status, might change */			\
+	e1->status = e2->status = SAME;				\
+	if (strcmp(e1->name, e2->name)) {			\
+		/* Rename mangled element */			\
+		log_debug("renaming %s to %s\n",		\
```
sm00th: Would be nice if we could somehow differentiate between sections and symbols here.

kpatch-build/create-diff-object.c
```diff
@@ -766,6 +766,12 @@ do {								\
 	e2->twin = e1;						\
 	/* set initial status, might change */			\
 	e1->status = e2->status = SAME;				\
+	if (strcmp(e1->name, e2->name)) {			\
+		/* Rename mangled element */			\
+		log_debug("renaming %s to %s\n",		\
+			  e2->name, e1->name);			\
+		e2->name = strdup(e1->name);			\
```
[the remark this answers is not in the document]
julien-thierry: The name of an element might point to a string in a elf file which might be closed when we still are using e2, so we need to dup (or not close the original elf nor the patched elf until we quit the program).

As for memory leaking, I must admit I followed the way the code Josh deleted used to handle things. The issue is that at this point a section or symbol name could point to memory mapped from an elf file or memory allocated through malloc.
Since kpatch is just a "one-off" program, having this memory allocated until the end of the program should be too much of a big deal. But if we want to avoid that we can strdup all names and free them in the kpatch_elf_teardown().

Let me know if we shoudl go with that.

kpatch-build/create-diff-object.c
julien-thierry: So, have two separate static functions?

kpatch-build/create-diff-object.c
```diff
@@ -760,11 +760,39 @@ static void kpatch_compare_symbols(struct list_head *symlist)
...
+	if (sec->sym)
+		CORRELATE_ELEMENTS(sec->sym, sec->twin->sym);
+}
+
 static void kpatch_correlate_sections(struct list_head *seclist1, struct list_head *seclist2)
 {
 	struct section *sec1, *sec2;
 
 	list_for_each_entry(sec1, seclist1, list) {
+		if (sec1->twin)
```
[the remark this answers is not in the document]
julien-thierry: The `kpatch_process_section()` might already have done the correlation for an associated rela/base section. In which case, looking for another element with a matching mangled name can match to a different object (since the correct one will be already correlated, i.e. `sec2->twin != NULL`, we can't match to it again).

kpatch-build/create-diff-object.c
```diff
@@ -800,6 +828,8 @@ static void kpatch_correlate_symbols(struct list_head *symlist1, struct list_hea
 	struct symbol *sym1, *sym2;
 
 	list_for_each_entry(sym1, symlist1, list) {
+		if (sym1->twin)
```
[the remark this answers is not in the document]
julien-thierry: More or less the same story as above. `kpatch_process_section()` perform the correlation for the section symbol or a symbol that is bundled.

kpatch-build/create-diff-object.c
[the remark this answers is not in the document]
sm00th: Well, I personally try to keep leaks to minimum even for "one-offs", but I did see a comment somewhere in the same vein of "we don't care since its a one-off", so I'll leave it for @jpoimboe to decide.

kpatch-build/create-diff-object.c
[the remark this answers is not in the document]
sm00th: Do you mean `kpatch_correlate_sections()`? I don't see how that would happen since it goes through each list only once.

kpatch-build/create-diff-object.c
sm00th: Well every way of doing this seems to be as inconvenient as separate functions... But since those are bundled together maybe a `log_debug` at the beginning of `kpatch_correlate_symbols` and `kpatch_correlate_sections` will be enough.

kpatch-build/create-diff-object.c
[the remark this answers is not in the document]
julien-thierry: No, sorry I meant, `kpatch_process_section_correlation()` which I introduce in this commit.

Maybe its introduction is being a bit overzealous, but taking multiple objects  with matching mangled names into consideration (which was the original motivation for @jpoimboe 's change), I don't know how e.g. we guarantee we would avoid the following scenario:
orig:
- sym_a -> bundled into sec_a

patched:
- sym_b -> bundled into sec_b
- sym_c -> bundled into sec_c

If both sym_b and sym_c have names matching sym_a through mangling, we could end up with sym_a being correlated with sym_b but sec_a being correlated with sec_c.
Same thing with rela/base sections.

So basically I introduced the `kpatch_process_section_correlation()` to avoid that. But the consequence is that when you correlate a section, you might correlate another one (and a symbol) before having treated in in the list_foreach. If that happens, it needs to be skipped.

kpatch-build/create-diff-object.c
julien-thierry: Oh, so you meant differentiate the message printed?

Yes, that makes sense, I'll do that in the next version.

kpatch-build/create-diff-object.c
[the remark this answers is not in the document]
julien-thierry: I could only see two ways to solve this:
- Always use malloc for the name of symbols and sections, but that would make us always use more memory than we need to avoid a leak that the OS will cleanup once the program finishes.
- Have a way to distinguish whether the name was allocated or not and check at the end.

First solution doesn't seems a bit wasteful just to avoid a leak that isn't that much of an issue (in my opinion) and I would prefer to keep the leak.
Second solution can be a bit awkward, but after sleeping on it we could pretend to make the distinction a bit more "useful". We could have a field `orig_name` in section and symbol structures, which by default would stay `NULL`. And if an element gets renamed `orig_name` would point to the string in the ELF while `name` would point to freeable memory. So we can know there is a need to free `name` if `orig_name != NULL` or something like that. While there is no big need for keeping the original name, I can at least see some use to it for debugging (e.g. could access it through GDB) or logging purposes.

Otherwise we can also keep the leak.

Let me know which you prefer.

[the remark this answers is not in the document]
julien-thierry: Also, I just check, we already have the same leak when renaming static local variables.

kpatch-build/create-diff-object.c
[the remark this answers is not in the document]
sm00th: Ok, that makes more sense now. Not sure how correct is this approach but I don't have a better solution on hand.

kpatch-build/create-diff-object.c
joe-lawrence: Ah, I see.  I don't have a strong preference either way then.

kpatch-build/create-diff-object.c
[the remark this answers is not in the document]
joe-lawrence: FWIW, some time in the past I wrapped kpatch-build with valgrind and tried to get the non-paths to be mostly silent.  For a userspace program, I wasn't going to pull my hair out getting all the various error paths to be perfect... but at the very least a clean successful build is helpful to find any new bad behavior.  /two-cents

julien-thierry:
I updated according to the comments.

I also added some RFC commits that are not needed for the functionality but that prevent the leaks related to renaming.

kpatch-build/create-diff-object.c
```diff
@@ -760,11 +760,72 @@ static void kpatch_compare_symbols(struct list_head *symlist)
...
+
+static void make_correlated_sections(struct section *sec1, struct section *sec2)
+{
+	CORRELATE_ELEMENTS(sec1, sec2, "section");
+}
+
+static void make_correlated_symbols(struct symbol *sym1, struct symbol *sym2)
+{
+	CORRELATE_ELEMENTS(sym1, sym2, "symbol");
+}
```
jpoimboe: Naming suggestions: make_correlated_sections() almost makes it sound like it's creating a section - how about renaming to kpatch_correlate_section()?   That also gives it a name analagous to its caller kpatch_correlate_sections().  And a similar comment for make_correlated_symbols().

And that would also imply a CORRELATE_ELEMENTS -> CORRELATE_ELEMENT rename for consistency - yes, it's correlating two elements, but depending on one's perspective they might be considered the same element :-)

kpatch-build/create-diff-object.c
```diff
@@ -1063,12 +1117,7 @@ static void kpatch_correlate_static_local_variables(struct kpatch_elf *base,
 				ERROR("sections %s and %s aren't correlated for symbol %s",
 				      sym->sec->name, patched_sym->sec->name, sym->name);
 
-			log_debug("renaming and correlating static local %s to %s\n",
-				  patched_sym->name, sym->name);
-
-			patched_sym->name = strdup(sym->name);
-			sym->twin = patched_sym;
-			patched_sym->twin = sym;
+			CORRELATE_ELEMENTS(sym, patched_sym, "static local");
```
jpoimboe: Here you might as well call a kpatch_correlate_static_local_variable() wrapper which hides the ugly macro, similar to the other two helpers.

kpatch-build/create-diff-object.c
```diff
@@ -1063,12 +1117,7 @@ static void kpatch_correlate_static_local_variables(struct kpatch_elf *base,
...
 
 			/* set initial status, might change */
 			sym->status = patched_sym->status = SAME;
```
jpoimboe: I think this is no longer needed since CORRELATE_ELEMENTS does this?

kpatch-build/create-diff-object.c
julien-thierry: I was a bit worried about the potential kpatch_correlate_section/sections confusion.
But now that you mention it, I agree it makes more sense the way you suggest.
I'll do the upate.

kpatch-build/create-diff-object.c
julien-thierry: Will do.

kpatch-build/create-diff-object.c
julien-thierry: Yes, well spotted!

kpatch-build/create-diff-object.c
```diff
@@ -1063,27 +1114,14 @@ static void kpatch_correlate_static_local_variables(struct kpatch_elf *base,
...
 
-				sym->sec->secsym->twin = patched_sym->sec->secsym;
-				patched_sym->sec->secsym->twin = sym->sec->secsym;
-
-				if (sym->sec->rela && patched_sym->sec->rela) {
-					sym->sec->rela->twin = patched_sym->sec->rela;
-					patched_sym->sec->rela->twin = sym->sec->rela;
-				}
+				if (sym->sec->rela && patched_sym->sec->rela)
+					kpatch_correlate_section(sym->sec->rela, patched_sym->sec->rela);
```
jpoimboe: This is all quite similar to what kpatch_process_section_correlation() does... could it all be replaced with
```
if (bundled) {
        kpatch_correlate_section(sym->sec, patched_sym->sec);
        kpatch_process_section_correlation(sym->sec);
}
```
?

And for that matter I wonder if it would be better to have kpatch_correlate_section() automatically do all the work in kpatch_process_section_correlation()?  Then kpatch_correlate_section() could call a lower-level __kpatch_correlate_section() version which doesn't do that extra work.  So kpatch_correlate_section() would be a higher-level interface which just does the right thing, and then this could just be
```
if (bundled)
        kpatch_correlate_section(sym->sec, patched_sym->sec);
```
And similarly kpatch_correlate_sections() wouldn't need to call kpatch_process_section_correlation().

julien-thierry: Yes, having kpatch_correlate_section() do all the needed work seems like a good suggestion.
Thanks.

jpoimboe:
@julien-thierry once the test PR is merged, you can rebase this PR to update the test/unit/objs ref to reference the new test PR.  Then feel free to merge this one.

PR #1052, opened 2019-10-23
Make symvers reading code more flexible.

sm00th:
Kernel commit cb9b55d21fe0 modpost: add support for symbol namespaces
adds a new namespace column to Module.symvers file which can be blank.
fscanf is no longer a viable solution to parse that. Switch to the way
scripts/mod/modpost.c handles this and try to support both versions with
and without namespace column.


sm00th:
Added a new comment above symvers_read() prototype.

PR #1053, opened 2019-11-01
kpatch-build: Make local static twining stricter

julien-thierry:
The current code to find the twin of a local static variable allows two
variables of the same name to be wrongly matched with the other's twin.

While there isn't a magic formula to avoid this, make stricter
requirements for twining static local from the original object with
a symbol from the patched object. This reduces the risk of erroneous
matches.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -975,18 +975,26 @@ static struct symbol *kpatch_find_static_twin(struct section *sec,
 
 	/* find the patched object's corresponding variable */
 	list_for_each_entry(rela, &sec->twin->relas, list) {
+		struct symbol *patched_sym;
 
 		rela_toc = toc_rela(rela);
 		if (!rela_toc)
 			continue; /* skip toc constants */
 
-		if (rela_toc->sym->twin)
+		patched_sym = rela_toc->sym;
+
+		if (patched_sym->twin)
 			continue;
 
-		if (kpatch_mangled_strcmp(rela_toc->sym->name, sym->name))
+		if (sym->type != patched_sym->type ||
+		    (sym->type == STT_OBJECT &&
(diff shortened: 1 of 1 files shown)


julien-thierry:
This fixes the error on RHEL8.0 test macro-printk:

ERROR: fib_semantics.o: object size mismatch: __msg.66237

Mentioned in PR #993 .

jpoimboe:
Looks good.  Would it be possible to create a unit test for this one?

julien-thierry:
> Looks good. Would it be possible to create a unit test for this one?

Sure, I'll do that.

jpoimboe:
Thanks.  We want the unit tests to be as comprehensive as possible.  I think some overlap between unit tests and integration tests is ok.

kpatch-build/create-diff-object.c
```diff
@@ -975,15 +975,23 @@ static struct symbol *kpatch_find_static_twin(struct section *sec,
...
+		if (sym->type != patched_sym->type ||
+		    (sym->type == STT_OBJECT &&
+		     sym->sym.st_size != patched_sym->sym.st_size))
 			continue;
 
-		if (kpatch_mangled_strcmp(rela_toc->sym->name, sym->name))
+		if (kpatch_mangled_strcmp(patched_sym->name, sym->name))
 			continue;
 
 		return rela_toc->sym;
```
sm00th: Might as well return `patched_sym` here for readability.

julien-thierry: You're right, I forgot to replace this reference.
Thanks for spotting it.

I'll update this along with the submodule update once dynup/kpatch-unit-test-objs#20 is merged.

joe-lawrence:
I think all the requested changes were done for this PR and that it's approved.  Any thing left before merging it?

PR #1055, opened 2019-11-01
create-diff-object: process debug sections last

jpoimboe:
Process the debug sections only after all the other inclusion logic has
finished, since it makes decisions based on what else has already been
included.


jpoimboe:
This was found via code inspection, so I don't really have a unit test for it.

joe-lawrence:
This PR looks pretty trivial, was there any reason we held off merging?

PR #1056, opened 2019-11-02
dynrela cleanups

jpoimboe:
Some long overdue cleanups to the dynrela and lookup code.  In theory this should be no functional change.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2719,9 +2719,12 @@ static void kpatch_create_patches_sections(struct kpatch_elf *kelf,
 
 	/* count patched functions */
 	nr = 0;
-	list_for_each_entry(sym, &kelf->symbols, list)
-		if (sym->type == STT_FUNC && sym->status == CHANGED && !sym->parent)
-			nr++;
+	list_for_each_entry(sym, &kelf->symbols, list) {
+		if (sym->type != STT_FUNC || sym->status != CHANGED ||
+		    sym->parent)
+			continue;
+		nr++;
+	}
 
 	/* create text/rela section pair */
 	sec = create_section_pair(kelf, ".kpatch.funcs", sizeof(*funcs), nr);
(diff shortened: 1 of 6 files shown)


kpatch-build/create-diff-object.c
```diff
@@ -2562,9 +2562,12 @@ static void kpatch_create_patches_sections(struct kpatch_elf *kelf,
...
 	/* count patched functions */
 	nr = 0;
-	list_for_each_entry(sym, &kelf->symbols, list)
-		if (sym->type == STT_FUNC && sym->status == CHANGED && !sym->parent)
-			nr++;
+	list_for_each_entry(sym, &kelf->symbols, list) {
+		if (sym->type != STT_FUNC || sym->status != CHANGED ||
+		    sym->parent)
+			continue;
+		nr++;
```
julien-thierry: I don't really feel this is improving readability in this particular instance.
The change here just adds a statement rather than simplify overall indentation.

(I do agree the other change in this commit makes the code nicer to read).

jpoimboe: In isolation I would agree, but the benefit is that this makes the condition identical to the 2nd loop's conditional, so the reader can verify the conditions are the same and 'nr' is valid.

kpatch-build/kpatch-intermediate.h
```diff
@@ -26,7 +26,7 @@
 
 struct kpatch_symbol {
 	unsigned long src;
-	unsigned long pos;
+	unsigned long sympos;
```
julien-thierry: I don't have a very strong opinion on this, but adding "sym" to an element of a symbol feels a bit redundant.
I think the change is good for lookup_result however.

kpatch-build/lookup.c
```diff
@@ -414,19 +414,18 @@ bool lookup_local_symbol(struct lookup_table *table, char *name,
 			break;
 
 		if (sym->bind == STB_LOCAL && !strcmp(sym->name, name)) {
-			match = 1;
-			break;
+
+			if (result->objname)
+				ERROR("duplicate local symbol found for %s", name);
```
julien-thierry: Could we dump a bit more information here?
Like the address of both symbols? (And maybe the size)

kpatch-build/lookup.c
```diff
@@ -439,15 +438,18 @@ bool lookup_global_symbol(struct lookup_table *table, char *name,
...
 		if ((sym->bind == STB_GLOBAL || sym->bind == STB_WEAK) &&
 		    !strcmp(sym->name, name)) {
-			result->objname = table->objname;
-			result->addr = sym->addr;
-			result->size = sym->size;
-			result->sympos = 0; /* always 0 for global symbols */
-			return true;
+
+			if (result->objname)
+				ERROR("duplicate global symbol found for %s", name);
```
julien-thierry: Same remark as for looup_local_symbol().

kpatch-build/create-diff-object.c
julien-thierry: I see, fair point.

kpatch-build/kpatch-intermediate.h
jpoimboe: Good point.  I also don't have a strong feeling either way but I might prefer to leave it as 'sympos' despite the redundancy, because that's what we call the value upstream.

jpoimboe:
Rebased onto master.

jpoimboe:
This is a prerequisite for an actual functional change which is coming soon which will

- add jump label support
- remove .klp.arch sections and arch-specific code
- finish Peter Z's upstream quest to disallow making executable mappings writable

kpatch-build/create-diff-object.c
joe-lawrence: Initially I thought the same as @julien-thierry, but then the code copy part made sense.  I guess if it were copied again, an inline helper function/macro could made it easier on the eyes.

kpatch-build/kpatch-intermediate.h
joe-lawrence: IMHO sympos is clearer (coming from someone familiar with the upstream code)

joe-lawrence:
> This is a prerequisite for an actual functional change which is coming soon which will
> 
>     * add jump label support
>     * remove .klp.arch sections and arch-specific code
>     * finish Peter Z's upstream quest to disallow making executable mappings writable

Feel free to take this to email, but can you elaborate on these?  Thanks.

kpatch-build/create-diff-object.c
yhcote: Might as well add my tidbit, reviewing from a checked out copy of the file made the intention apparent.

jpoimboe:
@joe-lawrence The short answer is that this is related to some work Peter Z did about 6 months ago to make module executable regions non-writeable.  Livepatch was the last holdout for allowing that to happen.  Peter Z convinced me that that .klp.arch sections are actually unnecessary.  The thread started here:

https://lkml.kernel.org/r/20191021135312.jbbxsuipxldocdjk@treble

Other things popped up since then and I'm just now getting the chance to fix it up.

The patches should hopefully be available soon.  They will have more details.

joe-lawrence:
> @joe-lawrence The short answer is that this is related to some work Peter Z did about 6 months ago to make module executable regions non-writeable. Livepatch was the last holdout for allowing that to happen. Peter Z convinced me that that .klp.arch sections are actually unnecessary. The thread started here:
> 
> https://lkml.kernel.org/r/20191021135312.jbbxsuipxldocdjk@treble
> 

Ah right, that thread.  Looks like some new restrictions coming for klp_write_relocations()?  This might intersect with the klp-convert fallout I was researching with respect to Petr's split livepatch PoC:

https://lore.kernel.org/lkml/20200117150323.21801-1-pmladek@suse.com/ 

> Other things popped up since then and I'm just now getting the chance to fix it up.
> 
> The patches should hopefully be available soon. They will have more details.

I don't know about you, but corona-quarantine has given me a chance to dust off some of those long-deferred upstream patches.  Looking forward to taking a look when they're ready.

jpoimboe:
> > @joe-lawrence The short answer is that this is related to some work Peter Z did about 6 months ago to make module executable regions non-writeable. Livepatch was the last holdout for allowing that to happen. Peter Z convinced me that that .klp.arch sections are actually unnecessary. The thread started here:
> > https://lkml.kernel.org/r/20191021135312.jbbxsuipxldocdjk@treble
> 
> Ah right, that thread. Looks like some new restrictions coming for klp_write_relocations()? This might intersect with the klp-convert fallout I was researching with respect to Petr's split livepatch PoC:
> 
> https://lore.kernel.org/lkml/20200117150323.21801-1-pmladek@suse.com/

Yeah, I was thinking that might be the case.  Hopefully things will be simpler with .klp.arch gone.

> > Other things popped up since then and I'm just now getting the chance to fix it up.
> > The patches should hopefully be available soon. They will have more details.
> 
> I don't know about you, but corona-quarantine has given me a chance to dust off some of those long-deferred upstream patches. Looking forward to taking a look when they're ready.

For those of us with kids around 24x7, it's not quite the same ;-)  But I do seem to have the time lately to finish up some of these old patches, which is nice.

Issue #1064, opened 2020-01-17
ZERO_PAGE(0) returns a wrong value in the patched functions


euspectre:
On x86_64, GCC generates the following instruction to compute 'empty_zero_page - __START_KERNEL_map' (__phys_addr_nodebug(), used in the implementation of ZERO_PAGE()):

    48 ba 00 00 00 00 00 00 00 00   movabs $0x0,%rdx
        R_X86_64_64  empty_zero_page+0x80000000

__START_KERNEL_map is 0xffffffff80000000.

However, the relocation addend becomes wrong in the patch module:

    48 ba 00 00 00 00 00 00 00 00   movabs $0x0,%rdx
        R_X86_64_64  empty_zero_page-0x80000000

Note the sign of the addend.

As a result, ZERO_PAGE(0) returns a wrong value in any function touched by the patch, which may lead to memory corruption and difficult-to-debug kernel crashes.

The cause is that 'struct rela' uses 'int' for the addend, which is not enough to store such values. r_addend from Elf64_Rela is int64_t (Elf64_Sxword) for that.

----------------

Here are the files for testing, created for the kernel 3.10.0-1062.9.1.el7 from CentOS 7:
[zero_page_reloc_bug.tar.gz](https://github.com/dynup/kpatch/files/4077733/zero_page_reloc_bug.tar.gz)

```
$ tar xf zero_page_reloc_bug.tar.gz
$ cd zero_page_reloc_bug
$ create-diff-object orig.o patched.o vmlinux vmlinux.symtab Module.symvers something output.o
orig.o: changed function: meminfo_proc_open
$ objdump -r patched.o | grep empty_zero_page
0000000000000008 R_X86_64_64       empty_zero_page+0x0000000080000000
000000000000001a R_X86_64_32S      empty_zero_page
$ objdump -r output.o | grep empty_zero_page
0000000000000008 R_X86_64_64       empty_zero_page-0x0000000080000000
000000000000001a R_X86_64_32S      empty_zero_page
```

If I use 'long' for addend in 'struct rela', the relocation becomes correct.

PR #1065, opened 2020-01-17
create-diff-object: fix relocations used for ZERO_PAGE(0)

euspectre:
On x86_64, GCC generates the following instruction to compute
'empty_zero_page - __START_KERNEL_map' (__phys_addr_nodebug(), used in
the implementation of ZERO_PAGE()):

    48 ba 00 00 00 00 00 00 00 00   movabs $0x0,%rdx
          R_X86_64_64  empty_zero_page+0x80000000

__START_KERNEL_map is 0xffffffff80000000.

However, the relocation addend becomes wrong in the patch module:

    48 ba 00 00 00 00 00 00 00 00   movabs $0x0,%rdx
          R_X86_64_64  empty_zero_page-0x80000000

Note the sign of the addend.

As a result, ZERO_PAGE(0) returns a wrong value in any function touched
by the patch, which may lead to memory corruption and difficult-to-debug
kernel crashes.

The cause is that 'struct rela' uses 'int' for the addend, which is not
enough to store such values. r_addend from Elf64_Rela is int64_t
(Elf64_Sxword) for that.

Let us use 'long' instead of 'int' for the addend in 'struct rela'.

Fixes https://github.com/dynup/kpatch/issues/1064.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -374,14 +374,14 @@ static int rela_equal(struct rela *rela1, struct rela *rela2)
 		 */
 		memcpy(&toc_data1, rela1->sym->sec->data->d_buf + rela1->addend, sizeof(toc_data1));
 		if (!toc_data1)
-			ERROR(".toc entry not found %s + %x", rela1->sym->name, rela1->addend);
+			ERROR(".toc entry not found %s + %lx", rela1->sym->name, rela1->addend);
 	}
 
 	rela_toc2 = toc_rela(rela2);
 	if (!rela_toc2) {
 		memcpy(&toc_data2, rela2->sym->sec->data->d_buf + rela2->addend, sizeof(toc_data2));
 		if (!toc_data2)
-			ERROR(".toc entry not found %s + %x", rela2->sym->name, rela2->addend);
+			ERROR(".toc entry not found %s + %lx", rela2->sym->name, rela2->addend);
 	}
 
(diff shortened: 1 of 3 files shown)


jpoimboe:
Ouch, nice find...

yhcote:
This change looks good to me also. Should we not also change `unsigned int offset` for `unsigned long offset` in the same structure? Mostly for the form for when `rela->offset = rela->r_offset (Elf64_Addr)` since otherwise relocatable objects use file offsets for that field which an int should be able to contain.

euspectre:
> This change looks good to me also. Should we not also change `unsigned int offset` for `unsigned long offset` in the same structure? Mostly for the form for when `rela->offset = rela->r_offset (Elf64_Addr)` since otherwise relocatable objects use file offsets for that field which an int should be able to contain.

Is it possible for the kernel or livepatches to have relocations with 'r_offset' for which 'unsigned int' is not enough?

jpoimboe:
To catch other issues like this I wonder if we should enable `-Wconversion`.

jpoimboe:
Or maybe `-Wconversion -Wno-sign-conversion` which reduces the number of warnings quite a bit compared to just `-Wconversion`.

yhcote:
Is there a blocker for enabling -Wall/-Wextra combined with -Wno-* 's as needed?

yhcote:
> Is there a blocker for enabling -Wall/-Wextra combined with -Wno-* 's as needed?

-Wall seems to be there, missed it when I first checked.

PR #1068, opened 2020-01-23
Integration test stabilization

sm00th:
These commits disable/fix a bunch of test and small issues in order to make integration tests on all of the supported distros pass.


README.md
```diff
@@ -80,7 +80,7 @@ sudo debuginfo-install kernel-${UNAME%.*}
 sudo yum install gcc-plugin-devel
 
 # optional, but highly recommended
-sudo yum install https://dl.fedoraproject.org/pub/epel/7/x86_64/Packages/c/ccache-3.3.4-1.el7.x86_64.rpm
+sudo yum install https://dl.fedoraproject.org/pub/epel/7/x86_64/Packages/c/ccache-3.7.6-1.el7.x86_64.rpm
```
joe-lawrence: I have a kpatch setup script that just installs the epel, ccache and then removes epel to avoid this issue.   Feel free to steal it if you are tired of updating these.  (It should probably check to see if epel was installed originally, but I only use the script on test boxes.)
``` shell
# ccache from EPEL
yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-${RHEL}.noarch.rpm
yum install -y ccache
ccache --max-size=5G
yum erase -y epel-release
```

test/integration/vm-integration-run
```diff
@@ -63,6 +63,12 @@ if [[ -d "${HOME}/src" && -f "${HOME}/src/version" ]]; then
   mv "${HOME}/src" "${HOME}/.kpatch/"
 fi
 
+source /etc/os-release
+
+if [[ "${NAME}" == "Fedora" ]] && [[ "${VERSION_ID}" -le 27 ]]; then
```
jpoimboe: Why <= 27?  I think CONFIG_LIVEPATCH is disabled for all versions of Fedora?

sm00th: Ah, I didn't realize that. This is not entirely true though, starting with fc30 we do have it enabled: https://src.fedoraproject.org/rpms/kernel/c/3792ab811f3d097b7883cf43235c75ea410b3ebc

I'll fix this up.

README.md
sm00th: That is a great suggestion, I'll apply that.

sm00th:
v2:
 - use epel repo instead of specifying ccache url directly
 - kmod build for fedora <30 instead of <=27
 - fixed typo in one of commit subjects

test/integration/vm-integration-run
jpoimboe: Ah, nice... I'm glad Fedora enabled it, that will make it easier to test with new kernels.

Eventually we may want to have a rolling "latest Fedora" set of integration tests to catch regressions with new kernels.

PR #1071, opened 2020-02-13
integration: update gcc-static-local-var-4.test

### LINKED ISSUE #1069: test/integration/gcc-static-local-var-4.patch fails on RHEL-7.7 ppc64le
The gcc-static-local-var-4.test script checks that it can't find a "free_ioctx" symbol in the resulting kpatch .ko, but on ppc64le it finds klp-relocation symbols that do match:
```
(x86_64)
  % nm ~/.kpatch/tmp/patch/livepatch-gcc-static-local-var-4.ko | grep 'free_ioctx'
  (nothing)

(ppc64le)
  % nm ~/.kpatch/tmp/patch/livepatch-gcc-static-local-var-4.ko | grep 'free_ioctx'
  0000000000000000 a .klp.sym.vmlinux.free_ioctx,1
  0000000000000000 a .klp.sym.vmlinux.free_ioctx_rcu,1
```
So perhaps we should narrow the grep pattern to filter them out `-v .klp.sym.` and add word boundaries `\<free_ioctx\>` to be even more specific.

At the same time, I'm not entirely clear how this test works.  The [patch](https://github.com/dynup/kpatch/blob/master/test/integration/rhel-7.7/gcc-static-local-var-4.patch) adds a small function with a printk and calls it from put_aio_ring_file().  But the [test](https://github.com/dynup/kpatch/blob/master/test/integration/rhel-7.7/gcc-static-local-var-4.test) looks for the existence of a free_ioctx symbol.  Is the test checking that unmodified functions that contain dynamic debugging printk's aren't incorrectly included in the resulting livepatch .ko?

joe-lawrence:
The gcc-static-local-var-4.patch and test check that unmodified functions
that contain dynamic debugging printk's aren't incorrectly included in
the resulting livepatch .ko.

Note that on RHEL-7.7 ppc64le, the gcc-static-local-var-4.ko livepatch
module may include klp-relocation references to free_ioctx(), a function
that would match the above criteria.  These klp-relocation symbols are
OK as they are only used for referencing said functions, they are not
copies that the check should FAIL on.

Modify gcc-static-local-var-4.test in a few ways:

  - Convert it to a LOADED test so that we can ...

  - Search /proc/kallsyms instead of using binutils or elfutils (which
    may have issues parsing klp-relocations).

  - Instead of a grep regex, use awk to string compare exact function
    and module names.  This avoids matching on .klp.sym prefixed
    klp-relocation symbols.

Fixes #1069.

### DIFF
--- a/test/integration/rhel-8.0/gcc-static-local-var-4.test
+++ b/test/integration/rhel-8.0/gcc-static-local-var-4.test
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-if $(nm test-gcc-static-local-var-4.ko | grep -q free_ioctx); then
+set -o pipefail
+if ! $(eu-readelf --wide --symbols test-gcc-static-local-var-4.ko | awk '$NF == "free_ioctx" { exit 1 }'); then
 	exit 1
 else
 	exit 0
--- a/test/integration/centos-7/gcc-static-local-var-4.test
+++ b/test/integration/centos-7/gcc-static-local-var-4.test
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-if $(nm test-gcc-static-local-var-4.ko | grep -q free_ioctx); then
+set -o pipefail
+if ! $(eu-readelf --wide --symbols test-gcc-static-local-var-4.ko | awk '$NF == "free_ioctx" { exit 1 }'); then
 	exit 1
 else
 	exit 0
--- a/test/integration/rhel-7.4/gcc-static-local-var-4.test
+++ b/test/integration/rhel-7.4/gcc-static-local-var-4.test
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-if $(nm test-gcc-static-local-var-4.ko | grep -q free_ioctx); then
+set -o pipefail
+if ! $(eu-readelf --wide --symbols test-gcc-static-local-var-4.ko | awk '$NF == "free_ioctx" { exit 1 }'); then
 	exit 1
 else
 	exit 0
--- a/test/integration/rhel-7.5/gcc-static-local-var-4.test
+++ b/test/integration/rhel-7.5/gcc-static-local-var-4.test
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-if $(nm test-gcc-static-local-var-4.ko | grep -q free_ioctx); then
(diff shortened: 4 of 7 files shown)


joe-lawrence:
I wasn't happy about converting this to a LOADED test, but the only reliably klp-relocation compatible symbol dumps are from the kernel (/proc/kallsym) or to use elfutils programs.  Since there was an effort to remove elfutils usage, that leaves the kernel.   We could alternatively only fail this test iff we can read the symbol table (so silently failing cases in which binutils chokes on klp-relocs).

test/integration/rhel-8.0/gcc-static-local-var-4.test
```diff
@@ -1,6 +1,11 @@
 #!/bin/bash
 
-if $(nm test-gcc-static-local-var-4.ko | grep -q free_ioctx); then
```
sm00th: I assume this should be renamed to .LOADED as well.

joe-lawrence: Oops, I'll go rename that one, too

joe-lawrence:
v2: Rename test/integration/rhel-8.0/gcc-static-local-var-4.test as well (sm00th)

jpoimboe:
Why not just use eu-readelf and fail the test if the user doesn't have it installed?  I suppose it would be nice to standardize on binutils only, but if binutils is broken...

joe-lawrence:
> Why not just use eu-readelf and fail the test if the user doesn't have it installed? I suppose it would be nice to standardize on binutils only, but if binutils is broken...

I read this as judging this change too ugly to avoid broken binutils, yeah?

I don't think there is a perfect solution to this problem, so we're picking the least worst fix.  Adding elfutils to the integration tests makes it at least an integration test dependency.  I'm not sure what originally motivated #969, as I don't even know what that script is used for.

jpoimboe:
I don't see a problem with adding a dependency on elfutils, if certain versions of binutils are broken.  The test can automatically fail if the dependency is missing anyway.  If others disagree they can speak up, but I don't think there are many people running the integration tests.

joe-lawrence:
v3: use eu-readelf and awk to handle this test case

PR #1076, opened 2020-02-27
kpatch.service: make sure it starts before networking services

euspectre:
A binary patch may be used to fix network-related issues, so it is better to
apply it before networking services have started.

We encountered a situation in Virtuozzo 7, when the older kernels
conflicted with a new NetworkManager, ip utility and other system
components (https://www.mail-archive.com/devel@openvz.org/msg35123.html).

Binary patches were provided for these kernels to fix the issue but were
loaded after networking services in some cases. As a result, NetworkManager
and some other system components failed to work properly.

Let us make sure the patches are applied earlier during boot.

### DIFF
--- a/contrib/kpatch.service
+++ b/contrib/kpatch.service
@@ -1,6 +1,8 @@
 [Unit]
 Description="Apply kpatch kernel patches"
 ConditionKernelCommandLine=!kpatch.enable=0
+Before=network-pre.target
+Wants=network-pre.target
 
 [Service]
 Type=oneshot


jpoimboe:
The patch looks ok to me, but I'm curious what's the point of live patching if you have to apply the patch during boot?

euspectre:
> what's the point of live patching if you have to apply the patch during boot?

Many of our customers are very reluctant to update the kernels on their systems to get stability fixes ("new kernel = new bugs", they think) and prefer that we deliver such bug fixes via live patches instead.

Another scenario is when the customers' nodes are rebooted for some unrelated reason (system hang or crash). Loading live patches at boot helps a lot to keep the system operational in such cases.

euspectre:
For some kernel versions, the focus of our live kernel patches is shifting more and more from security fixes towards stability & performance fixes, it seems.

jpoimboe:
> Many of our customers are very reluctant to update the kernels on their systems to get stability fixes ("new kernel = new bugs", they think) and prefer that we deliver such bug fixes via live patches instead.

This seems backwards to me, though I guess it's a testament to how reliable live patching can be (or at least perceived to be).  The fixes delivered via live patching could instead be delivered as part of a stable branched kernel, which would have less risk than live patching.

> Another scenario is when the customers' nodes are rebooted for some unrelated reason (system hang or crash). Loading live patches at boot helps a lot to keep the system operational in such cases.

Yes, but then the patch also has to work in the case when it's loaded after boot, which doesn't seem to be the case here?

euspectre:
> Yes, but then the patch also has to work in the case when it's loaded after boot, which doesn't seem to be the case here?

That particular patch for VZ kernel worked both at boot and after boot. It just hid support for NETLINK_EXT_ACK command from any userspace app, except "criu". This hack worked well as a temporary measure and allowed the customers' systems last till their scheduled maintenance periods.

PR #1079, opened 2020-02-28
create-diff-object: Handle ppc64le toc with only constants

### LINKED ISSUE #1078: ppc64le kpatch-build crashes when .rela.toc is missing
The following kernel patch causes create-diff-object to segfault on ppc64le

```
diff --git a/net/ceph/pagelist.c b/net/ceph/pagelist.c
index c7c220a736e..c748e45be48 100644
--- a/net/ceph/pagelist.c
+++ b/net/ceph/pagelist.c
@@ -5,6 +5,8 @@
 #include <linux/highmem.h>
 #include <linux/ceph/pagelist.h>
 
+/* Dummy change */
+
 static void ceph_pagelist_unmap_tail(struct ceph_pagelist *pl)
 {
 	if (pl->mapped_tail) {
```

The reason is that the .toc section of `net/ceph/pagelist.o` only contains constants and in such cases the compiler does not create a corresponding .rela.toc section. The current function toc_rela() expect this section to alway be present and tries to access it.

julien-thierry:
When a ppcle64 ".toc" section contains only constants, the compiler
might not (won't?) create a corresponding ".rela.toc" section.

In such cases, create-diff-object crashes, assuming ".rela.toc" exists
whenever .toc exists. Simply report that no rela are available when
looking up possible relocations in .toc.

Fixes #1078.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -177,6 +177,10 @@ static struct rela *toc_rela(const struct rela *rela)
 	    rela->type != R_PPC64_TOC16_LO_DS)
 		return (struct rela *)rela;
 
+	/* Only constants in toc */
+	if (!rela->sym->sec->rela)
+		return NULL;
+
 	/* Will return NULL for .toc constant entries */
 	return find_rela_by_offset(rela->sym->sec->rela,
 				   (unsigned int)rela->addend);


kpatch-build/create-diff-object.c
```diff
@@ -177,6 +177,10 @@ static struct rela *toc_rela(const struct rela *rela)
 	    rela->type != R_PPC64_TOC16_LO_DS)
 		return (struct rela *)rela;
 
+	/* Only constants in toc */
```
kamalesh-babulal: Should we re-phrase the comment, saying older GCC has this issue?

julien-thierry: Are we 100% sure that having ".toc" section without a ".rela.toc" section never happens for new versions of GCC?

If so I can update the comment. Otherwise, I'd prefer keeping it as is.

kamalesh-babulal: Let me try other GCC before conforming about it. We have always documented such powerpc quirks with comments, here I would suggest something like:
```
index c9d2cc6..ded5fc1 100644
@@ -177,6 +177,22 @@ static struct rela *toc_rela(const struct rela *rela)
            rela->type != R_PPC64_TOC16_LO_DS)
                return (struct rela *)rela;
 
+       /*
+        * it observed with certain objects, the .toc section's rela pair
+        * .rela.toc is not generated, when  the .toc section entries are
+        * only constants:
+        * objddump -dr net/ceph/pagelist.o:
+        * ...
+        * 0000000000000000 <.toc>:
+        *    0:   e0 ff ff ff     .long 0xffffffe0
+        *    4:   ff ff ff 0f     twui    r31,-1
+        *
+        * readelf -rW net/ceph/pagelist.o:
+        * ...
+        * [20] .data             PROGBITS        0000000000000000 000aa8 000000 00  WA  0   0  1
+        * [21] .toc              PROGBITS        0000000000000000 000aa8 000008 00  WA  0   0  8
+        * [22] ___kcrctab+ceph_pagelist_append PROGBITS        0000000000000000 000ab0 000008 00  WA  0   0  8
+        */
```

kamalesh-babulal:
It's not specific to any GCC version, the earlier observation was wrong. It looks like the copy and replace mess up with the .o file version between RHEL 7 and upstream kernel. It's seen with the  upstream kernel < `v4.10` and  `GCC 9.2` too:
```
Disassembly of section .toc:

0000000000000000 <.toc>:
   0:   0f ff ff ff     twui    r31,-1
   4:   ff ff ff e0     .long 0xffffffe0
```   

With the commit `d6a9996e84a` ("powerpc/mm: vmalloc abstraction in preparation for radix")
a relocation entry gets generated in `.rela.toc` in place of the `.toc` constants
  
```
Disassembly of section .toc:

0000000000000000 <.toc>:
        ...
                        0: R_PPC64_ADDR64       vmemmap

Relocation section '.rela.toc' at offset 0x1d000 contains 1 entry:
0000000000000000  0000002300000026 R_PPC64_ADDR64         0000000000000000 vmemmap + 0
```

PR #1081, opened 2020-03-10
testing/integration: update rebase_patches tool

joe-lawrence:
Update the rebase_patches hack^H^H^H tool to make it easier to define
environment variables to drive it, rather than sourcing source
/etc/os-release for everything.  Fix a bunch of shellcheck warnings
along the way.

### DIFF
--- a/test/integration/rebase-patches
+++ b/test/integration/rebase-patches
@@ -1,50 +1,56 @@
 #!/bin/bash
 #
-# rebase a set of patches, assumes the kernel has already been downloaded into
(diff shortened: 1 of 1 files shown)


joe-lawrence:
(rhel-7.8 rebase testing currently underway)  I dunno if anyone else actually uses this script, but I hacked it a bit to make it a few degrees more flexible to use against a given source rpm.  I didn't think it was worth the time and effort to polish it up with a getopt interface, so I just fed it environment vars.

kamalesh-babulal:
Minor nitpick in the commit message, ```hack^H^H^H tool``` has non-printable characters.

PR #1083, opened 2020-03-11
test/integration: add rhel-7.8 rebased patches

joe-lawrence:
Rebased against the RHEL-7.8 GA kernel-3.10.0-1127.el7.


joe-lawrence:
The same patches needed to be disabled as in RHEL-7.7, and in doing so all of the enabled patches plus the combined patch build and run/test okay.

joe-lawrence:
@sm00th : does internal CI testing run these when we commit to master?  if so, what is required to add rhel-7.8 to that test matrix?

sm00th:
> @sm00th : does internal CI testing run these when we commit to master? if so, what is required to add rhel-7.8 to that test matrix?

It will only run against preset distros. I'll run 7.8 manually for this PR and we will enable 7.8 permanently after this is merged.

sm00th:
> will enable 7.8 permanently after this is merged.

Scratch that, we'll need to wait till it is properly released.

joe-lawrence:
Okay, cool.  Looks like I'm a few weeks ahead, I had figured it was already released by now.  We can leave this PR open until the GA date if that makes it easier to verify.

yhcote:
:+1:

PR #1085, opened 2020-03-17
test/integration: add rhel-8.3 rebased patches

yhcote:
Rebased against the RHEL-8.3 DEVEL kernel.

Start working on an updated version of the integration tests for the upcoming RHEL8.3 kernel.

This will get updated as necessary until it's GA'ed.

### DIFF
--- a/test/integration/rhel-8.3/gcc-static-local-var-6.patch
+++ b/test/integration/rhel-8.3/gcc-static-local-var-6.patch
@@ -1,0 +1,22 @@
+diff -Nupr src.orig/net/ipv6/netfilter.c src/net/ipv6/netfilter.c
+--- src.orig/net/ipv6/netfilter.c	2020-10-21 11:19:40.261740227 -0400
++++ src/net/ipv6/netfilter.c	2020-10-21 11:21:09.055020262 -0400
+@@ -86,6 +86,8 @@ static int nf_ip6_reroute(struct sk_buff
+ 	return 0;
+ }
+ 
++#include "kpatch-macros.h"
++
+ int __nf_ip6_route(struct net *net, struct dst_entry **dst,
+ 		   struct flowi *fl, bool strict)
+ {
+@@ -99,6 +101,9 @@ int __nf_ip6_route(struct net *net, stru
+ 	struct dst_entry *result;
+ 	int err;
+ 
++	if (!jiffies)
++		printk("kpatch nf_ip6_route foo\n");
++
+ 	result = ip6_route_output(net, sk, &fl->u.ip6);
+ 	err = result->error;
+ 	if (err)
--- a/test/integration/rhel-8.3/module.patch
+++ b/test/integration/rhel-8.3/module.patch
@@ -1,0 +1,85 @@
+From 08078d00ab1749a6f84148a00d8d26572af4ec97 Mon Sep 17 00:00:00 2001
+Message-Id: <08078d00ab1749a6f84148a00d8d26572af4ec97.1586900628.git.jpoimboe@redhat.com>
+From: Josh Poimboeuf <jpoimboe@redhat.com>
+Date: Tue, 14 Apr 2020 15:17:51 -0500
+Subject: [PATCH] kpatch module integration test
+
+This tests several things related to the patching of modules:
+
+- 'kpatch_string' tests the referencing of a symbol which is outside the
+  .o, but inside the patch module.
+
+- alternatives patching (.altinstructions)
+
+- paravirt patching (.parainstructions)
+
+- jump labels (5.8+ kernels only) -- including dynamic printk
+
+Signed-off-by: Josh Poimboeuf <jpoimboe@redhat.com>
+---
+ fs/nfsd/export.c         | 30 ++++++++++++++++++++++++++++++
+ net/netlink/af_netlink.c |  5 +++++
+ 2 files changed, 35 insertions(+)
+
+diff -Nupr src.orig/fs/nfsd/export.c src/fs/nfsd/export.c
+--- src.orig/fs/nfsd/export.c	2020-05-12 11:14:29.230792719 -0400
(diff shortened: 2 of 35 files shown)


sm00th:
Oh, 8.3 already. We are also missing 8.1/8.2 which would be nice to have, especially 8.1.

kamalesh-babulal:
> Oh, 8.3 already. We are also missing 8.1/8.2 which would be nice to have, especially 8.1.

Coincidently I have pushed the test cases into my repo yesterday for rhel 8.1 kernel 4.18.0-147.el8 ( https://github.com/kamalesh-babulal/kpatch/tree/rhel8-1-testcases-rebase).  I will raise a PR for it.

yhcote:
> > Oh, 8.3 already. We are also missing 8.1/8.2 which would be nice to have, especially 8.1.
> 
> Coincidently I have pushed the test cases into my repo yesterday for rhel 8.1 kernel 4.18.0-147.el8 ( https://github.com/kamalesh-babulal/kpatch/tree/rhel8-1-testcases-rebase). I will raise a PR for it.

Sounds good Kamalesh.

Yes I created the WIP RHEL8.3 update out of necessity to cover more ground for other internal testing.

joe-lawrence:
RHEL-8.3 is GA with 4.18.0-240, so this PR could be rebased and retested.

yhcote:
> RHEL-8.3 is GA with 4.18.0-240, so this PR could be rebased and retested.

True, I'll give it a shot :)

test/integration/rhel-8.3/README
```diff
@@ -0,0 +1 @@
+4.18.0-187.2.el8
```
julien-thierry: This file is created here and deleted in a later commit. I guess it should not have been created at all?

julien-thierry (commented):
I don't understand the 4th commit. Why not start the rebase from the rhel8.2 status rather than rebasing and then adding rhel8.2 updates?

Also, nit: contrary to what the first commit implies, RHEL-8.3 has kernel-4.18.0-240.el8

test/integration/rhel-8.3/README
yhcote: I guess this sequence looks a mess at this point if you review that PR commit by commit. The first 3 commits have been done months ago, while 8.3 was in full development. That README as part of the original WIP PR (deleted by the recent update) was meant to remind me on which build I started that port, which was not 240 at that time. The first 3 commits were rebased to current code and the last 2 commits were added just recently. The reason why it's not technically rebased from RHEL-8.2 is because "most" of the work was already done back then and before we created to port for 8.2, but I wanted to refresh that work with changes we've made to the integration tests since then (it ends up being the same in the end I suppose).

I will squash those commits since the history of what was done in this case, especially way back then doesn't add much value I think.

Hope this makes sense.

test/integration/rhel-8.3/gcc-static-local-var-6.patch
```diff
@@ -0,0 +1,20 @@
...
++#include "kpatch-macros.h"
++
+ static int nf_ip6_reroute(struct sk_buff *skb,
+ 			  const struct nf_queue_entry *entry)
+ {
+ 	struct ip6_rt_info *rt_info = nf_queue_entry_reroute(entry);
+ 
++	if (!jiffies)
++		printk("kpatch nf_ip6_reroute foo\n");
++
```
jpoimboe, line 17: I think this printk() should be in __nf_ip6_route() instead of nf_ip6_reroute().  8.3's __nf_ip6_route() is a renamed version of 8.2's nf_ip6_route().  (Also based on the name of the test it probably tests a change to a function which has a static local variable, which __nf_ip6_route() does).

yhcote: This is updated in the squashed version I am pushing. I realized that I made that change too quick.

Thanks for confirming.

yhcote:
This should make that PR less ambiguous and fix issues pointed out.

Ran slow-integration on amd64 and ppc64le successfully.

jpoimboe (approved):

test/integration/rhel-8.3/module.patch
```diff
@@ -0,0 +1,85 @@
...
++	alternative("call yield", "ud2", X86_FEATURE_IA64);
++
++	sched_clock = paravirt_sched_clock();
++	if (!jiffies)
++		printk("kpatch: sched_clock: %llu\n", sched_clock);
++#endif
++
++	pr_debug("kpatch: pr_debug() test\n");
++
++#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 8, 0)
```
joe-lawrence, line 52: I happened to notice this today and it looks weird to check for `KERNEL_VERSION` from the rhel-8.{1,2,3} test directories, should we be using `RHEL_RELEASE_VERSION` here?

test/integration/rhel-8.3/module.patch
```diff
@@ -0,0 +1,85 @@
...
++{
++	static DEFINE_STATIC_KEY_TRUE(kpatch_key);
++
++	if (static_branch_unlikely(&mcsafe_key))
```
joe-lawrence, line 56: also, sanity check me:  in a future 8.4 integration test rebase,
- since we've backported the klp.arch relocation removal code, this static_branch_unlikely test should be removed (when we have rhel-8.4 dir of tests)
- but the .altinstructions and .parainstructions tests are still valid

test/integration/rhel-8.3/module.patch
joe-lawrence: Actually, n/m I didn't see `mcsafe_key` defined here so I assumed that nfsd defined it.  (Looks like belongs to vmlinux, so all should be well.)

test/integration/rhel-8.3/module.patch
joe-lawrence: (n/m 2021 brain not fully engaged yet)

PR #1087, opened 2020-03-30
test/integration: rhel-8.1 Add missing *.test files

kamalesh-babulal:
Add the missing *.test files in test/integration/rhel-8.1

### DIFF
--- a/test/integration/rhel-8.1/cmdline-string-LOADED.test
+++ b/test/integration/rhel-8.1/cmdline-string-LOADED.test
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+grep kpatch=1 /proc/cmdline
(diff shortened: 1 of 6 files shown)


kamalesh-babulal:
> One thing: gcc-static-local-var-4.test needs to be disabled because it's parent module is disabled. Renaming it to gcc-static-local-var-4.test.disabled should work I think.

@sm00th Thanks for the review. Pushed the updated patch.

**v2 Changes:**
- Disabled the `gcc-static-local-var-4.test`

yhcote:
@sm00th, @kamalesh-babulal: Has it run through the CI or a local run for verification?

kamalesh-babulal:
@yhcote previous PR to add the test cases went through the CI runs by Artem and some of the tests were locally build by me, not this PR. @sm00th Thanks again for the review and the runs.

yhcote:
> @yhcote previous PR to add the test cases went through the CI runs by Artem and some of the tests were locally build by me, not this PR. @sm00th Thanks again for the review and the runs.

Thanks! @sm00th, @kamalesh-babulal

PR #1091, opened 2020-04-22
create-diff-object: Ignore kpatch_ignored functions/sections missing …

julien-thierry:
…ftrace hook

Some theoretically unchanged functions can have undesired changes if the
compiler decides to perform inlining in a different way (e.g. because of
newly added references). In such a case, it can be useful to discard
changes to functions that don't actually need modification.

Sadly, this currently doesn't work for functions missing the ftrace hook
(e.g. notrace code) as presence of the hook is checked before
identifying elements to ignore.

Look for functions/sections to ignore earlier.


julien-thierry:
I can get a simple reproducer for unit test, just need some time to get it ready.

PR #1092, opened 2020-04-23
tests/integration/module.patch: fix ppc64le build

sm00th:
ppc64le build currently fails dues to unused sched_clock variable. Move
it's declaration into ifdef block.


sm00th:
For those wondering why this wasn't reported by ci - that's because 8.1 has postfix disabled by default so it failed to send those errors out. This should be fixed now.

Issue #1093, opened 2020-04-30
kmod/relocations: Missing + r_addend


stettberger:
I think in the calculation of the relocation, a + addend is missing to implement R_X86_64_64 correctly.

https://github.com/dynup/kpatch/blob/8e8707ed723adf8226e99563b88c5490ca51a7e0/kmod/core/core.c#L745

joe-lawrence:
Yeah, the [AMD64 ABI draft doc]( http://refspecs.linuxfoundation.org/elf/x86_64-abi-0.99.pdf ), Table 4.10: Relocation Types table lists that relocation type as: `R_X86_64_64 ... S + A`, so it looks like we're missing the addend for this type.  Do you happen to have an input kpatch that generates one of these relocations?

stettberger:
Unfortunately not. I just read the source code and stumpled upon this issue.

jpoimboe:
Nicely spotted 6-year-old bug :-)  This code is deprecated but it wouldn't hurt to fix it (just a one-line change).

euspectre:
Thanks for the fix!

Looks like such dynrelas are extremely rare but, in theory, are possible. I haven't seen these in the wild yet but they can appear if the patch module uses, say, a non-exported global variable from vmlinux with an appropriate addend.

I came up with the following (quite contrived) source patch to reproduce the issue:

```
diff --git a/fs/proc/meminfo.c b/fs/proc/meminfo.c
index 32512c4fa9c0..2f2f320950fe 100644
--- a/fs/proc/meminfo.c
+++ b/fs/proc/meminfo.c
@@ -17,6 +17,7 @@
 #include <asm/page.h>
 #include <asm/pgtable.h>
 #include <bc/beancounter.h>
+#include <asm/desc.h>
 #include "internal.h"
 
 void __attribute__((weak)) arch_report_meminfo(struct seq_file *m)
@@ -283,6 +284,7 @@ static int meminfo_proc_show(struct seq_file *m, void *v)
 
 static int meminfo_proc_open(struct inode *inode, struct file *file)
 {
+	pr_info("[DBG] idt_table+offset=%lu\n", (unsigned long)&idt_table+0x80000000);
 	return single_open(file, meminfo_proc_show, NULL);
 }
 
```
Checked the patch with a VZ kernel, which is based on 3.10.0-1062.12.1.el7 from RHEL.

[...]

PR #1094, opened 2020-05-01
kmod/core: apply dynrela addend for R_X86_64_64

joe-lawrence:
User stettberger noticed that the kpatch support module does not
apply the addend for R_X86_64_64 in kpatch_write_relocations().

The AMD64 ABI draft doc [1], Table 4.10: Relocation Types lists that
relocation type as:

  Name         Value  Field   Calculation
  R_X86_64_64  1      word64  S + A

where:

  S : Represents the value of the symbol whose index resides in the
      relocation entry.

  A : Represents the addend used to compute the value of the relocatable
      field.

[1] http://refspecs.linuxfoundation.org/elf/x86_64-abi-0.99.pdf

Fixes: #1093
Reported-by: Christian Dietrich <stettberger@dokucode.de>

### DIFF
--- a/kmod/core/core.c
+++ b/kmod/core/core.c
@@ -742,7 +742,7 @@ static int kpatch_write_relocations(struct kpatch_module *kpmod,
 			break;
 		case R_X86_64_64:
 			loc = dynrela->dest;
-			val = dynrela->src;
+			val = dynrela->src + dynrela->addend;
 			size = 8;
 			break;
 		default:


joe-lawrence:
@euspectre : Hi Evgenii, just a heads up for this bug and fix.  We don't build the legacy kpatch.ko support module any longer, but you may have a larger set of modules that this bug affects.

euspectre:
@joe-lawrence 
Hi Joe, thanks for the heads up!

Although I haven't seen R_X86_64_64 dynrelas in our production patches yet, they are possible. I have provided an example source patch in #1093.

PR #1096, opened 2020-05-05
integration/centos-7: update tests to current kernel

sm00th:
Since the kernel we were testing against before is no longer available
in centos repos I've updated our test vagrant image bringing the kernel
version to 3.10.0-1127.el7. Since this is basically the rhel-7.8 kernel
and we don't want extra work maintaining extra sets of patches just
symlink centos-7 patch directory to rhel-7.8. If(when) we are in this
situation again we'll just switch the symlink to the appropriate minor
again.


jpoimboe:
Just wondering if we still think it's worth it to keep maintaining the centos image since the old kernels can disappear, and since the code is identical to rhel anyway?

sm00th:
> Just wondering if we still think it's worth it to keep maintaining the centos image since the old kernels can disappear, and since the code is identical to rhel anyway?

I think yes, so that theoretical contributors that don't have access to rhel still can run integration tests against a rhel-like kernel?

It is also not centos-specific, we hit the same issue with fedora before.

jpoimboe:
True, but I question whether these theoretical contributors running centos integration tests exist.  Also RHEL is now publicly available for download at developers.redhat.com.

For Fedora, I wonder if we should instead do a rolling test, i.e. always test the latest kernel.  It's more likely to break, but would be useful for finding breaking changes upstream.

sm00th:
> True, but I question whether these theoretical contributors running centos integration tests exist.

I don't disagree but this is also true for fedora.

> Also RHEL is now publicly available for download at developers.redhat.com.
> 
> For Fedora, I wonder if we should instead do a rolling test, i.e. always test the latest kernel. It's more likely to break, but would be useful for finding breaking changes upstream.

If we are talking fedora kernels - we tried that and after couple of times of hitting a situation when gcc and kernel version is out of reach in the repo switched to the current method.

I'd love to do this on upstream kernels but that is certainly way more maintenance and I don't think we'll be able to run those on every commit/pull request.

sm00th:
Given there were no strong opinions during our conversation on this topic and that all the work is already done at this point - I propose we merge this one and revisit the topic later.

PR #1098, opened 2020-05-08
Added Table of contents

TheEvilSkeleton:
The README was really big at the point that it was hard to navigate, so I decided a Table of contents. Enjoy

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,10 +60,7 @@ Install the dependencies for compiling kpatch:
 
 ```bash
 UNAME=$(uname -r)
- # Standard kernel
 sudo dnf install gcc kernel-${UNAME%.*} elfutils elfutils-devel
- # Developer kernel
-sudo dnf install gcc kernel-devel-${UNAME%.*} elfutils elfutils-devel
 ```
 
 Install the dependencies for the "kpatch-build" command:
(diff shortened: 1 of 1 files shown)


TheEvilSkeleton:
> Assuming the links are correct, I can't get them to work in the review interface :-)
> 
> Thanks!

I tried the links before sending the merge request, and they all worked.

joe-lawrence:
> Assuming the links are correct, I can't get them to work in the review interface :-)
> 
> Thanks!

For markdown pages, I always go to the code browser to view them.  In this case,
https://github.com/TheEvilSkeleton/kpatch should work.

TheEvilSkeleton:
@joe-lawrence @yhcote @jpoimboe I also added this as a fix for Fedora users: https://github.com/dynup/kpatch/pull/1098/commits/2eeed7bde10601b20af401713e1dfb9abc0b8e00

jpoimboe:
@TheEvilSkeleton kernel-devel isn't the developer kernel, it's the kernel headers and makefiles needed to build kpatch.ko.  Since kpatch.ko is now deprecated, I think we can just remove kernel-devel from the instructions altogether.

jpoimboe:
Looks good to me, thanks.

kamalesh-babulal:
> kernel-devel isn't the developer kernel, it's the kernel headers and makefiles needed to build kpatch.ko. 

Agree, the term developer kernel doesn't represent the kernel-devel correctly.

PR #1100, opened 2020-05-09
kpatch: Fix kernel live patch and kpatch core module co-existing

xiejingf:
for kernels which support klp, the old kpatch module way is also allowed to
use, we try to support this

### DIFF
--- a/test/integration/linux-5.10.11/module-call-external.patch
+++ b/test/integration/linux-5.10.11/module-call-external.patch
@@ -1,16 +1,18 @@
-diff -Nupr src.orig/fs/nfsd/export.c src/fs/nfsd/export.c
---- src.orig/fs/nfsd/export.c	2021-01-28 04:47:10.905473196 -0500
-+++ src/fs/nfsd/export.c	2021-01-28 04:55:55.117394790 -0500
-@@ -1234,6 +1234,8 @@ static void exp_flags(struct seq_file *m
+diff --git a/fs/nfsd/export.c b/fs/nfsd/export.c
+index 21e404e7cb68..3a282338d6d9 100644
+--- a/fs/nfsd/export.c
++++ b/fs/nfsd/export.c
+@@ -1234,6 +1234,9 @@ static void exp_flags(struct seq_file *m, int flag, int fsid,
  	}
  }
  
 +extern char *kpatch_string(void);
 +
++__attribute__((optimize("-fno-optimize-sibling-calls")))
  static int e_show(struct seq_file *m, void *p)
  {
  	struct cache_head *cp = p;
-@@ -1243,6 +1245,7 @@ static int e_show(struct seq_file *m, vo
+@@ -1243,6 +1246,7 @@ static int e_show(struct seq_file *m, void *p)
  	if (p == SEQ_START_TOKEN) {
  		seq_puts(m, "# Version 1.1\n");
  		seq_puts(m, "# Path Client(Flags) # IPs\n");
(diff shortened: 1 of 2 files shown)


xiejingf:
when klp and kpatch are both used on the machine,  the kpatch script will only match only one, either klp or kpatch , klp is prioritized over kpatch,   my patch tries to  fix this issue.

xiejingf:
> when klp and kpatch are both used on the machine, the kpatch script will only match only one, either klp or kpatch , klp is prioritized over kpatch, my patch tries to fix this issue.

live patch enabled，end user  choose kpatch.ko based hotfix， kpatch script  will also fail to deal with such case

jpoimboe:
Instead can we just check core_loaded() in init_sysfs_var()?

Also please try to follow the existing coding style by using tabs instead of spaces.

joe-lawrence:
@euspectre - if this looks interesting to support for kpatch.ko, do you mind reviewing?  Thanks.

euspectre:
We had a similar problem with livepatches for Virtuozzo some time ago but used a bit different approach: https://github.com/euspectre/kpatch/commit/5b54fe4b

@joe-lawrence Yes, I'll take a look.

euspectre:
> Instead can we just check core_loaded() in init_sysfs_var()?

I don't think so. If I am not mistaken, core_loaded() returns true if Livepatch is supported as well:
```
core_loaded () {
	grep -q -e "T klp_enable_patch" -e "T kpatch_register" /proc/kallsyms
}
```

euspectre:
@xiejingf : Do you really need to use both the old-style patches (those that rely on kpatch.ko) and klp-based patches for the same kernel at the same time?

So far, I have only encountered situations where either kpatch.ko-based patches or klp-based patches were used for each given kernel, but not both kinds of patches simultaneously.

If it is not required to allow using both kinds of patches at the same time, support for that is easier to add.

euspectre:
@xiejingf : in_transition(), is_stalled(), show_stalled_processes() and related functions only make sense for klp-based patches. Traversing all sysfs directories there could be a waste.

xiejingf:
> @xiejingf : Do you really need to use both the old-style patches (those that rely on kpatch.ko) and klp-based patches for the same kernel at the same time?
> 
> So far, I have only encountered situations where either kpatch.ko-based patches or klp-based patches were used for each given kernel, but not both kinds of patches simultaneously.
> 
> If it is not required to allow using both kinds of patches at the same time, support for that is easier to add.

Yes，because of the different feature provided by klp and kpatch.ko and some other reasons,  we might use kpatch.ko-based  and klp-based patches for the same kernel at the same time, so I think it would be better to support this use case

On the other hand, I will rework in_transition(), is_stalled(), show_stalled_processes().

euspectre:
> Yes，because of the different feature provided by klp and kpatch.ko and some other reasons, we might use kpatch.ko-based and klp-based patches for the same kernel at the same time

Strange. kpatch.ko-based patches are deprecated here, in upstream kpatch project. They are still used a lot, esp. for older kernels, but, I suppose, their usage should slowly decrease over time.

Could you explain a bit more why it is needed to use both KLP and kpatch.ko-based patches on the same system? Do you use these this way?

What are these "different features" that do not allow picking one livepatching technology for each given system?

By the way, using 2 potentially conflicting live patching technologies on the same system is often not the best idea: for example, you'll have to ensure the sets of patched functions in KLP and kpatch.ko-based patches do not intersect to avoid undefined behaviour.

If you use cumulative patches, which is often quite convenient and easier to maintain, you can just pick KLP or kpatch.ko-based ones for each given system, and that should do it. No obvious need for co-existence of the patches.

Please elaborate.

PR #1102, opened 2020-05-12
create-diff-object: Add ppc64le mcount support

joe-lawrence:
Add the __mcount_loc section on ppc64le.  It has pointers to all the
mcount calls.  This will enable the ftrace hook to be used for patched
functions.

Signed-off-by: Joe Lawrence <joe.lawrence@redhat.com> (rebased)


joe-lawrence:
Note: I'm opening a PR for this only to see what automated integration tests think of Josh's patch.  He had closed #1001 sometime last year and implementing this would help us avoid a current ppc64le issue that we're hitting.

**DO NOT MERGE**  thanks.

yhcote:
> Note: I'm opening a PR for this only to see what automated integration tests think of Josh's patch. He had closed #1001 sometime last year and implementing this would help us avoid a current ppc64le issue that we're hitting.
> 
> **DO NOT MERGE** thanks.

Good stuff, as a convention, I'll edit the PR title with [wip], that's usually a good way of tracking PRs that are not ready to be merged, even with approvals.

joe-lawrence:
Hits this problem on a sample integration test module load:
```
  test_meminfo_string: loading out-of-tree module taints kernel.
  test_meminfo_string: tainting kernel with TAINT_LIVEPATCH
  test_meminfo_string: module verification failed: signature and/or required key missing - tainting kernel
  ftrace-powerpc: Not expected bl: opcode is fa61ff98
  WARNING: CPU: 39 PID: 6165 at kernel/trace/ftrace.c:2060 ftrace_bug+0x88/0x390
  Modules linked in: i2c_dev at24 ofpart powernv_flash mtd uio_pdrv_genirq uio ibmpowernv xts opal_prd ipmi_powernv ipmi_devintf ipmi_msghandler vmx_crypto ip_tables xfs libcrc32c sd_mod sg ast i2c_algo_bit drm_vram_helper ttm drm_kms_helper syscopyarea sysfillrect sysimgblt fb_sys_fops drm ahci libahci libata tg3 drm_panel_orientation_quirks dm_mirror dm_region_hash dm_log dm_mod
  CPU: 39 PID: 6165 Comm: insmod Kdump: loaded Tainted: G           OE K  --------- -  - 4.18.0-193.el8.ppc64le #1
  NIP:  c0000000002db50c LR: c0000000002d1e00 CTR: 0000000000000000
  REGS: c000003f875c7750 TRAP: 0700   Tainted: G           OE K  --------- -  -  (4.18.0-193.el8.ppc64le)
  MSR:  900000000282b033 <SF,HV,VEC,VSX,EE,FP,ME,IR,DR,RI,LE>  CR: 28008242  XER: 00000000
  CFAR: c0000000002db4f0 IRQMASK: 0
  GPR00: c0000000002d1e00 c000003f875c79d0 c000000001920a00 ffffffffffffffea
  GPR04: c000003f9f090060 0000000000000000 00000000000003fb c000003ffffd3200
  GPR08: 0000000000000007 c00000000007c8b0 0000000000000005 9000000002803033
  GPR12: 0000000000008000 c000003ffffd3200 c00800000e760080 0000000000000000
  GPR16: 000000000000ff20 000000000000fff1 000000000000fff2 c000003fcf70d060
  GPR20: 0000000000000005 c00800000dc72880 c000003fcf70d060 0000000000000001
  GPR24: 0000000000000000 c0000000019527a8 0000000002000000 c00000000007c6b8
  GPR28: 0000000000000005 c000000001033d60 c000003f9f090060 c00800000dc7080c
  NIP [c0000000002db50c] ftrace_bug+0x88/0x390
  LR [c0000000002d1e00] ftrace_process_locs+0x540/0x5d0
  Call Trace:
  [c000003f875c79d0] [c000003f875c7a00] 0xc000003f875c7a00 (unreliable)
  [c000003f875c7a60] [c0000000002d1e00] ftrace_process_locs+0x540/0x5d0
  [c000003f875c7b30] [c00000000025b584] load_module+0x36e4/0x3d40
  [c000003f875c7d20] [c00000000025bef0] sys_finit_module+0xd0/0x150
  [c000003f875c7e30] [c00000000000b388] system_call+0x5c/0x70
  Instruction dump:
  419e0018 2f83ffff 419e00c0 2f83ffea 409e00d0 4800001c 0fe00000 3c62ff71
  39000001 39400000 38633340 480000c8 <0fe00000> 3ce20003 39000001 3c62ff71
  ---[ end trace 0e2b3bd401674254 ]---
  ftrace failed to modify
  [<c00800000dc7080c>] 0xc00800000dc7080c
   actual:   98:ff:61:fa                                << what we found at reloc address
  Initializing ftrace call sites
  ftrace record flags: 2000000
   (0)
   expected tramp: c00000000007c8cc
  livepatch: enabling patch 'test_meminfo_string'
  livepatch: failed to set ftrace filter for function 'meminfo_proc_show' (-19)
  livepatch: failed to patch object 'vmlinux'
  livepatch: failed to enable patch 'test_meminfo_string'
  livepatch: 'test_meminfo_string': unpatching complete
```

A look at the __mcount_loc relocations:
```
  % objdump -Dr test-meminfo-string.ko

  Disassembly of section __mcount_loc:

  0000000000000000 <__mcount_loc>:
          ...
                          0: R_PPC64_ADDR64       patch_find_object_by_name+0xc
                          8: R_PPC64_ADDR64       patch_find_object_by_name+0x15c
                          10: R_PPC64_ADDR64      patch_find_object_by_name+0x2ac
                          18: R_PPC64_ADDR64      init_module+0xc
                          20: R_PPC64_ADDR64      meminfo_proc_show+0x14    << off by 0x8?


  Disassembly of section .text.meminfo_proc_show:

  0000000000000000 <meminfo_proc_show-0x8>:                      << note the function symbol
                                                                    is 0x8 bytes ahead of the
                                                                    section start
          ...
                          0: R_PPC64_REL64        .TOC.-0x8

  0000000000000008 <meminfo_proc_show>:
     8:   f8 ff 4c e8     ld      r2,-8(r12)
                          8: R_PPC64_ENTRY        *ABS*
     c:   14 62 42 7c     add     r2,r2,r12
    10:   a6 02 08 7c     mflr    r0
    14:   01 00 00 48     bl      14 <meminfo_proc_show+0xc>
                          14: R_PPC64_REL24       _mcount
    18:   a6 02 08 7c     mflr    r0
    1c:   98 ff 61 fa     std     r19,-104(r1)                   << @ meminfo_proc_show+0x14 [wrong]
          ^^^^^^^^^^^
          actual op-code
```

At least the ordinary module-build-generated _mcount relocations appear to be at their correct offsets:
```
  Disassembly of section .text.unlikely:

  0000000000000000 <patch_find_object_by_name-0x8>:
          ...
                          0: R_PPC64_REL64        .TOC.-0x8

  0000000000000008 <patch_find_object_by_name>:
     8:   f8 ff 4c e8     ld      r2,-8(r12)
                          8: R_PPC64_ENTRY        *ABS*
     c:   14 62 42 7c     add     r2,r2,r12
    10:   a6 02 08 7c     mflr    r0
    14:   01 00 00 48     bl      14 <patch_find_object_by_name+0xc>
                          14: R_PPC64_REL24       _mcount             << @ patch_find_object_by_name+0xc [ok]


  0000000000000158 <patch_free_scaffold>:
   158:   f8 ff 4c e8     ld      r2,-8(r12)
                          158: R_PPC64_ENTRY      *ABS*
   15c:   14 62 42 7c     add     r2,r2,r12
   160:   a6 02 08 7c     mflr    r0
   164:   01 00 00 48     bl      164 <patch_free_scaffold+0xc>
                          164: R_PPC64_REL24      _mcount             << @ patch_find_object_by_name+0x15c [ok]

  00000000000002a8 <patch_free_livepatch>:
   2a8:   f8 ff 4c e8     ld      r2,-8(r12)
                          2a8: R_PPC64_ENTRY      *ABS*
   2ac:   14 62 42 7c     add     r2,r2,r12
   2b0:   a6 02 08 7c     mflr    r0
   2b4:   01 00 00 48     bl      2b4 <patch_free_livepatch+0xc>
                          2b4: R_PPC64_REL24      _mcount             << @ patch_find_object_by_name+0x2ac [ok]
```

joe-lawrence:
Though seemingly correct for leaf / non-global TOC setup functions:
```
(from another .ko file)

Disassembly of section __mcount_loc:

0000000000000000 <__mcount_loc>:
        ...
                        20: R_PPC64_ADDR64      netlbl_catmap_getlong+0x4

Disassembly of section .text.netlbl_catmap_getlong:

0000000000000000 <netlbl_catmap_getlong>:
   0:   a6 02 08 7c     mflr    r0
   4:   01 00 00 48     bl      4 <netlbl_catmap_getlong+0x4>
                        4: R_PPC64_REL24        _mcount
```

jpoimboe:
@sm00th any idea why integration test CI didn't catch that for this PR?

sm00th:
> @sm00th any idea why integration test CI didn't catch that for this PR?

It never ran for this PR, that is our jenkins instance never fired off the test job. Maybe @lulinqing will have a clue why.

lulinqing:
After a quick look, the job triggered by last PR seems still running after 24h and blocking any future polling:
https://platform-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/kpatch-integration-el7-pr/232/

Thought the job is configured to allow concurrent builds but doesn't seem to work as expected.
Possible solutions I can think of right now:
1- limit the period of time per PR test job (e.g. 1~2 hours?) and/or set time out
2- decouple polling step from testing job so they'll be asynchronized/non-blocking

sm00th:
> Thought the job is configured to allow concurrent builds but doesn't seem to work as expected.

Hm, I saw multiple jobs running concurrently before so its strange it got stuck on this one.

lulinqing:
The jobs for this and next (#1104 ) PRs took around 4~5 hours and finished.
No job queued at this moment but still wonder if this issue is worth mitigated for future benefit? or it's just a one-time bump? @sm00th

kamalesh-babulal:
> Though seemingly correct for leaf / non-global TOC setup functions:

That's right if the section does not access any global variable then the TOC setup is not required and in the case of the leaf functions the TOC of the caller will not clobber, so setup is not required.

Looking at the readelf -rw of `livepatch-meminfo.ko`
```
Relocation section '.rela__mcount_loc' at offset 0x2d90 contains 5 entries:
0000000000000000  0000002f00000026 R_PPC64_ADDR64         0000000000000008 patch_find_object_by_name + c
0000000000000008  0000002f00000026 R_PPC64_ADDR64         0000000000000008 patch_find_object_by_name + 15c
0000000000000010  0000002f00000026 R_PPC64_ADDR64         0000000000000008 patch_find_object_by_name + 2ac
0000000000000018  0000004100000026 R_PPC64_ADDR64         0000000000000008 init_module + c
0000000000000020  0000003100000026 R_PPC64_ADDR64         0000000000000008 meminfo_proc_show + 14
```

Every rela entry's addend is mcount location - section symbol value, except for
one added by kpatch_create_mcount_sections:

i.e for the rela->offset(0x20) = should have been 0x14 - 0x8 = meminfo_proc_show + c

One way to fix is:
```
@@ -3427,7 +3427,7 @@ static void kpatch_create_mcount_sections(struct kpatch_elf *kelf)
                ALLOC_LINK(mcount_rela, &relasec->relas);
                mcount_rela->sym =sym;
                mcount_rela->type = ABSOLUTE_RELA_TYPE;
-               mcount_rela->addend = (insn_offset);
+               mcount_rela->addend = (insn_offset - sym->sym.st_value);
                mcount_rela->offset = (unsigned int) (index * sizeof(*funcs));
 
                index++;
```

With the above patch, the addend is adjusted:
```
Relocation section '.rela__mcount_loc' at offset 0x2d90 contains 5 entries:
0000000000000000  0000002f00000026 R_PPC64_ADDR64         0000000000000008 patch_find_object_by_name + c
0000000000000008  0000002f00000026 R_PPC64_ADDR64         0000000000000008 patch_find_object_by_name + 15c
0000000000000010  0000002f00000026 R_PPC64_ADDR64         0000000000000008 patch_find_object_by_name + 2ac
0000000000000018  0000004100000026 R_PPC64_ADDR64         0000000000000008 init_module + c
0000000000000020  0000003100000026 R_PPC64_ADDR64         0000000000000008 meminfo_proc_show + c
```

joe-lawrence:
> One way to fix is:
> 
> ```
> --- a/kpatch-build/create-diff-object.c
> +++ b/kpatch-build/create-diff-object.c
> @@ -3427,7 +3427,7 @@ static void kpatch_create_mcount_sections(struct kpatch_elf *kelf)
>                 ALLOC_LINK(mcount_rela, &relasec->relas);
>                 mcount_rela->sym =sym;
>                 mcount_rela->type = ABSOLUTE_RELA_TYPE;
> -               mcount_rela->addend = (insn_offset);
> +               mcount_rela->addend = (insn_offset - sym->sym.st_value);
>                 mcount_rela->offset = (unsigned int) (index * sizeof(*funcs));
>  
>                 index++;
> ```

Hi @kamalesh-babulal ... 

@jpoimboe had suggested changing the reloc sym to match the section's secsym:

```
-               mcount_rela->sym = sym;
+               mcount_rela->sym = sym->sec->secsym;
```

which accounted for the wrong offsets (and passed all the integration tests, btw.)

I wonder though if leaving the symbol alone and correcting the addend like you suggested might be more consistent with ordinary .rela__mcount_loc relocations.  Otherwise, we'll see a bunch of .text.meminfo_proc_show + x  relocations instead of meminfo_proc_show + x relocations.

Any preference for either or?

jpoimboe:
I'd probably have a slight preference towards making the mcount_loc relas "look" like normal ones like in @kamalesh-babulal's suggestion, but it really shouldn't matter either way.

joe-lawrence:
Updated with @kamalesh-babulal's suggestion.  The update passed 8.2 integration tests on ppc64le and x86_64, so I think it's ready for merge review now.

kpatch-build/create-diff-object.c
```diff
@@ -3402,14 +3403,40 @@ static void kpatch_create_mcount_sections(struct kpatch_elf *kelf)
...
+#endif
+		/*
+		 * 'rela' points to the mcount/fentry call.
+		 *
+		 * Create a .rela__mcount_loc entry which also points to it.
+		 */
+		ALLOC_LINK(mcount_rela, &relasec->relas);
+		mcount_rela->sym = sym;
+		mcount_rela->type = ABSOLUTE_RELA_TYPE;
+		mcount_rela->addend = (insn_offset - sym->sym.st_value);
```
jpoimboe, line 3430: nit: unnecessary parentheses

joe-lawrence: removed

kamalesh-babulal:
> Any preference for either or?

The relocation entry output from both the approaches looks similar.  I did not have a strong preference :)

```
0000000000000020  0000003100000026 R_PPC64_ADDR64         0000000000000008 meminfo_proc_show + c   <= insn_offset - sym->sym.st_value
0000000000000020  0000000700000026 R_PPC64_ADDR64         0000000000000000 .text.meminfo_proc_show + 14 <= sym->sec->secsym
```

jpoimboe:
It would be nice to have a unit or integration test for the panic we originally saw which prompted this PR.

joe-lawrence:
> It would be nice to have a unit or integration test for the panic we originally saw which prompted this PR.

Yeah, I still have the object files to do that.  Is it worth adding them for both ppc64le and x86_64, or just the former?

jpoimboe:
While you're at it, might as well test the x86 side as well.

joe-lawrence:
One more force push update to include the new kpatch-unit-test-objs for this PR.

Issue #1099, opened 2020-05-08
FIX: "sudo dnf install gcc kernel-devel-${UNAME%.*} elfutils elfutils-devel" not working


TheEvilSkeleton:
So, I decided to use the `sudo dnf install gcc kernel-devel-${UNAME%.*} elfutils elfutils-devel` command from https://github.com/dynup/kpatch#fedora, and I got this error:

```bash
$ sudo dnf install gcc kernel-devel-${UNAME%.*} elfutils elfutils-devel

Last metadata expiration check: 1:58:42 ago on Fri 08 May 2020 09:35:31 AM EDT.
No match for argument: kernel-devel-5.6.8-300.fc32
Error: Unable to find a match: kernel-devel-5.6.8-300.fc32
```

The fix to this is to just run `sudo dnf install gcc kernel-${UNAME%.*} elfutils elfutils-devel` instead.

```bash
$ sudo dnf install gcc kernel-${UNAME%.*} elfutils elfutils-devel     

Last metadata expiration check: 2:00:13 ago on Fri 08 May 2020 09:35:31 AM EDT.
Package kernel-5.6.8-300.fc32.aarch64 is already installed.
Dependencies resolved.
==============================================================================================================================================================================================
 Package                                                       Architecture                         Version                                        Repository                            Size
==============================================================================================================================================================================================
Installing:
 elfutils                                                      aarch64                              0.179-2.fc32                                   updates                              528 k
 elfutils-devel                                                aarch64                              0.179-2.fc32                                   updates                               54 k
 gcc                                                           aarch64                              10.0.1-0.14.fc32                               updates                               28 M
Installing dependencies:
 binutils                                                      aarch64                              2.34-2.fc32                                    fedora                               5.8 M
 binutils-gold                                                 aarch64                              2.34-2.fc32                                    fedora                               1.0 M
 cpp                                                           aarch64                              10.0.1-0.14.fc32                               updates                              9.4 M
 elfutils-libelf-devel                                         aarch64                              0.179-2.fc32                                   updates                               27 k
 glibc-devel                                                   aarch64                              2.31-2.fc32                                    fedora                               1.0 M
 glibc-headers                                                 aarch64                              2.31-2.fc32                                    fedora                               449 k
 isl                                                           aarch64                              0.16.1-10.fc32                                 fedora                               831 k
 kernel-headers                                                aarch64                              5.6.10-300.fc32                                updates                              1.1 M
[...]

PR #1104, opened 2020-05-12
Added Supported Architectures category

TheEvilSkeleton:
- Added Supported Architectures category to avoid confusion (for example https://github.com/dynup/kpatch/issues/1099#issuecomment-625913876).
- Updated Prerequisites category

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ And a few more:
 Table of contents
 =================
 
+- [Supported Architectures](#supported-architectures)
 - [Installation](#installation)
 	- [Prerequisites](#prerequisites)
 		- [Fedora](#fedora)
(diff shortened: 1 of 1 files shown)


jpoimboe:
How about calling the section "Supported Architectures" and moving it before Installation (like the TOC shows)?

I would do the checklist something like:

- [x] x86-64
- [x] ppc64le
- [ ] arm64
- [ ] s390

Because I haven't heard of any plans to support other architectures beyond those four.

Also please squash the commits into a single commit with an appropriate subject and description.  

Thanks.

TheEvilSkeleton:
> How about calling the section "Supported Architectures" and moving it before Installation (like the TOC shows)?

Oops. User error

TheEvilSkeleton:
@jpoimboe Not really sure how I can squash commits.

I have found online that a member of the repository can squash commits https://github.blog/2016-04-01-squash-your-commits/

jpoimboe:
You can 'git rebase -i' to squash them locally and then 'git push --force' to your branch.

TheEvilSkeleton:
Seems good to me now

PR #1106, opened 2020-05-13
test/integration: add RHEL-8.2 integration tests

joe-lawrence:
For kernel-4.18.0-193.el8.

### DIFF
--- a/test/integration/rhel-8.2/bug-table-section.patch
+++ b/test/integration/rhel-8.2/bug-table-section.patch
@@ -1,0 +1,12 @@
+diff -Nupr src.orig/fs/proc/proc_sysctl.c src/fs/proc/proc_sysctl.c
+--- src.orig/fs/proc/proc_sysctl.c	2020-05-12 11:14:29.250791853 -0400
++++ src/fs/proc/proc_sysctl.c	2020-05-12 11:14:36.220489794 -0400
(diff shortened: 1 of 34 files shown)


jpoimboe:
Compared to the RHEL 8.1 tests, gcc-static-local-var-5.patch and shadow-newpid-LOADED.test are disabled.  It would be good to have the justification for those changes in the commit log.

joe-lawrence:
gcc-static-local-var-5.patch: I'll fetch that that info tomorrow
shadow-newpid-LOADED.test: in RHEL-8.1, it's shadow-newpid.patch.disabled so I disabled the the test to match

sm00th:
> shadow-newpid-LOADED.test: in RHEL-8.1, it's shadow-newpid.patch.disabled so I disabled the the test to match

Could you also please disable the test on 8.1 in a separate commit (but same pr)? It is not required (looking at the code it just detects that corresponding module is not there and just complains and skips the test) but won't hurt to be consistent.

joe-lawrence:
> Compared to the RHEL 8.1 tests, gcc-static-local-var-5.patch and shadow-newpid-LOADED.test are disabled. It would be good to have the justification for those changes in the commit log.

I enabled all the tests and re-ran them on the latest 8.2 z-stream build, disabled those that no longer built and noted them in the commit message.

PR #1109, opened 2020-06-02
Rhel8 deps

julien-thierry:
The `kpatch_depencies()` function does not work out of the box for RHEL-8.

These changes add the missing packages to install allowing to build & run kpatch-build on RHEL-8

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,9 +26,7 @@ Table of contents
 - [Supported Architectures](#supported-architectures)
 - [Installation](#installation)
 	- [Prerequisites](#prerequisites)
-		- [Fedora](#fedora)
-		- [RHEL 7](#rhel-7)
-		- [CentOS 7](#centos-7)
+		- [Fedora, RHEL, CentOS](#fedora-rhel-centos)
 		- [Oracle Linux 7](#oracle-linux-7)
 		- [Ubuntu](#ubuntu)
 		- [Debian 9 (Stretch)](#debian-9-stretch)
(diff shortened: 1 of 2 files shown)


test/integration/lib.sh
```diff
@@ -53,18 +53,39 @@ kpatch_rhel_dependencies()
 {
 	local kernel_version
 	local arch
+	local rhel_major
+	local py_version
 	kernel_version=$(uname -r)
 	arch=$(uname -m)
+	rhel_major=${VERSION_ID%%.*}
+
+	if [ $rhel_major -ge 8 ]; then
```
sm00th, line 62: We try to keep these free of [shellcheck](https://www.shellcheck.net/) warnings (although I see I've already failed at that:), so please enclose "$rhel_major" in quotes, also {}s to be consistent with the rest of the file.

julien-thierry: Good point, the warnings from these commits should be fixed now!

jpoimboe: Should this file be added to the list of files checked by "make check"?

jpoimboe:
I think it would be a good idea to update the README to recommend using this script instead of the manual steps listed.

test/integration/lib.sh
julien-thierry: @jpoimboe I've made a pull request to add this and other bash files to make check #1110

julien-thierry:
> I think it would be a good idea to update the README to recommend using this script instead of the manual steps listed.

Should that be part of this pull request?

jpoimboe:
> Should that be part of this pull request?

I don't have an opinion :-)

README.md
```diff
@@ -192,47 +112,16 @@ ccache --max-size=5G
...
-deb http://ddebs.ubuntu.com/ ${codename} main restricted universe multiverse
-deb http://ddebs.ubuntu.com/ ${codename}-security main restricted universe multiverse
-deb http://ddebs.ubuntu.com/ ${codename}-updates main restricted universe multiverse
-deb http://ddebs.ubuntu.com/ ${codename}-proposed main restricted universe multiverse
-EOF
-
-# add APT key
-wget -Nq http://ddebs.ubuntu.com/dbgsym-release-key.asc -O- | sudo apt-key add -
-apt-get update && apt-get install linux-image-$(uname -r)-dbgsym
+# Will request root priviledges
```
jpoimboe, line 122: should be "privileges"

PR #1110, opened 2020-06-03
Shellcheck fixes

julien-thierry:
After realizing that some of the bash/sh code used in the kpatch tree is not checked by `make check`, here's a bunch of fixes or ignores on shellcheck warnings.


yhcote:
> Everything looks reasonable. Some of the quoting complaints could be fixed, but shutting them off to avoid quotation mark madness is okay w/me.

I agree with that.

PR #1113, opened 2020-06-12
create-diff-object: ignore .altinstr_replacement when not included

haoren3696:
When .altinstructions is not included and .altinstr_replacement status
is CHANGED, we got error ".altinstr_replacement not selected for inclusion".

Signed-off-by: Zhipeng Xie <xiezhipeng1@huawei.com>

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2712,8 +2712,10 @@ static void kpatch_process_special_sections(struct kpatch_elf *kelf,
 		 * Only include .altinstr_replacement if .altinstructions
 		 * is also included.
 		 */
-		if (!altinstr)
+		if (!altinstr) {
+			sec->status = SAME;
 			break;
+		}
 
 		/* include base section */
 		sec->include = 1;


kpatch-build/create-diff-object.c
```diff
@@ -2712,8 +2712,10 @@ static void kpatch_process_special_sections(struct kpatch_elf *kelf,
 		 * Only include .altinstr_replacement if .altinstructions
 		 * is also included.
 		 */
-		if (!altinstr)
+		if (!altinstr) {
+			sec->status = SAME;
```
julien-thierry, line 2716: Do you have an example of patch triggering this?
If the patch modifies the replacement instructions and the running kernel uses the "unpatched" alternative instructions, it probably makes more sense to have the modified replacement instructions in the patch, no?
Or if we are not able to patch those instructions in such a situation, it is probably safer to keep a kpatch-build error when altinstr_replacement is modified but .altrinstruction isn't.

haoren3696: I am sure that my patch does not modify the replacement instructions. The patch add some if condition to return earlier which make .altinstr_replacement changed.

joe-lawrence: > I am sure that my patch does not modify the replacement instructions. The patch add some if condition to return earlier which make .altinstr_replacement changed.

Hi haoren3696,

If you can't share the patch, can you share the object and function that contains this alternative?  Just curious to see if this is a unique or common pattern that we need to consider.  Thanks.

haoren3696: Hi joe-lawrence

I met this problem on aarch64, here is my test patch(patch itself is meaningless,just to reproduce the problem):
```
diff --git a/mm/memcontrol.c b/mm/memcontrol.c
index db2b316b2..b4ee0d098 100644
--- a/mm/memcontrol.c
+++ b/mm/memcontrol.c
@@ -2640,6 +2640,10 @@ int __memcg_kmem_charge(struct page *page, gfp_t gfp, int order)
        struct mem_cgroup *memcg;
        int ret = 0;
 
+       if (!cgroup_memory_nokmem)
+               cgroup_memory_nokmem = true;
+       if (cgroup_memory_nokmem)
+               return 0;
        if (memcg_kmem_bypass())
                return 0;
 
@@ -3087,6 +3091,8 @@ static int memcg_online_kmem(struct mem_cgroup *memcg)
 {
        int memcg_id;
 
+       if (!cgroup_memory_nokmem)
+               cgroup_memory_nokmem = true;
        if (cgroup_memory_nokmem)
                return 0;
 
```

kpatch-build error:
ERROR: memcontrol.o: 1 unsupported section change(s)
memcontrol.o: changed function: mem_cgroup_css_alloc
memcontrol.o: changed function: __memcg_kmem_charge
memcontrol.o: changed section .altinstr_replacement not selected for inclusion
/usr/libexec/kpatch/create-diff-object: unreconcilable difference

Neither mem_cgroup_css_alloc nor __memcg_kmem_charge has relocation in .rela.altinstructions

Thanks

joe-lawrence: My naive x86_64 understanding of alternatives is that they produce something like the following:
```
  alternative("call foo1", "call foo2", FEATURE);

           .altinstructions                    .rela.altinstructions
     ----     old = (reloc) patch spot            old
    |  --     new = (reloc) alt instr spot        new
    | |       feature
    | |       old_len, new_len, pad_len
    | |
    | |
    | |    .text                               .rela.text
     -|-->    call (reloc) foo1                   foo1
      |       < nop pads >
      |       < nop pads >
      |
      |    .altinst_replacement                .rela.altinst_replacement
       -->    call (reloc) foo2                   foo2
```
So each `.altinstructions` entry has corresponding relocations in `.rela.altinstructions` that point to where the old code lives in `.text` and the new code in `.altinst_replacement` (these sections in turn may have their own relocations).  The entry also has bookkeeping data to describe the CPU feature and the lengths of the old and new code, as well as the pad bytes needed.

There are a lot of different scenarios to consider here: what if the old instruction sequence changes, or the new one, or the feature, or alternative directives are added or removed, etc.  The bottom lines in my mind:

* Consider the `.rela.altinstructions` relocations in \<old, new\> pairs with an associated `.altinstructions` entry and regions of `.text` and `.altinst_replacement`.  We will call all of this (A).

* If \<old\> points to a `.text` region that has changed, we should include all of (A) in the kpatch-patch.ko as that code will need alternatives processing on module load.  This is true even if (A) has remained unchanged.

* If \<new\> points to an `.altinst_replacement` region (new_len bytes long) that has changed, we should consider including (A) as the alternative code sequence has changed.

Now to the aarch64 changes Xie mentioned.  My kernel was slightly different, so I could only apply part of the diff, but I saw alternatives changes even with:
```
  diff --git a/mm/memcontrol.c b/mm/memcontrol.c
  index 5381afb23..95deb3b79 100644
  --- a/mm/memcontrol.c
  +++ b/mm/memcontrol.c
  @@ -3404,6 +3404,9 @@ static int memcg_online_kmem(struct mem_cgroup *memcg)
   {
          int memcg_id;

  +       if (!cgroup_memory_nokmem)
  +               cgroup_memory_nokmem = true;
  +
          if (cgroup_memory_nokmem)
                  return 0;
```
This effectively removes the rest of the code from `memcg_online_kmem()`.

Using some heavy sed filtering, we can compare the resulting `.altinstr_replacement` sections and see that it's smaller:
```
  % diff -u <(objdump -D -j .altinstr_replacement before/memcontrol.o | sed 's/^[ 0-9a-f]*:[\t][0-9a-f]*//g') \
            <(objdump -D -j .altinstr_replacement after/memcontrol.o | sed 's/^[ 0-9a-f]*:[\t][0-9a-f]*//g')
  --- /dev/fd/63  2020-06-16 15:05:27.308506434 -0400
  +++ /dev/fd/62  2020-06-16 15:05:27.308506434 -0400
  @@ -1,5 +1,5 @@

  -before/memcontrol.o:     file format elf64-littleaarch64
  +after/memcontrol.o:     file format elf64-littleaarch64


   Disassembly of section .altinstr_replacement:
  @@ -69,7 +69,6 @@
          mrs     x20, s3_0_c4_c6_0
          eor     w0, w20, #0xe0
          mrs     x0, tpidr_el2
  -       msr     s3_0_c4_c6_0, x0
          swpal   x0, x1, [x3]
          nop
          nop
```
(I'll leave it to the aarch64 experts to figure out which ALTERNATIVE this is traced back to from memcg_online_kmem().)

The `.altinstructions` entries are binary, but we can see that those section sizes have changed as well, suggesting fewer entries:
```
  File: before/memcontrol.o
  Section Headers:
    [Nr] Name                  Type            Address          Off    Size   ES Flg Lk Inf Al
    [44] .altinstructions      PROGBITS        0000000000000000 00b178 000300 00   A  0   0  1
    [45] .rela.altinstructions RELA            0000000000000000 0d3a38 000c00 18   I 85  44  8
                                                                       ^^^^^^
  File: after/memcontrol.o
  Section Headers:
    [Nr] Name                  Type            Address          Off    Size   ES Flg Lk Inf Al
    [44] .altinstructions      PROGBITS        0000000000000000 00b044 0002f4 00   A  0   0  1
    [45] .rela.altinstructions RELA            0000000000000000 0d3250 000bd0 18   I 85  44  8
                                                                       ^^^^^^
```
At the end of all of this, the patch removes a bunch of `.text`.  That should result in a smaller number of (A) data sets.  For the kpatch-patch.ko, theoretically we should regenerate:

  `.altinstructions` and `.rela.altinstructions`
  `.altinst_replacement` and `.rela.altinst_replacement`

based on the criteria for including (A) above.  In this case, entries are removed, so I would expect that the final kpatch-patch.ko wouldn't need any of those sections.

I'm not really familiar with how kpatch-build currently handles these sections, but I wonder if it is as granular about the individual parts of (A) as this description.

This is only my understanding of things from the perspective of how alternatives were implemented for x86_64... so perhaps I've missed a great simplification?  @jpoimboe @julien-thierry 

Another thing I would have expected is that kpatch-build to detect that `.altinstructions` is smaller and should be included in some way.  Perhaps it didn't change, I don't know.  Without seeing the aarch64 kpatch support it's hard to tell.   @haoren3696 it might be interesting to attach your before and after memcontrol.o here to see how things might have looked from your end.

julien-thierry: From what I can see, the replacement instruction getting removed is related to interrupt masking. The related instruction is in arch/arm64/include/asm/irqflags.h, part of the function `arch_local_irq_restore()` when the feature `ARM64_HAS_IRQ_PRIO_MASKING` is available.

Since the patch is in generic mm code, it probably means that a call to `local_irq_restore()` or `spin_unlock_irqrestore()` was removed as a result of the change (either really removed, or some caller no longer gets inlined).

But, in either case this should mean that an entry of the .altinstruction has been removed. And from what I can see, it seems like kpatch-build doesn't worry about a change of size in the section. Looking at `kpatch_regenerate_special_section()`, it seems to me that .altinstructions will only be included if it's rela section references changed or new symbols. I don't have the impression the case of removing an entry from that section was expected. Athough I'm not 100% sure of this.

haoren3696:
Hi, here is my before and after memcontrol.o
[testcase.tar.gz](https://github.com/dynup/kpatch/files/4796042/testcase.tar.gz)

kpatch-build/create-diff-object.c
haoren3696: Hi，julien-thierry 

>But, in either case this should mean that an entry of the .altinstruction has been removed. >And from what I can see, it seems like kpatch-build doesn't worry about a change of size in >the section. Looking at kpatch_regenerate_special_section(), it seems to me that >.altinstructions will only be included if it's rela section references changed or new symbols. I >don't have the impression the case of removing an entry from that section was expected. >Athough I'm not 100% sure of this.

Yes,kpatch_regenerate_special_section only include it's rela section references changed or new symbols, in my case, relocations referencing __memcg_kmem_charge in .rela.altinstructions has been deleted. So kpatch_regenerate_special_section keep nothing in .altinstructions so its sec->include = 0.

julien-thierry: Hi haoren3696,

I don't think it makes sense to force embed the `.altrinst_replacement` in that case. Without the `.altinstructions` section, the `.altinstr_replacement` is completely useless.

The `.altinstruction` and `.altrinst_replacement` that are embedded in the final .ko file created by kpatch-build are only used to apply alternatives to the functions provided by the patch object. When `.altinstruction` is not selected for inclusion, it means that none of the functions that will get inserted by the kpatch/livepatch have instruction portions that might be replaced by the alternative processing at module loading.

In the case of simple deletion, that shouldn't be an issue since, if a function was referencing it, it's text size & content should've changed as the number of original instructions is at least reduced.

However I do have some concern if a patch decide to only modify an alternative instruction with an instruction of the same size. This means that the original function's text and relocations are unmodified. So the `.altinstruction` only references unmodified symbols and isn't included. But we need to include it in the patch module since it will be required in case the patch function need to use the new alternatives.

So I wonder if a proper fix wouldn't be to refine the link made between the original instructions and the alternative instructions. If only the alternative instructions are modified, then the function containing the original instructions should be considered as CHANGED. Then the special section regeneration would work properly since `.altinstructions` will be referencing a changed symbol.

@joe-lawrence , @jpoimboe would that make sense?

haoren3696: Hi，julien-thierry

>I don't think it makes sense to force embed the .altrinst_replacement in that case. Without >the .altinstructions section, the .altinstr_replacement is completely useless.

>The .altinstruction and .altrinst_replacement that are embedded in the final .ko file created >by kpatch-build are only used to apply alternatives to the functions provided by the patch >object. When .altinstruction is not selected for inclusion, it means that none of the functions >that will get inserted by the kpatch/livepatch have instruction portions that might be >replaced by the alternative processing at module loading.

Agreed, that is why my PR ignore .altrinst_replacement when .altinstructions is not included.
But my PR did not cover case where user modify an alternative instruction with an instruction of the same size.

julien-thierry: I just realised that as is my suggestion does not solve your issue.

- For added `.altinstr_replacement` , this means that some original set of instructions was also added to a function that is either NEW or CHANGED and referenced by `.altrinstructions`. So both sections are included as required by the current code.
- For changed original instructions, the functions containing them counts as changed and same as above, the sections are included by the current code.
- For changed `.altinstr_replacement`, we need to make sure that the change is reflected on the function containing the instructions that might get replaced
- For deleted instructions I think there isn't much to do. If the function still exists, its size would've changed and everything works as expected. If the function no longer exists, it won't be referenced in `.altinstructions` and that section might not be included if there was no other change, causing your issue. But in such a case we don't need either section as they don't provide any relevant information for the patch module.

So I think the two things to do here are:
1) Properly handle a change in alternative instructions that isn't accompanied by a change in the original instructions.
2) If `.altinstructions` is not marked for inclusion, ignore that `.altinstr_replacement` might have changed, set it's state to SAME and not include it either.

The idea is to ensure that `.altinstructions` is always marked for inclusion when it is needed, this way we can rely on that information to know whether `.altrinstr_replacement` is required or not.

PR #1117, opened 2020-06-24
Ignore changes to .altinstr_aux

jpoimboe:
On x86, .altinstr_aux is used to store temporary code which allows static_cpu_has() to work before apply_alternatives() has run.  This code is completely inert for modules, because apply_alternatives() runs during module init, before the module is fully formed.  Any changed references to it (i.e. changed addend) can be ignored.  As long as they're both references to .altinstr_aux, they can be considered equal, even if the addends differ.


jpoimboe:
I still need to make a unit test.

PR #1118, opened 2020-06-30
Don't enable dynamic debug during integration tests on older rhels

sm00th:
RHELs older than 7.7 are missing 5279631271b3 "module: fix ddebug_remove_module()"
which leads to a crash if dynamic debug is used with livepatching.
Disable recent dynamic debug addition on those distros.

### DIFF
--- a/test/integration/kpatch-test
+++ b/test/integration/kpatch-test
@@ -50,6 +50,7 @@ KPATCHBUILD="$ROOTDIR"/kpatch-build/kpatch-build
 ERROR=0
 LOG=test.log
 DYNDEBUG_CONTROL=/sys/kernel/debug/dynamic_debug/control
+DYNDEBUG_ENABLED=1
 rm -f ./*.log
 
 PATCHDIR="${PATCHDIR:-$PWD}"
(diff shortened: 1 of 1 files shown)


test/integration/kpatch-test
```diff
@@ -282,6 +283,14 @@ run_combined_test() {
 
 }
 
+# shellcheck disable=SC1091
+source /etc/os-release
+if [[ "${ID}" == "rhel" && "${VERSION_ID%%.*}" == "7" && "${VERSION_ID##*.}" -le "6" ]]; then
+	DYNDEBUG_ENABLED=0
+	echo "Dynamic debug is not supported on current distro, disabling:"
+	echo "${PRETTY_NAME}"
```
kamalesh-babulal, line 291: Looks good, a minor nitpick on the preference, `${PRETTY_NAME}` can be inlined in to the warning, like  `echo "Dynamic debug is not supported on ${PRETTY_NAME}, disabling."`

sm00th: Thanks for the suggestion, I agree it will look better this way.

PR #1122, opened 2020-07-07
kpatch-test: make it easier to use custom kpatch and kpatch-build commands

euspectre:
The commands used to build the livepatches and to load or unload them
are currently hard-coded in kpatch-test.

This patch adds 2 options to kpatch-test to make it easier to use custom
kpatch and kpatch-build commands:

*  --system-kpatch-tools - if set, 'sudo kpatch' will be used to
load/unload the patches; 'kpatch-build' - to build them.

To use custom tools here, the user can adjust $PATH.

If the option is not set, kpatch-test will assume it is in KPatch source
tree, same as before this commit, and will use the tools from there.

* --kpatch-build-opts="..." - additional options to pass to
kpatch-build.

Example:
```
  ./kpatch-test \
    --system-kpatch-tools \
    --kpatch-build-opts="-s ./linux-src -c ./config -v ./vmlinux" \
    -d my_kpatch_tests/test/integration/v01
```

In this case, kpatch and kpatch-build installed in the system will be used,
and kpatch-build will look for the kernel source tree, configuration
file and vmlinux binary in the current directory.

### DIFF
--- a/test/integration/kpatch-test
+++ b/test/integration/kpatch-test
@@ -43,7 +43,6 @@ shopt -s nullglob
 # shellcheck disable=SC2046
 SCRIPTDIR=$(readlink -f $(dirname $(type -p "$0")))
 ROOTDIR=$(readlink -f "$SCRIPTDIR/../..")
-# TODO: option to use system-installed binaries instead
 KPATCH="sudo $ROOTDIR/kpatch/kpatch"
 unset CCACHE_HASHDIR
 KPATCHBUILD="$ROOTDIR"/kpatch-build/kpatch-build
@@ -64,9 +63,11 @@ usage() {
 	echo "		-c, --cached	Don't rebuild patch modules" >&2
 	echo "		-d, --directory	Patch directory" >&2
 	echo "		-q, --quick	Test combined patch and -FAIL patches only" >&2
+	echo "		--system-kpatch-tools	Use kpatch tools installed in the system" >&2
+	echo "		--kpatch-build-opts	Additional options to pass to kpatch-build" >&2
 }
 
-options=$(getopt -o hcd:q -l "help,cached,directory,quick" -- "$@") || exit 1
+options=$(getopt -o hcd:q -l "help,cached,directory,quick,system-kpatch-tools,kpatch-build-opts:" -- "$@") || exit 1
 
 eval set -- "$options"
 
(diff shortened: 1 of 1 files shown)


test/integration/kpatch-test
```diff
@@ -43,10 +43,9 @@ shopt -s nullglob
 # shellcheck disable=SC2046
 SCRIPTDIR=$(readlink -f $(dirname $(type -p "$0")))
 ROOTDIR=$(readlink -f "$SCRIPTDIR/../..")
-# TODO: option to use system-installed binaries instead
-KPATCH="sudo $ROOTDIR/kpatch/kpatch"
+KPATCH="${KPATCH:-sudo $ROOTDIR/kpatch/kpatch}"
```
kamalesh-babulal, line 46: Looks good, a minor nit-pick should the `sudo` prefix stay,  `KPATCH="sudo ${KPATCH:-sudo $ROOTDIR/kpatch/kpatch}"`.  I am not sure if the `TODO` is still in plan and it should be retained.

sm00th:
I think it would be cleaner if we implented that "TODO" point instead, i.e. add an option to switch to system-wide `kpatch` and `kpatch-build` (further configurable by altering `$PATH`) and another one to supply `kpatch-build` options.

euspectre:
> I think it would be cleaner if we implented that "TODO" point instead, i.e. add an option to switch to system-wide kpatch and kpatch-build (further configurable by altering $PATH) and another one to supply kpatch-build options.

Perhaps.

Something like this, may be:
`kpatch-test --system-kpatch-tools --kpatch-build-opts="-s linux-src -v vmlinux"`

?

`--system-kpatch-tools` would mean 
```
KPATCH="sudo kpatch"
KPATCHBUILD="kpatch-build"
```
without explicit paths.

sm00th:
> > I think it would be cleaner if we implented that "TODO" point instead, i.e. add an option to switch to system-wide kpatch and kpatch-build (further configurable by altering $PATH) and another one to supply kpatch-build options.
> 
> Perhaps.
> 
> Something like this, may be:
> `kpatch-test --system-kpatch-tools --kpatch-build-opts="-s linux-src -v vmlinux"`
> 
> ?
> 
> `--system-kpatch-tools` would mean
> 
> ```
> KPATCH="sudo kpatch"
> KPATCHBUILD="kpatch-build"
> ```
> 
> without explicit paths.

Yes, this looks good.

euspectre:
Pushed a new variant that relies on options rather than env. vars, updated the description.

test/integration/kpatch-test
```diff
@@ -64,9 +63,11 @@ usage() {
 	echo "		-c, --cached	Don't rebuild patch modules" >&2
 	echo "		-d, --directory	Patch directory" >&2
 	echo "		-q, --quick	Test combined patch and -FAIL patches only" >&2
+	echo "		--system-kpatch-tools	Use KPatch tools installed in the system" >&2
```
jpoimboe, line 66: "KPatch" -> "kpatch"

euspectre: Done.

PR #1123, opened 2020-07-10
fix compile error on aarch64

haoren3696:
We got compile error on aarch64:

[root@localhost kpatch]# make
make -C kpatch-build
make[1]: Entering directory '/home/opensource/kpatch/kpatch-build'
gcc -MMD -MP -I../kmod/patch -Iinsn -Wall -Wsign-compare -Wconversion -Wno-sign-conversion -g -Werror   -c -o create-diff-object.o create-diff-object.c
create-diff-object.c: In function ‘kpatch_replace_sections_syms’:
create-diff-object.c:1493:5: error: implicit declaration of function ‘rela_insn’; did you mean ‘rela_equal’? [-Werror=implicit-function-declaration]
     rela_insn(sec, rela, &insn);
     ^~~~~~~~~
     rela_equal
make[1]: *** [<builtin>: create-diff-object.o] Error 1
make[1]: Leaving directory '/home/opensource/kpatch/kpatch-build'
make: *** [Makefile:22: build-kpatch-build] Error 2

### DIFF
--- a/kpatch-build/Makefile
+++ b/kpatch-build/Makefile
@@ -10,11 +10,7 @@ SOURCES = create-diff-object.c kpatch-elf.c \
 		  create-kpatch-module.c \
 		  create-kpatch-module.c lookup.c
 
-ifeq ($(ARCH),x86_64)
-SOURCES += insn/insn.c insn/inat.c
-INSN     = insn/insn.o insn/inat.o
-insn/%.o: CFLAGS := $(filter-out -Wconversion, $(CFLAGS))
-else ifeq ($(ARCH),ppc64le)
+ifeq ($(ARCH),ppc64le)
 SOURCES += gcc-plugins/ppc64le-plugin.c
 PLUGIN   = gcc-plugins/ppc64le-plugin.so
 TARGETS += $(PLUGIN)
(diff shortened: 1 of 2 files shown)


jpoimboe:
I don't see a need to fix the compile error if aarch64 isn't supported anyway.  Otherwise it might be misleading.  For example, Linux arm64 doesn't yet have CONFIG_HAVE_LIVEPATCH, or objtool.

kamalesh-babulal:
Should we bail out during the make on the unsupported architectures, something like:
```
--- a/kpatch-build/Makefile
+++ b/kpatch-build/Makefile
@@ -22,6 +22,8 @@ GCC_PLUGINS_DIR := $(shell gcc -print-file-name=plugin)
 PLUGIN_CFLAGS := $(filter-out -Wconversion, $(CFLAGS))
 PLUGIN_CFLAGS += -shared -I$(GCC_PLUGINS_DIR)/include \
                   -Igcc-plugins -fPIC -fno-rtti -O2 -Wall
+else
+$(error "Unsupported architecture ${ARCH}. Check https://github.com/dynup/kpatch/#supported-architectures.")
 endif
 
 
```

yhcote:
> Should we bail out during the make on the unsupported architectures, something like:
> 
> ```
> --- a/kpatch-build/Makefile
> +++ b/kpatch-build/Makefile
> @@ -22,6 +22,8 @@ GCC_PLUGINS_DIR := $(shell gcc -print-file-name=plugin)
>  PLUGIN_CFLAGS := $(filter-out -Wconversion, $(CFLAGS))
>  PLUGIN_CFLAGS += -shared -I$(GCC_PLUGINS_DIR)/include \
>                    -Igcc-plugins -fPIC -fno-rtti -O2 -Wall
> +else
> +$(error "Unsupported architecture ${ARCH}. Check https://github.com/dynup/kpatch/#supported-architectures.")
>  endif
>  
>  
> ```

Would avoid some confusion I think.

joe-lawrence:
A lot more work would need to go into fully supporting aarch64, so for the time being these kind of fixups should accumulate in a developer branch somewhere.  Also PR #1127 is going to error out of the kpatch build for unsupported arches, so closing this one for now.

Issue #1120, opened 2020-07-01
Deadlock caused by scheduling in callbacks may appear when I use kpatch


zhenjiangma:
I found a bug in kpatch, which may cause deadlock. 
  I do some pre-patch preparation work in pre_patch_callback function, and in the preparation work the task named P1 will gave up CPU for some reason. At the same time, there is another task named P2 running which will send IPI.
  However, because the irq has been disabled in stop_machine function, and the P1 also goes to sleep due to pre_patch_callback function. Exactly, the P2 is scheduled. 
  Therefore, terrible things happen. P1 has disabled all irq and went to sleep, while P2 sends IPI and waits for responses. Inevitably the system trapped into the deadlock state. The reason for this is that the pre_patch/post_patch work is executed in stop_machine.  Is there a better way to avoid this scenario？
  In order to reproduce it, I specially designed the following scene.
/*** this is my patch file ***/
````diff --git a/fs/proc/meminfo.c b/fs/proc/meminfo.c
index 71d006c..6a93a52 100644
--- a/fs/proc/meminfo.c
+++ b/fs/proc/meminfo.c
@@ -22,6 +22,10 @@
 #include "internal.h"
 #include <linux/pid_namespace.h>
 
+#include <linux/module.h>
+#include "kpatch-macros.h"
+#include <linux/kernel.h>
+#include <linux/semaphore.h>
 void __attribute__((weak)) arch_report_meminfo(struct seq_file *m)
 {
 }
@@ -189,6 +193,39 @@ static int meminfo_proc_show(struct seq_file *m, void *v)
 
 	return 0;
 }
+static void callback_info(const char *callback, patch_object *obj)
+{   
+	    pr_info("ENTERING INTO CALLBACK_INFO");
+}
+
+static int pre_patch_callback(patch_object *obj)
+{    
+        struct semaphore sem;
+	    callback_info(__func__, obj);
+		sema_init(&sem, 1);
+	    down(&sem);
+		down(&sem);
+		return 0;
+}
+KPATCH_PRE_PATCH_CALLBACK(pre_patch_callback);
+
+static void post_patch_callback(patch_object *obj)
+{   
+	    callback_info(__func__, obj);
+}
+KPATCH_POST_PATCH_CALLBACK(post_patch_callback);
+
+static void pre_unpatch_callback(patch_object *obj)
+{
+	    callback_info(__func__, obj);
+}
[...]

Issue #1077, opened 2020-02-27
kpatch load look leads to deadlock in stop_machine context


xiejingf:
I registered a look hook using KPATCH_LOAD_HOOK(foo) and encounter the scenario below which leads to system deadlock

stop_machine -> foo,   in the foo function , acquire a mutex failed,   then schedule,  the new picked up task happens to send ipi across other CPUs and busy wait other CPUs to respond, which is impossible, because the other CPUs are stopped by stop_machine call,  so the system resulted in a deadlock state.

My suggestion is move the kpatch load/unload hook outside the stop_machine context, like module_init in kernel live patch.

My patch is based on v0.3.2

```
diff --git a/kmod/core/core.c b/kmod/core/core.c
index 77c7290..9ffa57b 100644
--- a/kmod/core/core.c
+++ b/kmod/core/core.c
@@ -413,13 +413,6 @@ static int kpatch_apply_patch(void *data)
 		return -EBUSY;
 	}
 
-	/* run any user-defined load hooks */
-	list_for_each_entry(object, &kpmod->objects, list) {
-		if (!kpatch_object_linked(object))
-			continue;
-		list_for_each_entry(hook, &object->hooks_load, list)
-			(*hook->hook)();
-	}
 
	return 0;
@@ -450,13 +443,6 @@ static int kpatch_remove_patch(void *data)
 		hash_del_rcu(&func->node);
 	} while_for_each_linked_func();
 
-	/* run any user-defined unload hooks */
-	list_for_each_entry(object, &kpmod->objects, list) {
-		if (!kpatch_object_linked(object))
-			continue;
-		list_for_each_entry(hook, &object->hooks_unload, list)
-			(*hook->hook)();
-	}
 
 	return 0;
 }
@@ -923,6 +909,7 @@ int kpatch_register(struct kpatch_module *kpmod, bool replace)
 	int ret, i, force = 0;
 	struct kpatch_object *object, *object_err = NULL;
 	struct kpatch_func *func;
+	struct kpatch_hook *hook;
 
 	if (!kpmod->mod || list_empty(&kpmod->objects))
 		return -EINVAL;
@@ -971,6 +958,16 @@ int kpatch_register(struct kpatch_module *kpmod, bool replace)
 
 	kpatch_state_updating();
 
+
+	printk("running user defined load hooks");
+	/* run any user-defined load hooks */
+	list_for_each_entry(object, &kpmod->objects, list) {
+		if (!kpatch_object_linked(object))
[...]

PR #1126, opened 2020-07-20
core: remove the callbacks from kpatch_apply_patch and kpatch_remove_patch

zhenjiangma:
We encountered a problem refered to #1077 and  #1120. After analysis, we believe that the main reason is that the callbacks are placed in the context of stop_machine. Because there is no data-dependy, so we remove the callbacks from the functions of kpatch_apply_patch and kpatch_remove_patch. We will avoid being trapped into a dead lock while a sleep appears in callbacks and exactly a process sending IPI was scheduled.

### DIFF
--- a/kmod/core/core.c
+++ b/kmod/core/core.c
@@ -365,7 +365,6 @@ static int kpatch_apply_patch(void *data)
 	struct kpatch_module *kpmod;
 	struct kpatch_func *func;
 	struct hlist_node *tmp;
-	struct kpatch_object *object;
 	int ret;
 	int i;
 
@@ -375,16 +374,6 @@ static int kpatch_apply_patch(void *data)
 	if (ret) {
 		kpatch_state_finish(KPATCH_STATE_FAILURE);
 		return ret;
-	}
-
-	/* run any user-defined pre-patch callbacks */
-	list_for_each_entry(object, &kpmod->objects, list) {
-		ret = pre_patch_callback(object);
-		if (ret) {
-			pr_err("pre-patch callback failed!\n");
-			kpatch_state_finish(KPATCH_STATE_FAILURE);
-			goto err;
-		}
 	}
 
 	/* tentatively add the new funcs to the global func hash */
@@ -409,7 +398,7 @@ static int kpatch_apply_patch(void *data)
 		} while_for_each_linked_func();
 
 		ret = -EBUSY;
-		goto err;
+		return ret;
 	}
 
 	/*
@@ -425,17 +414,7 @@ static int kpatch_apply_patch(void *data)
 		}
 	}
 
-	/* run any user-defined post-patch callbacks */
-	list_for_each_entry(object, &kpmod->objects, list)
-		post_patch_callback(object);
-
 	return 0;
-err:
-	/* undo pre-patch callbacks by calling post-unpatch counterparts */
-	list_for_each_entry(object, &kpmod->objects, list)
-		post_unpatch_callback(object);
-
-	return ret;
 }
 
 /* Called from stop_machine */
@@ -443,7 +422,6 @@ static int kpatch_remove_patch(void *data)
 {
 	struct kpatch_module *kpmod = data;
 	struct kpatch_func *func;
-	struct kpatch_object *object;
 	int ret;
 
 	ret = kpatch_verify_activeness_safety(kpmod, false);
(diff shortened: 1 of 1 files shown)


joe-lawrence:
@euspectre - if this looks interesting to support for kpatch.ko, do you mind reviewing?  Thanks.

euspectre:
@joe-lawrence Yes, I'll take a look.

kmod/core/core.c
```diff
@@ -742,7 +705,7 @@ static int kpatch_write_relocations(struct kpatch_module *kpmod,
 			break;
 		case R_X86_64_64:
 			loc = dynrela->dest;
```
euspectre: This part of the patch is unrelated. Looks like an artefact from rebase or something.
Please rebase the patch on top of the current master branch in dynup/kpatch.

euspectre:
@zhenjiangma :  Commit https://github.com/dynup/kpatch/pull/1126/commits/f912b4aea2913a8c447447838587c4f2fb5853e6 is included into the pull requst but is not related to this.

As for "remove the callbacks from kpatch_apply_patch and kpatch_remove_patch" (https://github.com/dynup/kpatch/pull/1126/commits/45a8fca17f1baf141146eb5802d99ca15eba2df9) - I cannot see the description in the commit. The change is non-trivial, so it should be described in the commit message too. Otherwise, even the author would have a hard time, in a few months, remembering why these changes were needed.

-------

As far as I can see, you moved all four kinds of callbacks out of stop_machine(). As I suggested in https://github.com/dynup/kpatch/issues/1077, it is probably safe for pre-patch and post-unpatch callbacks because no races with the patched code are possible there. Should be enough for the most cases, including https://github.com/dynup/kpatch/issues/1120.

I'd leave post-patch and pre-unpatch callbacks under stop_machine() for now, because these could run in parallel with the patched code otherwise, leading to potential bugs.

kmod/core/core.c
zhenjiangma: Yes, maybe I got an old version. I will rebase the patch on top of the current master branch in dynup/kpatch.

zhenjiangma:
> @zhenjiangma : Commit [f912b4a](https://github.com/dynup/kpatch/commit/f912b4aea2913a8c447447838587c4f2fb5853e6) is included into the pull requst but is not related to this.
> 
> As for "remove the callbacks from kpatch_apply_patch and kpatch_remove_patch" ([45a8fca](https://github.com/dynup/kpatch/commit/45a8fca17f1baf141146eb5802d99ca15eba2df9)) - I cannot see the description in the commit. The change is non-trivial, so it should be described in the commit message too. Otherwise, even the author would have a hard time, in a few months, remembering why these changes were needed.
> 
> As far as I can see, you moved all four kinds of callbacks out of stop_machine(). As I suggested in #1077, it is probably safe for pre-patch and post-unpatch callbacks because no races with the patched code are possible there. Should be enough for the most cases, including #1120.
> 
> I'd leave post-patch and pre-unpatch callbacks under stop_machine() for now, because these could run in parallel with the patched code otherwise, leading to potential bugs.

Yes, as you mentioned above, pre-patch and post-unpatch callbacks are safe.  I didn't considered the race between the patched code and post-patch/pre-unpatch callbacks when run in parallel. In action,  we use pre_patch and post_unpatch usually,  and it is enough for most cases. I will repull it as you suggested above.

zhenjiangma:
> @zhenjiangma : Commit [f912b4a](https://github.com/dynup/kpatch/commit/f912b4aea2913a8c447447838587c4f2fb5853e6) is included into the pull requst but is not related to this.
> 
> As for "remove the callbacks from kpatch_apply_patch and kpatch_remove_patch" ([45a8fca](https://github.com/dynup/kpatch/commit/45a8fca17f1baf141146eb5802d99ca15eba2df9)) - I cannot see the description in the commit. The change is non-trivial, so it should be described in the commit message too. Otherwise, even the author would have a hard time, in a few months, remembering why these changes were needed.
> 
> As far as I can see, you moved all four kinds of callbacks out of stop_machine(). As I suggested in #1077, it is probably safe for pre-patch and post-unpatch callbacks because no races with the patched code are possible there. Should be enough for the most cases, including #1120.
> 
> I'd leave post-patch and pre-unpatch callbacks under stop_machine() for now, because these could run in parallel with the patched code otherwise, leading to potential bugs.

I pull a new request #1134 to make modification according to your suggestion, because I don't know how to recommit on this request.

euspectre:
> I pull a new request #1134 to make modification according to your suggestion, because I don't know how to recommit on this request.

You could, for example, drop the old commit(s) from your branch (see 'git rebase -i'), add the new commits and force-push them here (git push -f <...>).

If you authored the patch entirely in GitHub's web UI (not sure if that is possible), then I also don't know how to replace it with a new one.

Anyway, I'll take a look at the new patch.

PR #1128, opened 2020-07-20
kpatch-build: avoid modpost .output.o.cmd complaints

### LINKED ISSUE #1125: KLP module fails to build with latest mainline [5.8-rc]
The current mainline tree `5.8-rc` fails to build the kernel livepatching module,
  during the module build stage, with the error:
 ```
make -C /root/objtool/linux M=/root/.kpatch/tmp/patch
make[1]: Entering directory '/root/objtool/linux'
  AR      /root/.kpatch/tmp/patch/built-in.a
  LD [M]  /root/.kpatch/tmp/patch/livepatch-meminfo.o
/root/.kpatch/tmp/patch/.output.o.cmd: No such file or directory
make[2]: *** [scripts/Makefile.modpost:111: /root/.kpatch/tmp/patch/Module.symvers] Error 1
make[1]: *** [Makefile:1669: modules] Error 2
make[1]: Leaving directory '/root/objtool/linux'
make: *** [Makefile:22: livepatch-meminfo.ko] Error 2
```
 
the commit 70f30cfe5b89 ("modpost: use read_text_file() and get_line() for reading
text files") would fail any `.o` used in the building a module without `*.cmd` file.
The `*.cmd` file, holds the information like gcc command line used to build the
`.o`, source file details and dependencies list. This information I guess its used
by kbuild to trigger rebuilding of `.o` or `.o.*` files in case of the dependencies
are updated.
 
`kpatch/kmod/patch/Makefile`:
```
  ...
  $(KPATCH_NAME)-objs += patch-hook.o output.o
 
  all: $(KPATCH_NAME).ko
 
  $(KPATCH_NAME).ko:
          $(KPATCH_MAKE)
 
          $(KPATCH_MAKE) patch-hook.o
```

`output.o` doesn't uses the kernel build system but is hand build using
`create-diff-object`, hence the `.output.o.cmd` is not generated. One way to
address this issue would be created an empty `.output.o.cmd`, so that the kernel
build system is happy and it technically has no dependency on the kernel to
trigger re-build over using make -i (ignore) flag to bypass any such issue.

```
  --- a/kpatch-build/kpatch-build
  +++ b/kpatch-build/kpatch-build
@@ -963,6 +963,8 @@ ld -r $KPATCH_LDFLAGS -o ../patch/tmp_output.o $(find . -name "*.o") 2>&1 | logg
  
 if [[ "$USE_KLP" -eq 1 ]]; then
        cp "$TEMPDIR"/patch/tmp_output.o "$TEMPDIR"/patch/output.o || die
+       # empty .cmd file is required to keep KBUILD happy
+       touch "$TEMPDIR"/patch/.output.o.cmd
 else
        # Add .kpatch.checksum for kpatch script
        md5sum ../patch/tmp_output.o | awk '{printf "%s\0", $1}' > checksum.tmp || die
```
Makefile magic is something I don't understand well, I guess there are better ways
to fix this issue.

### PR DESCRIPTION
**Pull request by joe-lawrence**
The modpost step complains about one of our generated files, output.o
and that it can't find a corresponding .cmd file for it (full path names
stripped):

	WARNING: could not find .output.o.cmd for output.o

This was turned into an error in v5.8:

	.output.o.cmd: No such file or directory

Avoid this by creating an empty .cmd file so that modpost acknowledges
that the file exists, but doesn't parse anything out of it.

Fixes #1125
Reported-by: Kamalesh Babulal <kamalesh@linux.vnet.ibm.com> (for v5.8+)

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -963,6 +963,8 @@ ld -r $KPATCH_LDFLAGS -o ../patch/tmp_output.o $(find . -name "*.o") 2>&1 | logg
 
 if [[ "$USE_KLP" -eq 1 ]]; then
 	cp "$TEMPDIR"/patch/tmp_output.o "$TEMPDIR"/patch/output.o || die
+	# Avoid MODPOST warning (pre-v5.8) and error (v5.8+) with an empty .cmd file
+	touch "$TEMPDIR"/patch/.output.o.cmd || die
 else
 	# Add .kpatch.checksum for kpatch script
 	md5sum ../patch/tmp_output.o | awk '{printf "%s\0", $1}' > checksum.tmp || die

### CODE REVIEW

File: kpatch-build/kpatch-build [hunk 1]
```diff
@@ -963,6 +963,8 @@ ld -r $KPATCH_LDFLAGS -o ../patch/tmp_output.o $(find . -name "*.o") 2>&1 | logg
 
 if [[ "$USE_KLP" -eq 1 ]]; then
 	cp "$TEMPDIR"/patch/tmp_output.o "$TEMPDIR"/patch/output.o || die
+	# Avoid MODPOST warning and error (v5.8+) with an empty .cmd file
+	touch "$TEMPDIR"/patch/.output.o.cmd
```
kamalesh-babulal, line 967: Thanks for working on it. A minor nitpick, should `|| die` be used like other commands to fail the build based on the return code after its run? One argument would be that if the previous command `cp` succeeds in creating `output.o`, theoretically creating `.output.o.cmd` should also succeed.  On other hand if the die is not used and if the touch commands fails to create the `.output.o.cmd` file, the build will fail at every next step of kernel module build.  I am little biased towards including `die` here.

joe-lawrence:
v2: clarify avoidance comment and `die` if the `touch` command fails.

PR #1130, opened 2020-07-21
Hook hacking helpers

joe-lawrence:
These commits resulted from a debugging session with `livepatch-patch-hook.c`.  In that case, the `output.o` file was consistent, but I was testing changes to the livepatch API hook code.

This does neuter the `clean` target a bit, however the Makefile is already incapable of recreating `output.o` on its own anyway, so I figured the gain in debug-ability was worth it.  AFAIK `export -p` is POSIX, so hopefully not just a bash-ism.

### DIFF
--- a/kmod/patch/Makefile
+++ b/kmod/patch/Makefile
@@ -2,6 +2,9 @@ KPATCH_BUILD ?= /lib/modules/$(shell uname -r)/build
 KPATCH_MAKE = $(MAKE) -C $(KPATCH_BUILD) M=$(PWD) CFLAGS_MODULE='$(CFLAGS_MODULE)'
 LDFLAGS += $(KPATCH_LDFLAGS)
 
+# object files that this Makefile can (re)build on its own
+BUILDABLE_OBJS=$(filter-out output.o, $(wildcard *.o))
+
 obj-m += $(KPATCH_NAME).o
 ldflags-y += -T $(src)/kpatch.lds
 targets += kpatch.lds
@@ -19,5 +22,5 @@ patch-hook.o: patch-hook.c kpatch-patch-hook.c livepatch-patch-hook.c
 	$(KPATCH_MAKE) patch-hook.o
 
 clean:
-	$(RM) -Rf .*.o.cmd .*.ko.cmd .tmp_versions *.o *.ko *.mod.c \
+	$(RM) -Rf .*.o.cmd .*.ko.cmd .tmp_versions $(BUILDABLE_OBJS) *.ko *.mod.c \
 	Module.symvers
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -45,6 +45,7 @@ SRCDIR="$CACHEDIR/src"
 RPMTOPDIR="$CACHEDIR/buildroot"
 VERSIONFILE="$CACHEDIR/version"
 TEMPDIR="$CACHEDIR/tmp"
+ENVFILE="$TEMPDIR/kpatch-build.env"
 LOGFILE="$CACHEDIR/build.log"
 RELEASE_FILE=/etc/os-release
 DEBUG=0
@@ -87,6 +88,10 @@ logger() {
 		# Log only to the logfile
 		cat >> "$LOGFILE"
 	fi
+}
+
+save_env() {
+	export -p | grep -wv -e 'OLDPWD=' -e 'PWD=' > "$ENVFILE"
 }
 
 verify_patch_files() {
@@ -432,6 +437,7 @@ find_special_section_data() {
 	[[ -z "$PRINTK_INDEX_STRUCT_SIZE" && "$CONFIG_PRINTK_INDEX" -ne 0 ]] && die "can't find special struct pi_entry size"
 	[[ -z "$STATIC_CALL_STRUCT_SIZE" ]] && kernel_version_gte 5.10.0 && die "can't find special struct static_call_site size"
 
+	save_env
 	return
 }
 
@@ -714,6 +720,7 @@ elif [[ "$DISTRO" = ubuntu ]] || [[ "$DISTRO" = debian ]]; then
 
 	export PATH="/usr/lib/ccache:$PATH"
 fi
+save_env
 
 find_dirs || die "can't find supporting tools"
 
@@ -927,6 +934,7 @@ find_special_section_data
 if [[ $DEBUG -ge 4 ]]; then
 	export KPATCH_GCC_DEBUG=1
 fi
+save_env
 
 echo "Building original source"
 [[ -n "$OOT_MODULE" ]] || ./scripts/setlocalversion --save-scmversion || die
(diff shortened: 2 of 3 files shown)


kmod/patch/Makefile
```diff
@@ -18,12 +18,15 @@ $(KPATCH_NAME)-objs += patch-hook.o output.o
 
 all: $(KPATCH_NAME).ko
 
-$(KPATCH_NAME).ko:
+build.env:
+	export -p > $@
```
sm00th, line 22: I also often find myself in need of *_SIZE variables when working on create-diff-object. Maybe we can move this to kpatch-build instead so it can be used in both scenarios?

kmod/patch/Makefile
```diff
@@ -18,12 +18,15 @@ $(KPATCH_NAME)-objs += patch-hook.o output.o
...
+
+$(KPATCH_NAME).ko: build.env
 	$(KPATCH_MAKE)
 
 patch-hook.o: patch-hook.c kpatch-patch-hook.c livepatch-patch-hook.c
 	$(KPATCH_MAKE) patch-hook.o
 
 clean:
-	$(RM) -Rf .*.o.cmd .*.ko.cmd .tmp_versions *.o *.ko *.mod.c \
+	$(RM) -Rf .*.o.cmd .*.ko.cmd .tmp_versions *[^output].o *.ko *.mod.c \
```
sm00th, line 31: This doesn't have intended result. Any .o file containing o, u, t, p in it's name will be skipped. I am afraid we'll have to use make magic. This should work: `$(filter-out output.o, $(wildcard *.o))`.

joe-lawrence: Whoops, yeah you're right about that.  Make magic it is then.

kmod/patch/Makefile
joe-lawrence: That's a good suggestion, I've had to manually set those up for `create-diff-object` in the past as well.

As for implementing that, `create-diff-object` is called before `KCFLAGS` and `KCPPFLAGS` are exported.  I think we can move those exports to before `create-diff-object` if needed.

Also, the Makefile approach dumps `build.env` into the livepatch-patch directory.  If we move `build.env` creation into `kpatch-build` do we drop it into the same location?

jpoimboe: Could it go in ~/.kpatch/tmp?  If the user runs with '-d' or --skip-cleanup it would automatically be where they need it.

kamalesh-babulal: > As for implementing that, create-diff-object is called before KCFLAGS and KCPPFLAGS are exported. I think we can move those exports to before create-diff-object if needed.

Agree, we should re-arrange capture them as well. While we plan to capture the build environment, let's capture all the exported variables.

> Also, the Makefile approach dumps build.env into the livepatch-patch directory. If we move build.env creation into kpatch-build do we drop it into the same location?

Like @jpoimboe suggested, `~/.kpatch/tmp/` sounds like a nature fit to place the `build.env` file.

joe-lawrence: Is it worth adding a `do_export()` wrapper function that does the export and also runs the `export -p > "$TEMPDIR"/build.env`.  This way the exports can stay where they are (close to related code) and maybe it's useful if other parts of the script happen to fail.

Also `kpatch-build.env` a better filename?

sm00th: > Is it worth adding a do_export() wrapper function that does the export and also runs the export -p > "$TEMPDIR"/build.env. This way the exports can stay where they are (close to related code) and maybe it's useful if other parts of the script happen to fail.

That'll be a lot of overwrites but makes it much easier, we also won't have to make sure new exports are added before dumping them. So it is probably worth it.

> Also kpatch-build.env a better filename?

Makes it clearer, yeah.

kpatch-build/kpatch-build
```diff
@@ -88,6 +89,17 @@ logger() {
 	fi
 }
 
+save_env() {
+	export -p | grep -wve 'OLDPWD=' -wve 'PWD=' > "$ENVFILE"
```
joe-lawrence, line 93: Now that the saved .env file is up in $CACHEDIR/tmp and not in the patch directory, we need to filter out the `{OLD,}PWD` variables to avoid some bash environment weirdness when sourcing the file.

kpatch-build/kpatch-build
```diff
@@ -808,10 +821,8 @@ CROSS_COMPILE="$TOOLSDIR/kpatch-gcc " make "-j$CPUS" $TARGETS 2>&1 | logger || d
 echo "Building patched source"
 apply_patches
 mkdir -p "$TEMPDIR/orig" "$TEMPDIR/patched"
-KPATCH_GCC_TEMPDIR="$TEMPDIR"
-export KPATCH_GCC_TEMPDIR
-KPATCH_GCC_SRCDIR="$SRCDIR"
-export KPATCH_GCC_SRCDIR
+do_export 'KPATCH_GCC_TEMPDIR' "$TEMPDIR"
+do_export 'KPATCH_GCC_SRCDIR' "$SRCDIR"
```
joe-lawrence, line 825: Wrapping the export behind a function + eval means shellcheck can't find a reference to these variables and complains.  That can be avoided by moving the temporary assignment directly into the do_export call.

kpatch-build/kpatch-build
```diff
@@ -978,9 +989,10 @@ if [[ -z "$OOT_MODULE" ]]; then
 else
 	KPATCH_BUILD="/lib/modules/$ARCHVERSION/build"
 fi
-KPATCH_BUILD="$KPATCH_BUILD" KPATCH_NAME="$MODNAME" \
+export KPATCH_BUILD="$KPATCH_BUILD" KPATCH_NAME="$MODNAME" \
 KBUILD_EXTRA_SYMBOLS="$KBUILD_EXTRA_SYMBOLS" \
-KPATCH_LDFLAGS="$KPATCH_LDFLAGS" \
+KPATCH_LDFLAGS="$KPATCH_LDFLAGS"
+save_env
```
joe-lawrence, line 995: Now the the .env is created in the parent script through the do_export wrapper, it misses this pattern:
`FOO=x my_prog` since `FOO` was not previously exported.  The solution is to export them first.

kpatch-build/kpatch-build
```diff
@@ -350,6 +362,7 @@ find_special_section_data() {
 	[[ -z "$ORC_STRUCT_SIZE" && "$CONFIG_UNWINDER_ORC" -ne 0 ]] && die "can't find special struct orc_entry size"
 	[[ -z "$JUMP_STRUCT_SIZE" && "$CONFIG_JUMP_LABEL" -ne 0 ]] && die "can't find special struct jump_entry size"
 
+	save_env
```
joe-lawrence, line 365: save_env() lets us capture the environment after exports we couldn't wrapper

sm00th (changes requested):
I think we still want these merged. But it badly needs a rebase and I had a couple of comments. @joe-lawrence let me know if you need help with any of this.

joe-lawrence:
> I think we still want these merged. But it badly needs a rebase and I had a couple of comments. @joe-lawrence let me know if you need help with any of this.

@sm00th : if you want to take a crack at rebasing and updating this PR, go ahead.

sm00th (commented):
Rebased, got rid of do_export and fixed up grep in save_env as mentioned in https://github.com/dynup/kpatch/pull/1130#discussion_r704160013

sm00th (approved):

kpatch-build/kpatch-build
joe-lawrence: > I don't think we need to specify `-w` and `-v` twice...
Yes, applying those flags once, separately from the `-e` expressions looks cleaner.

kpatch-build/kpatch-build
joe-lawrence: Sure. Those  `save_env` calls may be a moving target if some additional environment is added w/o updating the call, but it's probably easier to just update/move the `save_env` calls if that ever happens.

joe-lawrence (commented):
Update approved*.  Thanks @sm00th for updating this one... now I just need to remember that we implemented these.

\* Unlike GitLab, GitHub tells me, "Pull request authors can't approve their own pull request."

PR #1134, opened 2020-07-28
deal with the deadlock may occur in stop_machine context.

zhenjiangma:
we remove the pre_patch_callback/post_unpatch_callback from the
stop_machine context. If a schedule/sleep happend in callbacks while the
process to be scheuded later will send IPI, because all interrupts
are disabled, the machine will trap into a  deadlock in such situation.
So we remove the pre_patch_callback and post_unpatch_callback from
the stop_machine to avoid such situation. On the other hand, to avoid
the race between the patched code and post-patch/pre-unpatch callbacks when
run in parallel, we didn't remove  the post_patch_callback and
pre_unpatch_callback from stop_machine.

### DIFF
--- a/kmod/core/core.c
+++ b/kmod/core/core.c
@@ -377,16 +377,6 @@ static int kpatch_apply_patch(void *data)
 		return ret;
 	}
 
-	/* run any user-defined pre-patch callbacks */
-	list_for_each_entry(object, &kpmod->objects, list) {
-		ret = pre_patch_callback(object);
-		if (ret) {
-			pr_err("pre-patch callback failed!\n");
-			kpatch_state_finish(KPATCH_STATE_FAILURE);
-			goto err;
-		}
-	}
-
 	/* tentatively add the new funcs to the global func hash */
 	do_for_each_linked_func(kpmod, func) {
 		hash_add_rcu(kpatch_func_hash, &func->node, func->old_addr);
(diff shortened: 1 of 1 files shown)


zhenjiangma:
We have made modifications accoding to the suggestions based on the previous ones #1126.

euspectre:
I have tested the patch on my system - seems to work OK. I cannot see any major problems with the patch.

Please fix the minor issues I pointed out above and, I think, the patch will be acceptable after that.

euspectre:
Besides, you can close https://github.com/dynup/kpatch/pull/1126, I think, as the discussion has moved here.

You can remove the previous version of the fix from https://github.com/zhenjiangma/kpatch:master (using 'git rebase -i HEAD^', for example) and force-push the new fix there. This way, this pull request #1134 will be reused.

zhenjiangma:
> Besides, you can close #1126, I think, as the discussion has moved here.
> 
> You can remove the previous version of the fix from https://github.com/zhenjiangma/kpatch:master (using 'git rebase -i HEAD^', for example) and force-push the new fix there. This way, this pull request #1134 will be reused.

Thanks for your suggestions, I've fixed the minor issues according to your suggestions. 
I‘ve tried my best to learn 'git rebase -i HEAD^' to reuse the pull request #1134, but it didn't seem to work. And finally, I made a new commit b311112 base on last commit using 'git commit --amend' to reuse this request.

euspectre (approved):
Sorry for the delay.

The patch looks good to me now. The tests haven't shown any problems yet.

zhenjiangma:
> Sorry for the delay.
> 
> The patch looks good to me now. The tests haven't shown any problems yet.

You are welcome, I'm very glad that I can contribute to the community.

joe-lawrence:
Thanks @zhenjiangma and @euspectre, the changes look reasonable and thank you for testing.

PR #1136, opened 2020-08-27
lookup: table->exp_nr should be effective number of exported symbols

chengmingzhou:
Or for_each_exp_symbol may trigger SIGSEGV fault.

Reported-by: Jiachen Zhang <zhangjiachen.jaycee@bytedance.com>
Tested-by: Jiachen Zhang <zhangjiachen.jaycee@bytedance.com>

### DIFF
--- a/kpatch-build/lookup.c
+++ b/kpatch-build/lookup.c
@@ -311,7 +311,7 @@ static void symtab_read(struct lookup_table *table, char *path)
 static void symvers_read(struct lookup_table *table, char *path)
 {
 	FILE *file;
-	int i, column, mod_column = 0;
+	int i = 0, column, mod_column = 0, max_exp_nr = 0;
 	char line[4096];
 	char *tmp, *objname, *symname;
 
@@ -319,7 +319,7 @@ static void symvers_read(struct lookup_table *table, char *path)
 		ERROR("fopen");
 
 	while (fgets(line, 4096, file)) {
-		table->exp_nr++;
+		max_exp_nr++;
 
 		if (mod_column)
 			continue;
@@ -332,17 +332,17 @@ static void symvers_read(struct lookup_table *table, char *path)
 		}
 	}
 
-	if (table->exp_nr && !mod_column)
+	if (max_exp_nr && !mod_column)
 		ERROR("Module.symvers: invalid format");
 
-	table->exp_syms = malloc(table->exp_nr * sizeof(*table->exp_syms));
+	table->exp_syms = malloc(max_exp_nr * sizeof(*table->exp_syms));
 	if (!table->exp_syms)
 		ERROR("malloc table.exp_syms");
 	memset(table->exp_syms, 0,
-	       table->exp_nr * sizeof(*table->exp_syms));
+	       max_exp_nr * sizeof(*table->exp_syms));
 
 	rewind(file);
-	for (i = 0; fgets(line, 4096, file); i++) {
+	while(fgets(line, 4096, file)) {
 		char *name = NULL, *mod = NULL;
 
 		for (column = 1, tmp = line; (tmp = strchr(tmp, '\t')); column++) {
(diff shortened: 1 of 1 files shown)


kpatch-build/lookup.c
```diff
@@ -363,8 +364,9 @@ static void symvers_read(struct lookup_table *table, char *path)
 		objname = make_modname(mod);
 
 		table->exp_syms[i].name = symname;
-		table->exp_syms[i].objname = objname;
+		table->exp_syms[i++].objname = objname;
 	}
+	table->exp_nr = i;
```
julien-thierry, line 369: Looking at the loop starting at line 321, I don't see what this changes. The value of table->exp_nr should be the same as before since both loops perform the same number of iterations.

Can you provide more detail on how to trigger the issue you are observing?

sm00th: AFAICT if `if (!name || !mod)` holds we might have a situation where there are some NULL symbols in table->exp_syms and none of the users of for_each_exp_symbol check for non-NULL causing a SIGSEGV.

This will fix the issue but maybe we shouldn't over-allocate either.

chengmingzhou: Yes, previous code will leave some NULL symbols in table->exp_syms, so users of for_each_exp_symbol will trigger SIGSEGV when we test.
This patch will only inc i for each non-NULL symbol, so table->exp_nr will be the correct number.
And we can also avoid over-allocate in the first place.
I think both way is correct.

julien-thierry: Ah, you're right, I missed that.
In which case that that happen? Blank line?
How come the issue was not hit before?

chengmingzhou: @julien-thierry Well, I actually don't know much about the inside logic or in which case the condition "if (!name || !mod)" will be true. Maybe in most cases for_each_exp_symbol break out before reach NULL symbols ?

jpoimboe: @chenIs I don't ever remember seeing NULL symbols, which kernel version had that?  Or if the problem is just theoretical, we could just do a fatal ERROR on a NULL symbol.

Otherwise, if it's a real problem, the patch looks ok, though the earlier uses of table->exp_nr in that function should just be replaced with a local variable.  We don't need to worry about overallocation.

jpoimboe: @chengmingzhou, sorry, github wouldn't autocomplete your name for some reason, and I entered in the wrong name twice.  I finally got it right :-)

chengmingzhou: @jpoimboe Hi, we encountered this problem when we make a livepatch hotpatch for mm/mempolicy.c of kernel 4.14 using a bugfix source patch from our inside team.
I don't know how NULL symbols came out, create-diff-object just crash with SIGSEGV, so I use gdb to find this bug.
But today I pulled a new 4.14 kernel and kpatch-build again, this SIGSEGV didn't show up...
If NULL symbols should not be there theoretically, I think a fatal ERROR is ok too :-)
Thanks!

chengmingzhou: @jpoimboe Hi, I replace the use of table->exp_nr with a local variable in that function.  Are there plans for this PR to be merged?

Thanks.

jpoimboe: @chengmingzhou If we don't expect NULL symbols to be possible, I think I'd rather go with an ERROR when a NULL symbol is encountered.

joe-lawrence: @zhouchengming1 : any chance of attaching the symvers file that causes the crash?  Also, a gdb backtrace would be helpful (even in the commit msg) to better understand what conditions may lead to the crash.

sm00th (changes requested):
I think the conclusion was we'd rather ERROR out on NULL symbols, so I am explicitly marking this as 'changes requested'.

PR #1138, opened 2020-09-08
Bump to v0.9.2

joe-lawrence:
Interesting changes since v0.9.1:

- Integration test support for rhel-7.8, rhel-7.9, rhel-8.1, centos-8
- Better support for gcc child functions
- Batch jump label errors to report all instances
- Dynrela code cleanup
- Remove .klp.arch and add support for jump labels in v5.8+ kernels
- Mark ignored sections earlier to support functions missing ftrace hook
- Minor README.md improvements
- Add ppc64le mcount support to patched functions
- Show additional stalled process information in kpatch script
- Increased shellcheck coverage and fixes
- ppc64le plugin fixes for gcc v10
- Ignore __UNIQUE_ID_ symbol from tristate config objects
- Don't clear dmesg during integration tests
- Detect and report MODVERSIONS symbol version CRC changes

### DIFF
--- a/contrib/kpatch.spec
+++ b/contrib/kpatch.spec
@@ -6,7 +6,7 @@
 
 Name: kpatch
 Summary: Dynamic kernel patching
-Version: 0.9.1
+Version: 0.9.2
 License: GPLv2
 Group: System Environment/Kernel
 URL: http://github.com/dynup/kpatch
@@ -104,6 +104,22 @@ rm -rf %{buildroot}
 %{_mandir}/man1/kpatch-build.1*
 
 %changelog
+* Tue Sep 8 2020 Joe Lawrence <joe.lawrence@redhat.com> - 0.9.2
+- Integration test support for rhel-{7.8,7.9,8.1,8.2}, centos-8
+- Better support for gcc child functions
+- Batch jump label errors to report all instances
+- Dynrela code cleanup
+- Remove .klp.arch and add support for jump labels in v5.8+ kernels
+- Mark ignored sections earlier to support functions missing ftrace hook
+- Minor README.md improvements
+- Add ppc64le mcount support to patched functions
+- Show additional stalled process information in kpatch script
+- Increased shellcheck coverage and fixes
+- ppc64le plugin fixes for gcc v10
+- Ignore __UNIQUE_ID_ symbol from tristate config objects
+- Don't clear dmesg during integration tests
+- Detect and report MODVERSIONS symbol version CRC changes
+
 * Wed Mar 11 2020 Yannick Cote <ycote@redhat.com> - 0.9.1
 - Handle ppc64le toc with only constants
 - Don't strip callback section symbols
--- a/kpatch/kpatch
+++ b/kpatch/kpatch
@@ -25,7 +25,7 @@
 
 INSTALLDIR=/var/lib/kpatch
 SCRIPTDIR="$(readlink -f "$(dirname "$(type -p "$0")")")"
-VERSION="0.9.1"
+VERSION="0.9.2"
 POST_ENABLE_WAIT=15	# seconds
 POST_SIGNAL_WAIT=60	# seconds
 


joe-lawrence:
This release is to coincide with imminent RHEL-7.9 GA release.

Should we use v0.9.2, bump to v0.10.0 or stick to single digits and wait for v1.0.0?

sm00th:
Didn't we decide to do annotated tags? I'd expect a tag commit as well in this pull-request if we did.

joe-lawrence:
> Didn't we decide to do annotated tags? I'd expect a tag commit as well in this pull-request if we did.

Yep that is accurate. I am following the instructions listed in our internal **RHEL GA release actions** doc > **Create a kpatch github release**.  The instructions imply that these changes are a separate step from the signed annotation tag.  My thought was for this PR to be reviewed, merged and then immediately after do the signed annotation tag.  AFAIK tags, even annotated ones, are not pull request review-able are they?

sm00th:
Looks like they are not.  For some reason I thought they would behave just like normal commits, but I can't remember how I came to this conclusion. Sorry for the noise.

sm00th (approved):

joe-lawrence:
> Looks like they are not. For some reason I thought they would behave just like normal commits, but I can't remember how I came to this conclusion. Sorry for the noise.

I had to double check, too.  It seems intuitive that they should be review-able, but there must be some git internal data structure reasoning behind it.

disaster123:
I can't compile the current master branch on debian while using (-Werror=maybe-uninitialized): 

```
create-diff-object.c: In function 'kpatch_process_special_sections':
create-diff-object.c:2215:41: error: 'key' may be used uninitialized in this function [-Werror=maybe-uninitialized]
       code->sym->name, code->addend, key->sym->name);
                                         ^~
create-diff-object.c:2138:22: note: 'key' was declared here
  struct rela *code, *key, *rela;
                      ^~~
In file included from kpatch-elf.h:26,
                 from create-diff-object.c:53:
log.h:20:3: error: 'code' may be used uninitialized in this function [-Werror=maybe-uninitialized]
   ^~~~~~
create-diff-object.c:2138:15: note: 'code' was declared here
  struct rela *code, *key, *rela;
               ^~~~
```

disaster123:
If this is expected it could be easily fixed by:
```
diff --git a/kpatch-build/Makefile b/kpatch-build/Makefile
index b5b2ffd..2837ba8 100644
--- a/kpatch-build/Makefile
+++ b/kpatch-build/Makefile
@@ -1,7 +1,7 @@
 include ../Makefile.inc
 
 CFLAGS += -MMD -MP -I../kmod/patch -Iinsn -Wall -Wsign-compare \
-         -Wconversion -Wno-sign-conversion -g -Werror
+         -Wconversion -Wno-sign-conversion -g -Werror -Wno-maybe-uninitialized
 LDLIBS = -lelf
 
 TARGETS = create-diff-object create-klp-module create-kpatch-module
```

joe-lawrence:
@disaster123 it seems that you're building with -O2 or higher?  That was the only way I could reproduce these compiler complaints.  Anyway we should go ahead and clean them up, can you verify that ea6b01241ff755c931 (" create-diff-object: cleanup maybe-uninitialized compiler complaints") fixes the build for you?

disaster123:
Yes this fixes it. I've not specific O3 anyhwere but maybe  debian buster defaults to it.

joe-lawrence:
> Yes this fixes it. No i'm not using O3 - debian uses those flags:
> 
> ```
> gcc -g -O2 -fdebug-prefix-map=/build/kpatch=. -fstack-protector-strong -Wformat -Werror=format-security -MMD -MP -I../kmod/patch -Iinsn -Wall -Wsign-compare -Wconversion -Wno-sign-conversion -g -Werror -Wdate-time -D_FORTIFY_SOURCE=2
> ```

Thanks for the report and test.  We might as well include this in v0.9.2 so Debian can build the tagged release, so for this one I'll wait on PR #1143.

@sm00th : Maybe travis should also run some -O2 builds to find these sort of things earlier. What do you think?

joe-lawrence:
Updated to include 25f12681fa16 ("create-diff-object: cleanup maybe-uninitialized compiler complaints")

joe-lawrence:
Updated to add rhel-8.2 to changelog entry for integration test support

joe-lawrence:
One more time to update the specfile, too.

PR #1143, opened 2020-09-09
create-diff-object: cleanup maybe-uninitialized compiler complaints

joe-lawrence:
User disaster123 reports the following build errors:

  create-diff-object.c: In function 'kpatch_process_special_sections':
  create-diff-object.c:2215:41: error: 'key' may be used uninitialized in this function [-Werror=maybe-uninitialized]
         code->sym->name, code->addend, key->sym->name);
                                           ^~
  create-diff-object.c:2138:22: note: 'key' was declared here
    struct rela *code, *key, *rela;
                        ^~~
  In file included from kpatch-elf.h:26,
                   from create-diff-object.c:53:
  log.h:20:3: error: 'code' may be used uninitialized in this function [-Werror=maybe-uninitialized]
     ^~~~~~
  create-diff-object.c:2138:15: note: 'code' was declared here
    struct rela *code, *key, *rela;
                 ^~~~

These are reproducible when building with 9.3.1 and 8.3.1 when building
with optimization level > 2 ( CFLAGS=-O2 make ).  Fix them by
initializing the reported variables to NULL and verifying that they are
infact non-NULL after processing the __jump_table.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2135,7 +2135,7 @@ static bool should_keep_jump_label(struct lookup_table *lookup,
 				   unsigned int group_size,
 				   int *jump_labels_found)
 {
-	struct rela *code, *key, *rela;
+	struct rela *code = NULL, *key = NULL, *rela;
 	bool tracepoint = false, dynamic_debug = false;
 	struct lookup_result symbol;
 	int i = 0;
@@ -2156,7 +2156,7 @@ static bool should_keep_jump_label(struct lookup_table *lookup,
 		}
 	}
 
-	if (i != 3)
+	if (i != 3 || !key || !code)
 		ERROR("BUG: __jump_table has an unexpected format");
 
 	if (!strncmp(key->sym->name, "__tracepoint_", 13))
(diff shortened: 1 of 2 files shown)


sm00th (approved):
Added -O2/-O3 builds to travis' build matrix.

kpatch-build/create-diff-object.c
```diff
@@ -2156,7 +2156,7 @@ static bool should_keep_jump_label(struct lookup_table *lookup,
 		}
 	}
 
-	if (i != 3)
+	if (i != 3 || !key || !code)
```
kamalesh-babulal, line 2159: It's redundant to check for `key` and `code` value, when checking for `i != 3`. Which implicitly covers the NULL check of both variables.

kamalesh-babulal (approved):

kpatch-build/create-diff-object.c
joe-lawrence: Right, we can deduce that from the loop code, but apparently the compiler can't for some reason.  /shrugs

joe-lawrence:
> Added -O2/-O3 builds to travis' build matrix.

Very nice, thanks!  BTW, how did you add that commit to the pull request?  Pretty neat that you can push additional commits to a branch from my repository.

sm00th:
> BTW, how did you add that commit to the pull request?

When you create a pull-request there is this option 'Allow edits and access to secrets by maintainers' which now defaults to on. So anyone with merge rights to kpatch repo can push to pull-request branches of other people if they selected this.

kamalesh-babulal (approved):

PR #1144, opened 2020-09-25
kpatch-elf: pass new ELF output file mode to kpatch_write_output_elf()

joe-lawrence:
Currently all the callers of kpatch_write_output_elf() are creating
.o object files or .ko kernel modules.  Neither of these filetypes are
executable on their own, so enhance kpatch_write_output_elf() to accept
file creation flags and update its callers to pass 0664 to match
the expected permissions.

### DIFF
--- a/kpatch-build/kpatch-elf.c
+++ b/kpatch-build/kpatch-elf.c
@@ -759,7 +759,8 @@ void kpatch_rebuild_rela_section_data(struct section *sec)
 		ERROR("size mismatch in rebuilt rela section");
 }
 
-void kpatch_write_output_elf(struct kpatch_elf *kelf, Elf *elf, char *outfile)
+void kpatch_write_output_elf(struct kpatch_elf *kelf, Elf *elf, char *outfile,
+			     mode_t mode)
 {
 	int fd;
 	struct section *sec;
@@ -770,8 +771,7 @@ void kpatch_write_output_elf(struct kpatch_elf *kelf, Elf *elf, char *outfile)
 	Elf_Data *data;
 	GElf_Shdr sh;
 
-	/* TODO make this argv */
-	fd = creat(outfile, 0777);
+	fd = creat(outfile, mode);
 	if (fd == -1)
 		ERROR("creat");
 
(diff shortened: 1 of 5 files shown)


kamalesh-babulal (approved):

sm00th (approved):

kpatch-build/kpatch-elf.c
```diff
@@ -759,7 +759,8 @@ void kpatch_rebuild_rela_section_data(struct section *sec)
 		ERROR("size mismatch in rebuilt rela section");
 }
 
-void kpatch_write_output_elf(struct kpatch_elf *kelf, Elf *elf, char *outfile)
+void kpatch_write_output_elf(struct kpatch_elf *kelf, Elf *elf, char *outfile,
+			     int flags)
```
julien-thierry, line 763: So, this is going to be nitpicking a bit. But the `flags` argument will be only used for the `mode` argument of `creat`.
This might get confusing when you see that `open` (of which creat is just an specialised alias) also has a `flags` argument whose value don't match the `mode` argument.

So, could this be called `file_mode` or something mode indicative?

kpatch-build/kpatch-elf.c
```diff
@@ -770,8 +771,7 @@ void kpatch_write_output_elf(struct kpatch_elf *kelf, Elf *elf, char *outfile)
 	Elf_Data *data;
 	GElf_Shdr sh;
 
-	/* TODO make this argv */
```
julien-thierry, line 773: I'm not sure I understand this TODO. Make what argv? What this to actually remind to make the change you just did? (pass the permission as argument)

kpatch-build/kpatch-elf.c
joe-lawrence: Sure, let's rename it to be clear.  And good catch on flags vs mode, I copied the wrong name from the manpage.

kpatch-build/kpatch-elf.c
joe-lawrence: I'm not really sure what Jessica meant when adding that TODO comment.  My initial guess was this change, but maybe she thought about plumbing it all the way out to the program `argv[]`?  That seems like overkill to me, so I landed on this PR.

julien-thierry: Yes, there's no need to make it into a command line option. So, I guess removing that TODO makes sense.

joe-lawrence:
v2: rename "flags" to "mode" and update data type to mode_t

yhcote (approved):
with those changes, looks good to me.

joe-lawrence:
v3: it's friday and Joe forgot to actually `git add` the v2 changes  :facepalm:

julien-thierry (approved):

PR #1145, opened 2020-10-27
Linux 5.9 fixes

wcwxyz:
These two patches fixes kpatch for a custom 5.9 kernel build.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -296,7 +296,11 @@ static bool is_dynamic_debug_symbol(struct symbol *sym)
 {
 	if (sym->type == STT_OBJECT && !strcmp(sym->sec->name, "__verbose"))
 		return true;
+	if (sym->type == STT_OBJECT && !strcmp(sym->sec->name, "__dyndbg"))
+		return true;
 	if (sym->type == STT_SECTION && !strcmp(sym->name, "__verbose"))
+		return true;
+	if (sym->type == STT_SECTION && !strcmp(sym->name, "__dyndbg"))
 		return true;
 	return false;
 }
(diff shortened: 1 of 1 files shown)


joe-lawrence (commented):
Hi @wcwxyz ,

The create-diff-object changes look good, but I'm not so sure about the kpatch-build modification.  Just a heads up, we're planning on backporting the klp.arch patchset to RHEL-8.4, for which we might apply a kpatch-build [change like this](https://github.com/dynup/kpatch/commit/8fd4bfbd6c5379d1c16e45b68e89496e0f01f4cb) to handle RHEL kernels that do or do not support klp.arch relocations.  I think that change would break your custom kernels that are pre-v5.8, right?

Also, I assume your custom kernels are from the upstream tree, built with a RHEL-style configuration (and naming convention).  If so, I *think* that will play nice with the kernel version macros in `kmod/patch/`, but not the kpatch-build's `kernel_is_rhel()` function...  so long story short, maybe we should start with something like [this](https://github.com/dynup/kpatch/commit/8fd4bfbd6c5379d1c16e45b68e89496e0f01f4cb), but then add a kpatch-build `--no-klp-arch` command line option to force `use_klp_arch()`'s status?

wcwxyz:
Yes, I'm testing upstream kernel and I use RHEL-style configuration and kernel naming convention.

The change you mentioned https://github.com/dynup/kpatch/commit/8fd4bfbd6c5379d1c16e45b68e89496e0f01f4cb looks good to me. That works for both RHEL-8.4+ and my custom kernel.

I'll drop the second patch.

Thanks for the review.

joe-lawrence:
Thanks for the fix, @wcwxyz

PR #1147, opened 2020-11-17
kpatch: wait for module ref counts on unload

joe-lawrence:
There exists a very small timing window in which "kpatch unload" gets to
its "rmmod" step before the kpatch-patch module's reference count has
cleared and the "rmmod" fails.

This is only a transient problem, but we can adopt code from upstream
livepatch kselftests which wait for the module refcounts to settle
before moving onto "rmmod".

A small wrinkle is that this is not supported by the older kpatch.ko
core.  The price for circumventing the activeness safety check via
KPATCH_FORCE_UNSAFE is that it must leave the kpatch patch modules in
place (see e1890e627a9b ("prevent rmmod of forced modules")).

### DIFF
--- a/kpatch/kpatch
+++ b/kpatch/kpatch
@@ -28,6 +28,7 @@ SCRIPTDIR="$(readlink -f "$(dirname "$(type -p "$0")")")"
 VERSION="0.9.2"
 POST_ENABLE_WAIT=15	# seconds
 POST_SIGNAL_WAIT=60	# seconds
+MODULE_REF_WAIT=15	# seconds
 
 # How many times to try loading the patch if activeness safety check fails.
 MAX_LOAD_ATTEMPTS=5
@@ -123,6 +124,10 @@ find_core_module() {
 	[[ -f "$COREMOD" ]] && return
 
 	return 1
+}
+
+kpatch_core_loaded() {
+	grep -q -e "T kpatch_register" /proc/kallsyms
 }
 
 core_loaded () {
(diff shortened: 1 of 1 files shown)


joe-lawrence:
Tested by livepatching the livepatch core (hold my beer! :beers:) to introduce a noticeable delay between disable and refcnt=0.
[slow-refcount.patch.txt](https://github.com/dynup/kpatch/files/5554948/slow-refcount.patch.txt)

kpatch/kpatch
```diff
@@ -381,7 +411,13 @@ disable_patch_strict () {
...
+	local modname="$1"
+
+	if ! wait_for_zero_module_ref_count "$modname"; then
+		die "failed to unload module $modname (refcnt)"
+	fi
+
+	echo "unloading patch module: $modname"
 	# ignore any error here because rmmod can fail if the module used
 	# KPATCH_FORCE_UNSAFE.
 	rmmod "$1" 2> /dev/null || return 0
```
sm00th, line 423: `$1` should be `$modname` here for consistency.

sm00th (approved):

julien-thierry (approved):

kamalesh-babulal (approved):

joe-lawrence:
v2: use `$modname` vs. `$1` per @sm00th's suggestion

yhcote (approved):

kpatch/kpatch
```diff
@@ -265,6 +270,31 @@ wait_for_patch_transition() {
...
+	local i=0
+
+	# We can't rely on a zero refcount with kpatch.ko as it
+	# implements KPATCH_FORCE_UNSAFE with an additional reference on
+	# kpatch-patch modules to avoid potential crashes.
+	kpatch_core_loaded && return 0
+
+	module_ref_count "$modname" || return 0
+
+	echo "waiting (up to $MODULE_REF_WAIT seconds) for module refcount..."
```
jpoimboe, line 289: Does this usually happen instantaneously?  Just wondering if it would be better to print a message only if it needs to sleep.

joe-lawrence: The way `wait_for_zero_module_ref_count()` is implemented is that it first checks the refcount and 99.99% of the time it will exit early, only printing the message if it needs to sleep.  Suggestions on renaming `module_ref_count()` if that might make the pattern clearer?

jpoimboe: Oh, I just glossed over line 287.  Looks good then!

jpoimboe (approved):

Issue #1148, opened 2020-12-04
5.10 breaks kpatch module build


sm00th:
commit [6212804f2d78 kbuild: do not create built-in objects for external module builds](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=6212804f2d78e86f15dba5b46a4065cbf1403cde) breaks kpatch module builds. This happens because kbuild no longer builds built-in targets for external modules one of which is [extra-y](https://elixir.bootlin.com/linux/latest/source/scripts/Makefile.build#L79) used by kpatch's makefile to build kpatch.lds linker script.

sm00th:
[always-y](https://elixir.bootlin.com/linux/latest/source/scripts/Makefile.build#L490) seems to be a target we can use instead.

PR #1149, opened 2020-12-07
kmod/patch: add kpatch.lds as a dependency for $(KPATCH_NAME).o

sm00th:
Starting with v5.10 kbuild no longer builds built-in targets for
external modules (including extra-y). Further it wasn't guaranteed that
extra-y targets were going to be built before linking.
Do a proper thing and add kpatch.lds as a dependency for $(KPATCH_NAME).o.

Fixes: #1148

### DIFF
--- a/kmod/patch/Makefile
+++ b/kmod/patch/Makefile
@@ -12,7 +12,7 @@ endif
 
 obj-m += $(KPATCH_NAME).o
 ldflags-y += -T $(src)/kpatch.lds
-extra-y := kpatch.lds
+targets += kpatch.lds
 
 $(KPATCH_NAME)-objs += patch-hook.o output.o
 
@@ -20,6 +20,8 @@ all: $(KPATCH_NAME).ko
 
 $(KPATCH_NAME).ko:
 	$(KPATCH_MAKE)
+
+$(obj)/$(KPATCH_NAME).o: $(src)/kpatch.lds
 
 patch-hook.o: patch-hook.c kpatch-patch-hook.c livepatch-patch-hook.c
 	$(KPATCH_MAKE) patch-hook.o


joe-lawrence:
How does this square with this [recent thread](https://lore.kernel.org/linux-kbuild/20201103054425.59251-1-chao.wang@ucloud.cn) on the kbuild list?

[ cc: @wcwxyz ]

sm00th:
Oh, didn't occur to me to check that.
`always-y` did work for me on fedora32 and v5.10-r6. I wonder why ldflags-y is wrong and what are the alternatives. Will look into this.

kamalesh-babulal:
I tried the patch on Power (Ubuntu 20.04), GCC-9.3, v5.10-rc7 with the `meminfo.patch` and fails with 
```
make[1]: Entering directory '/tmp/linux-5.10-rc7'
  LD [M]  /root/.kpatch/tmp/patch/livepatch-meminfo.o
ld: cannot open linker script file /root/.kpatch/tmp/patch/kpatch.lds: No such file or directory
```

wcwxyz:
I'm sorry. I've been busy working on something else. Didn't got the time to follow this issue.

Is `always-y` working for you? Last time I checked, that don't work for me.  Anyway I'm CCing you guys in that kbuild thread. Feel free to chime in.

sm00th:
You are right, `always-y` doesn't work. Looks like my testing environment wasn't as prestine as it should have been. Changed it to be a dependency instead.

wcwxyz:
This fix looks good to me. Thanks.

kamalesh-babulal:
Thanks, this fixes the build failure.

kamalesh-babulal (approved):

joe-lawrence (approved):
Approved, but feels like something may still be missing as I cannot figure out how to express the kpatch.lds.S -> kpatch.lds dependency such that modifying kpatch.lds.S results in rebuilding the .lds and relinking $(obj)/$(KPATCH_NAME).o, etc.

jpoimboe:
> Approved, but feels like something may still be missing as I cannot figure out how to express the kpatch.lds.S -> kpatch.lds dependency such that modifying kpatch.lds.S results in rebuilding the .lds and relinking $(obj)/$(KPATCH_NAME).o, etc.

Doesn't that already happen?  The kernel scripts/Makefile.build has a cpp_lds_S rule which does that.

sm00th:
> Doesn't that already happen? The kernel scripts/Makefile.build has a cpp_lds_S rule which does that.

Yes, scripts/Makefile.build is the source of lds.S -> lds rule however we kinda circumvent it with this weird dependency overrides:
all -> ${KPATCH_NAME}.ko -> nothing

To be honest I am not sure what this achieves. I think we can just leave it at
```
all:
  $(KPATCH_MAKE)
```

and let kbuild handle the rest.

I also don't see not rebuilding on lds.S change as a big problem as this will never happen during normal kpatch use, I can only see it happening in development scenarios.

joe-lawrence:
> I also don't see not rebuilding on lds.S change as a big problem as this will never happen during normal kpatch use, I can only see it happening in development scenarios.

Agreed.  I missed that rule override and it was just a rabbit hole I found myself down, so I thought I'd share the confusion.  :D

sm00th:
Added kpatch.lds to 'targets' as per Masahiro's suggestion.

joe-lawrence:
:+1: already approved, feel free to merge if you're happy with the revision, thanks

Issue #1151, opened 2021-01-13
Document the behaviour of kpatch "signal" option


julien-thierry:
The `kpatch signal` operation relies on the presence of the livepatch signal sysfs interface (/sys/kernel/livepatch/$module/signal).

When that interface is not present, kpatch signal silently does nothing.

Since the interface is removed from linux-5.1 onwards, it would be good to display a message to let the user know that no signal will be sent and update the man page regarding this option.

PR #1152, opened 2021-01-14
kpatch: Inform user when signal subcommand is unnecessary

julien-thierry:
From Linux-5.1 onward, the sysfs interface to send signals to
processes stalling livepatch transitions is replace by a peridical
signal sent by the kernel.

On such systems, the kpatch signal subcommand does nothing. Just
let the user know they don't need to worry about it.

Fixes: #1151 

### DIFF
--- a/kpatch/kpatch
+++ b/kpatch/kpatch
@@ -237,6 +237,9 @@ signal_stalled_processes() {
 	if [[ -e "/sys/kernel/livepatch/$module/signal" ]] ; then
 		echo "signaling stalled process(es):"
 		echo 1 > "/sys/kernel/livepatch/$module/signal"
+        else
+            warn "Livepatch process signaling is performed automatically on your system."
+            warn "Skipping manual process signaling."
 	fi
 }
 
--- a/man/kpatch.1
+++ b/man/kpatch.1
@@ -36,7 +36,10 @@ list
        list installed patch modules
 
 signal
-       signal/poke any process stalling the current patch transition
+       signal/poke any process stalling the current patch transition.
+       This is only useful on systems that have the sysfs livepatch signal interface.
+       On other systems, the signaling should be done automatically by the OS and
+       this subcommand is a no-op.
 
 version
        display the kpatch version


sm00th (approved):

kpatch/kpatch
```diff
@@ -237,6 +237,9 @@ signal_stalled_processes() {
 	if [[ -e "/sys/kernel/livepatch/$module/signal" ]] ; then
 		echo "signaling stalled process(es):"
 		echo 1 > "/sys/kernel/livepatch/$module/signal"
+        else
+            warn "Livepatch process signaling is performed automatically on your system."
+            warn "Skipping manual process signaling."
```
yhcote, line 242: I would probably change the term 'manual' by 'explicit' or something, but it doesn't really bother me.

yhcote (approved):
Small nit, looks good otherwise.

jpoimboe (approved):

man/kpatch.1
```diff
@@ -36,7 +36,10 @@ list
        list installed patch modules
 
 signal
-       signal/poke any process stalling the current patch transition
+       signal/poke any process stalling the current patch transition.
+       This is only usefull on systems that have the sysfs livepatch signal interface.
```
joe-lawrence, line 40: spelling nit: s/usefull/useful

kpatch/kpatch
joe-lawrence: I think we've used "manual" in most places when talking about this feature, so my vote would be to keep it as is.

julien-thierry: I also have a slight preference for "manual". So unless there are more votes for another option I'll leave it as is. :)

man/kpatch.1
julien-thierry: Ah, good catch, I'll update that.
Thanks.

kpatch/kpatch
yhcote: With @joe-lawrence 's explanation, it makes sense to me. I haven't thought of the other references.

joe-lawrence:
👍 looks good, thanks

joe-lawrence (approved):

kamalesh-babulal (approved):

PR #1154, opened 2021-01-20
kpatch-gcc: powerpc - skip vdso files

kamalesh-babulal:
Starting v5.11-rc, kpatch-build fails on powerpc with the error:

ERROR: invalid ancestor arch/powerpc/kernel/vdso64/vdso64.so.dbg for arch/powerpc/kernel/vdso64/vgettimeofday.o

the upstream commit ab037dd87a2f(powerpc/vdso: Switch VDSO to generic
implementation) introduced this breakage, lets skip vdso files. They are
not compatible with kpatch.


sm00th (approved):

jpoimboe (approved):

joe-lawrence:
Thanks, @kamalesh-babulal

Issue #1153, opened 2021-01-19
CONFIG_DEBUG_INFO_BTF and targets with many sections


joe-lawrence:
Yulia hit this kpatch-build problem with simple kpatch-patch test and an internal kernel build:
```
...
  BTF     .btf.vmlinux.bin.o
scripts/link-vmlinux.sh: line 127: 718345 Segmentation fault      (core dumped) LLVM_OBJCOPY=${OBJCOPY} ${PAHOLE} -J ${1}
objcopy: --change-section-vma .BTF=0x0000000000000000 never used
objcopy: --change-section-lma .BTF=0x0000000000000000 never used
objcopy: error: the input file '.btf.vmlinux.bin' is empty
Failed to generate BTF for vmlinux
Try to disable CONFIG_DEBUG_INFO_BTF
make: *** [Makefile:1050: vmlinux] Error 1
```
Jiri Olsa noted the large number of vmlinux ELF sections and that pahole would need to be enhanced to support that:
```
AFAICS the generated vmlinux file has many sections:

  Number of section headers:         0 (101720)
  Section header string table index: 65535 (101719)

and that requires special handling when reading the elf object,
which pahole does not support at the moment
```
TODO:
- Which upstream kernel versions are affected?
- Which distro kernels 
- Did we only encounter this because we /just/ spilled over the non-extended limit?

euspectre:
I've also got such error when building a patch for Virtuozzo kernel 4.18.0-240.1.1.vz8.5.4. 
The kernel is based on 4.18.0-240.1.1.el8_3 from RHEL 8.3, so, I suspect the kernel from RHEL might be affected too.

```
  LD      vmlinux.o
  MODPOST vmlinux.o
  BTF     .btf.vmlinux.bin.o
scripts/link-vmlinux.sh: line 127: 177631 Segmentation fault      (core dumped) LLVM_OBJCOPY=${OBJCOPY} ${PAHOLE} -J ${1}
objcopy: --change-section-vma .BTF=0x0000000000000000 never used
objcopy: --change-section-lma .BTF=0x0000000000000000 never used
objcopy: error: the input file '.btf.vmlinux.bin' is empty
Failed to generate BTF for vmlinux
Try to disable CONFIG_DEBUG_INFO_BTF
make: *** [Makefile:1065: vmlinux] Error 1
```

joe-lawrence:
[...]

PR #1155, opened 2021-01-22
kpatch-build: workaround pahole and CONFIG_DEBUG_INFO_BTF

joe-lawrence:
kpatch-build requires gcc flags -f[function|data]-sections when building
original and patched targets.  These flags result in an ELF binary with
many sections, potentially requiring special extended ELF header
processing to parse correctly.

CONFIG_DEBUG_INFO_BTF invokes pahole as part of the kernel build and
unfortunately pahole cannot iterate through more than 65535 section
headers.  As result, the pahole program segfaults and fails the build
like so:

  ...
    BTF     .btf.vmlinux.bin.o
  scripts/link-vmlinux.sh: line 127: 718345 Segmentation fault      (core dumped) LLVM_OBJCOPY=${OBJCOPY} ${PAHOLE} -J ${1}
  objcopy: --change-section-vma .BTF=0x0000000000000000 never used
  objcopy: --change-section-lma .BTF=0x0000000000000000 never used
  objcopy: error: the input file '.btf.vmlinux.bin' is empty
  Failed to generate BTF for vmlinux
  Try to disable CONFIG_DEBUG_INFO_BTF
  make: *** [Makefile:1050: vmlinux] Error 1

Workaround this limitation by disabling CONFIG_DEBUG_INFO_BTF code in
scripts/vmlinux-link.sh during kpatch-build.  This leaves
CONFIG_DEBUG_INFO_BTF contingent kernel code in place, but skips the
problematic pahole .BTF typeinfo generation step (for which kpatch
doesn't care about anyway).

Fixes: #1153
Suggested-by: Josh Poimboeuf <jpoimboe@redhat.com>

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -148,6 +148,9 @@ cleanup() {
 	# restore original vmlinux if it was overwritten by sourcedir build
 	[[ -e "$TEMPDIR/vmlinux" ]] && mv -f "$TEMPDIR/vmlinux" "$SRCDIR/"
 
+	# restore original link-vmlinux.sh if we updated it for the build
+	[[ -e "$TEMPDIR/link-vmlinux.sh" ]] && mv -f "$TEMPDIR/link-vmlinux.sh" "$SRCDIR/scripts"
+
 	[[ "$DEBUG" -eq 0 ]] && rm -rf "$TEMPDIR"
 	rm -rf "$RPMTOPDIR"
 	unset KCFLAGS
@@ -777,6 +780,14 @@ grep -q "CONFIG_MODVERSIONS=y"  "$CONFIGFILE" && CONFIG_MODVERSIONS=1
 grep -q "CONFIG_DEBUG_INFO_SPLIT=y" "$CONFIGFILE" && die "kernel option 'CONFIG_DEBUG_INFO_SPLIT' not supported"
 grep -q "CONFIG_GCC_PLUGIN_LATENT_ENTROPY=y" "$CONFIGFILE" && die "kernel option 'CONFIG_GCC_PLUGIN_LATENT_ENTROPY' not supported"
 grep -q "CONFIG_GCC_PLUGIN_RANDSTRUCT=y" "$CONFIGFILE" && die "kernel option 'CONFIG_GCC_PLUGIN_RANDSTRUCT' not supported"
+
+# CONFIG_DEBUG_INFO_BTF invokes pahole, for which some versions don't
+# support extended ELF sections.  Disable the BTF typeinfo generation in
+# link-vmlinux.sh since kpatch doesn't care about that anyway.
+if grep -q "CONFIG_DEBUG_INFO_BTF=y" "$CONFIGFILE" ; then
+	cp "$SRCDIR/scripts/link-vmlinux.sh" "$TEMPDIR/"
+	sed -i 's/CONFIG_DEBUG_INFO_BTF/DISABLED_FOR_KPATCH_BUILD/g' "$SRCDIR"/scripts/link-vmlinux.sh || die
+fi
 
 echo "Testing patch file(s)"
 cd "$SRCDIR" || die
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -640,7 +640,7 @@ if [[ -n "$USERSRCDIR" ]]; then
 
 	# save original vmlinux before it gets overwritten by sourcedir build
 	if [[ -z "$OOT_MODULE" ]] && [[ "$VMLINUX" -ef "$SRCDIR"/vmlinux ]]; then
-		cp -f "$VMLINUX" "$TEMPDIR/vmlinux"
+		cp -f "$VMLINUX" "$TEMPDIR/vmlinux" || die
 		VMLINUX="$TEMPDIR/vmlinux"
 	fi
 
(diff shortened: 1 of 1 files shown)


kpatch-build/kpatch-build
```diff
@@ -778,6 +787,16 @@ grep -q "CONFIG_DEBUG_INFO_SPLIT=y" "$CONFIGFILE" && die "kernel option 'CONFIG_
 grep -q "CONFIG_GCC_PLUGIN_LATENT_ENTROPY=y" "$CONFIGFILE" && die "kernel option 'CONFIG_GCC_PLUGIN_LATENT_ENTROPY' not supported"
 grep -q "CONFIG_GCC_PLUGIN_RANDSTRUCT=y" "$CONFIGFILE" && die "kernel option 'CONFIG_GCC_PLUGIN_RANDSTRUCT' not supported"
 
+# older versions of pahole don't support extended sections,
+# disable BTF typeinfo generation step in link-vmlinux.sh if needed
+if grep -q "CONFIG_DEBUG_INFO_BTF=y" "$CONFIGFILE" && \
+   ! pahole_btf_version_check; then
+	echo "Backing up scripts/link-vmlinux.s to $TEMPDIR/"
+	cp "$SRCDIR/scripts/link-vmlinux.sh" "$TEMPDIR/"
+	echo "Disabling CONFIG_DEBUG_INFO_BTF in link-vmlinux.s for kpatch-build (incompatible pahole version)"
```
jpoimboe, line 796: I'm thinking these printed messages may be a bit technical and confusing.  How about:

echo "Detected broken version of 'pahole'.  Implementing workaround.'

Though actually maybe we just shouldn't print anything?  kpatch-build tends to print only high-level status messages, and I'm thinking this is really just a minute detail that can probably be ignored.

joe-lawrence:
- v2: removed technical msgs as suggested by @jpoimboe

julien-thierry (approved):

kamalesh-babulal (approved):

jpoimboe (approved):

yhcote (approved):
looks good to me

kpatch-build/kpatch-build
```diff
@@ -215,6 +218,12 @@ find_core_symvers() {
 	[[ -e "$SYMVERSFILE" ]]
 }
 
+# CONFIG_DEBUG_INFO_BTF invokes pahole, which needs to have extended ELF
+# header support (version TBD).
+pahole_btf_version_check() {
+	[[ ! "$(pahole --version 2>/dev/null)" < "vTBD" ]]
+}
```
joe-lawrence, line 225: I think pahole will fix this in its next v1.20 release, but does it make sense to universally disable this if CONFIG_DEBUG_INFO_BTF is set?  This would save everyone a little bit of kpatch-build time and I could take out this pahole version check (which may or may not reflect reality if a fix is backported to something like dwarves-1.19-2.el8).

jpoimboe: Yeah, maybe so

joe-lawrence:
v3: I was hoping to see pahole v1.20 tagged so I could update, but on second thought I just removed the check to simplify everything.  Upstream pahole patch links added to the commit msg for future reference.

kpatch-build/kpatch-build
```diff
@@ -778,6 +781,14 @@ grep -q "CONFIG_DEBUG_INFO_SPLIT=y" "$CONFIGFILE" && die "kernel option 'CONFIG_
 grep -q "CONFIG_GCC_PLUGIN_LATENT_ENTROPY=y" "$CONFIGFILE" && die "kernel option 'CONFIG_GCC_PLUGIN_LATENT_ENTROPY' not supported"
 grep -q "CONFIG_GCC_PLUGIN_RANDSTRUCT=y" "$CONFIGFILE" && die "kernel option 'CONFIG_GCC_PLUGIN_RANDSTRUCT' not supported"
 
+# CONFIG_DEBUG_INFO_BTF invokes pahole, for which some versions don't
+# support extended ELF sections.  Disable the BTF typeinfo generation in
+# link-vmlinux.sh since kpatch doesn't care about that anyway.
+if grep -q "CONFIG_DEBUG_INFO_BTF=y" "$CONFIGFILE" && \
+	cp "$SRCDIR/scripts/link-vmlinux.sh" "$TEMPDIR/"
+	sed -i 's/CONFIG_DEBUG_INFO_BTF/DISABLED_FOR_KPATCH_BUILD/g' "$SRCDIR"/scripts/link-vmlinux.sh
+fi
```
jpoimboe, line 790: Does "if" work without "then"?

joe-lawrence: Ahh, you spotted that just before I forced pushed the update :D

jpoimboe: :-)

kpatch-build/kpatch-build
```diff
@@ -778,6 +781,14 @@ grep -q "CONFIG_DEBUG_INFO_SPLIT=y" "$CONFIGFILE" && die "kernel option 'CONFIG_
 grep -q "CONFIG_GCC_PLUGIN_LATENT_ENTROPY=y" "$CONFIGFILE" && die "kernel option 'CONFIG_GCC_PLUGIN_LATENT_ENTROPY' not supported"
 grep -q "CONFIG_GCC_PLUGIN_RANDSTRUCT=y" "$CONFIGFILE" && die "kernel option 'CONFIG_GCC_PLUGIN_RANDSTRUCT' not supported"
 
+# CONFIG_DEBUG_INFO_BTF invokes pahole, for which some versions don't
+# support extended ELF sections.  Disable the BTF typeinfo generation in
+# link-vmlinux.sh since kpatch doesn't care about that anyway.
+if grep -q "CONFIG_DEBUG_INFO_BTF=y" "$CONFIGFILE" ; then
+	cp "$SRCDIR/scripts/link-vmlinux.sh" "$TEMPDIR/"
```
jpoimboe, line 788: Just to be extra robust, use "cp -f"?

kpatch-build/kpatch-build
```diff
@@ -778,6 +781,14 @@ grep -q "CONFIG_DEBUG_INFO_SPLIT=y" "$CONFIGFILE" && die "kernel option 'CONFIG_
...
+	sed -i 's/CONFIG_DEBUG_INFO_BTF/DISABLED_FOR_KPATCH_BUILD/g' "$SRCDIR"/scripts/link-vmlinux.sh
```
jpoimboe, line 789: Probably a good idea to add "|| die" at the end

joe-lawrence:
v4: force copy or die

jpoimboe:
How about "sed or die" as well?

joe-lawrence:
v5: Everything must die!

jpoimboe (approved):

sm00th (approved):

kamalesh-babulal (approved):

PR #1158, opened 2021-02-25
kpatch: cleanup install directory

sm00th:
On module installation kpatch utility creates a directory with kernel version as a name in /var/lib/kpatch which is never removed. To address this check if any files are left in this directory after each 'uninstall' call and if not - remove it.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1930108

### DIFF
--- a/kpatch/kpatch
+++ b/kpatch/kpatch
@@ -575,10 +575,12 @@ case "$1" in
 	fi
 
 	[[ ! -e "$MODULE" ]] && die "$PATCH is not installed for kernel $KVER"
-	
 
 	echo "uninstalling $PATCH ($KVER)"
 	rm -f "$MODULE" || die "failed to uninstall module $PATCH"
+	rmdir --ignore-fail-on-non-empty "$INSTALLDIR/$KVER" || die "failed to remove directory $INSTALLDIR/$KVER"
+	rmdir --ignore-fail-on-non-empty "$INSTALLDIR" || die "failed to remove directory $INSTALLDIR"
+
 	;;
 
 "list")


kpatch/kpatch
```diff
@@ -575,10 +575,14 @@ case "$1" in
 	fi
 
 	[[ ! -e "$MODULE" ]] && die "$PATCH is not installed for kernel $KVER"
-	
 
 	echo "uninstalling $PATCH ($KVER)"
 	rm -f "$MODULE" || die "failed to uninstall module $PATCH"
+
+	if [ -z "$(ls -A "$INSTALLDIR/$KVER")" ]; then
```
julien-thierry, line 582: you could just use
`rmdir --ignore-fail-on-non-empty "$INSTALLDIR/$KVER"`

I should only remove the directory when it's empty.

julien-thierry (commented):
Nit:
"creates a directory with kernel version as a nami"

s/nami/name

kpatch/kpatch
sm00th: Oh nice, didn't know this existed. This will make it much cleaner.

julien-thierry (approved):

kpatch/kpatch
```diff
@@ -575,10 +575,11 @@ case "$1" in
 	fi
 
 	[[ ! -e "$MODULE" ]] && die "$PATCH is not installed for kernel $KVER"
-	
 
 	echo "uninstalling $PATCH ($KVER)"
 	rm -f "$MODULE" || die "failed to uninstall module $PATCH"
+	rmdir --ignore-fail-on-non-empty "$INSTALLDIR/$KVER" || die "failed to remove direcroty $INSTALLDIR/$KVER"
```
jpoimboe, line 581: typo: "direcroty"

jpoimboe: Should it also remove $INSTALLDIR?

jpoimboe (approved):

kamalesh-babulal (approved):

PR #1159, opened 2021-02-26
kpatch-build: workaround pahole and CONFIG_DEBUG_INFO_BTF_MODULES

liu-song-6:
With CONFIG_DEBUG_INFO_BTF_MODULES, the kernel build process adds BTF to
each in-tree modules. However, this process is broken with kpatch, with
error message like:

  Failed to parse base BTF 'vmlinux': -4001

Unblock build with CONFIG_DEBUG_INFO_BTF_MODULES with similar workaround
as the one for CONFIG_DEBUG_INFO_BTF.

Signed-off-by: Song Liu <songliubraving@fb.com>

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -151,6 +151,9 @@ cleanup() {
 	# restore original link-vmlinux.sh if we updated it for the build
 	[[ -e "$TEMPDIR/link-vmlinux.sh" ]] && mv -f "$TEMPDIR/link-vmlinux.sh" "$SRCDIR/scripts"
 
+	# restore original Makefile.modfinal if we updated it for the build
+	[[ -e "$TEMPDIR/Makefile.modfinal" ]] && mv -f "$TEMPDIR/Makefile.modfinal" "$SRCDIR/scripts"
+
 	[[ "$DEBUG" -eq 0 ]] && rm -rf "$TEMPDIR"
 	rm -rf "$RPMTOPDIR"
 	unset KCFLAGS
@@ -809,10 +812,14 @@ grep -q "CONFIG_GCC_PLUGIN_RANDSTRUCT=y" "$CONFIGFILE" && die "kernel option 'CO
 
 # CONFIG_DEBUG_INFO_BTF invokes pahole, for which some versions don't
 # support extended ELF sections.  Disable the BTF typeinfo generation in
-# link-vmlinux.sh since kpatch doesn't care about that anyway.
+# link-vmlinux.sh and Makefile.modfinal since kpatch doesn't care about
+# that anyway.
 if grep -q "CONFIG_DEBUG_INFO_BTF=y" "$CONFIGFILE" ; then
 	cp -f "$SRCDIR/scripts/link-vmlinux.sh" "$TEMPDIR/link-vmlinux.sh" || die
 	sed -i 's/CONFIG_DEBUG_INFO_BTF/DISABLED_FOR_KPATCH_BUILD/g' "$SRCDIR"/scripts/link-vmlinux.sh || die
+
+	cp -f "$SRCDIR/scripts/Makefile.modfinal" "$TEMPDIR/Makefile.modfinal" || die
+	sed -i 's/CONFIG_DEBUG_INFO_BTF_MODULES/DISABLED_FOR_KPATCH_BUILD/g' "$SRCDIR"/scripts/Makefile.modfinal || die
 fi
 
 if [[ "$CONFIG_CC_IS_CLANG" -eq 1 ]]; then


liu-song-6:
Actually, if the system has a newer pahole (supports -J), removing the DISABLED_FOR_KPATCH_BUILD workaround also fixes the issue. Maybe we can add a check for pahole version? Something like

    `pahole -J create-diff-object && HAS_PAHOLE=1`

kpatch-build/kpatch-build
```diff
@@ -451,6 +451,12 @@ find_kobj() {
 	done
 }
 
+check_pahole_supports_btf() {
+	PAHOLE_SUPPORTS_BTF=0
+	cp "$TOOLSDIR"/create-diff-object "$TEMPDIR"
+	pahole -J "$TEMPDIR"/create-diff-object && PAHOLE_SUPPORTS_BTF=1
```
sm00th, line 457: This check is not enough. pahole may support -J but still have an issue with big number of sections. See commit message for https://github.com/dynup/kpatch/commit/e25450f5a80349b88607634a80a3fa4e02823caa for more info.

Anyway kpatch builds do not benefit from this at all, so instead of complicated error-prone checks of pahole state we can always disable this.

kpatch-build/kpatch-build
```diff
@@ -807,14 +813,24 @@ grep -q "CONFIG_DEBUG_INFO_SPLIT=y" "$CONFIGFILE" && die "kernel option 'CONFIG_
...
+if [[ $(grep -q "CONFIG_DEBUG_INFO_BTF=y" "$CONFIGFILE") ]] && [[ "$PAHOLE_SUPPORTS_BTF" -eq 0 ]]; then
 	cp -f "$SRCDIR/scripts/link-vmlinux.sh" "$TEMPDIR/link-vmlinux.sh" || die
 	sed -i 's/CONFIG_DEBUG_INFO_BTF/DISABLED_FOR_KPATCH_BUILD/g' "$SRCDIR"/scripts/link-vmlinux.sh || die
 fi
 
+# CONFIG_DEBUG_INFO_BTF_MODULES invokes pahole for kernel modules, which
+# doesn't work with kpatch-cc.  Disable this command by replacing the line
+# with ": \".
+if [[ $(grep -q "CONFIG_DEBUG_INFO_BTF_MODULES=y" "$CONFIGFILE") ]] && [[ "$PAHOLE_SUPPORTS_BTF" -eq 0]]; then
+	cp -f "$SRCDIR/scripts/Makefile.modfinal" "$TEMPDIR/Makefile.modfinal" || die
```
sm00th, line 830: This is copied to tempdir but never restored.

kpatch-build/kpatch-build
```diff
@@ -807,14 +813,24 @@ grep -q "CONFIG_DEBUG_INFO_SPLIT=y" "$CONFIGFILE" && die "kernel option 'CONFIG_
...
+	sed -i 's/.*LLVM_OBJCOPY.*/\t\t:\t\\/g' "$SRCDIR"/scripts/Makefile.modfinal || die
```
sm00th, line 831: Instead of trying to fix this command which can change why not do the same thing as `CONFIG_DEBUG_INFO_BTF` code does and just disable everything?

`sed -i 's/CONFIG_DEBUG_INFO_BTF_MODULES/DISABLED_FOR_KPATCH_BUILD/g' "$SRCDIR"/scripts/Makefile.modfinal || die`

kpatch-build/kpatch-build
```diff
@@ -807,14 +813,24 @@ grep -q "CONFIG_DEBUG_INFO_SPLIT=y" "$CONFIGFILE" && die "kernel option 'CONFIG_
...
-if grep -q "CONFIG_DEBUG_INFO_BTF=y" "$CONFIGFILE" ; then
+if [[ $(grep -q "CONFIG_DEBUG_INFO_BTF=y" "$CONFIGFILE") ]] && [[ "$PAHOLE_SUPPORTS_BTF" -eq 0 ]]; then
 	cp -f "$SRCDIR/scripts/link-vmlinux.sh" "$TEMPDIR/link-vmlinux.sh" || die
 	sed -i 's/CONFIG_DEBUG_INFO_BTF/DISABLED_FOR_KPATCH_BUILD/g' "$SRCDIR"/scripts/link-vmlinux.sh || die
 fi
 
+# CONFIG_DEBUG_INFO_BTF_MODULES invokes pahole for kernel modules, which
+# doesn't work with kpatch-cc.  Disable this command by replacing the line
+# with ": \".
+if [[ $(grep -q "CONFIG_DEBUG_INFO_BTF_MODULES=y" "$CONFIGFILE") ]] && [[ "$PAHOLE_SUPPORTS_BTF" -eq 0]]; then
```
sm00th, line 829: Separate if is unnecessary. `CONFIG_DEBUG_INFO_BTF_MODULES` depends on `CONFIG_DEBUG_INFO_BTF` so we can disable module parts in the same if clause.

sm00th: `[[ "$PAHOLE_SUPPORTS_BTF" -eq 0]]`

Missing space afte `0` makes bash unhappy.

sm00th (changes requested):

kpatch-build/kpatch-build
liu-song-6: Fixed this part and dropped the second commit. Thanks!

kpatch-build/kpatch-build
```diff
@@ -815,6 +818,14 @@ if grep -q "CONFIG_DEBUG_INFO_BTF=y" "$CONFIGFILE" ; then
 	sed -i 's/CONFIG_DEBUG_INFO_BTF/DISABLED_FOR_KPATCH_BUILD/g' "$SRCDIR"/scripts/link-vmlinux.sh || die
 fi
 
+# CONFIG_DEBUG_INFO_BTF_MODULES invokes pahole for kernel modules, which
+# doesn't work with kpatch-cc.  Disable this command by replacing the line
+# with ": \".
+if grep -q "CONFIG_DEBUG_INFO_BTF_MODULES=y" "$CONFIGFILE" ; then
```
sm00th, line 824: Let's get rid of this if and move the below code into `CONFIG_DEBUG_INFO_BTF` if. It will work just as well but will make things a bit cleaner.

Otherwise looks good, thank you.

jpoimboe (approved):

kamalesh-babulal (approved):

sm00th (approved):

sm00th:
Thank you.

PR #1161, opened 2021-03-03
kpatch-build: check existance of Makefile.modfinal

sm00th:
It turns out there is a stretch of time in kernel's history when CONFIG_DEBUG_INFO_BTF was already added, but Makefile.modfinal wasn't split off yet. To address those we need to either check the file's existance or, as @liu-song-6 suggested initially, check config for CONFIG_DEBUG_INFO_BTF_MODULES=y.


liu-song-6 (approved):
Looks good!

joe-lawrence (approved):

kamalesh-babulal (approved):

Issue #1160, opened 2021-03-01
Issue with .rela.text.unlikely.* ? 


liu-song-6:
Hi folks, 

I am testing kpatch on a very simple patch:

```
---
 drivers/md/md.h | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/drivers/md/md.h b/drivers/md/md.h
index bcbba1b5ec4a7..eda6f344bc91b 100644
--- a/drivers/md/md.h
+++ b/drivers/md/md.h
@@ -551,7 +551,7 @@ extern void mddev_unlock(struct mddev *mddev);

 static inline void md_sync_acct(struct block_device *bdev, unsigned long nr_sectors)
 {
-       atomic_add(nr_sectors, &bdev->bd_disk->sync_io);
+       atomic_add(nr_sectors + 1, &bdev->bd_disk->sync_io);
 }

 static inline void md_sync_acct_bio(struct bio *bio, unsigned long nr_sectors)
```
which hit errors like:
```
raid1.o: changed section .rela.text.unlikely.sync_request_write not selected for inclusion
raid10.o: changed section .rela.text.unlikely.raid10d not selected for inclusion
```

Is this expected or a bug? I am using gcc 9.1.1, and I have `CONFIG_MD_RAID1=m`

Thanks in advance.

sm00th:
Probably a bug. I failed to reproduce it, however. Can you invoke `kpatch-build` with `-d` option and then upload `${HOME}/.kpatch/tmp/orig/raid1.o` and `${HOME}/.kpatch/tmp/patched/raid1.o` somewhere?

jpoimboe:
Does this fix it? https://github.com/jpoimboe/kpatch/commit/066c4baeafe048d4db53d62007ee573a1a8089ac

liu-song-6:
> Does this fix it? [jpoimboe@066c4ba](https://github.com/jpoimboe/kpatch/commit/066c4baeafe048d4db53d62007ee573a1a8089ac)

It doesn't seem to work.

liu-song-6:
> Probably a bug. I failed to reproduce it however. Can you invoke `kpatch-build` with `-d` option and then upload `${HOME}/.kpatch/tmp/orig/drivers/md/raid1.o` and `${HOME}/.kpatch/tmp/patched/drivers/md/raid1.o` somewhere?

I added these files and build.log to https://github.com/liu-song-6/kpatch/tree/debug_1160 .

[...]

PR #1162, opened 2021-03-05
support for .cold functions with no id suffix

sm00th:
create-build-diff expects .cold functions to be suffixed by an id, which is not always the case. Drop the trailing '.' when searching for cold functions.

Fixes: #1160


jpoimboe (approved):

kamalesh-babulal (approved):

liu-song-6:
Thanks for the fix!

PR #1168, opened 2021-03-22
github actions: unit tests

sm00th:
Add unit tests through github actions. This is exactly the same as our travis tests but using github actions instead.

I noticed that latest pull requests didn't have a notice from travis also when logging into travis-ci.org there is the following message:

```
 Please be aware travis-ci.org will be shutting down in several weeks, with all accounts migrating to travis-ci.com. Please stay tuned here for more information. 
```

travis-ci.com used to be their commercial offering, so there are some changes going on and I am not exactly sure what those mean for us. This commits enable github "actions" which are the same thing but first party. We can run both in parallel and retire travis at some point since it is third-party and is more likely to have issues.

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -1,0 +1,26 @@
+name: build and unit tests
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        cflags: [ "", "-O2", "-O3" ]
+    runs-on: ubuntu-latest
+    env:
+      CFLAGS: ${{ matrix.cflags }}
+    steps:
+    - uses: actions/checkout@v2
+    - name: dependencies
+      run: sudo apt-get install -y libelf-dev linux-headers-$(uname -r) shellcheck elfutils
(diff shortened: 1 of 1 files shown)


joe-lawrence (commented):
The travis file also runs `make check` and `make install`.  Do we want those here, too?

sm00th:
> The travis file also runs make check and make install. Do we want those here, too?

Thanks. Added those as well.

sm00th:
btw you can look at how it works in  my personal tree: https://github.com/sm00th/kpatch/actions

kamalesh-babulal:
The workflow is pretty neat :)

kamalesh-babulal (approved):

joe-lawrence (approved):

jpoimboe:
So I suppose we need to remove .travis.yml at the same time?

sm00th:
> So I suppose we need to remove .travis.yml at the same time?

They can run in parallel so we don't need to. I wanted to keep both for some time to make sure everything works.

PR #1172, opened 2021-04-13
A couple of fixes for the out-of-range relocation check

jpoimboe:
A couple of minor fixes for the out-of-range relocation check.   These were found by inspection, so I don't have any unit tests.


sm00th (approved):
I think I encountered issues with symbols not being at the start of the function while trying to run create-diff-object against linux.o. Maybe I'll be able to get some unit tests (and more fixes) on the topic when I get back to that.

joe-lawrence (approved):

yhcote (approved):

kamalesh-babulal (approved):

PR #1183, opened 2021-05-15
kpatch-build: enable option -R|--replace to build replace klp

liu-song-6:
Since 5.1 kernel, klp_patch supports a "replace" option, which does atomic
replace of cumulative patches. Enable building such patch with an option.

### DIFF
--- a/doc/patch-author-guide.md
+++ b/doc/patch-author-guide.md
@@ -39,6 +39,13 @@ recommended that when patching a system which has already been patched, the
 second patch should be a cumulative upgrade which is a superset of the first
 patch.
 
+Since upstream kernel 5.1, livepatch supports a "replace" flag to help the
+management of cumulative patches. With the flag set, the kernel will load
+the cumulative patch and unload all existing patches in one transition.
+kpatch-build enables the replace flag by default. If replace behavior is
+not desired, the user can disable it with -R|--non-replace.
+
+
 Data structure changes
 ----------------------
 
--- a/kmod/patch/livepatch-patch-hook.c
+++ b/kmod/patch/livepatch-patch-hook.c
@@ -74,6 +74,18 @@
 # define HAVE_SIMPLE_ENABLE
 #endif
 
+#ifdef RHEL_RELEASE_CODE
+# if RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(8, 2)
+#  define HAVE_KLP_REPLACE
+# endif
+#elif LINUX_VERSION_CODE >= KERNEL_VERSION(5, 1, 0)
+# define HAVE_KLP_REPLACE
+#endif
+
+#ifndef KLP_REPLACE_ENABLE
+#define KLP_REPLACE_ENABLE true
+#endif
+
 /*
  * There are quite a few similar structures at play in this file:
  * - livepatch.h structs prefixed with klp_*
@@ -385,6 +397,9 @@ static int __init patch_init(void)
 		goto out;
 	lpatch->mod = THIS_MODULE;
 	lpatch->objs = lobjects;
+#ifdef HAVE_KLP_REPLACE
+	lpatch->replace = KLP_REPLACE_ENABLE;
+#endif
 #if defined(__powerpc64__) && defined(HAVE_IMMEDIATE)
 	lpatch->immediate = true;
 #endif
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -55,6 +55,7 @@ DEBUG_KCFLAGS=""
 declare -a PATCH_LIST
 APPLIED_PATCHES=0
 OOT_MODULE=
+KLP_REPLACE=1
 
 warn() {
 	echo "ERROR: $1" >&2
@@ -215,6 +216,15 @@ use_klp_arch()
 		! rhel_kernel_version_gte 4.18.0-284.el8
 	else
 		! kernel_version_gte 5.8.0
+	fi
+}
+
+support_klp_replace()
+{
+	if kernel_is_rhel; then
+		rhel_kernel_version_gte 4.18.0-193.el8
+	else
+		kernel_version_gte 5.1.0
 	fi
 }
 
@@ -514,12 +524,13 @@ usage() {
 	echo "		                        (can be specified multiple times)" >&2
 	echo "		-e, --oot-module        Enable patching out-of-tree module," >&2
 	echo "		                        specify current version of module" >&2
+	echo "		-R, --non-replace       Disable replace patch (replace is on by default)" >&2
 	echo "		--skip-cleanup          Skip post-build cleanup" >&2
 	echo "		--skip-compiler-check   Skip compiler version matching check" >&2
 	echo "		                        (not recommended)" >&2
 }
 
-options="$(getopt -o ha:r:s:c:v:j:t:n:o:de: -l "help,archversion:,sourcerpm:,sourcedir:,config:,vmlinux:,jobs:,target:,name:,output:,oot-module:,debug,skip-gcc-check,skip-compiler-check,skip-cleanup" -- "$@")" || die "getopt failed"
+options="$(getopt -o ha:r:s:c:v:j:t:n:o:de:R -l "help,archversion:,sourcerpm:,sourcedir:,config:,vmlinux:,jobs:,target:,name:,output:,oot-module:,debug,skip-gcc-check,skip-compiler-check,skip-cleanup,non-replace" -- "$@")" || die "getopt failed"
 
 eval set -- "$options"
 
@@ -581,6 +592,9 @@ while [[ $# -gt 0 ]]; do
 		[[ ! -f "$2" ]] && die "out-of-tree module '$2' not found"
 		OOT_MODULE="$(readlink -f "$2")"
 		shift
+		;;
+	-R|--non-replace)
+		KLP_REPLACE=0
 		;;
 	--skip-cleanup)
 		echo "Skipping cleanup"
@@ -821,6 +835,12 @@ if grep -q "CONFIG_LIVEPATCH=y" "$CONFIGFILE" && (kernel_is_rhel || kernel_versi
 		KPATCH_LDFLAGS="--unique=.parainstructions --unique=.altinstructions"
 		CDO_FLAGS="--klp-arch"
 	fi
+
+	if [[  "$KLP_REPLACE" -eq 1 ]] ; then
+		support_klp_replace || die "The kernel doesn't support klp replace"
+	else
+		export KBUILD_CFLAGS_MODULE="$KBUILD_CFLAGS_MODULE -DKLP_REPLACE_ENABLE=false"
+	fi
 else
 	# No support for livepatch in the kernel. Kpatch core module is needed.
 
@@ -828,6 +848,9 @@ else
 	# sections.  Use with caution!
 	echo "WARNING: Use of kpatch core module (kpatch.ko) is deprecated!  There may be bugs!" >&2
 
+	if [[ "$KLP_REPLACE" -eq 1 ]] ; then
+		die "kpatch core module (kpatch.ko) does not support replace, please add -R|--non-replace"
+	fi
 	find_core_symvers || die "unable to find Module.symvers for kpatch core module"
 	KBUILD_EXTRA_SYMBOLS="$SYMVERSFILE"
 fi
--- a/test/integration/kpatch-test
+++ b/test/integration/kpatch-test
@@ -50,6 +50,7 @@ ERROR=0
 LOG=test.log
 DYNDEBUG_CONTROL=/sys/kernel/debug/dynamic_debug/control
 DYNDEBUG_ENABLED=1
+ARCHVERSION="$(uname -r)"
 rm -f ./*.log
 
 PATCHDIR="${PATCHDIR:-$PWD}"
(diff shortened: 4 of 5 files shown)


kmod/patch/livepatch-patch-hook.c
```diff
@@ -74,6 +74,11 @@
 # define HAVE_SIMPLE_ENABLE
 #endif
 
+/* TODO: check RHEL_RELEASE_CODE for KLP_REPLACE*/
```
joe-lawrence, line 77: Atomic replace was backported to rhel-8's kernel-4.18.0-147.3.el8.  It was never backported to rhel-7.

kpatch-build/kpatch-build
```diff
@@ -884,6 +907,10 @@ remove_patches
 
 cp -LR "$DATADIR/patch" "$TEMPDIR" || die
 
+if [[ "$KLP_REPLACE" -eq 1 ]] ; then
+   sed -i 's/lpatch->replace = false/lpatch->replace = true/g' "$TEMPDIR/patch/livepatch-patch-hook.c"
+fi
+
```
joe-lawrence, line 913: I don't know how I feel about this approach -- in a way, atomic replace and other features that directly affect the `klp_patch` object are more than just build options, they may change the way the kpatch is coded.  I wonder if we should try to move this in the kpatch itself somehow?  This might be too difficult to implement (how to check that said feature is supported, etc.) but I ask in case others have any good ideas.

kmod/patch/livepatch-patch-hook.c
liu-song-6: I think we can do `rhel_kernel_version_gte 4.18.0-147.el8` in kpatch-build. But I am not sure how to check it here. Is `RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(8, 0)` sufficient?

kpatch-build/kpatch-build
liu-song-6: With current klp_patch->replace design, which unloads **all** loaded patch with the new patch, I think the decision to use replace or not is done at kpatch-build time. Therefore, we don't need much flexibility in kpatch. (I may have missed some meaningful scenarios, though. )

kmod/patch/livepatch-patch-hook.c
joe-lawrence: For the .c code, we can use `RHEL_RELEASE_CODE(8, 2)` since atomic replace support was only added in 8.2 GA kernels onward.

For the `kpatch-build` code, we could have used `rhel_kernel_version_gte 4.18.0-147.3.el8` (the trailing `.3` would have been important), but if the kmod code can only check RHEL release numbers, then let's match these two checks.  So then it would be `rhel_kernel_version_gte 4.18.0-193`  as -193 was the 8.2 GA kernel.

kpatch-build/kpatch-build
joe-lawrence: Right, the question I have is whether `--replace` might tempt somebody to use the same input patch without the flag.  If `klp_patch->replace` behavior is encoded in the kpatch itself, then the kpatch author be responsible for doing so.   That is to say, the patch author wrote it in the style for atomic replace and expects it to be built that way.

joe-lawrence: Song - for your kpatch infrastructure, would you be intending to build _all_ kpatches with atomic replace?  I ask, as maybe another way to look at this is to default all kpatches as `klp->replace` where available.  Just curious what your use case is.

liu-song-6: In our use cases, I would like to build all kpatches with atomic replace. Each host will have at most one kpatch load at anytime. I think this would give fewer confusions.

liu-song-6:
@joe-lawrence How does this version look to you? Do we need more work (in kpatch?) to move forward here? Thanks!

jpoimboe:
Since cumulative updates are the recommended and standard practice, I'd vote to make `klp->replace = true` the default.  The only problem is that this breaks backwards compatibility, so we should be sure to at least bump the minor version for the next release.  Or maybe it's time for 1.0!

Also, instead of the in-place 'sed' of the .c file, how about putting the default boolean value in a macro:
  ```
#ifndef KLP_REPLACE_ENABLE
#define KLP_REPLACE_ENABLE true
#endif
```
Then kpatch-build can change the default if needed by passing the -D option to gcc.

Also, I think some update to the patch author guide document is needed, with respect to callbacks.  There's a subtle difference between replace and non-replace.  With replace, a cumulative update will call the prior patches' pre-unpatch callback which could have an unintended effect if the patch author isn't careful.  Maybe we should consider adding a new pre-replace callback.

**Comment by liu-song-6** [comment]
> Since cumulative updates are the recommended and standard practice, I'd vote to make `klp->replace = true` the default. The only problem is that this breaks backwards compatibility, so we should be sure to at least bump the minor version for the next release. Or maybe it's time for 1.0!

If we use replace as default, shall we add -N|--non-replace? Any better name for this option? Personally, I still think non-replace should be the default (considering the pre-unpatch pre-replace issue). 

> 
> Also, instead of the in-place 'sed' of the .c file, how about putting the default boolean value in a macro:
> 
> ```
> #ifndef KLP_REPLACE_ENABLE
> #define KLP_REPLACE_ENABLE true
> #endif
> ```

Yes, this is better. Let me fix it. 

> 
> Then kpatch-build can change the default if needed by passing the -D option to gcc.
> 
> Also, I think some update to the patch author guide document is needed, with respect to callbacks. There's a subtle difference between replace and non-replace. With replace, a cumulative update will call the prior patch's pre-unpatch callback which could have an unintended effect if the patch author isn't careful. Maybe we should consider adding a new pre-replace callback.

Let me also take a look at documentation and pre-replace. Thanks!

**Comment by joe-lawrence** [comment]
> Since cumulative updates are the recommended and standard practice, I'd vote to make `klp->replace = true` the default. The only problem is that this breaks backwards compatibility, so we should be sure to at least bump the minor version for the next release. Or maybe it's time for 1.0!

Version 1.0 sounds good to me :partying_face:   Unless you want to reserve that for eventual kpatch.ko support removal.

> 
> Also, instead of the in-place 'sed' of the .c file, how about putting the default boolean value in a macro:
> 
> ```
> #ifndef KLP_REPLACE_ENABLE
> #define KLP_REPLACE_ENABLE true
> #endif
> ```
> 
> Then kpatch-build can change the default if needed by passing the -D option to gcc.
> 

Agreed.  A ` -N|--non-replace` option works for me.  If we decide to default to using atomic replace, we should call that out loudly in the release notes.  

> Also, I think some update to the patch author guide document is needed, with respect to callbacks. There's a subtle difference between replace and non-replace. With replace, a cumulative update will call the prior patch's pre-unpatch callback which could have an unintended effect if the patch author isn't careful. Maybe we should consider adding a new pre-replace callback.

Yeah, it could definitely use an update here.  The kernel docs on [cumulative-patches](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/livepatch/cumulative-patches.rst#n61) provide a good summary of callback behavior under atomic replace.  I think the major difference is that that only the new cumulative callbacks will ever be called.  See the atomic replace test in the [test-callbacks](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/testing/selftests/livepatch/test-callbacks.sh#n504) selftest for example.  So I'm not sure if a pre-replace callback is needed, unless we wanted to introduce a change in this behavior upstream (no old callbacks run, except this special one).  I think the new cumulative patch's pre-patch callback should fit most use cases, or am I missing something?

**Comment by jpoimboe** [comment]
> > Since cumulative updates are the recommended and standard practice, I'd vote to make `klp->replace = true` the default. The only problem is that this breaks backwards compatibility, so we should be sure to at least bump the minor version for the next release. Or maybe it's time for 1.0!
> 
> Version 1.0 sounds good to me partying_face Unless you want to reserve that for eventual kpatch.ko support removal.

kpatch.ko is already deprecated and, IIRC, broken with new kernels anyway, so its eventual removal may be a non-event.

> Agreed. A ` -N|--non-replace` option works for me. If we decide to default to using atomic replace, we should call that out loudly in the release notes.

Sounds good to me.  Another possibility would be a positively-worded option (`-i|--incremental`?) but either way works for me.

> 
> > Also, I think some update to the patch author guide document is needed, with respect to callbacks. There's a subtle difference between replace and non-replace. With replace, a cumulative update will call the prior patch's pre-unpatch callback which could have an unintended effect if the patch author isn't careful. Maybe we should consider adding a new pre-replace callback.
> 
> Yeah, it could definitely use an update here. The kernel docs on [cumulative-patches](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/livepatch/cumulative-patches.rst#n61) provide a good summary of callback behavior under atomic replace. I think the major difference is that that only the new cumulative callbacks will ever be called. See the atomic replace test in the [test-callbacks](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/testing/selftests/livepatch/test-callbacks.sh#n504) selftest for example. So I'm not sure if a pre-replace callback is needed, unless we wanted to introduce a change in this behavior upstream (no old callbacks run, except this special one). I think the new cumulative patch's pre-patch callback should fit most use cases, or am I missing something?

I guess I was wrong in thinking that the previous patch's pre-unpatch callback gets called.  So forget my pre-replace callback idea.

**Comment by liu-song-6** [comment]
I updated the PR with
-   -N|--replace option
-   use -DKLP_REPLACE_ENABLE instead of sed
-   a brief introduction of the option in patch-author-guide.md

Please let me know your thoughts on this. Thanks!

**Comment by liu-song-6** [comment]
@joe-lawrence @jpoimboe Could you please share your feedbacks on this version? Thanks!

**Review by jpoimboe — COMMENTED** [review]
Thinking about it again, the '-N' bothers me, since it stands for "non" it can mean disabling anything.  Can we change to '-R'?

I'd prefer not to print the "Disabling replace for KLP" message as we try to keep the kpatch-build output down to a bare minimum.

Otherwise it's looking good to me.

File: kpatch-build/kpatch-build [hunk 3]
```diff
@@ -821,13 +835,22 @@ if grep -q "CONFIG_LIVEPATCH=y" "$CONFIGFILE" && (kernel_is_rhel || kernel_versi
...
+	fi
 else
 	# No support for livepatch in the kernel. Kpatch core module is needed.
 
 	# There may be ordering bugs, with jump labels and other special
 	# sections.  Use with caution!
 	echo "WARNING: Use of kpatch core module (kpatch.ko) is deprecated!  There may be bugs!" >&2
 
+	if [[ "$KLP_REPLACE" -eq 1 ]] ; then
+		die "kpatch core module (kpatch.ko) does not support replace, please add -N|--non-replace"
```
jpoimboe, line 852: "-N" -> "-R"

jpoimboe (approved):

joe-lawrence:
Sorry on getting back to this one @liu-song-6 , swamped with work items at the moment.  I just want to pop in to say that changing default kpatch-build to atomic replace breaks the integration tests when they run for all the older (unsupported) kernels.

We could 1) only use atomic replace default if the target kernel supports it.  I think that would be confusing.  or 2 ) copy that `support_klp_replace()` kernel version check into the `kpatch-test` script so that the integration tests build with `-R` for all those older kernels.  This _should_ work since I don't believe any of our integration test patches are sensitive to atomic replace semantics.

Also, a really small nit:  I think your editor automatically clears any trailing whitespace from the files that were edited.  Not a big deal, but the diff stat would be cleaner without all those edits.

kpatch-build/kpatch-build
```diff
@@ -514,12 +524,13 @@ usage() {
 	echo "		                        (can be specified multiple times)" >&2
 	echo "		-e, --oot-module        Enable patching out-of-tree module," >&2
 	echo "		                        specify current version of module" >&2
+	echo "		-R, --non-replace       Disable replace patch (replace is on by default)" >&2
```
joe-lawrence, line 527: This needs a similar update to `man/kpatch-build.1`

joe-lawrence (commented):
@liu-song-6 : looks mostly good, I think updating the manpage and the integration test script should be all that's left

doc/patch-author-guide.md
```diff
@@ -231,7 +238,7 @@ index 9bff3b28c357..18374fd35bd9 100644
 @@ -1751,6 +1751,8 @@ struct task_struct *fork_idle(int cpu)
  	return task;
  }
- 
+
```
joe-lawrence, line 241: These whitespace changes are still scattered throughout the PR.  Is there a way to omit them so they don't add unrelated changes?

test/integration/kpatch-test
```diff
@@ -308,13 +310,30 @@ new_dmesg() {
...
 fi
 
+if [[ "${ID}" == "rhel" ]]; then
+	if [[ "${VERSION_ID%%.*}" == "9" || ( "${VERSION_ID%%.*}" == "8" && "${VERSION_ID##*.}" -ge "2" ) ]]; then
+		SUPPORT_KLP_REPLACE=1
+	fi
+elif kernel_version_gte 5.1.0 ; then
+	SUPPORT_KLP_REPLACE=1
+fi
+
```
joe-lawrence, line 331: Was there a reason not to copy kpatch-build's `support_klp_replace()` and associated functions?  If it's only a few more net LOC, but if it matches exactly, hopefully it will be easier to keep synchronized when any updates are needed.

doc/patch-author-guide.md
liu-song-6: It requires some manual work. But yeah, I can get rid of them.

test/integration/kpatch-test
liu-song-6: Will fix.

joe-lawrence:
Looks good, @liu-song-6 , thanks for working on this one!

joe-lawrence (approved):

sm00th (approved):

PR #1186, opened 2021-06-03
create-diff-object: Check that the section has a secsym

gwelymernans:
A STT_SECTION symbol is not needed if if it is not used as a relocation
target. Therefore, a section, in this case a debug section, may not have
a secsym associated with it.

Signed-off-by: Bill Wendling <morbo@google.com>

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2564,7 +2564,7 @@ static void kpatch_include_debug_sections(struct kpatch_elf *kelf)
 	list_for_each_entry(sec, &kelf->sections, list) {
 		if (is_debug_section(sec)) {
 			sec->include = 1;
-			if (!is_rela_section(sec))
+			if (!is_rela_section(sec) && sec->secsym)
 				sec->secsym->include = 1;
 		}
 	}


sm00th:
Hi, thanks for the patch. Is it something you found by code inspection or encountered in the wild? In latter case can you provide the object files and/or the patch that resulted in these? We try to add cases like these to our unit tests.

gwelymernans:
Encountered in the wild. I'll get you a `.o` file. You'll have to use clang's LLD for linking, because it omits the `STT_SECTION` symbol from the `.symtab`.

sm00th:
> Encountered in the wild. I'll get you a `.o` file. You'll have to use clang's LLD for linking, because it omits the `STT_SECTION` symbol from the `.symtab`.

Please get both the original and patched files (can be found in `~/.kpatch/tmp/orig` and `~/.kpatch/tmp/patched` directories). We won't be linking them, just running create-diff-object on them.

sm00th (approved):
The objectfiles/patch provided in #1193 are enough. There are more locations where we need to introduce a similar check, but this is good to go.

joe-lawrence (approved):
Since #1200 is already open, we'll handle all of them across two PRs.

PR #1189, opened 2021-06-07
kpatch: Sync signal subcmd usage output with manpage

theyoyojo:
Fixes: 1188

Signed-off-by: Joel Savitz <jsavitz@redhat.com>

### DIFF
--- a/kpatch/kpatch
+++ b/kpatch/kpatch
@@ -36,7 +36,7 @@ MAX_LOAD_ATTEMPTS=5
 RETRY_INTERVAL=2
 
 usage_cmd() {
-	printf '   %-20s\n      %s\n' "$1" "$2" >&2
+	printf '   %-20s\n%s\n' "$1" "$(fmt -w 80 <(echo "      $2"))" >&2
 }
 
 usage () {
@@ -57,7 +57,7 @@ usage () {
 	echo >&2
 	usage_cmd "list" "list installed patch modules"
 	echo >&2
-	usage_cmd "signal" "signal/poke any process stalling the current patch transition"
+	usage_cmd "signal" "signal/poke any process stalling the current patch transition. This is only useful on systems that have the sysfs livepatch signal interface. On other systems, the signaling should be done automatically by the OS and this subcommand is a no-op."
 	echo >&2
 	usage_cmd "version" "display the kpatch version"
 	exit 1


kpatch/kpatch
```diff
@@ -57,7 +57,10 @@ usage () {
 	echo >&2
 	usage_cmd "list" "list installed patch modules"
 	echo >&2
-	usage_cmd "signal" "signal/poke any process stalling the current patch transition"
+	usage_cmd "signal" "signal/poke any process stalling the current patch transition
+	This is only useful on systems that have the sysfs livepatch signal interface.
+	On other systems, the signaling should be done automatically by the OS and
+	this subcommand is a no-op."
 	echo >&2
```
joe-lawrence, line 64: nit: what does indentation turn out to look like when this is passed through `usage_cmd()`'s:
```
printf '   %-20s\n      %s\n' "$1" "$2" 
```

I can't think of a simple way to manage terminal widths, wordwrap and indentation, so maybe it would be better to just manually wrap/indent with the right number of spaces for 80 cols.   Something like:

```
        usage_cmd "signal" "signal/poke any process stalling the current patch transition
      This is only useful on systems that have the sysfs livepatch signal
      interface.  On other systems, the signaling should be done automatically
      by the OS and this subcommand is a no-op."
```
which hopefully renders like:
```
   ...
   info <module>       
      show information about a patch module

   list                
      list installed patch modules

   signal              
      signal/poke any process stalling the current patch transition
      This is only useful on systems that have the sysfs livepatch signal
      interface.  On other systems, the signaling should be done automatically
      by the OS and this subcommand is a no-op.
```

theyoyojo: yeah the new lines are very slightly more indented than the first line, I can add spaces manually to make them line up but the script looks like it was designed for short usage messages

joe-lawrence: I think the `fmt` command could potentially solve this problem...  If you give it a string like `"<indentation spaces> foo bar fritz..."` it will detect the leading indentation and then apply it as it word wraps the rest of the text.  It's part of the coreutils package, so it should be safe to assume that its installed.

theyoyojo: Alright I just pushed a potentially more future-proof fix

theyoyojo: output of `kpatch`:

```
[jsavitz@jsavitz kpatch]$ ./kpatch
usage: kpatch <command> [<args>]

Valid commands:
   install [-k|--kernel-version=<kernel version>] <module>
      install patch module to be loaded at boot
   uninstall [-k|--kernel-version=<kernel version>] <module>
      uninstall patch module

   load --all          
      load all installed patch modules into the running kernel
   load <module>       
      load patch module into the running kernel
   unload --all        
      unload all patch modules from the running kernel
   unload <module>     
      unload patch module from the running kernel

   info <module>       
      show information about a patch module

   list                
      list installed patch modules

   signal              
      signal/poke any process stalling the current patch transition. This is
      only useful on systems that have the sysfs livepatch signal interface. On
      other systems, the signaling should be done automatically by the OS
      and this subcommand is a no-op.

   version             
      display the kpatch version
```

joe-lawrence: Nice, that turned out a lot simpler than I initially expected.

This is splitting hairs, but we can preserve the old indentation formatting by moving the six spaces from the printf format string to just before `$2` (right now it's only indenting a single space).  I _really_ hope nobody is parsing the output of usage, but I rather not like to find out :)

joe-lawrence (approved):
:+1:  looks good, thanks!

sm00th (approved):

kamalesh-babulal (approved):

yhcote (approved):

PR #1190, opened 2021-06-07
kpatch: rmmod module of the same name before loading a module

liu-song-6:
With klp->replace, when a patch is replaced, it is no longer an active
patch, but the module is still loaded with a refcount of 0. If the user
tries to load the patch again, insmod will fail with EEXIT. To avoid
such errors, run a proactive rmmod before loading the module. This is a
no-op if the module is not loaded or is actually in use.

### DIFF
--- a/kpatch/kpatch
+++ b/kpatch/kpatch
@@ -275,7 +275,7 @@ wait_for_patch_transition() {
 
 module_ref_count() {
 	local modname="$1"
-	[[ $(cat "/sys/module/$modname/refcnt" 2>/dev/null) != "0" ]]
+	[[ $(cat "/sys/module/$modname/refcnt" 2>/dev/null) -gt "0" ]]
 }
 
 wait_for_zero_module_ref_count() {
@@ -338,6 +338,9 @@ load_module () {
 			echo "module named $modname already loaded and enabled"
 		fi
 	else
+		# Cleanup possibly loaded, but disabled patch.
+		remove_module "$modname" "quiet"
+
 		echo "loading patch module: $module"
 		local i=0
 		while true; do
@@ -420,7 +423,9 @@ remove_module () {
 		die "failed to unload module $modname (refcnt)"
 	fi
 
-	echo "unloading patch module: $modname"
+	if [[ "$#" -lt 2 || "$2" != "quiet" ]] ; then
+		echo "unloading patch module: $modname"
+	fi
 	# ignore any error here because rmmod can fail if the module used
 	# KPATCH_FORCE_UNSAFE.
 	rmmod "$modname" 2> /dev/null || return 0


joe-lawrence:
Hi @liu-song-6,

I understand the scenario, but not the use case.  Is the user trying to reload the initial kpatch after it's already been atomically replaced?  Something like:
- kpatch A loaded
- kpatch B replaces kpatch A
- kpatch A disabled, refcount 0, module left loaded
- user wants to load kpatch A again?

Or did the user also disable and remove kpatch B before trying to reload kpatch A?

Also perhaps another way to tackle this is for the kpatch utility's load subcommand to hang around try and cleanup any kpatches that have been atomically replaced by the incoming kpatch.  Maybe that has more "replace" than simple "load" semantics, but I only offer it as a brainstorming idea before fully understanding the use case.

liu-song-6:
> 
> * kpatch A loaded
> * kpatch B replaces kpatch A
> * kpatch A disabled, refcount 0, module left loaded
> * user wants to load kpatch A again?
> 
> Or did the user also disable and remove kpatch B before trying to reload kpatch A?

The plan is to build kpatch A also with replace. Therefore, when the user loads kpatch A again, it should replace kpatch B. The use case is like, we have kpatch A loaded, then we want to test kpatch B. If the test didn't go well, we would roll back to kpatch A. 

I agree it is better to do the clean up when we load kpatch B for the first time. But this version is a lot simpler...

liu-song-6:
Hi @joe-lawrence , 

Here are some more details about our plan. We would like to use the replace flag on all live live and deliver live patch (the .ko file) in rpm files. The rpm will install the .ko file in /var/lib/kpatch/$(uname -r)/; and calls `kpatch load --all` in %post. With this design, we can simply upgrade live patch as regular rpms. For example:
```
# install the first hotfix
rpm -i klp-xxx-hotfix1-1.rpm
# upgrade to the second hotfix
rpm -U klp-xxx-hotfix2-1.rpm
```
If we want to do downgrade back to hotfix1, the rpm syntax is like:
```
rpm -U klp-xxx-hotfix1-1.rpm --force
```
However, without this change, this may break because the hotfix1 module is loaded. 

I don't think adding rmmod would break anything. Shall we merge this change and simplify the downgrade operation?

Thanks!

joe-lawrence:
Hi @liu-song-6,

Thanks for the extra details, I understand the use case now.  My only reservation about the strategy is that downgrading from hotfix2 to hotfix1 requires hotfix1 to be compatible with any side effects of hotfix2.  For example, if hotfix2 creates or modifies a shadow variable or other data structure, hotfix1 would need to be ready to handle those cases, too.  With a little foresight, that could be possible, and in cases where it's not, I think you could just upgrade to hotfix3 (which would be hotfix1 + extra code to handle hotfix2's side effects).

A few follow ups out of curiosity:

- is reversion back to hotfix1 intended for production, or only in testing scenarios?
- do you plan on loading multiple hotfix kpatches at the same time (addressing different CVEs or subsystems), or do you plan on building a single (large) cumulative kpatch?

IIRC, @euspectre uses the atomic replace feature of the older kpatch.ko core module, so he may have some real experience and advice on patch revert vs. upgrade strategies.


As far as this change goes, I'm still on the fence about where it would best fit.  The simple kpatch unload case waits around for a zero-reference-count and finally removes said module, so the user could expect kpatch load to do the same for replaced modules... but this seems non-trivial to implement for multiple reasons.  If this looks like the "best" place for this functionality, maybe @jpoimboe or @sm00th have some ideas.

In the meantime, `load_module()` is a hairball of support for various livepatch kernel versions, not to mention legacy kpatch.ko.  There are a lot of weird corner cases (like aborting an unpatching transition and legacy kpatch.ko checksum stuff), but would this work and be a little more targeted to the intended use case (untested)?

``` patch
diff --git a/kpatch/kpatch b/kpatch/kpatch
index 4e46b124e492..fd453a7eaced 100755
--- a/kpatch/kpatch
+++ b/kpatch/kpatch
@@ -338,6 +338,10 @@ load_module () {
                        echo "module named $modname already loaded and enabled"
                fi
        else
+
+               # Cleanup possibly loaded, but disabled patch.
+               remove_module "$modname"
+
                echo "loading patch module: $module"
                local i=0
                while true; do
```

liu-song-6:
> Thanks for the extra details, I understand the use case now. My only reservation about the strategy is that downgrading from hotfix2 to hotfix1 requires hotfix1 to be compatible with any side effects of hotfix2. For example, if hotfix2 creates or modifies a shadow variable or other data structure, hotfix1 would need to be ready to handle those cases, too. With a little foresight, that could be possible, and in cases where it's not, I think you could just upgrade to hotfix3 (which would be hotfix1 + extra code to handle hotfix2's side effects).

We understand that there are cases where we cannot roll back to hotfix1, and further upgrading to hotfix3 is the better (and maybe the only viable) option. Rolling back is not a normal operation. It is more for emergencies when something is seriously wrong with hotfix2. 

> 
> A few follow ups out of curiosity:
> 
> * is reversion back to hotfix1 intended for production, or only in testing scenarios?

We may need it in production, but not often (emergency only)

> * do you plan on loading multiple hotfix kpatches at the same time (addressing different CVEs or subsystems), or do you plan on building a single (large) cumulative kpatch?

We plan to build a single large cumulative kpatch. 

> 
> In the meantime, `load_module()` is a hairball of support for various livepatch kernel versions, not to mention legacy kpatch.ko. There are a lot of weird corner cases (like aborting an unpatching transition and legacy kpatch.ko checksum stuff), but would this work and be a little more targeted to the intended use case (untested)?
> 
> ```diff
> diff --git a/kpatch/kpatch b/kpatch/kpatch
> index 4e46b124e492..fd453a7eaced 100755
> --- a/kpatch/kpatch
> +++ b/kpatch/kpatch
> @@ -338,6 +338,10 @@ load_module () {
>                         echo "module named $modname already loaded and enabled"
>                 fi
>         else
> +
> +               # Cleanup possibly loaded, but disabled patch.
> +               remove_module "$modname"
> +
>                 echo "loading patch module: $module"
>                 local i=0
>                 while true; do
> ```

Yes, this version also works.

liu-song-6:
> > In the meantime, `load_module()` is a hairball of support for various livepatch kernel versions, not to mention legacy kpatch.ko. There are a lot of weird corner cases (like aborting an unpatching transition and legacy kpatch.ko checksum stuff), but would this work and be a little more targeted to the intended use case (untested)?
> > ```diff
> > diff --git a/kpatch/kpatch b/kpatch/kpatch
> > index 4e46b124e492..fd453a7eaced 100755
> > --- a/kpatch/kpatch
> > +++ b/kpatch/kpatch
> > @@ -338,6 +338,10 @@ load_module () {
> >                         echo "module named $modname already loaded and enabled"
> >                 fi
> >         else
> > +
> > +               # Cleanup possibly loaded, but disabled patch.
> > +               remove_module "$modname"
> > +
> >                 echo "loading patch module: $module"
> >                 local i=0
> >                 while true; do
> > ```
> 
> Yes, this version also works.

Shall I update the pull request with this version?

joe-lawrence:
> Shall I update the pull request with this version?

I don't have a better solution for now, so yes, I think moving that code a little deeper in `load_module()` would better target this use case.  Also using `remove_module()` should handle a very small timing window where the livepatch sysfs interface has gone away but the its module reference may have not dropped to zero yet.

jpoimboe (approved):

sm00th:
Ugh. This now fails the case of initial module load because `module_ref_count()` can't find `/sys/module/$modname/refcnt`. So we'll need something like this:

```
diff --git a/kpatch/kpatch b/kpatch/kpatch
index 1707fdd..6936c34 100755
--- a/kpatch/kpatch
+++ b/kpatch/kpatch
@@ -275,7 +275,10 @@ wait_for_patch_transition() {
 
 module_ref_count() {
        local modname="$1"
-       [[ $(cat "/sys/module/$modname/refcnt" 2>/dev/null) != "0" ]]
+       local refcnt="/sys/module/$modname/refcnt"
+
+       [[ -f "$refcnt" ]] || return 1
+       [[ $(cat "$refcnt" 2>/dev/null) != "0" ]]
 }
 
 wait_for_zero_module_ref_count() {
```

joe-lawrence:
> Ugh. This now fails the case of initial module load because `module_ref_count()` can't find `/sys/module/$modname/refcnt`. So we'll need something like this:
> 
> ```
> diff --git a/kpatch/kpatch b/kpatch/kpatch
> index 1707fdd..6936c34 100755
> --- a/kpatch/kpatch
> +++ b/kpatch/kpatch
> @@ -275,7 +275,10 @@ wait_for_patch_transition() {
>  
>  module_ref_count() {
>         local modname="$1"
> -       [[ $(cat "/sys/module/$modname/refcnt" 2>/dev/null) != "0" ]]
> +       local refcnt="/sys/module/$modname/refcnt"
> +
> +       [[ -f "$refcnt" ]] || return 1
> +       [[ $(cat "$refcnt" 2>/dev/null) != "0" ]]
>  }
>  
>  wait_for_zero_module_ref_count() {
> ```

Ooo, good catch.

Should we be paranoid that the refcnt file could go away between the `-f` and `cat` calls and end up with the same issue?  How about updating the original condition slightly so that it _only_ succeeds when the refcnt is a number greater than 0.  Again, untested :D 

```bash
[[ $(cat "/sys/module/$modname/refcnt" 2>/dev/null) -gt 0 ]]
```

liu-song-6:
Thanks @joe-lawrence and @sm00th . I updated the commit with the feedback. I also added remove_module_quiet(), which do not have the confusing `echo "unloading patch module: $modname"`.

kpatch/kpatch
```diff
@@ -426,6 +429,18 @@ remove_module () {
 	rmmod "$modname" 2> /dev/null || return 0
 }
 
+remove_module_quiet () {
```
sm00th, line 432: Looks good. Silencing the echo is a good idea but it is hardly worth a full copy of remove_module_quiet() function. We can add an optional variable instead and only set it to "true" or something in the new call, leaving old ones as is.

sm00th (approved):

joe-lawrence (approved):
Thanks for the updates, @liu-song-6  :+1:

PR #1192, opened 2021-06-10
kpatch: use /sys/kernel/kpatch/ to check whether core module is loaded

### LINKED ISSUE #1187: kpatch: bug in core_loaded  function when determining whether kpatch.ko is loaded
kpatch.service exited abnormally during system boot. 

`systemd[1]: Starting "Apply kpatch kernel patches" ...
kpatch[1156]: loading core module: /usr/local/lib/kpatch/4.19.91.x86_64/kpatch.ko
kpatch[1156]: loading patch module: /var/lib/kpatch/4.19.91.x86_64/kpatch-3938029.ko
kpatch[1156]: loading patch module: /var/lib/kpatch/4.19.91.x86_64/kpatch-4148281.ko
kpatch[1156]: loading patch module: /var/lib/kpatch/4.19.91.x86_64/kpatch-4312241.ko
insmod core module:  /usr/local/lib/kpatch/4.19.91.x86_64/kpatch.ko
insmod : EROR: could not insert module  /usr/local/lib/kpatch/4.19.91.x86_64/kpatch.ko: File exists
kpatch: failed to load core module
kpatch.service: main process exited, code=exited, status=1/FAILURE
Failed to start "Apply kpatch kernel patches"
Unit kpatch.service entered failed state
kpatch.service failed
`

After analysis, we found that the core_loaded function in kpatch script can be wrong in some cases

core_loaded ()
{
      grep -q -e "T klp_enable_patch" -e "T kpatch_register" /proc/kallsyms
}

reading /proc/kallsyms can be inconsistent as this post mentioned ( [https://stackoverflow.com/questions/5713451/is-it-safe-to-parse-a-proc-file](url) )

In the other hand, I can reproduce the same issue when insmod & rmmod repeatedly, here is my test case:

one process, execute the folllowing command
 insert and remove bridge.ko repeatedly, just to manipulate the kallsyms file content
`
while true; do rmmod bridge; modprobe bridge; done
`

the other process, loop to check core_loaded value, following is my script:

!/bin/bash
BASE="$PWD"
COREMOD="/usr/local/lib/kpatch/$(uname -r)/kpatch.ko"
LOGFILE="$BASE/test.log"

core_loaded () {
        grep -q -e "T klp_enable_patch" -e "T kpatch_register" /proc/kallsyms
}

logger() {
    local to_stdout=${1:-0}
     tee -a "$LOGFILE"
}

warn() {
        echo "$(date) kpatch: $*" | logger
}

die() {
        warn "$@"
        exit 1
}

cleanup() {
        rmmod "$COREMOD" || die "failed to unload core module"
}

trap cleanup EXIT INT TERM HUP
while true
 do
         insmod "$COREMOD" || die "failed to load core module"
         if ! core_loaded; then
                die "core not loaded"
         fi
        rmmod "$COREMOD" || die "failed to unload core module"
done

the test script fails after printing 'core not loaded' within 1 min.

xiejingf:
Fixes #1187

### DIFF
--- a/kpatch/kpatch
+++ b/kpatch/kpatch
@@ -127,11 +127,11 @@ find_core_module() {
 }
 
 kpatch_core_loaded() {
-	grep -q -e "T kpatch_register" /proc/kallsyms
+	[[ -d "/sys/kernel/kpatch" ]]
 }
 
 core_loaded () {
-	grep -q -e "T klp_enable_patch" -e "T kpatch_register" /proc/kallsyms
+	[[ -d "/sys/kernel/kpatch" ]] || [[ -d "/sys/kernel/livepatch" ]]
 }
 
 get_module_name () {


euspectre:
Looks good to me in general.

2 nits though:

* Please provide some info about the problem in the commit message, in addition to '#1187'. GitHub issues may come and go, or the key maintainers might decide to move the git repository somewhere else in the future. It would be nice if one is able to understand the issue and the fix from the commit itself.

* I suggest using "/sys/kernel/kpatch", that is, without the trailing slash. Just for uniformity with the rest of the script.

xiejingf:
> 2 nits though:
> Please provide some info about the problem in the commit message, in addition to '#1187'. GitHub issues may come and go, > or the key maintainers might decide to move the git repository somewhere else in the future. It would be nice if one is able > > to understand the issue and the fix from the commit itself.

>I suggest using "/sys/kernel/kpatch", that is, without the trailing slash. Just for uniformity with the rest of the script.

thanks for the careful review ,  the commit is updated per you advice

euspectre:
Looks good to me.
And - it works on my machine (tm) :-)

xiejingf:
> Looks good to me.
> And - it works on my machine (tm) :-)


If it is ok, please approve the patch

jpoimboe (approved):

PR #1201, opened 2021-07-11
Kpatch analyze: core implementation

divyacote:
As discussed previously with a few developers, this pull request implements the first stage of the kpatch-analyze tool.

Briefly, the kpatch-analyze tool is a program based on Linus’s `sparse` semantic parser tool to study a regular kernel source code patch for its ability to be built into a kpatch patch. The long term goal is to implement as many complete and thorough checks as possible to help with the work of creating patches that will build, load and behave correctly when applied by Livepatch. For now this PR focuses on implementing a core feature that analyzes changes, instead of complete ‘C’ files which is the goal of most parsers like sparse.

The kpatch-analyze workflow is divided into two main components. 

- kpatch-analyze/kpatch-analyze: is the front-end script (ala kpatch-build) that prepares the analysis environment. In fact kpatch-analyze expects that “kpatch-build --skip-build <patchname>” is called previously so that a kernel source tree is available. This script is mostly responsible for applying the desired patch to the source tree in order to extract the line numbers that were changed by the patch. This information is mostly generated using “git blame”. A data file is created from that information and fed into the core component, kpatch-sparse.c

- kpatch-analyze/kpatch-sparse.c: is the main kpatch-analyze logic. Using the changed line mappings from the previous stage, the kpatch-sparse tokenizer emits information about changed tokens which is transmitted directly to the parse tree by the sparse parsing code. The program then is able to examine the parse tree and determine which symbols, statements and expressions were changed and in which scope. The “evaluate_change()” function is called with a list (small for now, hopefully will grow and improve in the future) of specified checks that look at different kpatch patch known pitfalls and scenarios to generate a report on screen.

A simple example:

$ kpatch-build/kpatch-build -r kernel-4.18.0-193.el8.src.rpm --skip-build test/unit/analyze/test-check-notrace-rhel8_2.patch
$ kpatch-analyze/kpatch-analyze test/unit/analyze/test-check-notrace-rhel8_2.patch


joe-lawrence:
Hi @divyacote : thanks for posting this project!

Here are my initial thoughts:

**Should this live in the kpatch repository (as this PR currently implements)?**
For related projects [kpatch-dnf](https://github.com/dynup/kpatch-dnf) and [kpatch-unit-test-objs](https://github.com/dynup/kpatch-unit-test-objs) we them gave their own repo / PRs / issues, etc.  In the case of [kpatch-unit-test-objs](https://github.com/dynup/kpatch-unit-test-objs), we import it as a git subproject in the kpatch tree as `test/unit/objs`.  Git subprojects are bit clunky to work with, but it let us cleave off the `test/unit/objs` directories and stash binary objects in a separate git repo.  On the other hand, [kpatch-dnf](https://github.com/dynup/kpatch-dnf) does not need to build or run under any kpatch project path, so it simply lives as another project repo.

I think kpatch-analyze could work in either model, it just needs `kpatch-build --skip-build` to have been run first?  (Aside: would`--prepare-kernel-tree` be more descriptive?)

**What checks are  currently supported?**
I see the `init` and `notrace` function annotation checks... how does `check_global_changes()` work and can we add a test case for that?

**What checks do you envision could be added?**
Were there some ideas that you had that you considered out of scope / time for the project?

**About the sparse code**
The [smatch](https://repo.or.cz/smatch.git) project similarly leverages sparse code.  Do you know if that project ever needs to update the underlying sparse code?  Also, would there be any (dis)advantages to have based kpatch-analyzer on smatch and not sparse directly.  (I don't know much about either project, so hopefully this question makes sense.)

divyacote:
Hi @joe-lawrence 

> **Should this live in the kpatch repository (as this PR currently implements)?**
> For related projects [kpatch-dnf](https://github.com/dynup/kpatch-dnf) and [kpatch-unit-test-objs](https://github.com/dynup/kpatch-unit-test-objs) we them gave their own repo / PRs / issues, etc. In the case of [kpatch-unit-test-objs](https://github.com/dynup/kpatch-unit-test-objs), we import it as a git subproject in the kpatch tree as `test/unit/objs`. Git subprojects are bit clunky to work with, but it let us cleave off the `test/unit/objs` directories and stash binary objects in a separate git repo. On the other hand, [kpatch-dnf](https://github.com/dynup/kpatch-dnf) does not need to build or run under any kpatch project path, so it simply lives as another project repo.
> 
> I think kpatch-analyze could work in either model, it just needs `kpatch-build --skip-build` to have been run first? (Aside: would`--prepare-kernel-tree` be more descriptive?)

It could probably be its own repo, but as you mentioned there is some codependency between kpatch-build and kpatch-analyze. For instance, kpatch-analyze uses the same workspace environment (CACHEDIR). Also, one can call the other. For simplicity in this PR I simply added the --skip-build option to kpatch-build because I wasn’t sure if we would like to call kpatch-analyze as a new step in kpatch-build and/or call kpatch-build directly from kpatch-analyze which in the future seems like a better workflow than having to call both scripts one after the other.

Yes, I can change --skip-build to --prepare-kernel-tree. 


> 
> **What checks are currently supported?**
> I see the `init` and `notrace` function annotation checks... how does `check_global_changes()` work and can we add a test case for that?

For the moment, I expect those checks to mostly work but I believe that deeper analysis would help to make sure that there are no corner cases creating false positives. I think my subsequent PRs will focus on better checks and include a decent set of tests. Regarding check_global_changes(), right now the check simply verifies whether statements are annotated as part of a function or not. 



> 
> **What checks do you envision could be added?**

I should come up with an inventory of those but my feeling is that the tool provides enough analysis capabilities to implement any interesting kpatch scenarios.


> Were there some ideas that you had that you considered out of scope / time for the project?

Well, the checks could be improved and should probably be moved to their own .c file. For the moment, kpatch-analyze will only be able to let the workflow know if a patch can be fed directly to kpatch-build. It could certainly be interesting to explore if some of the modifications could be generated by the tool. 


> 
> **About the sparse code**
> The [smatch](https://repo.or.cz/smatch.git) project similarly leverages sparse code. Do you know if that project ever needs to update the underlying sparse code? Also, would there be any (dis)advantages to have based kpatch-analyzer on smatch and not sparse directly. (I don't know much about either project, so hopefully this question makes sense.)

My initial intention was to have sparse or libsparse be installed as part of the kpatch setup. The first problem is that there are no packages available for easy installation, at least for the system I developed on (Fedora). The second problem is that because of our peculiar use case, that is analyzing patches instead of full c files, I had to modify few files from the sparse code and this could continue in the future. It would certainly be nice to refresh the code from time to time, but changes seem to be infrequent on sparse nowadays. BTW the sparse initial import is a stripdown set of files that only includes the needed core code. My understanding is that smatch was built to provide more sophisticated checks using the parse tree generated by sparse whereas our checks only need annotations from the tree. In other words, analyzing changed symbols in specific scopes and stuff of that nature is “simple enough” compared to what static analyzers are trying to do nowadays. 


Thanks a lot for your review.

PR #1203, opened 2021-08-02
S390x kpatch support

sumanthkorikkar:
Hi All,

Add s390x support for kpatch

Prerequisite gcc patches:
https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=a1c1b7a888ade6f21bc7c7f05a2cbff290273fcc
https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=0990d93dd8a4268bff5bbe48aa26748cf63201c7
https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=935b5226c385e34088c314374cbbe9e4995b9e44

Prerequisite kernel patches:
https://lore.kernel.org/live-patching/YObPhPkzRSqnzgK3@alley/T/#me6c2fcdbd4f582d5cc5c594bafb083ba814acdf9 - Yet to be upstream.
https://git.kernel.org/pub/scm/linux/kernel/git/s390/linux.git/commit/?h=for-next&id=7561c14d8a4d1a24a40b1839d927d488e2d6345a
Recommended Patches:
torvalds/linux@de5012b
torvalds/linux@67ccddf

 
Tested on kernel 5.14:
KPATCHBUILD_OPTS="-v /home/extend/vmlinux -s /home/extend/linux-devel/linux/  -d" ./kpatch-test  -d linux-5.14.0/ -q
build: combined module
load test: combined module
SUCCESS

Let me know your suggestions. 
Thank you. 

--
Sumanth

### DIFF
--- a/doc/s390-upstream-prerequisites.md
+++ b/doc/s390-upstream-prerequisites.md
@@ -1,0 +1,33 @@
+### s390 backporting
+
+**Prerequisite gcc patches (all backported to releases/gcc-11 branch):**
+- gcc-mirror/gcc@a1c1b7a IBM Z: Define NO_PROFILE_COUNTERS
+- gcc-mirror/gcc@0990d93 IBM Z: Use @PLT symbols for local functions in 64-bit mode
+- gcc-mirror/gcc@935b522 S/390: New option -mpic-data-is-text-relative
+- gcc-mirror/gcc@8753b13 IBM Z: fix section type conflict with -mindirect-branch-table
+
+**Prerequisite kernel patches:**
+**v5.19:**
+- 69505e3d9a39 bug: Use normal relative pointers in 'struct bug_entry'
+
+**v5.18:**
+- 602bf1687e6f s390/nospec: align and size extern thunks
+- 1d2ad084800e s390/nospec: add an option to use thunk-extern
+- eed38cd2f46f s390/nospec: generate single register thunks if possible
+- 2268169c14e5 s390: remove unused expoline to BC instructions
+- f0003a9e4c18 s390/entry: remove unused expoline thunk
+
+**v5.16:**
+- torvalds/linux@f6ac18f sched: Improve try_invoke_on_locked_down_task()
+- torvalds/linux@9b3c4ab sched,rcu: Rework try_invoke_on_locked_down_task()
+- torvalds/linux@00619f7 sched,livepatch: Use task_call_func()
+- torvalds/linux@8850cb6 sched: Simplify wake_up_*idle*()
+- torvalds/linux@5de62ea sched,livepatch: Use wake_up_if_idle()
+- torvalds/linux@96611c2 sched: Improve wake_up_all_idle_cpus() take #2
+
+**v5.15**
+- torvalds/linux@de5012b s390/ftrace: implement hotpatching
+- torvalds/linux@67ccddf ftrace: Introduce ftrace_need_init_nop()
+
+**v5.14:**
+- torvalds/linux@7561c14 s390/vdso: add .got.plt in vdso linker script
--- a/kpatch-build/Makefile
+++ b/kpatch-build/Makefile
@@ -21,7 +21,8 @@ GCC_PLUGINS_DIR := $(shell $(CROSS_COMPILE)gcc -print-file-name=plugin)
 PLUGIN_CFLAGS := $(filter-out -Wconversion, $(CFLAGS))
 PLUGIN_CFLAGS += -shared -I$(GCC_PLUGINS_DIR)/include \
 		   -Igcc-plugins -fPIC -fno-rtti -O2 -Wall
-else ifneq ($(ARCH),x86_64)
+endif
+ifeq ($(filter $(ARCH),s390x x86_64 ppc64le),)
 $(error Unsupported architecture ${ARCH}, check https://github.com/dynup/kpatch/#supported-architectures)
 endif
 
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -163,6 +163,8 @@ static bool is_gcc6_localentry_bundled_sym(struct kpatch_elf *kelf,
 			sym->sym.st_value == 8);
 	case X86_64:
 		return false;
+	case S390:
+		return false;
 	default:
 		ERROR("unsupported arch");
 	}
@@ -304,8 +306,7 @@ static bool is_dynamic_debug_symbol(struct symbol *sym)
 
 static bool is_string_literal_section(struct section *sec)
 {
-	return !strncmp(sec->name, ".rodata.", 8) &&
-	       strstr(sec->name, ".str1.");
+	return !strncmp(sec->name, ".rodata.", 8) && strstr(sec->name, ".str");
 }
 
 /*
@@ -2105,6 +2106,17 @@ static int fixup_barrier_nospec_group_size(struct kpatch_elf *kelf, int offset)
 }
 
 /*
+ * .s390_indirect_jump, .s390_indirect_call, .s390_indirect_branches,
+ * .s390_return_reg, .s390_return_mem contains indirect branch locations. This
+ * is an array of 32 bit elements. These sections could be used during runtime
+ * to replace the expolines with the normal indirect jump.
+ */
+static int s390_expolines_group_size(struct kpatch_elf *kelf, int offset)
+{
+	return 4;
+}
+
+/*
  * The rela groups in the .fixup section vary in size.  The beginning of each
  * .fixup rela group is referenced by the __ex_table section. To find the size
  * of a .fixup rela group, we have to traverse the __ex_table relas.
@@ -2157,27 +2169,27 @@ static int fixup_group_size(struct kpatch_elf *kelf, int offset)
 static struct special_section special_sections[] = {
 	{
 		.name		= "__bug_table",
-		.arch		= X86_64 | PPC64,
+		.arch		= X86_64 | PPC64 | S390,
 		.group_size	= bug_table_group_size,
 	},
 	{
 		.name		= ".fixup",
-		.arch		= X86_64 | PPC64,
+		.arch		= X86_64 | PPC64 | S390,
 		.group_size	= fixup_group_size,
 	},
 	{
 		.name		= "__ex_table", /* must come after .fixup */
-		.arch		= X86_64 | PPC64,
+		.arch		= X86_64 | PPC64 | S390,
 		.group_size	= ex_table_group_size,
 	},
 	{
 		.name		= "__jump_table",
-		.arch		= X86_64 | PPC64,
+		.arch		= X86_64 | PPC64 | S390,
 		.group_size	= jump_table_group_size,
 	},
 	{
 		.name		= ".printk_index",
-		.arch		= X86_64 | PPC64,
+		.arch		= X86_64 | PPC64 | S390,
 		.group_size	= printk_index_group_size,
 	},
 	{
@@ -2192,7 +2204,7 @@ static struct special_section special_sections[] = {
 	},
 	{
 		.name		= ".altinstructions",
-		.arch		= X86_64,
+		.arch		= X86_64 | S390,
 		.group_size	= altinstructions_group_size,
 	},
 	{
@@ -2229,6 +2241,31 @@ static struct special_section special_sections[] = {
 		.name		= "__barrier_nospec_fixup",
 		.arch		= PPC64,
 		.group_size	= fixup_barrier_nospec_group_size,
+	},
+	{
+		.name = ".s390_return_mem",
+		.arch = S390,
+		.group_size = s390_expolines_group_size,
+	},
+	{
+		.name = ".s390_return_reg",
+		.arch = S390,
+		.group_size = s390_expolines_group_size,
+	},
+	{
+		.name = ".s390_indirect_call",
+		.arch = S390,
+		.group_size = s390_expolines_group_size,
+	},
+	{
+		.name = ".s390_indirect_branches",
+		.arch = S390,
+		.group_size = s390_expolines_group_size,
+	},
+	{
+		.name = ".s390_indirect_jump",
+		.arch = S390,
+		.group_size = s390_expolines_group_size,
 	},
 	{},
 };
@@ -3024,6 +3061,11 @@ static bool kpatch_is_core_module_symbol(char *name)
 		!strcmp(name, "kpatch_shadow_get"));
 }
 
+static bool is_expoline(struct kpatch_elf *kelf, char *name)
+{
+	return kelf->arch == S390 && !strncmp(name, "__s390_indirect_jump_r", 22);
+}
+
 static int function_ptr_rela(const struct rela *rela)
 {
 	const struct rela *rela_toc = toc_rela(rela);
@@ -3080,6 +3122,13 @@ static bool need_dynrela(struct kpatch_elf *kelf, struct lookup_table *table,
 	 * relas.  They should be exported.
 	 */
 	if (kpatch_is_core_module_symbol(rela->sym->name))
+		return false;
+
+       /*
+        * Allow references to s390 expolines to remain as normal relas.  They
+        * will be generated in the module by the kernel module link.
+        */
+	if (is_expoline(kelf, rela->sym->name))
 		return false;
 
 	if (rela->sym->sec) {
@@ -3548,6 +3597,10 @@ static void kpatch_create_mcount_sections(struct kpatch_elf *kelf)
 			rela->type = R_X86_64_PC32;
 			break;
 		}
+		case S390: {
+			insn_offset = sym->sym.st_value;
+			break;
+		}
 		default:
 			ERROR("unsupported arch");
 		}
@@ -3713,6 +3766,7 @@ static void kpatch_find_func_profiling_calls(struct kpatch_elf *kelf)
 {
 	struct symbol *sym;
 	struct rela *rela;
+	unsigned char *insn;
 	list_for_each_entry(sym, &kelf->symbols, list) {
 		if (sym->type != STT_FUNC || !sym->sec || !sym->sec->rela)
 			continue;
@@ -3736,6 +3790,14 @@ static void kpatch_find_func_profiling_calls(struct kpatch_elf *kelf)
 				continue;
 
 			sym->has_func_profiling = 1;
+			break;
+		case S390:
+			/* Check for compiler generated fentry nop - jgnop 0 */
+			insn = sym->sec->data->d_buf;
+			if (insn[0] == 0xc0 && insn[1] == 0x04 &&
+				insn[2] == 0x00 && insn[3] == 0x00 &&
+				insn[4] == 0x00 && insn[5] == 0x00)
+				sym->has_func_profiling = 1;
 			break;
 		default:
 			ERROR("unsupported arch");
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -119,6 +119,16 @@ static bool is_bundleable(struct symbol *sym)
 		return true;
 
 	if (sym->type == STT_OBJECT &&
+	    !strncmp(sym->sec->name, ".data.rel.ro.local.", 19) &&
+	    !strcmp(sym->sec->name + 19, sym->name))
+		return 1;
+
+	if (sym->type == STT_OBJECT &&
+	    !strncmp(sym->sec->name, ".data.rel.local.", 16) &&
+	    !strcmp(sym->sec->name + 16, sym->name))
+		return 1;
+
+	if (sym->type == STT_OBJECT &&
 	   !strncmp(sym->sec->name, ".rodata.",8) &&
 	   !strcmp(sym->sec->name + 8, sym->name))
 		return true;
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -937,6 +937,10 @@ if [[ "$ARCH" = "ppc64le" ]]; then
 	ARCH_KCFLAGS="-mcmodel=large -fplugin=$PLUGINDIR/ppc64le-plugin.so"
 fi
 
+if [[ "$ARCH" = "s390x" ]]; then
+	ARCH_KCFLAGS="-mno-pic-data-is-text-relative -fno-section-anchors"
+fi
+
 export KCFLAGS="-I$DATADIR/patch -ffunction-sections -fdata-sections \
 		$ARCH_KCFLAGS $DEBUG_KCFLAGS"
 
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -343,12 +343,18 @@ find_special_section_data() {
 	check[e]=true						# exception_table_entry
 
 	# Arch-specific features, without kernel CONFIG_ toggle
-	if [[ "$ARCH" = "x86_64" ]]; then
-		check[a]=true					# alt_instr
-		kernel_version_gte 5.10.0 && check[s]=true	# static_call_site
-	elif [[ "$ARCH" = "ppc64le" ]]; then
-		check[f]=true					# fixup_entry
-	fi
+	case "$ARCH" in
+		"x86_64")
+			check[a]=true					# alt_instr
+			kernel_version_gte 5.10.0 && check[s]=true	# static_call_site
+			;;
+		"ppc64le")
+			check[f]=true					# fixup_entry
+			;;
+		"s390x")
+			check[a]=true					# alt_instr
+			;;
+	esac
 
 	# Kernel CONFIG_ features
 	[[ -n "$CONFIG_PRINTK_INDEX" ]] && check[i]=true	# pi_entry
@@ -856,6 +862,7 @@ trace_on
 	die "openEuler kernel doesn't have 'CONFIG_LIVEPATCH_PER_TASK_CONSISTENCY' enabled"
 
 [[ -z "$CONFIG_DEBUG_INFO" ]] && die "kernel doesn't have 'CONFIG_DEBUG_INFO' enabled"
+[[ "$ARCH" = "s390x" ]] && [[ -z "$CONFIG_EXPOLINE_EXTERN" ]] && [[ -n "$CONFIG_EXPOLINE" ]] && die "kernel doesn't have 'CONFIG_EXPOLINE_EXTERN' enabled"
 
 # Build variables - Set some defaults, then adjust features
 # according to .config and kernel version
--- a/kpatch-build/kpatch-elf.c
+++ b/kpatch-build/kpatch-elf.c
@@ -140,6 +140,8 @@ unsigned int absolute_rela_type(struct kpatch_elf *kelf)
 		return R_PPC64_ADDR64;
 	case X86_64:
 		return R_X86_64_64;
+	case S390:
+		return R_390_64;
 	default:
 		ERROR("unsupported arch");
 	}
@@ -220,6 +222,23 @@ long rela_target_offset(struct kpatch_elf *kelf, struct section *relasec,
 				  (long)sec->data->d_buf -
 				  rela->offset;
 		} else
+			ERROR("unhandled rela type %d", rela->type);
+		break;
+	case S390:
+		/*
+		 * For branch and relative load instructions,
+		 * add_off is -2.
+		 */
+		if (rela->type == R_390_GOTENT ||
+			rela->type == R_390_PLT32DBL ||
+			rela->type == R_390_PC32DBL)
+			add_off = -2;
+		else if (rela->type == R_390_32 ||
+			rela->type == R_390_64 ||
+			rela->type == R_390_PC32 ||
+			rela->type == R_390_PC64)
+			add_off = 0;
+		else
 			ERROR("unhandled rela type %d", rela->type);
 		break;
 	default:
@@ -478,6 +497,9 @@ struct kpatch_elf *kpatch_elf_open(const char *name)
 		break;
 	case EM_X86_64:
 		kelf->arch = X86_64;
+		break;
+	case EM_S390:
+		kelf->arch = S390;
 		break;
 	default:
 		ERROR("Unsupported target architecture");
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Supported Architectures
 - [x] x86-64
 - [x] ppc64le
 - [ ] arm64
-- [ ] s390
+- [x] s390 [upstream prerequisites](doc/s390-upstream-prerequisites.md)
 
 Installation
 ------------
--- a/kpatch-build/kpatch-cc
+++ b/kpatch-build/kpatch-cc
@@ -43,6 +43,9 @@ if [[ "$TOOLCHAINCMD" =~ ^(.*-)?gcc$ || "$TOOLCHAINCMD" =~ ^(.*-)?clang$ ]] ; th
 				drivers/firmware/efi/libstub/*|\
 				arch/powerpc/kernel/prom_init.o|\
 				arch/powerpc/kernel/vdso64/*|\
+				arch/s390/boot/*|\
+				arch/s390/purgatory/*|\
+				arch/s390/kernel/vdso64/*|\
 				lib/*|\
 				.*.o|\
 				*/.lib_exports.o)
--- a/kpatch-build/lookup.c
+++ b/kpatch-build/lookup.c
@@ -272,12 +272,14 @@ static void symtab_read(struct lookup_table *table, char *path)
 		 * tables.  .dynsym is just a subset of .symtab, so skip it to
 		 * avoid duplicates.
 		 */
-		if (strstr(line, ".dynsym")) {
-			skip = true;
-			continue;
-		} else if (strstr(line, ".symtab")) {
-			skip = false;
-			continue;
+		if (!strncmp(line, "Symbol table ", 13)) {
+			if (strstr(line, ".dynsym")) {
+				skip = true;
+				continue;
+			} else if (strstr(line, ".symtab")) {
+				skip = false;
+				continue;
+			}
 		}
 		if (skip)
 			continue;


joe-lawrence:
@sumanthkorikkar : Hi Sumanth, thanks for posting.  I'll try to give it a spin next week.

In the meantime, two quick questions:
- For the gcc patches, do you know what the minimum version is that includes all of them?
- For the kernel patch "[RFC PATCH] livepatch: Kick idle cpu's tasks to perform transition", did you ever follow up with the schedule folks that Petr suggested?

sumanthkorikkar:
Hi Joe,

Thank you.

1. I will check with Ilya. However, this should be back portable to rhel9.1.
2. Vasily is working on next version.

sumanthkorikkar:
Hi @joe-lawrence 

The gcc patches are back portable from gcc11. 

Thanks.

joe-lawrence:
> Hi @joe-lawrence
> 
> The gcc patches are back portable from gcc11.
> 

Great thanks for the update.  FWIW, I tried with the stock rhel-alpha gcc and was able to build about 60% of the integration tests successfully.  I haven't tried with an updated / backported gcc just yet, but I think that will fix the remaining cases.

Quick note about the github CI -- it ran `make check` and found a small nitpick in [kpatch-build/kpatch-cc](https://github.com/dynup/kpatch/pull/1203/checks?check_run_id=3253682775).  With `arch/s390/boot/compressed/*` removed from the pattern list, it should test cleanly.

joe-lawrence:
Hi Sumanth,

I had better luck with the integration tests from this branch (https://github.com/joe-lawrence/kpatch/tree/integration-tests-kernel-5.14.0-0.rc3.29.el9) and the gcc backports that you mentioned.  I did not try rebuilding the kernel with that version of gcc, so these may be false reports, but I found problems with two of the tests:

#### data_read_mostly.patch

```
[   73.060379] test_data_read_mostly: loading out-of-tree module taints kernel.
[   73.060389] test_data_read_mostly: tainting kernel with TAINT_LIVEPATCH
[   73.060502] test_data_read_mostly: module verification failed: signature and/or required key missing - tainting kernel
[   73.060681] module test_data_read_mostly: relocation error for symbol netdev_cmd_to_name (r_type 5, value 0xfffffc009b332194)

% grep netdev_cmd_to_name /proc/kallsyms 
000000001b783b20 T netdev_cmd_to_name
000000001bc54bd8 d __ksymtab_netdev_cmd_to_name
000000001bc64caf r __kstrtabns_netdev_cmd_to_name
000000001bc804e9 r __kstrtab_netdev_cmd_to_name

% readelf --wide --relocs test/integration/test-data-read-mostly.ko
Relocation section '.rela.s390_indirect_jump' at offset 0x95b0 contains 1 entry:
0000000000000000  000001a400000005 R_390_PC32             0000000000000000 netdev_cmd_to_name + 32
```

#### meminfo-string.patch

```
% kpatch-build ...
ERROR: no functional changes found. Check /root/.kpatch/build.log for more details.

% strings ~/.kpatch/tmp/orig/fs/proc/meminfo.o | grep -i VMALLOCCHUNK
VmallocChunk:   
% strings ~/.kpatch/tmp/patched/fs/proc/meminfo.o | grep -i VMALLOCCHUNK
VMALLOCCHUNK:

% diff -u <(objdump --no-addresses -D ~/.kpatch/tmp/orig/fs/proc/meminfo.o) <(objdump --no-addresses -D ~/.kpatch/tmp/patched/fs/proc/meminfo.o)
--- /dev/fd/63  2021-08-20 16:18:52.666565151 -0400
+++ /dev/fd/62  2021-08-20 16:18:52.666565151 -0400
@@ -1,5 +1,5 @@
 
-/root/.kpatch/tmp/orig/fs/proc/meminfo.o:     file format elf64-s390
+/root/.kpatch/tmp/patched/fs/proc/meminfo.o:     file format elf64-s390
 
 
 Disassembly of section .group:
@@ -345,9 +345,9 @@
        ...
 
 <.LC36>:
-       56 6d 61 6c             o       %r6,364(%r13,%r6)
-       6c 6f 63 43             md      %f6,835(%r15,%r6)
-       68 75 6e 6b             ld      %f7,3691(%r5,%r6)
+       56 4d 41 4c             o       %r4,332(%r13,%r4)
+       4c 4f 43 43             mh      %r4,835(%r15,%r4)
+       48 55 4e 4b             lh      %r5,3659(%r5,%r4)
        3a 20                   aer     %f2,%f0
        20 20                   lpdr    %f2,%f0
        ...
```
I did try putting the second one into gdb, and I think there may be something suspect with create-diff-object's `rela_equal()` for these symbols.  I'm not very well versed in ppc64 vs. s390x TOC indirection, so that whole routine is a bit of a mystery to me... but I thought that might be where the `.LC36` content would be compared.  (Is it in `rela_tocX->string`, `rela_tocX->sym->sec->data->d_buf`, or ... ?)

sumanthkorikkar:
Hi Joe, 

Thank you. I will check these cases.

kpatch-build/create-diff-object.c
```diff
@@ -2913,8 +2992,8 @@ static void kpatch_create_patches_sections(struct kpatch_elf *kelf,
 		if (sym->bind == STB_LOCAL && symbol.global)
 			ERROR("can't find local symbol '%s' in symbol table", sym->name);
 
-		log_debug("lookup for %s: obj=%s sympos=%lu size=%lu",
-			  sym->name, symbol.objname, symbol.sympos,
+		log_debug("lookup for %s: obj=%s sympos=%lu size=%lu\n",
+		          sym->name, symbol.objname, symbol.sympos,
```
joe-lawrence, line 2996: nit: adding `\n` is good, but the second line introduces spaces (in front of `sym->name`) instead of a tab

kpatch-build/create-diff-object.c
```diff
@@ -3234,6 +3313,13 @@ static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
 				special = true;
 
 		list_for_each_entry_safe(rela, safe, &sec->relas, list) {
+			/*
+			 * s390_indirect_jump is a global hidden symbol and is
+			 * present in each object file which references it.
+			 * They can be used as normal relas.
+			 */
+			if (strstr(rela->sym->name, "s390_indirect_jump"))
+				continue;
```
joe-lawrence, line 3322: Are there instances where `s390_indirect_jump` relocation would have `rela->need_dynrela` set?  FWIW, when building the integration tests, I didn't see any examples where this happened.

kpatch-build/kpatch-elf.c
```diff
@@ -193,7 +211,10 @@ void kpatch_create_rela_list(struct kpatch_elf *kelf, struct section *sec)
 			ERROR("could not find rela entry symbol\n");
 		if (rela->sym->sec &&
 		    (rela->sym->sec->sh.sh_flags & SHF_STRINGS)) {
-			rela->string = rela->sym->sec->data->d_buf + rela->addend;
+			if(lc_string(rela))
```
joe-lawrence, line 214: nit: whitespace: `s/if(/if (/`

joe-lawrence (commented):
Hi @sumanthkorikkar , I haven't found much time to review the PR code, but I listed a few nits and questions when taking a quick look.  I'll follow up with some more integration test results...

joe-lawrence:
Hi @sumanthkorikkar , I have been busy working on rebasing the integration tests for RHEL-9, which will be based on v5.14.  It's been slow since I've been fighting a few x86_64 and ppc64le problems at the same time, but I think I'm getting close to finishing.   I setup three branches:

- [integration-tests-kernel-5.14.0-1.el9](https://github.com/dynup/kpatch/compare/master...joe-lawrence:integration-tests-kernel-5.14.0-1.el9?expand=1)
- [integration-tests-kernel-5.14.0-1.el9+workarounds](https://github.com/dynup/kpatch/compare/master...joe-lawrence:integration-tests-kernel-5.14.0-1.el9+workarounds?expand=1) - above plus workarounds for x86_64 and ppc64le issues
- [integration-tests-kernel-5.14.0-1.el9+workarounds+s390x](https://github.com/dynup/kpatch/compare/master...joe-lawrence:integration-tests-kernel-5.14.0-1.el9+workarounds+s390x?expand=1) - above plus your s390x branch

With [integration-tests-kernel-5.14.0-1.el9+workarounds+s390x](https://github.com/dynup/kpatch/compare/master...joe-lawrence:integration-tests-kernel-5.14.0-1.el9+workarounds+s390x?expand=1), I can build all of the integration patches and load all but one with: `make integration-slow KPATCH_BUILD_OPTS="--non-replace"`

Two interesting things to report:

1 - bug-table-section.patch, gcc-isra.patch.build, gcc-mangled-3.patch, macro-callbacks.patch all build, but when inspecting their `~/.kpatch/build.log` file, they all include a create-klp-module error like this:
```
make[3]: Entering directory '/root/.kpatch/src'
  LD [M]  /root/.kpatch/tmp/patch/test-macro-callbacks.o
  CC [M]  /root/.kpatch/tmp/patch/test-macro-callbacks.mod.o
  LD [M]  /root/.kpatch/tmp/patch/test-macro-callbacks.ko
make[3]: Leaving directory '/root/.kpatch/src'
make[2]: Leaving directory '/root/.kpatch/tmp/patch'
/root/kpatch/kpatch-build/create-klp-module: ERROR: tmp.ko: kpatch_mark_grouped_sections: 803: symbol not found
```
However,  kpatch-build produces a .ko and doesn't report failure.  I don't believe this behavior is unique to this merge request, but we should definitely investigate why `ERROR()` doesn't seem to fail the build.

2 - bug-table-section.patch crashes on load once executing replacement code:
```
$ insmod ./livepatch-bug-table-section.ko
(wait for transition to finish)
$ sudo ls

[16986.671403] livepatch: enabling patch 'livepatch_bug_table_section'
[16986.672382] livepatch: 'livepatch_bug_table_section': starting patching transition
[16988.252413] livepatch: 'livepatch_bug_table_section': patching complete
[16996.297859] Unable to handle kernel pointer dereference in virtual kernel address space
[16996.297872] Failing address: c4a8000000005000 TEID: c4a8000000005403
[16996.297873] Fault in home space mode while using kernel ASCE.
[16996.297876] AS:000000018d274007 R3:00000001efff0007 S:00000001effef800 P:000000000000513d
[16996.297978] Oops: 0038 ilc:2 [#1] SMP
[16996.297982] Modules linked in: livepatch_bug_table_section(OEK) ...
[16996.298019] CPU: 1 PID: 1659633 Comm: sudo Kdump: loaded Tainted: G        W  OE K  --------- ---  5.14.0-1.el9.s390x #1
[16996.298021] Hardware name: IBM 8561 LT1 400 (KVM/Linux)
[16996.298022] Krnl PSW : 0704c00180000000 000003ff8045b706 (sysctl_head_grab+0x46/0xb0 [livepatch_bug_table_section])
[16996.298030]            R:0 T:1 IO:1 EX:1 Key:0 M:1 W:0 P:0 AS:3 CC:0 PM:0 RI:0 EA:3
[16996.298032] Krnl GPRS: 0000000180000000 0000000100000000 0000000100000002 00000000805047f8
[16996.298034]            0000000000000001 0000000000000074 00000380066e7c38 000000018ce3ba08
[16996.298036]            0000000000000000 000000018ce3ba08 c4a8000000005820 000000018cdfac60
[16996.298037]            000000017d9e2200 000000018c6cfd40 00000380066e7af8 00000380066e7ac0
[16996.298044] Krnl Code: 000003ff8045b6f8: c4a800000000        lgrl    %r10,000003ff8045b6f8
           000003ff8045b6fe: 582003ac           l       %r2,940
          #000003ff8045b702: a7180000           lhi     %r1,0
          >000003ff8045b706: ba12a000           cs      %r1,%r2,0(%r10)
           000003ff8045b70a: ec16001b007e       cij     %r1,0,6,000003ff8045b740
           000003ff8045b710: e310b0180002       ltg     %r1,24(%r11)
           000003ff8045b716: a774001c           brc     7,000003ff8045b74e
           000003ff8045b71a: eb01b008006a       asi     8(%r11),1
[16996.298056] Call Trace:
[16996.298058]  [<000003ff8045b706>] sysctl_head_grab+0x46/0xb0 [livepatch_bug_table_section]
[16996.298062]  [<000000018c1c3486>] proc_sys_permission+0x56/0xc0
[16996.298068]  [<000000018c11ed7c>] inode_permission+0x10c/0x1e0
[16996.298072]  [<000000018c123006>] link_path_walk.part.0.constprop.0+0x256/0x350
[16996.298075]  [<000000018c12387a>] path_openat+0xaa/0x2b0
[16996.298077]  [<000000018c125790>] do_filp_open+0x90/0x130
[16996.298079]  [<000000018c10bcb8>] do_sys_openat2+0xa8/0x160
[16996.298081]  [<000000018c10c362>] do_sys_open+0x62/0x90
[16996.298083]  [<000000018c6aafdc>] __do_syscall+0x1bc/0x1f0
[16996.298088]  [<000000018c6b84b8>] system_call+0x78/0xa0
[16996.298092] Last Breaking-Event-Address:
[16996.298093]  [<000000018c6cfd40>] __func__.4+0xdc/0xe4
[16996.298099] Kernel panic - not syncing: Fatal exception: panic_on_oops
```
Let me know if you would like a vmlinux/ vmcore for this one.

sumanthkorikkar:
Hi @joe-lawrence,

Thank you. I am able to reproduce livepatch_bug_table_section crash in one of my  machines. I will check these issues.

kpatch-build/create-diff-object.c
sumanthkorikkar: Yes. You are right. I removed it.

sumanthkorikkar:
Hi @joe-lawrence,

v3 Changes:

* Indentation changes
* Handle error, when create-klp-module / create-kpatch-module fails
* Handle group sections properly. The kpatch_mark_grouped_sections should be called before kpatch_reindex_elements() in create-klp-module. This ensures that the  links (sec->sh.sh_link, sec->sh.sh_info) are still intact
  * Reason for OOPS: One of the rela entries (sysctl_lock) for .text.sysctl_grab_head was missing which lead to specification exception. Now, error handling is checked in create-klp-module and rela entry sysctl_lock is present.
 
Thank you

sumanthkorikkar:
* Added Double quotes in checks

kpatch-build/Makefile
```diff
@@ -22,7 +22,7 @@ GCC_PLUGINS_DIR := $(shell gcc -print-file-name=plugin)
 PLUGIN_CFLAGS := $(filter-out -Wconversion, $(CFLAGS))
 PLUGIN_CFLAGS += -shared -I$(GCC_PLUGINS_DIR)/include \
 		   -Igcc-plugins -fPIC -fno-rtti -O2 -Wall
-else
+else ifneq ($(ARCH),s390x)
 $(error Unsupported architecture ${ARCH}, check https://github.com/dynup/kpatch/#supported-architectures)
```
joe-lawrence, line 26: Depending on how confident we feel after this PR, we could update the referenced README.md `#supported-architectures` section to include s390x.  Or would you prefer we merge and test internally before advertising support.

kpatch-build/kpatch-elf.c
```diff
@@ -726,6 +753,119 @@ void kpatch_reindex_elements(struct kpatch_elf *kelf)
...
+		 *    group_section. This is later used to compare if the reindexed
+		 *    section and group_section matches.
+		 * b. Maintain the secnames, symnames to later update
+		 *    group section members sh_link and sh_info.
+		 */
+		groupsec->groupindex = groupsec->index;
+		data = groupsec->data->d_buf;
+		end = groupsec->data->d_buf + groupsec->data->d_size;
+		data++; /* skip first flag word (e.g. GRP_COMDAT) */
+		gsec = malloc(sizeof(*gsec));
```
joe-lawrence, line 780: I may be the only one who resorts to using Valgrind now and then to debug problems in this space, but can we clean up these new allocations.  No need to worry about all the error path scenarios, but at least successful execution will run `kpatch_elf_teardown()` on its way out.

kpatch-build/kpatch-elf.c
```diff
@@ -726,6 +753,119 @@ void kpatch_reindex_elements(struct kpatch_elf *kelf)
...
+		i = 0;
+		list_for_each_entry(sec, &kelf->sections, list) {
+			/* section's groupindex is set, only if it is a group section */
+			if (sec->sh.sh_type != SHT_GROUP)
+				continue;
+			if (sec->groupindex != groupsec->groupindex)
+				continue;
+			data = sec->data->d_buf;
+			end = sec->data->d_buf + sec->data->d_size;
+			newdata = malloc(sec->data->d_size);
```
joe-lawrence, line 837: Likewise for these allocations.

kpatch-build/kpatch-elf.c
```diff
@@ -726,6 +753,119 @@ void kpatch_reindex_elements(struct kpatch_elf *kelf)
...
+				tsec = find_section_by_name(&kelf->sections, ".symtab");
+				if (!tsec)
+					ERROR("section not found");
+
+
+				sec->sh.sh_link = tsec->index;
+				sec->sh.sh_info = tsym->index;
+				data++;
+			}
+			sec->data->d_buf = newdata;
```
joe-lawrence, line 863: Also, I think this leaks the previous `sec->data->d_buf` ?

kpatch-build/kpatch-build
```diff
@@ -1149,6 +1155,7 @@ else
 	rm -f checksum.tmp
 	"$TOOLSDIR"/create-kpatch-module "$TEMPDIR"/patch/tmp_output.o "$TEMPDIR"/patch/output.o 2>&1 | logger 1
 	check_pipe_status create-kpatch-module
+	test "$rc" -ne 0 && die "create-kpatch-module: exited with return code: $rc"
```
joe-lawrence, line 1158: See note in review summary about pulling this out into its own pull request.  Also, just a style nit, but the rest of the script uses `[[ $rc -ne 0 ]] && ...` type syntax.

kpatch-build/create-klp-module.c
```diff
@@ -502,6 +535,8 @@ int main(int argc, char *argv[])
 	kpatch_create_shstrtab(kelf);
 	kpatch_create_strtab(kelf);
 	kpatch_create_symtab(kelf);
+	kpatch_dump_kelf(kelf);
```
joe-lawrence, line 538: Needed for future debug?

joe-lawrence (commented):
Hi @sumanthkorikkar, thanks for updating, this PR is coming along nicely:  I can build all and load all of the integration tests in [integration-tests-kernel-5.14.0-1.el9+workarounds](https://github.com/dynup/kpatch/compare/master...joe-lawrence:integration-tests-kernel-5.14.0-1.el9+workarounds?expand=1) with the latest set of changes.

See review comments in the code, but some higher level PR feedback:
* I opened #1223 and suggest that you spin off the bug fix for ("kpatch-build: Handle error in create-klp-module") into its own PR.  This will separate the s390x-specific work from incidental bugfixes along the way.
* Before we merge, shouldn't ("Enable kpatch build support for s390x" to end of list") move to the back of the commit order?  (Tacking them onto the end is good for review, but once we merge into the tree, we can flip on support as the final step.)
* One additional level of testing we haven't talked about yet are the unit tests.  Take a look at https://github.com/dynup/kpatch-unit-test-objs to see the git submodule that lives under this project's `test/unit/objs`.  @sm00th : what kind of arch-specific unit tests would you like to see when we add s390x support?

sm00th:
> @sm00th : what kind of arch-specific unit tests would you like to see when we add s390x support?

We can start off with the bulk of generic tests we have for x86 and add specific stuff later on as we encounter arch-specific issues.

kpatch-build/kpatch-build
sumanthkorikkar: ok sure. I will send out a separate pull request for this

kpatch-build/kpatch-elf.c
sumanthkorikkar: Thank you. Will do the cleanup of allocations.

kpatch-build/create-klp-module.c
sumanthkorikkar: Yes. I wanted to print the final dump coming out of create-klp-module. For debugging, It was beneficial for me.

kpatch-build/Makefile
sumanthkorikkar: After this is merged, we can update s390x as supported in README.md.

kpatch-build/kpatch-elf.c
sumanthkorikkar: I  couldnt free it (Elf_Data). Modifying the memory referenced by sec->data->d_buf or Freeing this results in segfault.

kpatch-build/kpatch-elf.c
sumanthkorikkar: Cleanup of this allocation performed in kpatch_free_groupsec.

sumanthkorikkar:
Hi @joe-lawrence,

v4 changes:
* Separate pull request for handling error in kpatch-build. Remove this commit here.
* Perform the group section handling cleanup operation in kpatch_free_groupsec.
  * kpatch_elf_teardown is being called for orig obj, patched obj and kelf_out. Hence introduced this function
* Reorganized the commits.
  * Enable build support for s390x is the last commit.
* Moved kpatch_dump_kelf to the end in create-klp-module.c

Thank you

iii-i:
I've committed the GCC patches mentioned in https://github.com/dynup/kpatch/pull/1203#issue-958203216 to the `releases/gcc-11` branch:

https://gcc.gnu.org/pipermail/gcc-patches/2021-September/580562.html
https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=4a62dfbb9a35db3031108db87569955728cc5f65
https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=2335aa8771acd06b082d3e15d9f21ae0a802afd7

joe-lawrence:
> I've committed the GCC patches mentioned in [#1203 (comment)](https://github.com/dynup/kpatch/pull/1203#issue-958203216) to the `releases/gcc-11` branch:
> 
> https://gcc.gnu.org/pipermail/gcc-patches/2021-September/580562.html https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=4a62dfbb9a35db3031108db87569955728cc5f65 https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=2335aa8771acd06b082d3e15d9f21ae0a802afd7

Cool, thanks for the update, Ilya.  Support for s390x is coming together nicely :+1: 

As far as the current outstanding kernel work is concerned, I think we're looking for PeterZ's recent patchset to help the large /  idle cpu cause:  [[PATCH v2 00/11] sched,rcu,context_tracking,livepatch: Improve livepatch task transitions for idle and NOHZ_FULL](https://lore.kernel.org/lkml/20210929151723.162004989@infradead.org/)  Any kernel updates, aside from torvalds/linux@7561c14d8a4d, the already merged linker script update?

sumanthkorikkar:
Hi Joe,

> Hi Sumanth,
> 
> Re: `kpatch_line_macro_change_only()` : looking a bit more into this, I think it may be interesting work for one of our new junior guys, do you mind if he takes a swing at implementing this? 

Ok. I could help with reviewing then.

>Also, do you know if this rule of thumb is accurate for the instruction set: top two MSB of instruction - 00 = 2 byte instruction, 01/10 = 4 bytes, 11 = 6 bytes.

Yes, correct
z/Architecture Principles of Operations:
section Program Execution -> Instructions -> Instruction Formats

The first two bits of the first or only byte of the **op
code** specify the length and format of the instruction,

\<c0\> 04 00 00 00 00 | jgnop   0                | 6 bytes
\<b9\> 04 00 ef            | lgr     %r14,%r15   | 4 bytes
\<07\> fe                      | br      %r14            | 2 bytes

> 
> Re: kernel commits: do you mind updating the PR description with these additional commits. It would be good to collect all recent recommended kernel commits for any distro maintainers out there who may need to backport them. Thanks!

ok sure. Thanks.

sumanthkorikkar:
Hi @joe-lawrence, @jpoimboe , Ping.

I will update the README as s390x supported in this PR. 

Request for review.

Thanks a lot.

joe-lawrence:
> Hi @joe-lawrence, @jpoimboe , Ping.
> Request for review.

Hi @sumanthkorikkar : I believe Josh just got back from PTO, so he's catching up on email/reviews/etc.

> I will update the README as s390x supported in this PR.

I think this would be an updated copy of the prerequisites that you've provided in the MR description:

Prerequisite gcc patches (all backported to **gcc-11** branch):
gcc-mirror/gcc@a1c1b7a888ad `IBM Z: Define NO_PROFILE_COUNTERS`
gcc-mirror/gcc@0990d93dd8a4 `IBM Z: Use @PLT symbols for local functions in 64-bit mode`
gcc-mirror/gcc@935b5226c385 `S/390: New option -mpic-data-is-text-relative`

Prerequisite kernel patches:
v5.16:
torvalds/linux@f6ac18fafcf6 `sched: Improve try_invoke_on_locked_down_task()`
torvalds/linux@9b3c4ab3045e `sched,rcu: Rework try_invoke_on_locked_down_task()`
torvalds/linux@00619f7c650e `sched,livepatch: Use task_call_func()`
torvalds/linux@8850cb663b5c `sched: Simplify wake_up_*idle*()`
torvalds/linux@5de62ea84abd `sched,livepatch: Use wake_up_if_idle()`
torvalds/linux@96611c26dc35 `sched: Improve wake_up_all_idle_cpus() take #2`
v5.14:
torvalds/linux@7561c14d8a4d `s390/vdso: add .got.plt in vdso linker script`

Recommended Patches:
v5.15:
torvalds/linux@de5012b41e5c `s390/ftrace: implement hotpatching`
torvalds/linux@67ccddf86621 `ftrace: Introduce ftrace_need_init_nop()`

I don't know the best spot for this list, but I feel like once the merge request closes, it will be buried in the archive.  Maybe we could link the "s390x" text in the README support list into a small markdown file that hosts an updated list of such patches or other arch-specific quirks?  Have you found anything else s390x-specific worth noting?

Also, I just merged https://github.com/dynup/kpatch-unit-test-objs/pull/36 which adds a bunch of s390x object files to the unit tests.  This MR should bump the `test/unit/objs/` subproject commit to include those new object files.  Then you can give `make unit` a spin and avoid long integration test build times when sanity testing :)  

Thanks again for all your work on implementing this arch.  And oh yeah, we will be starting to look at the follow up macro line change feature soon, so stay tuned for that.

sumanthkorikkar:
Hi Joe,

The list covers all the needed kernel patches. I added these in README now. Thanks for listing.

I applied those patches on top of linux 5.14. Both the kpatch tests and qa_test_klp tests passes.

To apply cleanly, I also applied these patches below, but these are not related to kpatch itself. Hence, havent included them in the list.

e6a901a44f76 rcu: Fix to include first blocked task in stall warning
dc87740c8a68 rcu: Fix stall-warning deadlock due to non-release of rcu_node ->lock

Thanks.

kpatch-build/create-diff-object.c
```diff
@@ -1642,12 +1635,42 @@ static void kpatch_check_func_profiling_calls(struct kpatch_elf *kelf)
...
+		seclink = find_section_by_index(&kelf->sections, *data);
+		if (!seclink)
+			ERROR("group section not found");
+		if (strstr(seclink->name, "s390_indirect_jump"))
+			return true;
+		data++;
+	}
+	return false;
+}
+
```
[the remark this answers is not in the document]
sumanthkorikkar: Hi Josh,

Ok sure. I will add the details regarding the .group and expolines sections in the commit log.

I thought about discarding the .group sections during link time. But Few things which led me to consider reindexing of group sections:

* When indirect branches are added, There could be addition of new expolines in the patched object. This means that the group sections are marked as modified and kpatch does not handle it well. Also new symbols (like __s390_indirect_jump_r14) in .rela.text.*
* Analysing the object files after create-diff-object will not be possible by objdump. (As group sections would not be present and objdump fails without proper indexing). 

So, I think this approach would be better.

Thanks.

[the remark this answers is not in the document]
jpoimboe: >     * When indirect branches are added, There could be addition of new expolines in the patched object. This means that the group sections are marked as modified and kpatch does not handle it well. Also new symbols (like __s390_indirect_jump_r14) in .rela.text.*

Isn't it easy to tell create-diff-object to ignore such changes, e.g. in kpatch_compare_correlated_section()?

>     * Analysing the object files after create-diff-object will not be possible by objdump. (As group sections would not be present and objdump fails without proper indexing).

I don't follow, why would objdump need .group sections for proper indexing?

sumanthkorikkar:
Thanks a lot for the review. I have tried to address your concerns/approaches which were mentioned in the pull request.

I would like to share the details regarding the two possible approaches which I have tried.



1. Explicit expolines generation:
======================
Expoline intro:
------------------
* Expolines is essential to prevent the indirect branch  from
  dynamic branch prediction. This is performed via execute instruction (ex/exrl) in s390x.
 
  - A indirect branch will be replaced with the following when compiled with -mindirect-branch=thunk
    - brasl %r14, __s390x_indirect_jump_r1
    - As per Z Architecture reference manual, The execute instruction causes one target instruction in
      main storage to be executed out of sequence without actually branching to the target instruction.
      This will be useful to prevent against spectre attack
    - Disassembly of __s390x_indirect_jump_r1 is as follows:
            exrl 0, 1f
            j .
            1f: br %r1
            
* The following combinations are needed in the expoline.S.
  * __s390_indirect_jump_r[0-15]
  * __s390_indirect_jump_r[0-15]use_r[0-15]
  * __s390_indirect_branch_[displacement]_r[0-15]use_r[0-15]
 
* This leads to many possible combinations in expoline.S and including that in every kpatch module would be a concern
* This would need inclusion of all the expolines irrespective of the detected expolines in group section references.
 
 
 
2. Group section reindexing approach:
===========================
Group sections intro:
---------------------------
Group sections are used in s390x to deal with the expolines. The expolines are referenced in the group section
and the linker ensures that only one instance of expolines are added from different object files.

Current group section handling in kpatch:
-------------------------------------------------------
* Maintain the group section info before reindexing
  * maintain the secnames and synames before reindexing. This info is later used to update sec->sh.sh_link and sec->sh.sh_info
    in the group section
  * After reindexing, group section indexes references are updated.
* Can rely on the expolines generated by the compiler
* Can avoid many unnecessary expolines addition in the final kpatch module.
* Group sections could also reference other sections in the future. Which means ignoring the group section might not be intended.
            
Possible solution : 
Better group section handling:
---------------------------------------
* Group sections are currently not handled and not correlated in kpatch. However, the pull request 1203 only handles s390x expoline specific group sections.
* However, this could change in the future and probably many other sections could be referenced in group sections(other arch as well?) . So, we could handle it better and make it arch independent. 
* We could also add group section handling testcases for x86, s390x. (After making it arch-independent)
* Also any thoughts on why this group section changes wasnt handled before? 
  
Thank you

jpoimboe:
> * The following combinations are needed in the expoline.S.
>   
>   * __s390_indirect_jump_r[0-15]
>   * __s390_indirect_jump_r[0-15]use_r[0-15]
>   * __s390_indirect_branch_[displacement]_r[0-15]use_r[0-15]
> 
> * This leads to many possible combinations in expoline.S and including that in every kpatch module would be a concern
> 
> * This would need inclusion of all the expolines irrespective of the detected expolines in group section references.

Is that really a problem?
```
# grep s390_indirect System.map
000000000011a4b0 T __s390_indirect_jump_r1
000000000011a4bc T __s390_indirect_jump_r9
000000000011a4c8 T __s390_indirect_jump_r14
000000000011a4d4 T __s390_indirect_jump_r4
000000000011a4e0 T __s390_indirect_jump_r1use_r9
000000000011a4ec T __s390_indirect_jump_r1use_r14
000000000011a4f8 T __s390_indirect_jump_r7
000000000011a504 T __s390_indirect_jump_r13use_r14
000000000011a510 T __s390_indirect_jump_r10
000000000011a51c T __s390_indirect_jump_r1use_r1
000000000011a528 T __s390_indirect_jump_r8
000000000011a534 T __s390_indirect_jump_r6
000000000011a540 T __s390_indirect_jump_r13
000000000011a54c T __s390_indirect_jump_r2
000000000011a558 T __s390_indirect_jump_r3
000000000011a564 T __s390_indirect_jump_r5
000000000011a570 T __s390_indirect_jump_r11
```
In vmlinux I count 17 functions at 12 bytes each, for a total of 204 bytes.  I don't see how it's worth all the added complexity to add group section support, only to save ~200 bytes in the patch module.  Not only does the complexity affect both create-diff-object and create-klp-module, it also gets spread to other arches unnecessarily.

In contrast, if you have a hard-coded expoline.S which is always linked in, it isolates most of the changes to a single file, and a single arch.  And it doesn't spread to create-diff-object and create-klp-module.

> * Group sections are currently not handled and not correlated in kpatch. However, the pull request 1203 only handles s390x expoline specific group sections.
> 
> * However, this could change in the future and probably other sections could be referenced in group sections (other arch  could have this as well?) . So, we could handle it better and make it arch independent.
> 
> * We could also add group section handling testcases for x86, s390x. (After making it arch-independent)
> 
> * Also any thoughts on why this group section changes wasn't handled before?

group sections are an obscure feature.  As far as I can tell, the s390 `__THUNK_PROLOG_NAME` macro is the only user of group sections in the entire kernel.  Even in that one case, their usage seems unnecessary.  The upstream s390 kernel could instead use `-mindirect-branch=thunk-extern` like x86 and other arches do, and have an expoline.S which defines all the expoline thunks in a single file.  For example see arch/x86/lib/retpoline.S.  In fact that would be preferable, because it would be good to converge s390's behavior with the other arches.  Group sections are odd and confusing and unwieldy, they should be avoided unless they actually provide a real benefit.

I'd be surprised if any other group section usages showed up in the kernel anytime soon, but if they did, then we could add generic support for them at the time if that becomes necessary.  But right now I don't see a good reason to start supporting them.  It adds a lot of unnecessary complexity.

kpatch-build already has a lot of inherent complexity due to the nature of supporting so many kernels, compilers and arches, so we have to be vigilant about simplifying as much as possible where we can.  If you remove the group section support in favor of hard-coded expoline.S which is added at link time, I think this patch set becomes quite simple.

sumanthkorikkar:
Hi All,

With the following commits, group section is avoided in the kernel and
the number of expolines is restricted to minimum. 

kernel commit:
- 602bf1687e6f s390/nospec: align and size extern thunks
- 1d2ad084800e s390/nospec: add an option to use thunk-extern
- eed38cd2f46f s390/nospec: generate single register thunks if possible
- 2268169c14e5 s390: remove unused expoline to BC instructions
- f0003a9e4c18 s390/entry: remove unused expoline thunk

gcc commit:
- gcc-mirror/gcc@8753b13 IBM Z: fix section type conflict with -mindirect-branch-table


The kernel should be compiled with -mindirect-branch=thunk-extern and -mfunction-return=thunk-extern.
When modules are far away from the kernel image, the expolines from kernel image cannot be used.
Hence expolines __s390_indirect_jump_r[1-15] is included in every module (which are done in the kernel commit)

This reduces the kpatch handling to minimum and no group section handling is needed altogether.

Thanks.

sumanthkorikkar:
Hi @joe-lawrence @jpoimboe

As per the discussion, I have integrated the suggestions. I request you to review it. Thank you all

jpoimboe:
Hi @sumanthkorikkar I will be reviewing shortly.

joe-lawrence:
Hi @sumanthkorikkar : sorry for the delay, I've been trying to line up backports for testing.  Once I have compiler and kernel in place, I'll give it a spin and let you know how it turns out.

kpatch-build/create-diff-object.c
```diff
@@ -3728,6 +3818,13 @@ static void kpatch_find_func_profiling_calls(struct kpatch_elf *kelf)
 
 			sym->has_func_profiling = 1;
 			break;
+		case S390:
+			insn = sym->sec->data->d_buf;
```
jpoimboe, line 3822: A short comment here with the name of the instruction would be helpful.

sumanthkorikkar:
Thank you for the review. Some of the Pending comments are now shown. Please ignore that. Sorry, Didnt know that I have to submit it as review in github

kpatch-build/create-diff-object.c
```diff
@@ -2576,8 +2630,24 @@ static void kpatch_check_relocations(struct kpatch_elf *kelf)
...
+			 * 0000000000000090  000000f90000001a R_390_GOTENT 0000000000000000 __key.18 + 2
+			 *
+			 *   8e:   c4 48 00 00 00 00       lgrl    %r4,8e <copy_signal+0x8e>
+			 *               90: R_390_GOTENT        __key.18+0x2
+			 * which is a valid relocation
+			 */
 			if (rela->sym->sec) {
 				sdata = rela->sym->sec->data;
+				if (!sdata->d_size)
+					continue;
```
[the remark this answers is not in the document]
jpoimboe: This `__key` thing is an interesting case.  In a way, kpatch_check_relocations() is right to warn about it.  It comes from:
```
copy_signal()
    init_waitqueue_head()
        static struct lock_class_key __key;
        __init_waitqueue_head((wq_head), #wq_head, &__key);
```
And notice that with CONFIG_LOCKDEP=n, `struct lock_class_key` is size zero:
```
/*
 * The class key takes no space if lockdep is disabled:
 */
struct lock_class_key { };
```
So this is the code intentionally sending a bad pointer to a function... :-/  Luckily the function doesn't actually use the pointer with CONFIG_LOCKDEP=n.

So is `__key` the only problem?  If so then we should look for another way to ignore that warning.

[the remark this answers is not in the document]
sumanthkorikkar: I could find many empty structures, but this __key pointer passed to a function is unique.
Then, I will  skip the reloc check when the rela->sym->name is __key && d_size = 0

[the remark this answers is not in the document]
sumanthkorikkar: Added the above mentioned comments in skip_check_relocation().

kpatch-build/kpatch-elf.c
```diff
@@ -206,7 +208,60 @@ void kpatch_create_rela_list(struct kpatch_elf *kelf, struct section *sec)
...
+			 * 	...
+			 * <.LC1>:
+			 *  	...
+			 *
+			 */
+			if ((rela->type == R_390_PC32 && strstr(sec->name, ".rela__bug_table")) ||
+				(rela->type == R_390_GOTENT && strstr(rela->sym->name, ".LC")))
+				rela->string = rela->sym->sec->data->d_buf + rela->sym->sym.st_value;
+			else
+				rela->string = rela->sym->sec->data->d_buf + rela->addend;
```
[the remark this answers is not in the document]
jpoimboe: The addend of 4 really didn't make sense to me, but after staring at it for a while, I now get it.  It's a quirk of the bug table and CONFIG_GENERIC_BUG_RELATIVE_POINTERS.
```
Relocation section [214] '.rela__bug_table' for section [213] '__bug_table' at offset 0x79c78 contains 20 entries:
  000000000000000000  390_PC32        000000000000000000    +350 .text.alloc_thread_stack_node
  0x0000000000000004  390_PC32        000000000000000000      +4 .L1^B14
```
That comes from the following code
```
#define __EMIT_BUG(x) do {					\
	asm_inline volatile(					\
		"0:	mc	0,0\n"				\
		".section .rodata.str,\"aMS\",@progbits,1\n"	\
		"1:	.asciz	\""__FILE__"\"\n"		\
		".previous\n"					\
		".section __bug_table,\"awM\",@progbits,%2\n"	\
		"2:	.long	0b-2b,1b-2b\n"			\
		"	.short	%0,%1\n"			\
		"	.org	2b+%2\n"			\
		".previous\n"					\
		: : "i" (__LINE__),				\
		    "i" (x),					\
		    "i" (sizeof(struct bug_entry)));		\
} while (0)
```
Specifically this line:
```
		"2:	.long	0b-2b,1b-2b\n"			\
```
`1b-2b` makes a PC- relative pointer to the `__FILE__` string.  Usually a PC-relative pointer is calculated starting from the address of the pointer itself.  i.e., if this were a "normal" PC32 relocation, it would be:
```
		"2:	.long	0b-2b\n"			\
		"3:	.long	1b-3b\n"			\
```
But instead, the calculation starts from the beginning of the `bug_entry` struct, which is **four** bytes before the file string pointer.  Thus the 4-byte addend.

I guess the kernel does it that way so that bug_get_file_line() works as expected for all arches.
```
  CONFIG_GENERIC_BUG_RELATIVE_POINTERS - use 32-bit pointers relative to
	the containing struct bug_entry for bug_addr and file.
```
It works, but it's still it's a weird way to do it.  Maybe I'll post a patch upstream to fix it so we can drop this change.

jpoimboe:
Here's the upstream patch to fix the bug table string +4 addend issue:  https://lkml.kernel.org/r/afddb4548e93f6458ec1d9ec185a834c348eda33.1651798983.git.jpoimboe@kernel.org

kpatch-build/create-diff-object.c
```diff
@@ -1774,8 +1803,8 @@ static void kpatch_include_symbol(struct symbol *sym)
 
 static bool is_string_literal_section(struct section *sec)
 {
-	return !strncmp(sec->name, ".rodata.", 8) &&
-	       strstr(sec->name, ".str1.");
+	return !strcmp(sec->name, ".rodata.str") ||
+		(!strncmp(sec->name, ".rodata.", 8) && strstr(sec->name, ".str1."));
```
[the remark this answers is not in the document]
sumanthkorikkar: Should we simplify this to !strncmp(sec->name, ".rodata.", 8) && strstr(sec->name, ".str") to cover both cases?

kpatch-build/create-diff-object.c
```diff
@@ -2642,6 +2707,22 @@ static void kpatch_regenerate_orc_sections(struct kpatch_elf *kelf)
...
+	 * zero. Pointer __key of type lock_class_key is passed to
+	 * __init_waitqueue_head(), eventhough the pointer is not used.
+	 * Ignore out-of-range relocation in this case.
+	 */
+	if (!sdata->d_size &&
+	    (!strncmp(rela->sym->name, "__key.", 6) ||
+	    !strncmp(rela->sym->name, ".bss.__key.", 11)))
+		return true;
+	else
+		return false;
```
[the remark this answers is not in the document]
sumanthkorikkar: It would work. But it would skip valid ones like:
rela->sym->name: __func__.0 sec->name: .rela.text.audit_log_start.part.0 rela->sym->sec->sh.sh_size: 16 sdata->size: 16  rela->sym->sym.st_value: 0,

Should we check for sym->type == STT_OBJECT && !sdata->d_size instead ?

kpatch-build/create-diff-object.c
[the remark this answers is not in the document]
jpoimboe: Sounds good

kpatch-build/create-diff-object.c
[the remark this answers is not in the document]
jpoimboe: Sorry, typo.  I meant to check the symbol size, something like `sym->type == STT_OBJECT && !sym->sym.st_size`

kpatch-build/kpatch-elf.c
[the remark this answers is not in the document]
sumanthkorikkar: I tried out the bug table fix patch. Thanks, However, I still believe, the check for NOTYPE vs SECTION would be required. Correct me if I am wrong.
 

Example usecase on s390x:
<details>
<summary>
When type is NOTYPE: -> Need to consider the rela->sym->sym.st_value
</summary>
sec->name: .rela.init.text rela->sym->name : .LC24 rela->sym->sec->name: .rodata.str1.2 rela->addend: 2 rela->sym->sym.st_value:32 rela->sym->type: 0 rela->type: 26 rela->string: sighand_cache.

</details>

<details>
<summary>
When type is STT_SECTION -> Need to consider the rela->addend.
</summary>
sec->name: .rela.data.rel.event_exit__clone rela->sym->name : .rodata.str1.2 rela->sym->sec->name: .rodata.str1.2 rela->addend: 270 rela->sym->sym.st_value:0 rela->sym->type: 3 rela->type: 22 rela->string: sys_exit_clone
</details>

kpatch-build/create-diff-object.c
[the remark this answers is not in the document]
sumanthkorikkar: Added.

kpatch-build/create-diff-object.c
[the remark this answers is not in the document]
sumanthkorikkar: Added

kpatch-build/kpatch-elf.c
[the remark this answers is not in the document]
jpoimboe: With the bug_entry fix is it possible now to combine them into a single case?
```
rela->string = rela->sym->sec->data->d_buf + rela->sym->st.value + rela->addend + rela_add_off(rela);
```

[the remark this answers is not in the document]
sumanthkorikkar: * I think, the first approach keeps it simple. 

* I tried out the second proposed approach (Please see the attached patch):  [rela_add_off_patch.txt](https://github.com/dynup/kpatch/files/8667543/rela_add_off_patch.txt)
  - Involves code duplication in kpatch_replace_sections_syms() and rela_add_off()
  - Duplicated it because of continue statement in kpatch_replace_sections_syms().
  - Need to export rela_insn(), because it is being used in rela_add_off().
  - Also 3 args needed for rela_add_off(), As it calls rela_insn() for x86 case.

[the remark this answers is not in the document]
jpoimboe: There's no need for the code duplication.  The 'continue' is probably a bug, it should instead be an ERROR for an unhandled type.  There are several cleanups needed, I'll open a pull request.

jpoimboe:
Hi @sumanthkorikkar, the cleanups have been merged with #1269, and the bug_entry patch has been accepted upstream in the tip tree: https://git.kernel.org/pub/scm/linux/kernel/git/tip/tip.git/commit/?h=x86/core&id=69505e3d9a39a988aaed9b58aa6b3482238f6516

Can you rebase on master and also maybe mention the bug_entry patch in the dependencies (assuming it's still needed in light of the latest changes)?

Hopefully this PR will be ready to merge soon.  Thanks!

sumanthkorikkar:
Thank you for the review. Rebased the changes.

doc/s390-upstream-prerequisites.md
```diff
@@ -0,0 +1,32 @@
+### s390 backporting
+
+**Prerequisite gcc patches (all backported to releases/gcc-11 branch):**
+- gcc-mirror/gcc@a1c1b7a IBM Z: Define NO_PROFILE_COUNTERS
+- gcc-mirror/gcc@0990d93 IBM Z: Use @PLT symbols for local functions in 64-bit mode
+- gcc-mirror/gcc@935b522 S/390: New option -mpic-data-is-text-relative
+- gcc-mirror/gcc@8753b13 IBM Z: fix section type conflict with -mindirect-branch-table
+
+**Prerequisite kernel patches:**
+- 69505e3d9a39 bug: Use normal relative pointers in 'struct bug_entry'
```
jpoimboe, line 10: Most likely this will end up in 5.19, so it's probably safe to add a v5.19 header for this one.

sumanthkorikkar: Added

jpoimboe:
@sumanthkorikkar Thanks for your patience!  This looks good to me.

jpoimboe (approved):

joe-lawrence:
I know there are still some upstream kernel patches that need to make their way down to distro kernels, but shall we merge this now before the main branch drifts any further?  @sumanthkorikkar @jpoimboe

jpoimboe:
@joe-lawrence do we have the integration testing CI setup for testing s390 with a new enough kernel/toolchain?  I think that would be helpful, maybe a separate issue can be opened to track that (for whatever "tracking" means considering our unfortunately large number of neglected open issues).

But we at least have unit tests, and these patches are very nicely self contained, so yeah, I think it would be a good idea to go ahead and merge this now before it starts getting conflicts again.

sumanthkorikkar:
> I know there are still some upstream kernel patches that need to make their way down to distro kernels, but shall we merge this now before the main branch drifts any further? @sumanthkorikkar @jpoimboe

Yes, Looking forward to get this into the main branch. Thanks

joe-lawrence:
> @joe-lawrence do we have the integration testing CI setup for testing s390 with a new enough kernel/toolchain? I think that would be helpful, maybe a separate issue can be opened to track that (for whatever "tracking" means considering our unfortunately large number of neglected open issues).
> 

Compiler support should be available, we're still waiting on the external expolines MR and a subsequent ftrace backport (probably targeting 9.2 at this point). Since CI is run internally, we can always add it to one of our TODO lists.

joe-lawrence:
> > I know there are still some upstream kernel patches that need to make their way down to distro kernels, but shall we merge this now before the main branch drifts any further? @sumanthkorikkar @jpoimboe
> 
> Yes, Looking forward to get this into the main branch. Thanks

Great, thanks for all your work on this functionality.

Regarding kernel backports, I posted this question [1] the other day.  It popped up as we noticed a missing expolines.o when building the tools/testing/selftests/bpf/bpf_testmod.ko (which builds as a sort of OOT module).

[1] https://lore.kernel.org/linux-s390/8569429d-57f8-a0cf-8b17-1291f6973d32@redhat.com/T/#u

sumanthkorikkar:
> Regarding kernel backports, I posted this question [1] the other day. It popped up as we noticed a missing expolines.o when building the tools/testing/selftests/bpf/bpf_testmod.ko (which builds as a sort of OOT module).
> 
> [1] https://lore.kernel.org/linux-s390/8569429d-57f8-a0cf-8b17-1291f6973d32@redhat.com/T/#u

Noted. Thanks. We are looking into that.

jpoimboe:
> Compiler support should be available, we're still waiting on the external expolines MR and a subsequent ftrace backport (probably targeting 9.2 at this point). Since CI is run internally, we can always add it to one of our TODO lists.

The CI could be on a v5.18 kernel which I think has all the prereqs.

joe-lawrence:
> > Compiler support should be available, we're still waiting on the external expolines MR and a subsequent ftrace backport (probably targeting 9.2 at this point). Since CI is run internally, we can always add it to one of our TODO lists.
> 
> The CI could be on a v5.18 kernel which I think has all the prereqs.

@sm00th : have we ever setup a CI instance running w/upstream kernel?

sm00th:
> @sm00th : have we ever setup a CI instance running w/upstream kernel?

The beaker jobs have `x86_64/ppc64le upstream kernel` recipes. They test 5.10.11 atm, so will need config updates and maybe something else to switch over to 5.18.

borntraeger:
Any update on this pull request? Is anything blocking this from being merged?

joe-lawrence:
> Any update on this pull request? Is anything blocking this from being merged?

Nothing more to do for this PR, however we would like to add s390x support to our internal integration tests -- these automatically run whenever a PR is posted or updated and when the master branch is updated.  Unfortunately the rhel kernel has not yet backported all the needed kernel patches (internal build/packaging is blocked on [1]), so that means we need to target an upstream v5.18 kernel.  For that, the integration tests needed to be rebased [2] and tested for all three arches (x86_64, ppc64le, s390x).  While doing so, I regenerated the unit test object files [3] so that they now include the external expolines.  There is a bit more to it (bumping the unit test submodule reference, updating our internal integration test jobs, etc.) that are independent from this PR, but I was hoping to line all that up to run when we merge it.  If the stars don't align by next week, I'll just merge the PR and manually kick off those internal jobs.

[1] https://lore.kernel.org/linux-s390/8569429d-57f8-a0cf-8b17-1291f6973d32@redhat.com/T/#u
[2] https://github.com/dynup/kpatch/pull/1275
[3] https://github.com/dynup/kpatch-unit-test-objs/pull/41

joe-lawrence:
5.18 integration tests merged, let's get this one in.  I'll figure out internal testing later.

PR #1204, opened 2021-08-03
Support for multiple source files

sm00th:
This patchset allows create-diff-object to work on objectfiles built from multiple source files.


kpatch-build/lookup.c
```diff
@@ -407,36 +292,35 @@ void lookup_close(struct lookup_table *table)
...
+		if (sym->type == STT_FILE) {
+			file = sym->name;
 			continue;
 		}
 
-		if (!in_file)
-			continue;
-
-		if (sym->type == STT_FILE)
-			break;
```
jpoimboe, line 434: I'm not quite convinced this won't cause a regression.  It's possible to have multiple STT_FILE symbols with the same name.  How can we be sure this is the right file?  For example, issue #604.

sm00th: I guessed that shouldn't be that easy but failed to find this case. Pushed v2 that does locals_match() for every STT_FILE section in the file and stores lookup_table->local_syms in each symbol.

kpatch-build/lookup.c
```diff
@@ -108,9 +107,15 @@ static int locals_match(struct lookup_table *table, int idx,
 			continue;
 
 		found = 0;
-		for (child = child_locals; child->name; child++) {
-			if (child->type == sym->type &&
-			    !strcmp(child->name, sym->name)) {
+		iter_sym = file_sym;
```
jpoimboe, line 110: The variable names were already fairly confusing in this function, but the addition of `struct symbol` to this file confuses the naming even more.  There's no logical difference between `sym` and `iter_sym` when reading the code.

I'd suggest making the following naming changes here, as well as in other places with `struct object_symbol *sym`.

- rename `sym` to `table_sym`.
- rename `iter_sym` to `sym`.

That way, `sym` has its traditional meaning (an instance of `struct symbol *`) and `table_sym` is more appropriately named.

kpatch-build/lookup.c
```diff
@@ -108,9 +107,15 @@ static int locals_match(struct lookup_table *table, int idx,
...
+		list_for_each_entry_continue(iter_sym, sym_list, list) {
+			if (iter_sym->type == STT_FILE)
+				break;
+			if (iter_sym->bind != STB_LOCAL)
+				continue;
+
```
jpoimboe, line 116: To match the previous functionality I think we also want to make sure it's either an STT_FUNC or an STT_OBJECT (and otherwise `continue`).

kpatch-build/lookup.c
```diff
@@ -135,12 +147,11 @@ static int locals_match(struct lookup_table *table, int idx,
 				break;
 			if (sym->bind != STB_LOCAL)
 				continue;
-			if (sym->type != STT_FUNC && sym->type != STT_OBJECT)
-				continue;
```
jpoimboe, line 139: If my above suggestion to check for STT_{FUNC,OBJECT} is taken, this hunk probably doesn't need removed.

kpatch-build/lookup.c
```diff
@@ -153,32 +164,46 @@ static int locals_match(struct lookup_table *table, int idx,
...
+		      file_sym->name, table->objname);
+
+	list_for_each_entry_continue(file_sym, sym_list, list) {
+		if (file_sym->type == STT_FILE)
+			break;
+		file_sym->lookup_table_sym = lookup_table_sym;
+	}
+}
+
+static void link_local_symbols(struct lookup_table *table, struct kpatch_elf *kelf)
```
jpoimboe, line 199: `link_local_symbols()` makes it sound like something different than finding, when in fact it's just a loop around `find_local_syms()`.  Maybe call it `find_local_syms_multiple()` or so?  Also I think this function deserves a comment describing why it's doing this.  I assume it's for analyzing a bigger .o file like a built-in.o or vmlinux.o.

Similarly, the commit message needs more detail, as "support object-files built from multiple sources" didn't make sense to me until after I studied the code.  The commit message should also give more info about what problem is being solved and why.

kpatch-build/kpatch-elf.h
```diff
@@ -76,6 +76,7 @@ struct symbol {
 	struct section *sec;
 	GElf_Sym sym;
 	char *name;
+	struct object_symbol *lookup_table_sym;
```
jpoimboe, line 79: Technically this should probably be called `lookup_table_file_sym`, since it points to the lookup table's FILE symbol rather than the corresponding symbol itself.

That said, maybe it would be simpler to just set it to the exact corresponding symbol to begin with.  locals_match() already finds the corresponding symbol -- twice, in fact.  Then, later, lookup_local_symbol() finds it a third time.  Can we just set `sym->lookup_table_sym` to the corresponding symbol to begin with, in find_local_syms() and/or locals_match()?  Then lookup_local_symbol() becomes redundant.

kpatch-build/lookup.c
```diff
@@ -92,9 +91,9 @@ static int maybe_discarded_sym(const char *name)
 }
 
 static int locals_match(struct lookup_table *table, int idx,
```
jpoimboe, line 93: While you're at it, can you change locals_match() and maybe_discarded_sym() to return bool?  Their int return codes are opposite than normal kernel error code expectations, making them extra confusing.  Might as well clean those up while we're touching the code.

kpatch-build/create-diff-object.c
```diff
@@ -925,6 +925,10 @@ static void __kpatch_correlate_section(struct section *sec1, struct section *sec
 static void kpatch_correlate_symbol(struct symbol *sym1, struct symbol *sym2)
 {
 	CORRELATE_ELEMENT(sym1, sym2, "symbol");
+	if (sym1->lookup_table_sym && !sym2->lookup_table_sym)
+		sym2->lookup_table_sym = sym1->lookup_table_sym;
+	else if (!sym1->lookup_table_sym && sym2->lookup_table_sym)
+		sym1->lookup_table_sym = sym2->lookup_table_sym;
```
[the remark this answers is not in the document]
sm00th: Yes, all current usage of kpatch_correlate_symbol() follows this logic, I was safeguarding against possible future code, but maybe making expectations clear through argument names will be enough.

kpatch-build/lookup.c
sm00th: Thanks, good suggestion. I was struggling with variable naming in this patchset, at some point I had lookup_sym->lookup_sym in lookup_local_symbol().

kpatch-build/lookup.c
sm00th: This is actually intentional, so is the removal in the next comment. The very next statement is `iter_sym->type == sym->type` which does make sure the type is one of those possibilities so the check was redundant. Or do you think it is worth keeping it for clarity?

kpatch-build/lookup.c
sm00th: Agreed, will do.

kpatch-build/kpatch-elf.h
sm00th: Ah, I see. That makes sense, I'll see if it will work.

kpatch-build/lookup.c
sm00th: Ok

kpatch-build/lookup.c
jpoimboe: Ah I see now.  Never mind...

kpatch-build/kpatch-elf.h
sm00th: After looking into this I currently don't see a straightforward way of dropping `lookup_local_symbol()`. The only reason is `sympos` calculation it does. Because of that we'll either have to traverse the whole lookup table for each symbol in `find_local_syms()` (which will make things worse) or keep a list of seen symbol names and write down sympos in `symtab_read()` but I am not sure it is worth doing that. Any thoughts? Can we leave it as is for now?

kpatch-build/create-diff-object.c
```diff
@@ -917,87 +917,95 @@ do {								\
 	elem->twin = NULL;					\
 } while (0)
 
-static void __kpatch_correlate_section(struct section *sec1, struct section *sec2)
+static void __kpatch_correlate_section(struct section *sec_base,
+		struct section *sec_patched)
 {
-	CORRELATE_ELEMENT(sec1, sec2, "section");
+	CORRELATE_ELEMENT(sec_base, sec_patched, "section");
```
jpoimboe, line 923: This is a nice cleanup IMO.  The new naming helps to codify certain existing assumptions.

While making such a broad change, I'd suggest one more naming tweak (but arguments to the contrary are welcome).  I've long regretted the use of the term "base" in this code.  I'd argue that "orig" is more descriptive than "base" (and matches the directory names in ~/.kpatch/tmp/.)  Since you're already doing a mass renaming which adds more "base", I'd suggest going ahead and doing s/base/orig/g pretty much everywhere -- if there are no objections.

And sorry about pulling this PR into the weeds :-)

jpoimboe (commented):
Would it be possible to add a unit test?  vmlinux.o is probably too big, but either a built-in.o file or some manually linked version of multiple .o's might be useful.

sm00th:
v3: base->orig rename. I've also created https://github.com/dynup/kpatch-unit-test-objs/pull/34 that will be added to this PR once merged.

jpoimboe:
Looks good to me.

sm00th:
v3: included unit-tests and rebased on top of #1207

jpoimboe (approved):

joe-lawrence:
> This patchset allows create-diff-object to work on objectfiles built from multiple source files.

This will eventually help support kernel-LTO, right?  
Are there  circumstances today, without LTO, where we would see such files from the kernel build?
And finally, what is inside `multifile-pr-1204.*.o` and could they be created for ppc64le as well?

The code looks good to me, including all the cleanups.  (Btw, if you think this PR is ready for merge, ok, if not ... would it be easier to spin off the integration test and cleanup commits to make future rebases easier?)

sm00th:
> This will eventually help support kernel-LTO, right?

Right, with LTO we won't have the luxury to work with individual objectfiles.

> Are there circumstances today, without LTO, where we would see such files from the kernel build?

No, not to my knowledge at least. This series doesn't make single source file case any worse, just makes create-diff-object more flexible.

> And finally, what is inside `multifile-pr-1204.*.o`

From the [test-objs commit message](https://github.com/dynup/kpatch-unit-test-objs/commit/d39146043aa3e3f8f585bb22da303e04fd397e78):

```
Adding unit tests for the case when the objectfile is built from
multiple source files. Added objectfiles were created by building our
standard test-patches for meminfo and cmdline and then linking
fs/proc/meminfo.o and fs/proc/cmdline.o together with "ld -r".
```

So this is not exactly a real-life scenario.

> and could they be created for ppc64le as well?

Do you expect ppc64le elfs to have something special about them in this case? If not I don't see any reason to add them since they will be running through the same code. I think we should focus mostly on x86_64 since they are ran more frequently (travis/github runs only on x86_64) and only add ppc64le ones when that case is different.
 
> The code looks good to me, including all the cleanups. (Btw, if you think this PR is ready for merge, ok, if not ... would it be easier to spin off the integration test and cleanup commits to make future rebases easier?)

It is ready, just wanted to give you a chance to take a look. Will merge it now and we can add unit tests later if it comes to that.

PR #1209, opened 2021-08-17
Always set CONFIG_DEBUG_SECTION_MISMATCH for Ubuntu/Debian

luckycat889:
I think I have concluded that stock Ubuntu kernels are built with CONFIG_DEBUG_SECTION_MISMATCH=y.

This CONFIG controls whether or not '-fno-inline-functions-called-once' is set during compilation.
For 'kpatch' this is naturally important; I see instances where there is a long c file with a lot of 'static' functions in it called only once from a public function.
And if it's not too complex, the compiler will inline the function, UNLESS the above flag is on.
As anyone knows, this could pose problems for the kpatch tool

However, the perhaps strange thing is that it's not on in the .config file provided in the 'modules' package.
Instead look in the following files, using Ubuntu 21.04 as an example:

linux-5.11.0/debian/rules.d/0-common-vars.mk:255
...
kmake = make ARCH=$(build_arch)
CROSS_COMPILE=$(CROSS_COMPILE)
KERNELVERSION=$(abi_release)-$(target_flavour)
CONFIG_DEBUG_SECTION_MISMATCH=y
KBUILD_BUILD_VERSION="$(uploadnum)"
LOCALVERSION= localver-extra=
CFLAGS_MODULE="-DPKG_ABI=$(abinum)"
PYTHON=$(PYTHON)
ifneq ($(LOCAL_ENV_CC),)
kmake += CC="$(LOCAL_ENV_CC)" DISTCC_HOSTS="$(LOCAL_ENV_DISTCC_HOSTS)"
endif

linux-5.11.0/Makefile:914
...
ifdef CONFIG_DEBUG_SECTION_MISMATCH
KBUILD_CFLAGS += $(call cc-option, -fno-inline-functions-called-once)
endif

If one only uses linux-5.11.0/Makefile to build as kpatch does it, then this is not set.
However, the debian-style scripts used during official builds, see https://wiki.ubuntu.com/Kernel/BuildYourOwnKernel for a good reference, the build is kicked off with:
$ LANG=C fakeroot debian/rules binary
This is debian's way of building the kernel + a boatload of DKMS modules and debian packaging (This takes forever to build....thus not desired for kpatch)

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -713,6 +713,7 @@ elif [[ "$DISTRO" = ubuntu ]] || [[ "$DISTRO" = debian ]]; then
 	fi
 
 	export PATH="/usr/lib/ccache:$PATH"
+	export CONFIG_DEBUG_SECTION_MISMATCH=y
 fi
 
 find_dirs || die "can't find supporting tools"


sm00th:
I don't think this is a good idea. The patch doesn't take into account the kernels not built with debian scripts. It also also assumes that this always has and will be like that on every version of debian/ubuntu, which I don't think is a fair assumption.

If debian's scripts fiddle with compiler options regardless of kernel config (which I find alarming) we'll need to build the kernel using those to guarantee they are the same.

luckycat889:
The patch is within the 'if debian or ubuntu' enclosure, so it's only active when building debian or ubuntu, which use the debian script harness. Well, it's on for all ubuntu from 16.04 and newer, which is everything livepatch/kpatch works on (kpatch doesn't work on v4.4 on 14.04 anyway). If now the scripts change in the future, and this bizarre way of enabling a config option is removed, then it's just a matter of refining kpatch scripts.
Point is, basically kpatch doesn't work hooking functions where the function is calling file-local static functions where inlining changes depending on the compile-time flag discussed.

Yes, it is alarming and bizarre, but it is also reality.....
And yes, if there is a better way of enabling the option, that would be fine. The patch suggested aimed to mimic the debian scripts

nunojpg:
@luckycat889 I found this after I tracked down some missing static functions on the object file in Ubuntu 22.04.

It took me over 1 day.

This patch does not have any effect if you are not under Debian / Ubuntu and applies the best guess option for Ubuntu.

I can't think of any rational to not have this merged.

Please merge it.

jpoimboe:
Right, we used to have some folks from Canonical contributing to kpatch, but that dried up years ago.

It's very unfortunate that CONFIG_DEBUG_SECTION_MISMATCH gets hacked in like this and doesn't even show up in the .config file.

Plus the option isn't very useful for a production kernel and slows down performance.  Seems really odd to have it enabled at all.

That said I'd be ok with this patch if somebody can verify it doesn't break Debian.  Do they have the same hack?

john-cabaj:
I’m coming into this issue fresh, but I can take a look at which kernels use this config option and cross-reference to which we may livepatch.

nunojpg:
With this patch I managed to generate a livepatch, but it wouldn't load due to modversions.
I come to the conclusion that there are so many issues with Ubuntu to use kpatch that I just abandoned the idea.
So I will pull back on what I said before. It only makes sense to merge this if someone can get it to actually generate a valid patch.

john-cabaj:
I've looked into the main Ubuntu kernels, and agree with @luckycat889's assessment:

CONFIG_DEBUG_SECTION_MISMATCH=y is used when building a kernel through the Debian scripts, and has been for the last 12 years or so. This is how Ubuntu kernels should be built. However, generally all architectures have CONFIG_DEBUG_SECTION_MISMATCH=n in the .config. I'll continue chasing this down because having this configuration overwritten seems like an error-prone approach. At present, this patch makes sense to merge.

PR #1211, opened 2021-08-22
create-diff-object: fix use after free in kpatch-check-relocations()

joe-lawrence:
Building data-read-mostly.patch on rhel-9.0-beta for ppc64le leads to a
segmentation fault:

    Program received signal SIGSEGV, Segmentation fault.
    kpatch_check_relocations (kelf=0x10040490) at create-diff-object.c:2571
    2571                                    sdata = rela->sym->sec->data;
    (gdb) bt
    (gdb) p rela->sym->sec->data
    Cannot access memory at address 0x160000007e

Valgrind narrows the problem down to invalid reads through rela->sym in
kpatch-check-relocations().

The culprits are kpatch_create_intermediate_sections(), which marks
symbols referenced by rela sections that are now dynrelas to be
stripped, and kpatch_strip_unneeded_syms(), which removes and frees
them.

The problem with the symbol stripping is that multiple relas may
reference the same ELF symbol.  If any remaining relocation references a
shared symbol, we must keep it.

Replace the symbol->strip boolean with an enumeration:

  SYMBOL_DEFAULT - initial value, symbol usage unknown
  SYMBOL_USED    - symbol is definitely used by a rela
  SYMBOL_STRIP   - symbol was only referenced by dynrela(s)

Allow transitions from SYMBOL_DEFAULT to SYMBOL_* and SYMBOL_STRIP to
SYMBOL_USED, but _not_ SYMBOL_USED to SYMBOL_*.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1940,7 +1940,7 @@ static void kpatch_migrate_included_elements(struct kpatch_elf *kelf, struct kpa
 		list_del(&sym->list);
 		list_add_tail(&sym->list, &out->symbols);
 		sym->index = 0;
-		sym->strip = 0;
+		sym->strip = SYMBOL_DEFAULT;
 		if (sym->sec && !sym->sec->include)
 			/* break link to non-included section */
 			sym->sec = NULL;
@@ -3234,8 +3234,10 @@ static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
 				special = true;
 
 		list_for_each_entry_safe(rela, safe, &sec->relas, list) {
-			if (!rela->need_dynrela)
+			if (!rela->need_dynrela) {
+				rela->sym->strip = SYMBOL_USED;
 				continue;
+			}
 
 			/*
 			 * Starting with Linux 5.8, .klp.arch sections are no
@@ -3332,8 +3334,8 @@ static void kpatch_create_intermediate_sections(struct kpatch_elf *kelf,
 			 * later (for example, they may have relocations
 			 * of their own which should be processed).
 			 */
-			if (!rela->sym->sec)
-				rela->sym->strip = 1;
+			if (!rela->sym->sec && rela->sym->strip != SYMBOL_USED)
+				rela->sym->strip = SYMBOL_STRIP;
 			list_del(&rela->list);
 			free(rela);
 
@@ -3519,7 +3521,7 @@ static void kpatch_strip_unneeded_syms(struct kpatch_elf *kelf,
 	struct symbol *sym, *safe;
 
 	list_for_each_entry_safe(sym, safe, &kelf->symbols, list) {
-		if (sym->strip) {
+		if (sym->strip == SYMBOL_STRIP) {
 			list_del(&sym->list);
 			free(sym);
 		}
--- a/kpatch-build/kpatch-elf.h
+++ b/kpatch-build/kpatch-elf.h
@@ -67,6 +67,12 @@ struct section {
 	};
 };
 
+enum symbol_strip {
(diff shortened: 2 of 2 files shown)


joe-lawrence:
I added dbg printf's while debugginf this (see `kpatch-build.diff` in attached repro tarball):
```
$ ./repro &> repro.out

$ grep '^dbg:' repro.out | tail -n1
dbg: check symbol: section: .rela__jump_table, rela @ 0x100288f5d80 rela->sym @ 0x1002886b270

$ grep -e 0x1002886b270 -e 0x100288f5d80 repro.out 
dbg: section: .rela.toc, rela @ 0x100288e7f80 rela->sym @ 0x1002886b270 netstamp_needed_key
dbg: section: .rela__jump_table, rela @ 0x100288f5d80 rela->sym @ 0x1002886b270 netstamp_needed_key
dbg: strip symbol: section: .rela.toc, rela @ 0x100288e7f80 rela->sym @ 0x1002886b270 netstamp_needed_key
dbg: free stripped sym @ 0x1002886b270 netstamp_needed_key
dbg: check symbol: section: .rela__jump_table, rela @ 0x100288f5d80 rela->sym @ 0x1002886b270
```
and can see that the problem is that the symbol for `netstamp_needed_key` was used after it was freed.

I encountered a bug like this in upstream klp-convert development (see https://github.com/joe-lawrence/linux/commit/e2ec7b6117aa517b5280db67727ef2dd85d04001 for example), where multiple relocation symbols are share the same symbol.  One must be careful when throwing away such shared symbols.

Comments welcome.  An alternative approach I considered was some kind of rela<->symbol reference count, but thought that might be overkill.

[repro.tar.gz](https://github.com/dynup/kpatch/files/7027440/repro.tar.gz)

sm00th:
Looks ok, but the objectfiles from the repro tarball are prime candidates for addition to unit tests.

joe-lawrence:
> Looks ok, but the objectfiles from the repro tarball are prime candidates for addition to unit tests.

Good idea.  I'm a bit surprised we haven't seen this in real world CVE kpatches.. but it took a long while to encounter it in klp-convert work, so perhaps just another weird corner case.  I'll prep a unit test PR later.  In the meantime, do you think the enum in kpatch-elf.h needs a description of its intent?  Or any other comments to clarify the code if you think it may hide some subtlety.

sm00th:
> In the meantime, do you think the enum in kpatch-elf.h needs a description of its intent? Or any other comments to clarify the code if you think it may hide some subtlety.

The commit message is good enough for me personally, but if you worry about intent being clear I think we can leave sym->strip as boolean and just make `SYMBOL_USED` a local flag in that `list_for_each_entry_safe` loop.

joe-lawrence:
> > In the meantime, do you think the enum in kpatch-elf.h needs a description of its intent? Or any other comments to clarify the code if you think it may hide some subtlety.
> 
> The commit message is good enough for me personally, but if you worry about intent being clear I think we can leave sym->strip as boolean and just make `SYMBOL_USED` a local flag in that `list_for_each_entry_safe` loop.

Would that be saved into another field or list somehow?

I'm stuck on how to keep a boolean value that represents three ideas:

1. initial condition, okay to set to any value
2. definitely used, can't strip it
3. may be unused, suitable to strip for now

so an additional flag could help, but how would that work?

sm00th:
> so an additional flag could help, but how would that work?

You are right, it won't, nvm

sm00th (changes requested):
Requires unit-tests added so marking as such.

joe-lawrence:
(updated with unit test objects subproject update)

kamalesh-babulal (approved):

sm00th (approved):

Issue #1210, opened 2021-08-19
kmod/patch/livepatch-patch-hook.c: Is the for loop condition 'object->funcs' in patch_free_livepatch neccessary?


he7850:
https://github.com/dynup/kpatch/blob/1ca8e8fc1fa34751a5c7067b82a8319958680a60/kmod/patch/livepatch-patch-hook.c#L230-L247
Is the for loop condition 'object->funcs' in livepatch-patch-hook.c#L235 neccessary?
1. object->funcs is checked null inside the loop. double checking is redundant.
2. if object->funcs could be null, then it will break the for loop and may cause memory leak, bcz remaining parts of patch->objs will not be released.
I am new to kpatch, if I misunderstand the for loop condition, pls correct me, thx.

jpoimboe:
I think you're right that the two checks are redundant.

The only way I see object->funcs can be NULL is if patch_init() fails to allocate memory for an instance of lobject->funcs.  In that case I don't think there would be a memory leak if patch_free_livepatch() broke out of the loop early because all the subsequent objects in 'patch->objs' would already have NULL funcs and relocs.

So I don't see a memory leak, but you're welcome to submit a pull request to get rid of the redundancy.

he7850:
@jpoimboe 
Thanks for your answer. I have submitted a PR.

PR #1213, opened 2021-08-24
livepatch-patch-hook: remove redundant loop condition in patch_free_livepatch

he7850:
Inside the for loop of releasing patch->objs, each object->funcs is checked
null, thus there is no need to check it twice.

See issue: https://github.com/dynup/kpatch/issues/1210

Signed-off-by: Hu Bin <im.bin.hu@gmail.com>

### DIFF
--- a/kmod/patch/livepatch-patch-hook.c
+++ b/kmod/patch/livepatch-patch-hook.c
@@ -232,7 +232,7 @@ static void patch_free_livepatch(struct klp_patch *patch)
 	struct klp_object *object;
 
 	if (patch) {
-		for (object = patch->objs; object && object->funcs; object++) {
+		for (object = patch->objs; object; object++) {
 			if (object->funcs)
 				kfree(object->funcs);
 #ifndef HAVE_ELF_RELOCS


sm00th:
Ok, got a bit of time to look into this. The actually useless condition in that statement is 'object' as it is never NULL, but all of the structure members for the last object in the array are zeroed out and `->funcs` is the member original author decided to use to find the end of the array as it is always set. So we can't get rid of the check in the for loop, but we can remove the if inside the loop. @jpoimboe correct me if I am wrong.

jpoimboe:
@sm00th Yes, you are right of course.  Thanks for debugging the issue.

@he7850 Apologies for my wrong answer in #1210.  As @sm00th said, `object->funcs == NULL` means the last entry of the `objects` array has been reached.  So:

- The `object` check in the `for` statement is unnecessary, since it will never be NULL.
- The `object->funcs` check in the `for` statement is necessary to end the loop.
- The `object->funcs` check *inside* the loop is redundant -- and unnecessary for another reason: it's fine to pass a NULL pointer to kfree().
- The `object->relocs` check is also unnecessary, since it's fine to pass a NULL pointer to kfree().

sm00th (changes requested):
See https://github.com/dynup/kpatch/pull/1213#issuecomment-913053835

PR #1217, opened 2021-09-02
kpatch-build: skip system_certificates.o

xiejingf:
system_certificates.o: WARNING: FILE symbol not found in base. Stripped object file or assembly source?
since kpatch does not support patch assembly file cert/system_certificates.S, we can simply skip it and
avoid the warning.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -1059,6 +1059,7 @@ for i in $FILES; do
 		[[ "$i" = arch/x86/lib/copy_user_64.o ]] && continue
 
 	[[ "$i" = usr/initramfs_data.o ]] && continue
+	[[ "$i" = certs/system_certificates.o ]] && continue
 
 	mkdir -p "output/$(dirname "$i")"
 	cd "$SRCDIR" || die


jpoimboe:
Seems reasonable.  What kind of patch triggered this warning?  Was it a change to a header file?

xiejingf:
> Seems reasonable. What kind of patch triggered this warning? Was it a change to a header file?

Honestly, it is a very simple patch, no header file is changed at the first glance.

[patch.log](https://github.com/dynup/kpatch/files/7104007/patch.log)

jpoimboe:
Hm, I wonder if this could be a problem with your kpatch-build setup?  I don't see how changing fs/ext4/super.c could trigger a rebuild of certs/system_certificates.o.

```
$ make -s -j8 vmlinux
$ ls -l certs/system_certificates.o
-rw-rw-r--. 1 jpoimboe jpoimboe 2440 Sep  4 10:45 certs/system_certificates.o
$ touch fs/ext4/super.c
$ sleep 120
$ make -s -j8 vmlinux
$ ls -l certs/system_certificates.o
-rw-rw-r--. 1 jpoimboe jpoimboe 2440 Sep  4 10:45 certs/system_certificates.o
$ ls -l fs/ext4/super.o
-rw-rw-r--. 1 jpoimboe jpoimboe 2859680 Sep  4 10:47 fs/ext4/super.o
```

xiejingf:
@jpoimboe ，yeah， I tried the your steps and the result is the same as yours.  After checking my kpatch-build script, I finally confirmed that  I copy a  xx.pem file to certs directory to overwrite the original one, which cause the certs/system_certificates.S rebuilt.

Thanks for your tests.

jpoimboe:
@xiejingf Are you you using CONFIG_LIVEPATCH or kpatch.ko?

For livepatch, I'd guess that signing the module during the module kernel linking build wouldn't actually work, since create-klp-module edits the module after the build.  Instead I think you'd need to manually call sign-file after kpatch-build finishes.

xiejingf:
I use kpatch.ko and sign the built ko manually every time after kpatch-build finishes.

jpoimboe:
@xiejingf Do you still need this patch?  If so, I think the change probably belongs in `kpatch-cc` with most of the other skipped files.

xiejingf:
> @xiejingf Do you still need this patch? If so, I think the change probably belongs in `kpatch-cc` with most of the other skipped files.

this patch is not needed,  the issue is caused by my customized pre-kpatch-build precedure, anyway, thanks for the discussion ~

PR #1220, opened 2021-09-10
kpatch-build: Do not check KLP_REPLACE for kpatch.ko-based patches

euspectre:
After commit 17dcebf077d69 ("kpatch-build: enable klp with replace option by default"),
building the old-style (kpatch.ko-based) patches fails with the following
error:

> "kpatch core module (kpatch.ko) does not support replace, please add -R|--non-replace"

kpatch.ko actually supports atomic replacement of patches but KLP_REPLACE
has nothing to do with that. Let us not break such builds and remove that check.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -858,9 +858,6 @@ else
 	# sections.  Use with caution!
 	echo "WARNING: Use of kpatch core module (kpatch.ko) is deprecated!  There may be bugs!" >&2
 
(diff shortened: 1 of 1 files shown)


joe-lawrence (approved):
Good catch, it looks like we used a 'replace' module parameter back then,  so the build time assertion is obviously wrong.

jpoimboe (approved):

euspectre:
>  it looks like we used a 'replace' module parameter back then

Yes, exactly.

sm00th (approved):

Issue #1212, opened 2021-08-22
ppc64l4 and R_PPC64_REL64 relocations on rhel-9-beta


joe-lawrence:
While debugging #1211, I ran into a subsequent kpatch-build problem:
```
...
dev.o: changed function: __netif_receive_skb_core.constprop.0
make -C /root/.kpatch/src M=/root/.kpatch/tmp/patch CFLAGS_MODULE=''
make[1]: Entering directory '/root/.kpatch/src'
LDS     /root/.kpatch/tmp/patch/kpatch.lds
CC [M]  /root/.kpatch/tmp/patch/patch-hook.o
LD [M]  /root/.kpatch/tmp/patch/livepatch-data-read-mostly.o
MODPOST /root/.kpatch/tmp/patch/Module.symvers
ERROR: modpost: "__tracepoint_netif_receive_skb" [/root/.kpatch/tmp/patch/livepatch-data-read-mostly.ko] undefined!
```
which was strange as I thought we had supported tracepoints.  (I've seen this in a few other ppc64le patches on rhel-9-beta as well.)

More on this relocation:
```
$ readelf --wide --relocs ~/.kpatch/tmp/patch/tmp_output.o

Relocation section '.rela__jump_table' at offset 0x14c6f0 contains 21 entries:
    Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
...
0000000000000018  000000830000002c R_PPC64_REL64          0000000000000000 __tracepoint_netif_receive_skb + 8
...
```

And then from added dbg code in create-diff-object's `need_dynrela()` (see kpatch-build.diff in tarball attached):
```
dev.o.orig: changed function: __netif_receive_skb_core.constprop.0
dbg: skipping sec: .rela.text.__netif_receive_skb_core.constprop.0      rela: R_PPC64_REL64     .TOC.
dbg: skipping sec: .rela.text.__netif_receive_skb_core.constprop.0      rela: R_PPC64_ENTRY
dbg: skipping sec: .rela__jump_table    rela: R_PPC64_REL64     netstamp_needed_key
dbg: skipping sec: .rela__jump_table    rela: R_PPC64_REL64     __tracepoint_netif_receive_skb
dbg: skipping sec: .rela__jump_table    rela: R_PPC64_REL64     generic_xdp_needed_key
dbg: skipping sec: .rela__jump_table    rela: R_PPC64_REL64     ingress_needed_key
[...]

PR #1222, opened 2021-09-13
create-diff-object: support ppc64le relative jump labels

joe-lawrence:
RHEL-9 integration tests revealed that the kernel now makes use of
R_PPC64_REL64 relocations in the jump table:

  Relocation section '.rela__jump_table' at offset 0x14c6f0 contains 21 entries:
  ...
  0000000000000018  000000830000002c R_PPC64_REL64          0000000000000000 __tracepoint_netif_receive_skb + 8
  ...

but need_dynrela() contains code to specifically skip any R_PPC64_REL64
type when determining if a relocation should be turned into dynrela.

Kamalesh Babulal explains:

  I tried digging a little deeper and the upstream Kernel commit
  b0b3b2c78ec (powerpc: Switch to relative jump labels) in v5.13,
  introduced the change of generating relocation entries of type
  R_PPC64_REL64, instead of absolute relocation type R_PPC64_ADDR64:

  Relocation section '.rela__jump_table' at offset 0x1a87d8 contains 303 entries:
  ...
  00000000000003c8  000007910000002c R_PPC64_REL64          0000000000000000 __tracepoint_netif_receive_skb + 8
  ...

Relax the existing check in need_dynrela() for .rela__jump_table
R_PPC64_REL64 relocations in case we need dynrelas for them.

Fixes: #1212


kamalesh-babulal (approved):

sm00th (approved):

yhcote (approved):

joe-lawrence:
Thanks for the reviews, guys, but let's  to put this on hold -- for some reason when I build and load module.patch via the integration tests, it crashes the machine.  However directly building w/kpatch-build (replace and non-replace), no crashes.  It's gotta be something really silly in my environment, so I'll check back if and when I figure that out.

joe-lawrence:
> Thanks for the reviews, guys, but let's to put this on hold -- for some reason when I build and load module.patch via the integration tests, it crashes the machine. However directly building w/kpatch-build (replace and non-replace), no crashes. It's gotta be something really silly in my environment, so I'll check back if and when I figure that out.

Finally got back to this one: I'm confident that the module.patch crash is not related.  Testing differences attributed to whether a klp-relocation target module (nfsd) was loaded or not.  I'll open a separate issue for that.

For this one, I made two small updates:
* The conditional in `need_dynrela()` should check `sec->name` and not `rela->sym->sec->name` for ".rela__jump_table" as the latter wasn't set for the relocations in question.  Pipe the section pointer through to the function.
* Remove the redundant readelf relocation output from the commit message and leave the one Kamalesh quoted.

Also for future reference, some object files and scaffolding to reproduce.  
[repro-1222.tar.gz](https://github.com/dynup/kpatch/files/7246420/repro-1222.tar.gz)

sm00th (approved):

PR #1224, opened 2021-09-24
add cross-compilation support

omatiusha:
This PR upstreams two patches already present in OpenEmbedded source tree: https://git.openembedded.org/meta-openembedded/tree/meta-oe/recipes-kernel/kpatch/kpatch, ported to the latest version of kpatch. They allow livepatches to be compiled for target that differs from the host.


sm00th:
Hi, this has been discussed previously in https://github.com/dynup/kpatch/pull/811 and there were some adjustments original author agreed  to come back with.

PR #1226, opened 2021-09-25
kpatch-build: Handle error in create-klp-module

### LINKED ISSUE #1223: kpatch-build: ignores ERROR() from create-kpatch-module and create-klp-module
As discovered in #1203, **kpatch-build** continues executing if **create-klp-module** or **create-kpatch-module** sets error status.  

In both cases, `check_pipe_status()` is called to try and catch a segfault condition (code 139), but the `ERROR()` macro in log.h sets `EXIT_STATUS_ERROR` which is not the same.  This can lead to **kpatch-build** reporting success when an error actually occurred.

sumanthkorikkar:
create-klp-module could error out and in this case ensure that
kpatch-build does not proceed further.

Fixes #1223


sumanthkorikkar:
Added same style check instead of test command

kamalesh-babulal (approved):

joe-lawrence (approved):

joe-lawrence:
Thanks, @sumanthkorikkar

PR #1231, opened 2021-10-13
Basic Clang PGO support

liu-song-6:
This series adds basic Clang PGO support. 
The first commit fixes issue that function section name in original and patched kernel has different prefixes (e.g., .text vs. .text.unlikely). Such behavior is expected with PGO. 
The secondd commit adds -p|--profile-data option to kpatch-build.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -455,6 +455,14 @@ static bool rela_equal(struct rela *rela1, struct rela *rela2)
 		return true;
 
 	/*
+	 * __llvm_prf_cnts is used by clang PGO to store counters. Ignore
+	 * these to void unnecessary "changed function".
+	 */
+	if (!strcmp(rela1->sym->name, "__llvm_prf_cnts") &&
+	    !strcmp(rela2->sym->name, "__llvm_prf_cnts"))
+		return true;
+
+	/*
 	 * With -mcmodel=large on ppc64le, GCC might generate entries in the .toc
 	 * section for relocation symbol references.   The .toc offsets may change
 	 * between the original and patched .o, so comparing ".toc + offset" isn't
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -986,6 +986,8 @@ static void kpatch_correlate_section(struct section *sec_orig,
 		kpatch_correlate_symbol(sec_orig->sym, sec_patched->sym);
 }
 
+static char *kpatch_section_function_name(struct section *sec);
+
 static void kpatch_correlate_sections(struct list_head *seclist_orig,
 		struct list_head *seclist_patched)
 {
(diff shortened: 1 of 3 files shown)


sm00th:
Hi, thanks for the pullrequest. 

It seems it hasn't been merged to upstream kernel yet, so is this work based on ["pgo: add clang's Profile Guided Optimization infrastructure"](https://lore.kernel.org/all/20210407211704.367039-1-morbo@google.com/) patch? If so I see documentation suggesting to use `KCFLAGS=-fprofile-use=...`, but you use `CFLAGS_PGO_CLANG` instead, is there a reason for this?

Also do you know if there is any way to tell if the original kernel was built with the same profile data kpatch-build gets? Maybe clang notes it down somewhere in resulting binaries and we can check it the same way we do with compiler version?

Overall I think this looks good, but we'll have to wait until CONFIG_PGO_CLANG gets merged to the kernel before merging this.

liu-song-6:
> It seems it hasn't been merged to upstream kernel yet, so is this work based on ["pgo: add clang's Profile Guided Optimization infrastructure"](https://lore.kernel.org/all/20210407211704.367039-1-morbo@google.com/) patch? If so I see documentation suggesting to use `KCFLAGS=-fprofile-use=...`, but you use `CFLAGS_PGO_CLANG` instead, is there a reason for this?

Yes, our version is based on this patch. We use CFLAGS_PGO_CLANG flag here to only pass `-fprofile-use` to the compiler. If we use `KCFLAGS`, we will pass in both `-fprofile-use` and `-fprofile-generate`. 

> Also do you know if there is any way to tell if the original kernel was built with the same profile data kpatch-build gets? Maybe clang notes it down somewhere in resulting binaries and we can check it the same way we do with compiler version?

I am not aware of such information shipped with the kernel.

sm00th (approved):

liu-song-6:
> Hi @liu-song-6, I think your force pushes were just rebasing this development branch?

Yes, this is to address some conflict. 

> 
> Also, on the subject of the original commit msg that "fixes [the] issue that function section name in original and patched kernel has different prefixes (e.g., .text vs. .text.unlikely)"... I noticed [similar](https://github.com/dynup/kpatch/pull/1236#issuecomment-1050459138) behavior with gcc on arm64. Do you have any information on why or how the compiler decides that this can occur?

In PGO case, the compiler compares the function signature of current version and the profiled version. If the signature changed (due to patch here), the compiler will ignore the profile data, and thus made different likely/unlikely decision. In none PGO case, I guess it is some other change in likely/unlikely decision?

> 
> And then, it looks like your fix was dropped from this PR. How was it fixed? Were such changes ignored (or always added) if it was determined that only section names changed? Thanks.

It was a mistake. Thanks for catching it! I will add the fix back.

joe-lawrence:
Hi @liu-song-6 
In PGO cases, can the profile data change _only_ .text vs. .text.unlikely, or is that change always accompanied by some code update in the function?  (I'm trying to understand if correlating a non-code update change between .text.foo and .text.unlikely can be safely ignored in the non-PGO cases.)

And then did upstream kernel progress on adding clang PGO stall?  I see Kees note here: https://lore.kernel.org/lkml/202106140921.5E591BD@keescook/  but is this any closer to merging?  Thanks.

PR #1234, opened 2021-10-28
kpatch-build: building OOT livepatch modules for kernels built from source

omatiusha:
Ability to build out-of-tree modules for non-host kernel is currently
missing. This commit introduces that ability for any kernel built
from source, including cross-compiled ones. The workaround checks for
existence of /lib/modules/$ARCHVERSION/build, and if nonexistent,
assumes the kernel source directory to be a build directory.

This PR is subsequent to #1233

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -540,13 +540,14 @@ usage() {
 	echo "		                        (can be specified multiple times)" >&2
 	echo "		-e, --oot-module        Enable patching out-of-tree module," >&2
 	echo "		                        specify current version of module" >&2
+	echo "		-p, --oot-module-src    Set OOT module source directory" >&2
 	echo "		-R, --non-replace       Disable replace patch (replace is on by default)" >&2
 	echo "		--skip-cleanup          Skip post-build cleanup" >&2
 	echo "		--skip-compiler-check   Skip compiler version matching check" >&2
 	echo "		                        (not recommended)" >&2
 }
 
-options="$(getopt -o ha:r:s:c:v:j:t:n:o:de:R -l "help,archversion:,sourcerpm:,sourcedir:,config:,vmlinux:,jobs:,target:,name:,output:,oot-module:,debug,skip-gcc-check,skip-compiler-check,skip-cleanup,non-replace" -- "$@")" || die "getopt failed"
+options="$(getopt -o ha:p:r:s:c:v:j:t:n:o:de:R -l "help,archversion:,oot-module-src:,sourcerpm:,sourcedir:,config:,vmlinux:,jobs:,target:,name:,output:,oot-module:,debug,skip-gcc-check,skip-compiler-check,skip-cleanup,non-replace" -- "$@")" || die "getopt failed"
 
 eval set -- "$options"
 
@@ -609,6 +610,11 @@ while [[ $# -gt 0 ]]; do
 		OOT_MODULE="$(readlink -f "$2")"
 		shift
 		;;
+	-p|--oot-module-src)
+		[[ ! -d "$2" ]] && die "OOT module source dir '$2' not found"
+		OOT_SOURCE="$(readlink -f "$2")"
+		shift
+		;;
 	-R|--non-replace)
 		KLP_REPLACE=0
 		;;
@@ -662,6 +668,16 @@ if [[ -n "$OOT_MODULE" ]] &&  [[ -z "$USERSRCDIR" ]]; then
 	exit 1
 fi
 
+
+if [[ -n "$OOT_MODULE" ]] &&  [[ -z "$OOT_SOURCE" ]]; then
+	if [[ -e "$(dirname "$OOT_MODULE")/Makefile" ]]; then
+		OOT_SOURCE="$(dirname "$OOT_MODULE")"
+	else 
+		warn "--oot-module requires --oot-module-src"
+		exit 1
+	fi
+fi
+
 # ensure cachedir and tempdir are setup properly and cleaned
 mkdir -p "$TEMPDIR" || die "Couldn't create $TEMPDIR"
 rm -rf "${TEMPDIR:?}"/*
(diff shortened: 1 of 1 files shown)


kpatch-build/kpatch-build
```diff
@@ -1164,7 +1167,12 @@ cd "$TEMPDIR/patch" || die
 if [[ -z "$OOT_MODULE" ]]; then
 	KPATCH_BUILD="$SRCDIR"
 else
```
sm00th, line 1169: `$(dirname "$CONFIGFILE")` doesn't sound great, I think we'll be better off using `$SRCDIR` instead.

```suggestion
if [[ -z "$OOT_MODULE" || ! -e "/lib/modules/$ARCHVERSION/build" ]]; then
	KPATCH_BUILD="$SRCDIR"
else
	KPATCH_BUILD="KPATCH_BUILD="/lib/modules/$ARCHVERSION/build"
fi
```

omatiusha: I tried `$SRCDIR` and it didn't work because if script is invoked from directory with OOT module, it does invoke `make -C $SRCDIR`, which does nothing, because `-C` must have **kernel** source directory as its argument.

sm00th: `SRCDIR` should contain the path to kernel source directory here. Here where it gets assigned: https://github.com/dynup/kpatch/blob/master/kpatch-build/kpatch-build#L675. And here we make sure `USERSRCDIR` is set when we use `OOT_MODULE`: https://github.com/dynup/kpatch/blob/master/kpatch-build/kpatch-build#L660.

omatiusha: But this repo's README has the invocation example, and it states that SRCDIR is assigned module's source directory: https://github.com/dynup/kpatch/blob/61fc3f10776017f21249e2cbd4b0639cbc325d9e/README.md?plain=1#L588

sm00th: Yes, you are right. This semantic change is rather confusing. So are you relying on user-specified config then (through `--config` option)? Because if it is not specified that way it will be set to `/boot/config-xxx` (https://github.com/dynup/kpatch/blob/master/kpatch-build/kpatch-build#L738).

It seems to me that instead of changing `--source` semantics it'll be better to add another option for `OOT_MODULE` case that will define `OOT_MODULE_PATH` or something in that vein. We then'll be able to keep using `SRCDIR` normally

omatiusha: Yes, I thought about that. Will be working on it next week and remove confusing semantics. Thanks for reviewing!

omatiusha: Added `-p | --oot-module-src` option, allowing to specify module source and use `-s` in conventional way, only for kernel. Should I change the README of the repo too?

kpatch-build/kpatch-build
```diff
@@ -1161,8 +1183,9 @@ else
 fi
 
 cd "$TEMPDIR/patch" || die
-if [[ -z "$OOT_MODULE" ]]; then
-	KPATCH_BUILD="$SRCDIR"
+
+if [[ -z "$OOT_MODULE" || ! -e "/lib/modules/$ARCHVERSION/build" ]]; then
+	KPATCH_BUILD="$USERSRCDIR"
```
sm00th, line 1188: Hi, thanks for the update.

Unfortunately this change breaks the builds that don't specify `$USERSRCDIR`.

omatiusha: But, if I understood correctly, that was the whole point of 
> It seems to me that instead of changing `--source` semantics it'll be better to add another option for `OOT_MODULE` case that will define `OOT_MODULE_PATH` or something in that vein. We then'll be able to keep using `SRCDIR` normally

In what cases the build will break?

sm00th: > In what cases the build will break?

Sorry, I should have been more clear. This breaks the normal build (without oot_module) without `USERSRCDIR` In that case `SRCDIR` should be used for kpatch-build.  I think ideally we would keep SRCDIR untouched by OOT_MODULE code and only use OOT_SOURCE directly when needed. Can't tell if that will be easy to do though, we don't normaly build out-of-tree modules so I am not very familiar with that code.

omatiusha: Unfortunately, if we do not set SRCDIR to OOT_SOURCE and use it directly, that will add to many if's and that is not good. I tried to mitigate the issue in easier way, please check my last update.

kpatch-build/kpatch-build
```diff
@@ -672,7 +688,9 @@ if [[ -n "$USERSRCDIR" ]]; then
 		warn "--archversion is incompatible with --sourcedir"
 		exit 1
 	fi
-	SRCDIR="$USERSRCDIR"
+
+	[[ -z "$OOT_MODULE" ]] && SRCDIR="$USERSRCDIR" || SRCDIR="$OOT_SOURCE"
```
sm00th, line 692: One last question: does `OOT_MODULE` still need to depend on `USERSRCDIR` after these changes?

kpatch-build/kpatch-build
sm00th: Ok, this makes sense.

One last question: does `OOT_MODULE` still need  to depend on `USERSRCDIR`, or can we now decouple it?

omatiusha: At what line does `OOT_MODULE` depend on `USERSRCDIR`?

sm00th: > At what line does OOT_MODULE depend on USERSRCDIR?

https://github.com/dynup/kpatch/pull/1234/files#diff-0c4525f1e83472a84e8ac916c7860ab840bb98069cc6bb021f18675f1f3d9092R666 is the main place but there are also others like https://github.com/dynup/kpatch/pull/1234/files#diff-0c4525f1e83472a84e8ac916c7860ab840bb98069cc6bb021f18675f1f3d9092R666 which is inside `if [[ -n "$USERSRCDIR" ]]; then` clause.

omatiusha: Well, i don't think it shouldn't depend on `USERSRCDIR`. Where else can we get working Kbuild?

sm00th: Normally (for non-oot cases) if `USERSRCDIR` is not specified kpatch-build would download running distro kernel's sources and prepare them for build in `SRCDIR` (`~/.kpatch/src` by default). There is [this whole section](https://github.com/dynup/kpatch/blob/master/kpatch-build/kpatch-build#L744) handling it. I think that can work for OOT modules as well.

sm00th:
Hi,

Since there haven't been any updates for a while I took a stab at it in #1242. Hopefully that solution works for your usecase, please take a look.

I am closing this PR as duplicate for now, but it can be reopened if needed.

omatiusha:
Hi, I was trying to get it to work but you did it first, thank you so much!

Issue #1045, opened 2019-10-04
dump-ipa-clones reporting enhancement


joe-lawrence:
I found some time to hack up kpatch-build to see if I could use the new `-fdump-ipa-clones` flag to generate a before and after report comparison.  For those not familiar, this gcc option leaves behind a bunch of .ipa-clones files that describe inlining decisions by the compiler.  Comparing these reports around kpatching could be helpful in patch review or kpatch-build bug triaging. 

Here's a really simple demo patch:
```diff
Force show_val_kb calls to be inlined into its lone caller, meminfo_proc_show

--- src.orig/fs/proc/meminfo.c  2019-10-04 14:23:27.301741298 -0400
+++ src/fs/proc/meminfo.c       2019-10-04 14:25:14.920930984 -0400
@@ -24,7 +24,7 @@ void __attribute__((weak)) arch_report_m
 {
 }
 
-static void show_val_kb(struct seq_file *m, const char *s, unsigned long num)
+static __always_inline void show_val_kb(struct seq_file *m, const char *s, unsigned long num)
 {
        seq_put_decimal_ull_width(m, s, num << (PAGE_SHIFT - 10), 8);
        seq_write(m, " kB\n", 4);
```

My hacked-up kpatch-build saves off orig and patched ipa-clones files.  Then manually run commands (for now) from `~/.kpatch/tmp` through the [kgraft-analysis-tool](https://github.com/marxin/kgraft-analysis-tool):
```shell
ANALYSIS=/root/kgraft-analysis-tool/kgraft-ipa-analysis.py
FILES=$(diff -Nupr orig/ipa-clones/ patched/ipa-clones/ | diffstat -l -p0)

for file in $FILES; do

        orig="orig/${file#*/}"
        patched="patched/${file#*/}"

        diff -U0 \
                --label "$orig"    <($ANALYSIS <(echo $orig)    | sed 's/object file: .*ipa-clones\///') \
                --label "$patched" <($ANALYSIS <(echo $patched) | sed 's/object file: .*ipa-clones\///')
done
```

Generates this report:
```
--- orig/ipa-clones/fs/proc/meminfo.c.000i.ipa-clones
+++ patched/ipa-clones/fs/proc/meminfo.c.000i.ipa-clones
@@ -94,0 +95,6 @@
+Function: show_val_kb/3169 (fs/proc/meminfo.c:27:29) [REMOVED] [fs/proc/meminfo.c.000i.ipa-clones]
[...]

PR #1236, opened 2021-11-03
Add Arm64 Support

surajjs95:
Implement support for building livepatches for arm64

Kernel Support:
- Currently being worked on upstream
- My tree here: https://github.com/amazonlinux/linux/tree/amazon-arm-livepatch-v5.10

With the above kernel tree and the changes in this pull request I have been able to build and apply patches using kpatch.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -3401,6 +3401,21 @@ static void kpatch_create_callbacks_objname_rela(struct kpatch_elf *kelf, char *
 	}
 }
 
+static void kpatch_alloc_mcount_sections(struct kpatch_elf *kelf, struct kpatch_elf *kelfout)
+{
+	int nr;
+	struct symbol *sym;
+
+	nr = 0;
+	list_for_each_entry(sym, &kelfout->symbols, list)
+		if (sym->type == STT_FUNC && sym->status != SAME &&
+		    sym->has_func_profiling)
+			nr++;
+
+	/* create text/rela section pair */
+	create_section_pair(kelfout, "__mcount_loc", sizeof(void*), nr);
+}
+
 /*
  * This function basically reimplements the functionality of the Linux
  * recordmcount script, so that patched functions can be recognized by ftrace.
@@ -3408,7 +3423,7 @@ static void kpatch_create_callbacks_objname_rela(struct kpatch_elf *kelf, char *
  * TODO: Eventually we can modify recordmount so that it recognizes our bundled
  * sections as valid and does this work for us.
  */
-static void kpatch_create_mcount_sections(struct kpatch_elf *kelf)
+static void kpatch_populate_mcount_sections(struct kpatch_elf *kelf)
 {
 	int nr, index;
 	struct section *sec, *relasec;
@@ -3417,15 +3432,10 @@ static void kpatch_create_mcount_sections(struct kpatch_elf *kelf)
 	void **funcs;
 	unsigned long insn_offset;
 
-	nr = 0;
-	list_for_each_entry(sym, &kelf->symbols, list)
-		if (sym->type == STT_FUNC && sym->status != SAME &&
-		    sym->has_func_profiling)
-			nr++;
 
-	/* create text/rela section pair */
-	sec = create_section_pair(kelf, "__mcount_loc", sizeof(void*), nr);
+	sec = find_section_by_name(&kelf->sections, "__mcount_loc");
 	relasec = sec->rela;
+	nr = (int) (sec->data->d_size / sizeof(void *));
 
 	/* populate sections */
 	index = 0;
@@ -3788,6 +3798,9 @@ int main(int argc, char *argv[])
 	/* this is destructive to kelf_patched */
 	kpatch_migrate_included_elements(kelf_patched, &kelf_out);
 
+	/* this must be done before kelf_patched is torn down */
+	kpatch_alloc_mcount_sections(kelf_patched, kelf_out);
+
 	/*
 	 * Teardown kelf_patched since we shouldn't access sections or symbols
 	 * through it anymore.  Don't free however, since our section and symbol
@@ -3806,7 +3819,7 @@ int main(int argc, char *argv[])
 	kpatch_create_callbacks_objname_rela(kelf_out, parent_name);
 	kpatch_build_strings_section_data(kelf_out);
 
-	kpatch_create_mcount_sections(kelf_out);
+	kpatch_populate_mcount_sections(kelf_out);
 
 	/*
 	 *  At this point, the set of output sections and symbols is
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1635,7 +1635,7 @@ static void kpatch_check_func_profiling_calls(struct kpatch_elf *kelf)
 		    (sym->parent && sym->parent->status == CHANGED))
 			continue;
 		if (!sym->twin->has_func_profiling) {
-			log_normal("function %s has no fentry/mcount call, unable to patch\n",
+			log_normal("function %s doesn't have patchable function entry, unable to patch\n",
 				   sym->name);
 			errs++;
 		}
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -62,7 +62,9 @@
 
 #ifdef __powerpc64__
 #define ABSOLUTE_RELA_TYPE R_PPC64_ADDR64
-#else
+#elif defined(__aarch64__)
+#define ABSOLUTE_RELA_TYPE R_AARCH64_ABS64
+#elif defined(__x86_64__)
 #define ABSOLUTE_RELA_TYPE R_X86_64_64
 #endif
 
@@ -213,6 +215,28 @@ static struct rela *toc_rela(const struct rela *rela)
 	return find_rela_by_offset(rela->sym->sec->rela,
 				   (unsigned int)rela->addend);
 }
+
+#ifdef __aarch64__
+/*
+ * Mapping symbols are used to mark and label the transitions between code and
+ * data in elf files. They begin with a "$" dollar symbol. Don't correlate them
+ * as they often all have the same name either "$x" to mark the start of code
+ * or "$d" to mark the start of data.
+ */
+static bool kpatch_is_mapping_symbol(struct symbol *sym)
+{
+	if (sym->name && sym->name[0] == '$'
+		&& sym->type == STT_NOTYPE \
+		&& sym->bind == STB_LOCAL)
+		return 1;
+	return 0;
+}
+#else
+static int kpatch_is_mapping_symbol(struct symbol *sym)
+{
+	return 0;
+}
+#endif
 
 /*
  * When compiling with -ffunction-sections and -fdata-sections, almost every
@@ -560,6 +584,13 @@ static void kpatch_compare_correlated_section(struct section *sec)
 	/* Short circuit for mcount sections, we rebuild regardless */
 	if (!strcmp(sec->name, ".rela__mcount_loc") ||
 	    !strcmp(sec->name, "__mcount_loc")) {
+		sec->status = SAME;
+		goto out;
+	}
+
+	/* As above but for aarch64 */
+	if (!strcmp(sec->name, ".rela__patchable_function_entries") ||
+	    !strcmp(sec->name, "__patchable_function_entries")) {
 		sec->status = SAME;
 		goto out;
 	}
@@ -1028,6 +1059,9 @@ static void kpatch_correlate_symbols(struct list_head *symlist_orig,
 			 */
 			if (sym_orig->type == STT_NOTYPE &&
 			    !strncmp(sym_orig->name, ".LC", 3))
+				continue;
+
+			if (kpatch_is_mapping_symbol(sym_orig))
 				continue;
 
 			/* group section symbols must have correlated sections */
@@ -1539,7 +1573,7 @@ static void kpatch_replace_sections_syms(struct kpatch_elf *kelf)
 				continue;
 			}
 
-#ifdef __powerpc64__
+#if defined(__powerpc64__) || defined(__aarch64__)
 			add_off = 0;
 #else
 			if (rela->type == R_X86_64_PC32 ||
@@ -1571,7 +1605,8 @@ static void kpatch_replace_sections_syms(struct kpatch_elf *kelf)
 				end = sym->sym.st_value + sym->sym.st_size;
 
 				if (!is_text_section(sym->sec) &&
-				    rela->type == R_X86_64_32S &&
+				    (rela->type == R_X86_64_32S ||
+				     rela->type == R_AARCH64_ABS64) &&
 				    rela->addend == (long)sym->sec->sh.sh_size &&
 				    end == (long)sym->sec->sh.sh_size) {
 
@@ -2047,21 +2082,6 @@ static int parainstructions_group_size(struct kpatch_elf *kelf, int offset)
 	return size;
 }
 
-static int altinstructions_group_size(struct kpatch_elf *kelf, int offset)
-{
-	static int size = 0;
-	char *str;
-
-	if (!size) {
-		str = getenv("ALT_STRUCT_SIZE");
-		if (!str)
-			ERROR("ALT_STRUCT_SIZE not set");
-		size = atoi(str);
-	}
-
-	return size;
-}
-
 static int smp_locks_group_size(struct kpatch_elf *kelf, int offset)
 {
 	return 4;
@@ -2076,6 +2096,22 @@ static int static_call_sites_group_size(struct kpatch_elf *kelf, int offset)
 		str = getenv("STATIC_CALL_STRUCT_SIZE");
 		if (!str)
 			ERROR("STATIC_CALL_STRUCT_SIZE not set");
+		size = atoi(str);
+	}
+
+	return size;
+}
+#endif
+#if defined(__x86_64__) || defined(__aarch64__)
+static int altinstructions_group_size(struct kpatch_elf *kelf, int offset)
+{
+	static int size = 0;
+	char *str;
+
+	if (!size) {
+		str = getenv("ALT_STRUCT_SIZE");
+		if (!str)
+			ERROR("ALT_STRUCT_SIZE not set");
 		size = atoi(str);
 	}
 
@@ -2190,12 +2226,14 @@ static struct special_section special_sections[] = {
 		.group_size	= parainstructions_group_size,
 	},
 	{
-		.name		= ".altinstructions",
-		.group_size	= altinstructions_group_size,
-	},
-	{
 		.name		= ".static_call_sites",
 		.group_size	= static_call_sites_group_size,
+	},
+#endif
+#if defined(__x86_64__) || defined(__aarch64__)
+	{
+		.name		= ".altinstructions",
+		.group_size	= altinstructions_group_size,
 	},
 #endif
 #ifdef __powerpc64__
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -3401,6 +3401,11 @@ static void kpatch_create_callbacks_objname_rela(struct kpatch_elf *kelf, char *
 	}
 }
 
+/*
+ * Allocate the mcount/patchable_function_entry sections which must be done
+ * before the patched object is torn down so that the section flags can be
+ * copied.
+ */
 static void kpatch_alloc_mcount_sections(struct kpatch_elf *kelf, struct kpatch_elf *kelfout)
 {
 	int nr;
@@ -3413,10 +3418,30 @@ static void kpatch_alloc_mcount_sections(struct kpatch_elf *kelf, struct kpatch_
 			nr++;
 
 	/* create text/rela section pair */
+#ifdef __aarch64__
+{
+	struct section *sec, *tmp;
+
+	sec = create_section_pair(kelfout, "__patchable_function_entries", sizeof(void*), nr);
+
+	/*
+	 * Depending on the compiler the __patchable_function_entries section
+	 * can be ordered or not, copy this flag to the section we created to
+	 * avoid:
+	 * ld: __patchable_function_entries has both ordered [...] and unordered [...] sections
+	 */
+	tmp = find_section_by_name(&kelf->sections, "__patchable_function_entries");
+	sec->sh.sh_flags |= (tmp->sh.sh_flags & SHF_LINK_ORDER);
+	sec->sh.sh_flags = 1;
+}
+#else /* !__aarch64__ */
 	create_section_pair(kelfout, "__mcount_loc", sizeof(void*), nr);
+#endif
 }
 
 /*
+ * Populate the mcount sections allocated by kpatch_alloc_mcount_sections()
+ * previously.
  * This function basically reimplements the functionality of the Linux
  * recordmcount script, so that patched functions can be recognized by ftrace.
  *
@@ -3428,12 +3453,16 @@ static void kpatch_populate_mcount_sections(struct kpatch_elf *kelf)
 	int nr, index;
 	struct section *sec, *relasec;
 	struct symbol *sym;
-	struct rela *rela, *mcount_rela;
+	struct rela *mcount_rela;
 	void **funcs;
 	unsigned long insn_offset;
 
 
+#ifdef __aarch64__
+	sec = find_section_by_name(&kelf->sections, "__patchable_function_entries");
+#else /* !__aarch64__ */
 	sec = find_section_by_name(&kelf->sections, "__mcount_loc");
+#endif
 	relasec = sec->rela;
 	nr = (int) (sec->data->d_size / sizeof(void *));
 
@@ -3450,8 +3479,8 @@ static void kpatch_populate_mcount_sections(struct kpatch_elf *kelf)
 		}
 
 #ifdef __x86_64__
-
-		rela = list_first_entry(&sym->sec->rela->relas, struct rela, list);
+{
+		struct rela *rela = list_first_entry(&sym->sec->rela->relas, struct rela, list);
 
 		/*
 		 * For "call fentry", the relocation points to 1 byte past the
@@ -3491,9 +3520,37 @@ static void kpatch_populate_mcount_sections(struct kpatch_elf *kelf)
 
 			rela->type = R_X86_64_PC32;
 		}
+}
+#elif defined(__aarch64__)
+{
+		unsigned char *insn;
+		int i;
 
+		insn = sym->sec->data->d_buf;
+		insn_offset = 0;
+
+		/*
+		 * If BTI (Branch Target Identification) is enabled then there
+		 * might be an additional 'BTI C' instruction before the two
+		 * patchable function entry 'NOP's.
+		 * i.e. 0xd503245f (little endian)
+		 */
+		if (insn[0] == 0x5f) {
+			if (insn[1] != 0x24 || insn[2] != 0x03 || insn[3] != 0xd5)
+				ERROR("%s: unexpected instruction in patch section of function", sym->name);
+			insn_offset += 4;
+			insn += 4;
+		}
+		for (i = 0; i < 8; i += 4) {
+			/* We expect a NOP i.e. 0xd503201f (little endian) */
+			if (insn[i] != 0x1f || insn[i + 1] != 0x20 ||
+			    insn[i + 2] != 0x03 || insn [i + 3] != 0xd5)
+				ERROR("%s: unexpected instruction in patch section of function", sym->name);
+		}
+}
 #else /* __powerpc64__ */
 {
+		struct rela *rela;
 		bool found = false;
 
 		list_for_each_entry(rela, &sym->sec->rela->relas, list)
--- a/kpatch-build/kpatch-elf.c
+++ b/kpatch-build/kpatch-elf.c
@@ -332,7 +332,24 @@ static void kpatch_find_func_profiling_calls(struct kpatch_elf *kelf)
 				break;
 			}
 		}
-#else
+#elif defined(__aarch64__)
+{
+		struct section *sec = find_section_by_name(&kelf->sections,
+				      "__patchable_function_entries");
+		/*
+		 * If we can't find the __patchable_function_entries section or
+		 * there are no relocations in it then not patchable.
+		 */
+		if (!sec || !sec->rela)
+			return;
+		list_for_each_entry(rela, &sec->rela->relas, list) {
+			if (rela->sym->sec && sym->sec == rela->sym->sec) {
+				sym->has_func_profiling = 1;
+				break;
+			}
+		}
+}
+#else /* x86_64 */
 		rela = list_first_entry(&sym->sec->rela->relas, struct rela,
 					list);
 		if ((rela->type != R_X86_64_NONE &&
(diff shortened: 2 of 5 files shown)


kpatch-build/create-diff-object.c
```diff
@@ -214,6 +216,22 @@ static struct rela *toc_rela(const struct rela *rela)
 				   (unsigned int)rela->addend);
 }
 
+#ifdef __aarch64__
+static int kpatch_is_mapping_symbol(struct symbol *sym)
```
jpoimboe, line 220: This could use a comment to describe what a mapping symbol is and why we wouldn't correlate it.

jpoimboe: Also, probably best to make it return a bool so the return code semantics are clear.

kpatch-build/create-diff-object.c
```diff
@@ -1568,7 +1596,11 @@ static void kpatch_replace_sections_syms(struct kpatch_elf *kelf)
 				end = sym->sym.st_value + sym->sym.st_size;
 
 				if (!is_text_section(sym->sec) &&
+#ifdef __x86_64__
 				    rela->type == R_X86_64_32S &&
+#elif defined(__aarch64__)
+				    rela->type == R_AARCH64_ABS64 &&
+#endif
```
jpoimboe, line 1603: No need to '#ifdef' this, the defines exist in /usr/include/elf.h regardless of arch.

kpatch-build/create-diff-object.c
```diff
@@ -1632,8 +1664,13 @@ static void kpatch_check_func_profiling_calls(struct kpatch_elf *kelf)
 		    (sym->parent && sym->parent->status == CHANGED))
 			continue;
 		if (!sym->twin->has_func_profiling) {
+#ifdef __aarch64__
+			log_normal("function %s doesn't have patchable_function_entry,"
+				   " unable to patch\n", sym->name);
+#else
 			log_normal("function %s has no fentry/mcount call, unable to patch\n",
 				   sym->name);
+#endif
```
jpoimboe, line 1673: Another unneeded '#ifdef', instead the existing message can be made arch-independent, e.g. "function %s doesn't have patchable function entry".

kpatch-build/create-diff-object.c
```diff
@@ -3479,6 +3548,37 @@ static void kpatch_create_mcount_sections(struct kpatch_elf *kelf)
...
+		if (!patchable_sec)
+			ERROR("%s: unexpected missing __patchable_function_entries section", __func__);
+
+		/* We search the patchable_sec to ensure that the symbol is patchable. */
+		list_for_each_entry(rela, &patchable_sec->relas, list) {
+			if (rela->sym && rela->sym->sec && sym->sec ==
+							   rela->sym->sec) {
+				found = true;
+				break;
+			}
```
jpoimboe, line 3566: Why is this needed?  The check to ensure the symbol is patchable should have already been done already in kpatch_check_func_profiling_calls().  So I could be missing something, but I don't see the need to preserve `patchable_sec` or to do this check here.

I do notice that `rela->addend` is needed below for the `insn_offset` calculation, but if the offset isn't always at the same location, can the function entry point be discovered by just looking for it at the beginning of the function?  If so, I'd much rather do that because it seems less complex than the hacks needed to preserve the contents of the table.

kpatch-build/create-diff-object.c
```diff
@@ -3421,7 +3486,11 @@ static void kpatch_create_mcount_sections(struct kpatch_elf *kelf)
 			nr++;
 
 	/* create text/rela section pair */
+#ifdef __aarch64__
+	sec = create_section_pair(kelf, "__patchable_function_entries", sizeof(void*), nr);
```
jpoimboe, line 3490: When trying to kpatch-build a change to meminfo_proc_show(), I got the following error:
```
make[1]: Entering directory '/root/linux'
  LD [M]  /root/.kpatch/tmp/patch/livepatch-a.o
ld: __patchable_function_entries has both ordered [`__patchable_function_entries' in /root/.kpatch/tmp/patch/patch-hook.o] and unordered [`__patchable_function_entries' in /root/.kpatch/tmp/patch/output.o] sections
ld: final link failed: bad value
make[2]: *** [scripts/Makefile.build:431: /root/.kpatch/tmp/patch/livepatch-a.o] Error 1
```
The problem is that the compiler-created version of the section has the `SHF_LINK_ORDER` flag set.  I "fixed" it with the following hack.  `sh_link` just points to a random section but maybe that's ok, AFAICT its value doesn't matter for the kernel's use of `__patchable_function_entries`.
```
index 11a5f94..bd34764 100644
@@ -3488,6 +3488,8 @@ static void kpatch_create_mcount_sections(struct kpatch_elf *kelf,
        /* create text/rela section pair */
 #ifdef __aarch64__
        sec = create_section_pair(kelf, "__patchable_function_entries", sizeof(void*), nr);
+       sec->sh.sh_flags |= SHF_LINK_ORDER;
+       sec->sh.sh_link = 1;
 #else /* !__aarch64__ */
        sec = create_section_pair(kelf, "__mcount_loc", sizeof(void*), nr);
 #endif
```

kpatch-build/create-diff-object.c
```diff
@@ -1933,6 +1970,30 @@ static void kpatch_migrate_included_elements(struct kpatch_elf *kelf, struct kpa
...
+	patchable_sec = find_section_by_name(&kelf->sections,
+					     "__patchable_function_entries");
+	if (patchable_sec) {
+		list_for_each_entry(rela, &patchable_sec->rela->relas, list) {
+			if (!rela->sym->include)
+				rela->sym = NULL;
+		}
+	}
+}
+#endif
```
jpoimboe, line 1996: This feels like a major hack and I'm not really convinced it's needed.  I have a more detailed comment below.

kpatch-build/create-diff-object.c
```diff
@@ -3785,6 +3885,14 @@ int main(int argc, char *argv[])
...
 	kpatch_migrate_included_elements(kelf_patched, &kelf_out);
 
+	/* Preserve the patchable_function_entries sec before it's torn down */
+	patchable_sec = find_section_by_name(&kelf_patched->sections,
+					     "__patchable_function_entries");
+	if (patchable_sec) {
+		patchable_sec = patchable_sec->rela;
+		list_del(&patchable_sec->list);
+	}
+
```
jpoimboe, line 3895: As decsribed above, I don't think this is needed.

kpatch-build/kpatch-elf.c
```diff
@@ -332,7 +335,27 @@ static void kpatch_find_func_profiling_calls(struct kpatch_elf *kelf)
...
+			struct section *sec = find_section_by_name(&kelf->sections,
+					      "__patchable_function_entries");
+			/*
+			 * If we can't find the __patchable_function_entries
+			 * section or there are no relocations in it then not
+			 * patchable.
+			 */
+			if (!sec || !sec->rela)
+				return;
+			patchable_sec = sec->rela;
```
jpoimboe, line 349: The `patchable_sec` variable is confusingly named, since it's really the relocation section for the patchable sec.  I don't think you need a separate variable for it anyway, it can just continue to be referred to by `sec->rela`.

kpatch-build/kpatch-elf.c
```diff
@@ -332,7 +335,27 @@ static void kpatch_find_func_profiling_calls(struct kpatch_elf *kelf)
...
+		}
+		list_for_each_entry(rela, &patchable_sec->relas,
+				    list) {
```
jpoimboe, line 352: No need for the line break here.

kpatch-build/create-diff-object.c
surajjs95: Yep, noted

kpatch-build/kpatch-elf.c
surajjs95: This was an optimisation to avoid searching for the "__patchable_function_entries" section each time. Although I'm sure it makes almost zero difference to the run time of the script.

kpatch-build/kpatch-elf.c
surajjs95: Excellent point :D

kpatch-build/create-diff-object.c
sm00th: I guess what @jpoimboe is saying is that it can be written down as `(rela->type == R_X86_64_32S || rela->type == R_AARCH64_ABS64)` which looks nice. We also might want to get rid of all ifdefs at some point (see #1179) so unless it is necessary we are reluctant to add new ones.

surajjs95: Ah yes, that makes sense

kpatch-build/kpatch-elf.c
jpoimboe: Oh, I see what you mean.  Maybe do the `find_section_by_name()` before the loop, and give the rela sec a more descriptive name like `patchable_rela_sec`.

kpatch-build/create-diff-object.c
surajjs95: Ok so I have the opposite problem in that patch-hook.o is unordered and output.o becomes ordered with the change you suggested which and the linking fails.

Which version of gcc are you using?

sm00th: I've seen this on both rhel9 and fedora34 (`gcc version 11.2.1 20210728 (Red Hat 11.2.1-1) (GCC)`). We might want to copy `sh_flags`/`sh_link` from `__patchable_function_entries` of patched objectfile.

surajjs95: Yeah that sounds like the best option

kpatch-build/create-diff-object.c
```diff
@@ -1536,7 +1564,7 @@ static void kpatch_replace_sections_syms(struct kpatch_elf *kelf)
 				continue;
 			}
 
-#ifdef __powerpc64__
+#if defined(__powerpc64__) || defined(__aarch64__)
```
ivechan, line 1567: Just curious, why `add_off ` is `0` in aarch64?
I see different handling in x86 below. 
:)

surajjs95:
Updated with V2 addressing review comments

joe-lawrence:
When building a small patch [test1.tar.gz](https://github.com/dynup/kpatch/files/8042548/test1.tar.gz), I notice a modpost warning:
```
WARNING: modpost: /root/.kpatch/tmp/patch/livepatch-test.o (__patchable_function_entries): unexpected non-allocatable section.
Did you forget to use "ax"/"aw" in a .S file?
Note that for example <linux/init.h> contains
section definitions for use in .S files.
```

And see that there are multiple sections in the livepatch.ko file with that name (the second one missing the flags modpost complains about):
```
[Nr]  Name                          Type      Address           Off     Size    ES  Flg  Lk  Inf  Al
[32]  __patchable_function_entries  PROGBITS  0000000000000000  001a40  000028  00  WAL  4   0    8
[60]  __patchable_function_entries  PROGBITS  0000000000000000  0d8558  000008  00  W    0   0    8
```

Q: can the modpost warning be ignored?
Q: is there expected to be multiple `__patchable_function_entries` (and associated rela) sections in the livepatch.ko? 

gcc: gcc version 11.2.1 20220127 (Red Hat 11.2.1-9) (GCC)
ld: GNU ld version 2.35.2-17.el9
kernel tree: amazon-arm-livepatch-v5.10
kpatch-build: surajjs95-arm64

kpatch-build/create-diff-object.c
```diff
@@ -3402,30 +3440,69 @@ static void kpatch_create_callbacks_objname_rela(struct kpatch_elf *kelf, char *
...
+
+	/*
+	 * Depending on the compiler the __patchable_function_entries section
+	 * can be ordered or not, copy this flag to the section we created to
+	 * avoid:
+	 * ld: __patchable_function_entries has both ordered [...] and unordered [...] sections
+	 */
+	tmp = find_section_by_name(&kelf->sections, "__patchable_function_entries");
+	sec->sh.sh_flags |= (tmp->sh.sh_flags & SHF_LINK_ORDER);
+	sec->sh.sh_flags = 1;
```
keiya-nobuta, line 3473: sec->sh.sh_link = 1 ?

sm00th: Good catch, this resolves the issue @joe-lawrence reported above. I also thought we decided to copy it over from patched objectfile instead of always setting it.

keiya-nobuta:
> Hi @surajjs95 , @keiya-nobuta : some of my recent merges to the master branch introduced a bunch of merge conflicts for this patchset, mainly those where I removed #ifdef __arch__ checks in create-diff-object. Please see my rebase at https://github.com/joe-lawrence/kpatch/tree/arm64-rebase-aaaebaf25895. It builds, but I have not tested it. Can you guys test and if it looks good, incorporate those updates this PR and we can resume reviewing. Thanks.

Hi @joe-lawrence, thanks your work.
I tested your branch and got some errors in bug-table-section.patch, data-read-mostly.patch, fixup-section.patch, etc:

```
proc_sysctl.o: changed section __bug_table not selected for inclusion
proc_sysctl.o: changed section .rela__bug_table not selected for inclusion

dev.o: changed section .rodata.str not selected for inclusion
dev.o: changed section .rela__jump_table not selected for inclusion

readdir.o: changed section .rela.fixup not selected for inclusion
readdir.o: changed section .rela__ex_table not selected for inclusion
```

It has appeared in other patches, but it is mostly summarized for the above reasons.
It looks like good with the below fix:
```
index 2fc030c..50ca7a5 100644
@@ -1813,6 +1813,7 @@ static void kpatch_include_standard_elements(struct kpatch_elf *kelf)
 		    !strcmp(sec->name, ".strtab") ||
 		    !strcmp(sec->name, ".symtab") ||
 		    !strcmp(sec->name, ".toc") ||
+		    !strcmp(sec->name, ".rodata.str") ||
 		    !strcmp(sec->name, ".rodata") ||
 		    (!strncmp(sec->name, ".rodata.", 8) &&
 		     strstr(sec->name, ".str1."))) {
@@ -2210,22 +2211,22 @@ static int fixup_group_size(struct kpatch_elf *kelf, int offset)
 static struct special_section special_sections[] = {
 	{
 		.name		= "__bug_table",
-		.arch		= X86_64 | PPC64,
+		.arch		= AARCH64 | X86_64 | PPC64,
 		.group_size	= bug_table_group_size,
 	},
 	{
 		.name		= ".fixup",
-		.arch		= X86_64 | PPC64,
+		.arch		= AARCH64 | X86_64 | PPC64,
 		.group_size	= fixup_group_size,
 	},
 	{
 		.name		= "__ex_table", /* must come after .fixup */
-		.arch		= X86_64 | PPC64,
+		.arch		= AARCH64 | X86_64 | PPC64,
 		.group_size	= ex_table_group_size,
 	},
 	{
 		.name		= "__jump_table",
-		.arch		= X86_64 | PPC64,
+		.arch		= AARCH64 | X86_64 | PPC64,
 		.group_size	= jump_table_group_size,
 	},
 	{
```

And warn-detect-FAIL.patch was successfully built, I think this is because line_macro_change_only for aarch64 is not implemented.
```
ERROR: warn-detect-FAIL: build succeeded when it should have failed
```

joe-lawrence:
> Hi @joe-lawrence, thanks your work. I tested your branch and got some errors in bug-table-section.patch, data-read-mostly.patch, fixup-section.patch, etc:

Thanks for testing!  By the way, what is your gcc / kernel version combo that you tested with?  I was hoping to try the PR on a 5.14-based kernel (aka, rhel-9), but I only saw @surajjs95 's tree: [amazon-arm-livepatch-v5.10](https://github.com/amazonlinux/linux/tree/amazon-arm-livepatch-v5.10), do you have a kernel branch is closer to up to date?

> It looks like good with the below fix: [ snip ]

Great, thanks for those fixups, I'll incorporate those too.

> And warn-detect-FAIL.patch was successfully built, I think this is because line_macro_change_only for aarch64 is not implemented.
> 

Yep, that is still on the TODO list for the arm implementation.

keiya-nobuta:
> Thanks for testing! By the way, what is your gcc / kernel version combo that you tested with?

I'm using:
gcc (Ubuntu 10.3.0-1ubuntu1~20.04) 10.3.0 /
[v5.15.0 based kernel with Madhavan's v10](https://github.com/keiya-nobuta/linux/commits/v5.15_livepatch), and [my testset](https://github.com/keiya-nobuta/kpatch/tree/test_for_linux-5.15/test/integration/linux-5.15)

> You can use mine https://github.com/sm00th/linux/tree/kpatch/aarch64 it is v13 with dropped STACK_VALIDATION dependency on top of v5.17-rc4

Thanks, I'll try it.

joe-lawrence:
Some more incremental work here: https://github.com/joe-lawrence/kpatch/tree/arm64-rebase
* Using @sm00th 's [aarch64](https://github.com/sm00th/linux/tree/kpatch/aarch64) kernel tree and my [integration-tests-kernel-5.14.0-17.el9](https://github.com/joe-lawrence/kpatch/tree/integration-tests-kernel-5.14.0-17.el9) branch, created an initial set of unit tests (currently pointing to a dev branch in my repro.  When we get closer to full support, we should rebuild these and push to the official repo.)
  * Run with `make unit`, with recent updates to master branch, can test aarch64 from x86 :sunglasses: 
* First pass at line change detection in the last commit
  * warn-detect-FAIL.patch now fails with "ERROR: no changed objects found" as expected

Finally, if I try this change:
``` diff
@@ -3567,7 +3567,7 @@ static void kpatch_alloc_mcount_sections(struct kpatch_elf *kelf, struct kpatch_
                 */
                tmp = find_section_by_name(&kelf->sections, "__patchable_function_entries");
                sec->sh.sh_flags |= (tmp->sh.sh_flags & SHF_LINK_ORDER);
-               sec->sh.sh_link = 1;
+               sec->sh.sh_link = tmp->sh.sh_link;
                break;
        }
        case PPC64:
```
The integration tests fail with "nm: gcc-static-local-var-2.OUTPUT.o: sh_link [2] in section `__patchable_function_entries' is incorrect".  Did I implement the suggestion correctly?

keiya-nobuta:
> The integration tests fail with "nm: gcc-static-local-var-2.OUTPUT.o: sh_link [2] in section `__patchable_function_entries' is incorrect". Did I implement the suggestion correctly?

I checked by cross-compile on x86 using gcc-linaro-11.2.1, and I got same error.

On the PATCHED.o, `sh_link [2]` to point to `.text.<...>` section, but on the OUTPUT.o, maybe it's incorrect for `sh_link [2]` to point to the `RELA` section. 

patched/mm/mmap.o
```
Section Headers:
  [Nr] Name                               Type            Address          Off    Size   ES Flg Lk Inf Al
  [ 0]                                    NULL            0000000000000000 000000 000000 00      0   0  0
  [ 1] .text                              PROGBITS        0000000000000000 000040 000000 00  AX  0   0  1
  [ 2] .text.__traceiter_vm_unmapped_area PROGBITS        0000000000000000 000040 00006c 00  AX  0   0 16
  [185] __patchable_function_entries      PROGBITS        0000000000000000 0062b0 0001f8 00 WAL  2   0  8

Relocation section '.rela__patchable_function_entries' at offset 0x5cc50 contains 63 entries:
0000000000000000  0000000200000101 R_AARCH64_ABS64        0000000000000000 .text.__traceiter_vm_unmapped_area + 4
<...>
00000000000001b0  0000003c00000101 R_AARCH64_ABS64        0000000000000000 .text.mmap_region + 4
<...>
```

output/mm/mmap.o
```
Section Headers:
  [Nr] Name                         Type            Address          Off    Size   ES Flg Lk Inf Al
  [ 0]                              NULL            0000000000000000 000000 000000 00      0   0  0
  [ 1] .text.mmap_region            PROGBITS        0000000000000000 000040 000608 00  AX  0   0 16
  [ 2] .rela.text.mmap_region       RELA            0000000000000000 000648 000120 18   I 28   1  8
  [31] __patchable_function_entries PROGBITS        0000000000000000 089a50 000008 08  AL  2   0  8

Relocation section '.rela__patchable_function_entries' at offset 0x89a58 contains 1 entry:
0000000000000000  0000000200000101 R_AARCH64_ABS64        0000000000000000 mmap_region + 4
```

In most cases `Section [1]` will be the `.text` section, so I think `sh_link = 1` is fine. 
I don't know how to use sh_link, but I think it's safer to set sh_link to point to the first-entry of __patchable_function_entries (after reindex sections). I fixed it with workaround:
```
index ede5ae3..4303794 100644
@@ -3567,7 +3567,6 @@ static void kpatch_alloc_mcount_sections(struct kpatch_elf *kelf, struct kpatch_
 		 */
 		tmp = find_section_by_name(&kelf->sections, "__patchable_function_entries");
 		sec->sh.sh_flags |= (tmp->sh.sh_flags & SHF_LINK_ORDER);
-		sec->sh.sh_link = tmp->sh.sh_link;
 		break;
 	}
 	case PPC64:
@@ -4096,6 +4095,18 @@ int main(int argc, char *argv[])
 	kpatch_strip_unneeded_syms(kelf_out, lookup);
 	kpatch_reindex_elements(kelf_out);
 
+	/*
+	 * At __patchable_function_entries, sh_link points to first-entry.
+	 */
+	sec = find_section_by_name(&kelf_out->sections,
+				   "__patchable_function_entries");
+	if (sec) {
+		struct rela *rela = list_first_entry(&sec->rela->relas,
+						     struct rela, list);
+		if (rela)
+			sec->sh.sh_link = rela->sym->sec->index;
+	}
+
 	/*
 	 * Update rela section headers and rebuild the rela section data
 	 * buffers from the relas lists.
```

joe-lawrence:
Re: sh_link:  I don't know what sh_link should be either.  It seems that ordinary module builds end up with sh_link set to the first non-zero length .text section.  Perhaps for now, we should just leave it set to 1?

Re: integration tests: I built the [integration-tests-kernel-5.14.0-17.el9](https://github.com/joe-lawrence/kpatch/tree/integration-tests-kernel-5.14.0-17.el9) tests and almost all succeeded, save for macro-printk.patch, due to diff context shifts (from 5.14 to 5.17), and [gcc-mangled-3.patch](https://github.com/joe-lawrence/kpatch/blob/integration-tests-kernel-5.14.0-17.el9/test/integration/rhel-9.0/gcc-mangled-3.patch):

```gcc-mangled-3.log:
ERROR: slub.o: 2 function(s) can not be patched
slub.o: function percpu_ref_put_many.constprop.0 doesn't have patchable function entry, unable to patch
slub.o: function memcg_slab_free_hook doesn't have patchable function entry, unable to patch
```

Interestingly both percpu_ref_put_many() and memcg_slab_free_hook() are defined in header files as static inlines, the compiler gives them their own .text.<name> sections, but they don't have the usual `bti c,  nop,  nop` opening sequence and therefore no  spot in rela__patchable_function_entries.

I modified the patch to only add nops to get_slabinfo() and the build succeeds, reporting expected `slub.o: changed function: get_slabinfo`... I wonder if the real version of the patch somehow provokes the compiler into creating new versions of the static inline code (but not directly affected by the patch).  Attached here for someone more qualified at reading aarch64 asm:  [gcc-mangled-3.tar.gz](https://github.com/dynup/kpatch/files/8074565/gcc-mangled-3.tar.gz)

Also, I remember reporting a similar problem trying to build for PR #1244.  Maybe this is new gcc-11 behavior?

joe-lawrence:
Some more interesting behavior to report.

Using a rhel-9 kernel and [gcc-mangled-3.tar.gz](https://github.com/dynup/kpatch/files/8134506/gcc-mangled-3.tar.gz):
```
slub.ORIG.o: function __list_add doesn't have patchable function entry, unable to patch
ERROR: slub.ORIG.o: 1 function(s) can not be patched
../kpatch-build/create-diff-object: unreconcilable difference
```
In this case, the entire **__list_add** function has moved from `.text.__list_add` to `.text.unlikely.__list_add`:
```
$ readelf --wide --sections slub.ORIG.o | grep 'list_add'
  [21] .text.__list_add  PROGBITS        0000000000000000 000620 00004c 00  AX  0   0 16
  [22] .rela.text.__list_add RELA            0000000000000000 0a52a8 000018 18   I 520  21  8
$ readelf --wide --sections slub.PATCHED.o | grep 'list_add'
  [21] .text.unlikely.__list_add PROGBITS        0000000000000000 000618 00004c 00  AX  0   0  4
  [22] .rela.text.unlikely.__list_add RELA            0000000000000000 0a5298 000018 18   I 521  21  8
```
Yet their compiled code is exactly the same:
```
$ diff -u <(objdump -Dr -j .text.__list_add slub.ORIG.o) \
          <(objdump -Dr -j .text.unlikely.__list_add slub.PATCHED.o)
--- /dev/fd/63  2022-02-24 11:57:42.497384380 -0500
+++ /dev/fd/62  2022-02-24 11:57:42.497384380 -0500
@@ -1,8 +1,8 @@
 
-slub.ORIG.o:     file format elf64-littleaarch64
+slub.PATCHED.o:     file format elf64-littleaarch64
 
 
-Disassembly of section .text.__list_add:
+Disassembly of section .text.unlikely.__list_add:
 
 0000000000000000 <__list_add>:
    0:  d503233f        paciasp
```
I googled a bit and found gcc [Bug 61958](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61958) - functions arbitrarily placed in .text.unlikely section, from @jpoimboe way back in 2014.  The gcc-managled-3.patch is similarly adding a printk and functions are moving from .text to .text.likely, however I don't believe that **__list_add()** is called by **get_slabinfo()**, nor is **__list_add()** inlined into its calling functions.

For an additional data point, with the rhel-9 kernel, I see the same section change for `.text.irda_usb_dump_class_desc.isra.0` in @keiya-nobuta 's data-reloc.patch.  ([data-reloc.tar.gz](https://github.com/dynup/kpatch/files/8137967/data-reloc.tar.gz))

keiya-nobuta:
Sorry for late reply.

I couldn't prepare RHELL9, so I tried it with fedora 35 aarch64.
gcc: gcc (GCC) 11.2.1 20220127 (Red Hat 11.2.1-9)
ld: GNU ld version 2.37-10.fc35

I didn't get the same result but I get similar result that static inline functions were changed which independent of the patched functions. It seems that arm64 gcc11 has optimization features that I don't understand well.

I didn't get this result with defconfig based config, so I tried turning off some configs from fedora's config. When CONFIG_DEBUG_SECTION_MISMATCH=n was set, these static inline function did not change. CONFIG_DEBUG_SECTION_MISMATCH is an option to add -fno-inline-functions-called-once to KBUILD_CFLAGS. I'm not sure what this is used for. 

It looks like it is solved, but the non-patched functions seems to be changed :(
```
slub.o: changed function: process_slab
slub.o: changed function: __kmem_cache_shutdown
slub.o: changed function: __check_heap_object
slub.o: changed function: get_slabinfo
```

keiya-nobuta:
I posted this question to GCC Bugzilla.
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105326

joe-lawrence:
> I posted this question to GCC Bugzilla. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105326

Thanks for raising the question to the gcc folks.  Maybe resurrecting https://github.com/dynup/kpatch/issues/1045 to at least dump out some of the IPA inlining results could help us cope.  (Ideally we could feed such data back to the compiler to essentially repeat the same decisions, but I would assume that is simply not possible.)

keiya-nobuta:
I think the Arm64 specific gcc issue needs to be resolved separately, but I don't think it's this pull-request issue.
Is it difficult to continue the review?

swine:
I've rebased your arm64 support into v0.9.6 and addressed a few issues which clang-built linux-5.15 throws up, and it builds working patches for most test cases.

I'll clean up and post my tree shortly, after merging with the recent s/390 changes, which cause much incidental merge conflict, as they both add to the switch(kelf->arch) cases, and special_sections[] entries

I think I've solved the problem with recent clang, where a separate __patchable_function_entries section is created for *each* patchable function, and it's required that create-diff-object can correctly read this different arrangement, and write_elf with the same many-sections format.

One remaining issue I haven't yet solved with 5.15 is insmod failure logging "overflow in relocation type 261 val 4" where clang (or kpatch) is using type 261 (R_AARCH64_PREL32) in a way that's now deprecated due to increased security

joe-lawrence:
[keiya-nobuta](https://github.com/keiya-nobuta) says,
> I think the Arm64 specific gcc issue needs to be resolved separately, but I don't think it's this pull-request issue. Is it difficult to continue the review?

The most difficult part is finding time during the day to work through it :) But yes, now with the initial s390x code out of the way, let's try continuing with arm64. :)


[swine](https://github.com/swine) says,
> I've rebased your arm64 support into v0.9.6 and addressed a few issues which clang-built linux-5.15 throws up, and it builds working patches for most test cases.
> I'll clean up and post my tree shortly, after merging with the recent s/390 changes, which cause much incidental merge conflict, as they both add to the switch(kelf->arch) cases, and special_sections[] entries

Perfect.  Apologies for the code churn.  We refactored a lot of the architecture-specific code from #ifdef to switch statements.  Hopefully it's a little more obvious to fit in new arches and then run cross-arch unit testing.  Rebasing would be a good restart for reviewing... I don't know if you can directly push that here, or if @surajjs95 would have to manually import those changes to his branch that the PR was open for.

> I think I've solved the problem with recent clang, where a separate __patchable_function_entries section is created for each patchable function, and it's required that create-diff-object can correctly read this different arrangement, and write_elf with the same many-sections format.

Was the many __patchable_function_entries sections issue limited to clang or did it affect gcc, too?

> One remaining issue I haven't yet solved with 5.15 is insmod failure logging "overflow in relocation type 261 val 4" where clang (or kpatch) is using type 261 (R_AARCH64_PREL32) in a way that's now deprecated due to increased security 

Can you elaborate on that?  Also, is this no longer an issue on newer kernels?  Thanks.

t-msn:
Hello, is anyone currently working on this?
I'd like to help this PR to be merged.

To see the current status, I also rebased the code to current master : [this branch](https://github.com/t-msn/kpatch/tree/arm64_rebase_202209)
I run unit & integration tests(linux-5.18) on fedora36 and ubuntu22 with using this version of [kernel](https://github.com/madvenka786/linux/tree/orc_v2) (based on 5.18) and it basically looks fine. Some notes:
 - I see kpatch-build failure by gcc problem mentioned above in data-reloc test (#1244).  
 -  module-load test can be built but failed to load module. It seems the kernel side's problem.
 -  I added one fix to keep ubsan section if CONFIG_UBSAN is enabled (for ubuntu) but I'm not sure if this is correct.

I didn't test clang build which @swine mentioned.
Also, I think it'd be better to @surajjs95 update this PR.

On the other hand, I believe kernel side needs to/will be updated. I'm not familiar with kernel side at the moment and try to understand the current situation as well. Thanks.

swine:
thanks @t-msn, I wish I'd pushed [my arm64 branch](https://github.com/swine/kpatch/tree/arm64) earlier to save you duplicated effort, but was hunting clang issues.

your rebase is possibly cleaner (certainly smaller, as it didn't need to address the clang-vs-gcc issues that I've explored in https://github.com/dynup/kpatch/pull/1302).

BTW, the "overflow in relocation" issue I mentioned earlier is a non-issue, typo in earlier version of the code.

t-msn:
@swine Thank you very much for sharing your reabase code. It looks like the clang support needs some effort... Maybe is it better to support(merge) gcc only part for the starting point? Anyway, I will try to look the code.

> BTW, the "overflow in relocation" issue I mentioned earlier is a non-issue, typo in earlier version of the code.

That's good. Thanks.

PR #1237, opened 2021-11-18
kpatch-build: refactor kernel config and special section parsing

joe-lawrence:
Previous commits added support for CONFIG_PRINTK_INDEX, but do not
actually read the option from the kernel configuration.  Implement that
now so the feature support will be active.

Fixes: 56471ffc7c0c ("kpatch-build: Support CONFIG_PRINTK_INDEX, part 2")
Fixes: 6cf50a6fca84 ("create-diff-object: Support CONFIG_PRINTK_INDEX, part 1")

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -333,8 +333,8 @@ clang_version_check() {
 
 find_special_section_data_ppc64le() {
 
-	[[ "$CONFIG_JUMP_LABEL" -eq 0 ]] && AWK_OPTIONS="-vskip_j=1"
-	[[ "$CONFIG_PRINTK_INDEX" -eq 0 ]] && AWK_OPTIONS="$AWK_OPTIONS -vskip_i=1"
+	[[ -z "$CONFIG_JUMP_LABEL" ]] && AWK_OPTIONS="-vskip_j=1"
+	[[ -z "$CONFIG_PRINTK_INDEX" ]] && AWK_OPTIONS="$AWK_OPTIONS -vskip_i=1"
 
 	SPECIAL_VARS="$(readelf -wi "$VMLINUX" |
 		gawk --non-decimal-data '
@@ -369,8 +369,8 @@ find_special_section_data_ppc64le() {
 	[[ -z "$FIXUP_STRUCT_SIZE" ]] && die "can't find special struct fixup_entry size"
 	[[ -z "$BUG_STRUCT_SIZE" ]] && die "can't find special struct bug_entry size"
 	[[ -z "$EX_STRUCT_SIZE" ]]  && die "can't find special struct exception_table_entry size"
-	[[ -z "$JUMP_STRUCT_SIZE" && "$CONFIG_JUMP_LABEL" -ne 0 ]]  && die "can't find special struct jump_entry size"
-	[[ -z "$PRINTK_INDEX_STRUCT_SIZE" && "$CONFIG_PRINTK_INDEX" -ne 0 ]]  && die "can't find special struct pi_entry size"
+	[[ -z "$JUMP_STRUCT_SIZE" ]] && [[ -n "$CONFIG_JUMP_LABEL" ]] && die "can't find special struct jump_entry size"
+	[[ -z "$PRINTK_INDEX_STRUCT_SIZE" ]] && [[ -n "$CONFIG_PRINTK_INDEX" ]] && die "can't find special struct pi_entry size"
 
 	return
 }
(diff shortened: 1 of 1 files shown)


joe-lawrence:
(I think we missed this since the corresponding unit test has to set PRINTK_INDEX_STRUCT_SIZE=44 itself, rather than parsing a vmlinux file.)

joe-lawrence:
Hmm, maybe the commit msg isn't entirely accurate.  The current state is that the previous commits _are active_, but we're missing: 1) verification that we found the size when the kernel config option is set, and 2) don't look for it (ie, `skip_i`) if it's not set in the kernel config.  Does that sound correct?

joe-lawrence:
v2:
- Updated the first commit msg with better explanation
- Added `AWK_OPTIONS` for ppc64le...  this reduced the awk script wall clock time from over 15minutes to 50seconds!

joe-lawrence:
v3:
- Appease the shellcheck bot :fire: :robot: :fire:

jpoimboe:
Looks good to me.  The first commit says that it fixes both "part 1" and "part 2" but I believe it's only fixing "part 2".

One idea, could we avoid this class of bug altogether by just sourcing the .config file?  Then all the config options will automatically become bash environment variables.

joe-lawrence:
> Looks good to me. The first commit says that it fixes both "part 1" and "part 2" but I believe it's only fixing "part 2".

With the first commit, "part 2" will take effect for `find_special_section_data()` as it always did apply `AWK_OPTIONS`, including this bit: `[[ "$CONFIG_PRINTK_INDEX" -eq 0 ]] && AWK_OPTIONS="$AWK_OPTIONS -vskip_i=1"`)

The second commit just copies the `AWK_OPTIONS` code to the ppc64le version.

I suppose the commit msgs could be even more accurate if I were to reorder the commits? 

> One idea, could we avoid this class of bug altogether by just sourcing the .config file? Then all the config options will automatically become bash environment variables.

I thought about this when adding yet another case to the config reading.  I'll take a look tomorrow to give it a try.  Any idea what the upper limit is on the number of environment variables? 
```
% grep ^CONFIG /boot/config-5.11.19-100.fc32.x86_64 | wc -l
6225
```

sm00th:
> I thought about this when adding yet another case to the config reading. I'll take a look tomorrow to give it a try. Any idea what the upper limit is on the number of environment variables?

AFAIR environment is limited by [ARG_MAX](https://www.in-ulm.de/~mascheck/various/argmax/), which is quite big nowadays. On my F34:
```
$ getconf ARG_MAX
2097152
```

It also seems variables need to be exported to count towards that:
```
$ echo "" | xargs --show-limits
Your environment variables take up 5312 bytes
POSIX upper limit on argument length (this system): 2089792
POSIX smallest allowable upper limit on argument length (all systems): 4096
Maximum length of command we could actually use: 2084480
Size of command buffer we are actually using: 131072
Maximum parallelism (--max-procs must be no greater): 2147483647

$ source redhat/configs/kernel-4.18.0-x86_64.config
$ echo "" | xargs --show-limits
Your environment variables take up 5312 bytes
POSIX upper limit on argument length (this system): 2089792
POSIX smallest allowable upper limit on argument length (all systems): 4096
Maximum length of command we could actually use: 2084480
Size of command buffer we are actually using: 131072
Maximum parallelism (--max-procs must be no greater): 2147483647

$ set -a
$ source redhat/configs/kernel-4.18.0-x86_64.config
$ echo "" | xargs --show-limits
Your environment variables take up 117503 bytes
POSIX upper limit on argument length (this system): 1977601
POSIX smallest allowable upper limit on argument length (all systems): 4096
Maximum length of command we could actually use: 1860098
Size of command buffer we are actually using: 131072
Maximum parallelism (--max-procs must be no greater): 2147483647
```

So even if we export the whole config we are nowhere near the limit and should be fine.

jpoimboe:
> With the first commit, "part 2" will take effect for `find_special_section_data()` as it always did apply `AWK_OPTIONS`, including this bit: `[[ "$CONFIG_PRINTK_INDEX" -eq 0 ]] && AWK_OPTIONS="$AWK_OPTIONS -vskip_i=1"`)
> 
> The second commit just copies the `AWK_OPTIONS` code to the ppc64le version.
> 
> I suppose the commit msgs could be even more accurate if I were to reorder the commits?

What I meant was that your patch `kpatch-build: read CONFIG_PRINTK_INDEX from kernel config` references two commits (Support CONFIG_PRINTK_INDEX "part 1" and "part 2") and has two "Fixes" tags.  But as far as I can tell, only "part 2" (commit 56471ffc7c0c) had a bug.

The other one, "part 1" (commit 6cf50a6fca84), only adds `_entry` to is_special_static(), which seems correct and isn't changed by your patch anyway.

joe-lawrence:
@sm00th 
> > I thought about this when adding yet another case to the config reading. I'll take a look tomorrow to give it a try. Any idea what the upper limit is on the number of environment variables?
> 
> AFAIR environment is limited by [ARG_MAX](https://www.in-ulm.de/~mascheck/various/argmax/), which is quite big nowadays...

Cool thanks, that's exactly what my google searches came up with and I didn't have time to really vet those sources.

@jpoimboe 
> What I meant was that your patch kpatch-build: read CONFIG_PRINTK_INDEX from kernel config references two commits (Support CONFIG_PRINTK_INDEX "part 1" and "part 2") and has two "Fixes" tags. But as far as I can tell, only "part 2" (commit 56471ff) had a bug.
> 
> The other one, "part 1" (commit 6cf50a6), only adds _entry to is_special_static(), which seems correct and isn't changed by your patch anyway.

Ah I get it now, yeah you are right about the "Fixes" tags. 

I think sourcing the kernel configuration is a good idea, so this PR will move to more of a refactoring theme.  The changes will still fix the above bugs, but hopefully be a bit more future proof in terms of new kernel options and future arch support.

kpatch-build/kpatch-build
```diff
@@ -233,6 +233,12 @@ support_klp_replace()
 	fi
 }
 
+kconfig_set()
+{
+	local config_setting="$1"
+	[[ "$config_setting" == "y" || "$config_setting" == "Y" ]]
+}
+
```
jpoimboe, line 241: I think helper function isn't really needed, and IMO hurts readability, when instead we can just do `[[ -n "$CONFIG_PRINTK_INDEX" ]]` or `[[ -z "$CONFIG_DEBUG_INFO" ]]`.

joe-lawrence:
-- updated to remove `kconfig_set()` helper

sm00th (approved):
Looks good. `find_special_section_data` still gives me headaches but at least it is now all in a single function.

kpatch-build/kpatch-build
joe-lawrence: @jpoimboe : the last push dropped the helper, look good?

jpoimboe (approved):

joe-lawrence:
(last push resolved minor merge conflict)

Issue #1238, opened 2021-11-25
KBUILD_EXTRA_SYMBOLS is needed when module dependencies exist


william-ye999:
Is this a **bug**? Assume that there are two kernel modules A and B. When module A uses the symbols exported by module B, kernel module of livepatch can be created for module A because KBUILD_EXTRA_SYMBOLS has been added to the Makefile of module A.

        1. SymB is defined in module B and is exported.
        2. Module A uses SymB.
        3. KBUILD_EXTRA_SYMBOLS is added in Makefile of module A.
        4. Make a livepatch to module A.

However, if the symbol exported by module B is not used in module A but the symbol is used in the livepatch, the livepatch fails to be created even if KBUILD_EXTRA_SYMBOLS is added to the Makefile of module A.

        1. SymB is defined in module B and is exported.
        2. Module A does not use SymB.
        3. KBUILD_EXTRA_SYMBOLS is added in Makefile of module A.
        4. Make a livepatch with SymB to module A.

This is because KBUILD_EXTRA_SYMBOLS needs to be added to kmod/patch/Makefile of kpatch too. If there are many livepatches for similar situations that need to be made, the contents of the kmod/patch/Makefile need to be constantly modified. How do we solve this problem?

Note: The dependency symbols are removed from Module.symvers in commit 52c3416db00d970c91a6992ab6e5ff48e077ad29 (modpost: track if the symbol origin is a dump file or ELF object) in the Linux community.

sm00th:
We actually already relay `KBUILD_EXTRA_SYMBOLS` via an env var [here](https://github.com/dynup/kpatch/blob/master/kpatch-build/kpatch-build#L1176). The problem is we currently [unset it by default](https://github.com/dynup/kpatch/blob/master/kpatch-build/kpatch-build#L835). Could you try removing that line and running kpatch-build as `KBUILD_EXTRA_SYMBOLS=... kpatch-build...` to see if it works properly with your usecase?

william-ye999:
Thank you for your reply.

As you said, I removed the KBUILD_EXTRA_SYMBOLS initialization part and tested it, but the result still failed. 

After I reverted the patch from the community and added KBUILD_EXTRA_SYMBOLS to the Makefile in module A and kmod/patch, the test was successful.

sm00th:
> As you said, I removed the KBUILD_EXTRA_SYMBOLS initialization part and tested it, but the result still failed.

[...]

PR #1242, opened 2021-12-06
OOT-related updates

sm00th:
This has two commits: one dealing with building OOT modules with non-distro kernels, this also includes some refactoring disambiguating `SRCDIR` since we now have 2 SRCDIRS - one for kernel and another one for oot-module. This superceeds #1234.

Second commit tries to deal with #1238.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -41,7 +41,7 @@ SCRIPTDIR="$(readlink -f "$(dirname "$(type -p "$0")")")"
 ARCH="$(uname -m)"
 CPUS="$(getconf _NPROCESSORS_ONLN)"
 CACHEDIR="${CACHEDIR:-$HOME/.kpatch}"
-SRCDIR="$CACHEDIR/src"
+KERNEL_SRCDIR="$CACHEDIR/src"
 RPMTOPDIR="$CACHEDIR/buildroot"
 VERSIONFILE="$CACHEDIR/version"
 TEMPDIR="$CACHEDIR/tmp"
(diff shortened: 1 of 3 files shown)


joe-lawrence (approved):
Changes look good and I like the SRCDIR distinctions.

kpatch-build/kpatch-build
```diff
@@ -540,13 +540,14 @@ usage() {
 	echo "		                        (can be specified multiple times)" >&2
 	echo "		-e, --oot-module        Enable patching out-of-tree module," >&2
 	echo "		                        specify current version of module" >&2
+	echo "		-p, --oot-module-src    Specify out-of-tree module source directory" >&2
```
joe-lawrence, line 543: Will -p conflict with #1231 ?   (FWIW, I'd say -p for --profile-data is more intuitive... For the OOT options maybe we just drop the short options?  What do you think?)

sm00th: Agreed, I wasn't sure about it either. Its just every other option has a short version but I guess you need to start somewhere. I'll remove short versions from both oot-module-* options.

joe-lawrence: Update looks good, thanks.

kpatch-build/kpatch-build
```diff
@@ -1161,11 +1172,7 @@ else
 fi
 
 cd "$TEMPDIR/patch" || die
-if [[ -z "$OOT_MODULE" ]]; then
-	KPATCH_BUILD="$SRCDIR"
-else
-	KPATCH_BUILD="/lib/modules/$ARCHVERSION/build"
-fi
+KPATCH_BUILD="$SRCDIR"
```
jpoimboe, line 1175: This line can be removed in favor of changing the below `export KPATCH_BUILD="$KPATCH_BUILD"` to `export KPATCH_BUILD="$SRCDIR"`.

kpatch-build/kpatch-build
```diff
@@ -41,7 +41,7 @@ SCRIPTDIR="$(readlink -f "$(dirname "$(type -p "$0")")")"
 ARCH="$(uname -m)"
 CPUS="$(getconf _NPROCESSORS_ONLN)"
 CACHEDIR="${CACHEDIR:-$HOME/.kpatch}"
-SRCDIR="$CACHEDIR/src"
+KERNEL_SRCDIR="$CACHEDIR/src"
```
jpoimboe, line 44: This SRCDIR/KERNEL_SRCDIR split is unfortunate, and is most likely going to be a future source of bugs (but I don't have any better ideas).

At the very least, a comment would help to describe the difference between the two for future code readers trying to make sense of it.

kpatch-build/kpatch-build
sm00th: Will do

kpatch-build/kpatch-build
sm00th: How about we rename the leftover `SRCDIR` to `BUILDDIR`? I think it will better convey it's purpose and then `KERNEL_SRCDIR`/`OOT_MODULE_SRCDIR` are easier to understand as well.

jpoimboe: Yes, that's much better.  Then there would be no comment needed since the variables are more self-descriptive.

sm00th:
v2: `SRCDIR`->`BUILDDIR` and cleaner `KPATCH_BUILD` assignment.

jpoimboe (approved):
Looks good, the variable rename helped a lot.

This does introduce user-visible breakage due to the changing behavior of `--sourcedir` with OOT modules.  We should try to remember to bump the .y version in the next release.

PR #1241, opened 2021-12-06
create-diff-object: update for __already_done

joe-lawrence:
Upstream v5.14+ kernel change a358f40600b3 ("once: implement
DO_ONCE_LITE for non-fast-path "do once" functionality") consolidated a
bunch of do-once macros into a common macro:

  #define DO_ONCE_LITE_IF(condition, func, ...)				\
  	({								\
  		static bool __section(".data.once") __already_done;	\
  		...

which replaced static local variable __warned with __already_done.

Update any __warned static local checks to also look for the new
__already_done variable as well.

Signed-off-by: Joe Lawrence <joe.lawrence@redhat.com>

PR #1243, opened 2021-12-08
Detect S390 line macro updates

C-Erastus:
This pull request implements kpatch_line_macro_change_only() for s390x. Since this statcks on top of #1203, ignore all commits but the last one.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2616,8 +2616,24 @@ static void kpatch_check_relocations(struct kpatch_elf *kelf)
 		if (!is_rela_section(sec))
 			continue;
 		list_for_each_entry(rela, &sec->relas, list) {
+			/*
+			 * The size of the sdata->d_size could be zero.
+			 * Usecase:
+			 * [327] .bss.__key.18
+			 *        NOBITS 0000000000000000 007284 000000 00   0   0  2
+			 *
+			 * Relocation section '.rela.text.copy_signal' at offset 0x7f9c8 contains 19 entries:
+			 * (...)
+			 * 0000000000000090  000000f90000001a R_390_GOTENT 0000000000000000 __key.18 + 2
+			 *
+			 *   8e:   c4 48 00 00 00 00       lgrl    %r4,8e <copy_signal+0x8e>
+			 *               90: R_390_GOTENT        __key.18+0x2
+			 * which is a valid relocation
+			 */
 			if (rela->sym->sec) {
 				sdata = rela->sym->sec->data;
+				if (!sdata->d_size)
+					continue;
 				if ((long)rela->sym->sym.st_value + rela->addend > (long)sdata->d_size) {
 					ERROR("out-of-range relocation %s+%lx in %s", rela->sym->name,
 							rela->addend, sec->name);
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -48,7 +48,11 @@
 
 #include "list.h"
 #include "lookup.h"
+
+#ifdef __x86_64__
 #include "asm/insn.h"
+#endif
+
 #include "kpatch-patch.h"
 #include "kpatch-elf.h"
 #include "kpatch-intermediate.h"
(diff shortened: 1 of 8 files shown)


kpatch-build/create-diff-object.c
```diff
@@ -771,6 +771,88 @@ static bool kpatch_line_macro_change_only(struct section *sec)
 
 	return true;
 }
+
+#elif __s390x__
+
+static int _insn_get_length(unsigned char *inst)
```
sm00th, line 777: If I understand correctly the only reason you are prefixing this function with a `_` is that it is already defined in `insn/asm/insn.h`. We only build insn dir for x86_64 so I think that define can be safely moved under `#ifdef __x86_64__` and then we'll be able to use common `insn_get_length` name.

kpatch-build/create-diff-object.c
```diff
@@ -771,6 +771,88 @@ static bool kpatch_line_macro_change_only(struct section *sec)
...
+	for (offset = 0; offset < size; offset += length) {
+		length = _insn_get_length(start1 + offset);
+
+		if (_insn_get_length(start1 + offset) != _insn_get_length(start2 + offset))
+			return false;
+
+		if (!memcmp(start1 + offset, start2 + offset, length))
+			continue;
+
+		if ((start1[offset] != 0xa7) || (start1[offset+1] != 0x49) ||
```
joe-lawrence, line 824: I know the disassembly is in the commit msg, but a quick one line comment (like the x86_64 version) before the instruction comparison would be nice:  `/* verify lghi %r4, <line number> */`

kpatch-build/create-diff-object.c
```diff
@@ -771,6 +771,88 @@ static bool kpatch_line_macro_change_only(struct section *sec)
...
+		    (start2[offset] != 0xa7) || (start2[offset+1] != 0x49))
+			return false;
+
+		/*
+		 * Verify zero or more string relas followed by a
+		 * warn_slowpath_* or another similar rela.
+		 */
+		found = 0;
+		list_for_each_entry(rela, &sec->rela->relas, list) {
```
joe-lawrence, line 833: For the other arches, we look for a relocation to static local variable `__warned` (and soon `__already_done` per #1241).  In this commit msg's disassembly listing, the compiled order throughout the object file was:

1. `R_390_PC32DBL -> .data.once+0x2` is static local  `__warned.xxxx` or `__already_done.xxxxx`
2. `lghi %r4, <line number>` 
3. `R_390_PC32DBL -> .LC1+0x2` is a string rela
4. `R_390_PC32DBL -> .LC2+0x2` is a string rela
5. `R_390_PLT32DBL-> __warn_printk+0x2` is function rela

When testing this change, @C-Erastus and I only focused on pattern parts 2-5 above.  Can we make any assumptions about the relocation ordering, or should we add back checks for the static local variables just in case they land after the `lghi` instruction?

kpatch-build/create-diff-object.c
joe-lawrence: +1

kpatch-build/create-diff-object.c
```diff
@@ -771,6 +771,88 @@ static bool kpatch_line_macro_change_only(struct section *sec)
...
+		case 0b10:
+			length = 4;
+			break;
+		case 0b11:
+			length = 6;
+			break;
+	}
+
+	return length;
+}
```
joe-lawrence, line 795: Minor style suggestion: we can save a few lines by skipping the local variable and breaks without sacrificing readability:

``` C
static int _insn_get_length(unsigned char *inst)
{
	switch (inst[0] >> 6) {
	case 0b00:
		return 2;
	case 0b01:
	case 0b10:
		return 4;
	case 0b11:
		return 6;
	}
}
```

kpatch-build/create-diff-object.c
sm00th: I have no knowledge of whether this ordering can be guaranteed but it seems to me that having the check 'just in case' has more upsides than downsides.

joe-lawrence:
Hi @C-Erastus : we don't typically put commit version changelog into the commit messages.  For github / gitlab pull requests, we just add a comment to the PR conversation.  Some folks ignore all the git push updates, so leaving version info as an update to the review thread makes it obvious that you are requesting a follow up review.  Thanks!

joe-lawrence:
> Hi @C-Erastus : we don't typically put commit version changelog into the commit messages. For github / gitlab pull requests, we just add a comment to the PR conversation. Some folks ignore all the git push updates, so leaving version info as an update to the review thread makes it obvious that you are requesting a follow up review. Thanks!

@C-Erastus : bump

kpatch-build/create-diff-object.c
```diff
@@ -771,6 +775,86 @@ static bool kpatch_line_macro_change_only(struct section *sec)
...
+	for (offset = 0; offset < size; offset += length) {
+		length = insn_get_length(start1 + offset);
+
+		if (insn_get_length(start1 + offset) != insn_get_length(start2 + offset))
+			return false;
+
+		if (!memcmp(start1 + offset, start2 + offset, length))
+			continue;
+
+		/* Verify lghi %r4 <line number> */
```
sumanthkorikkar, line 822: When the line number is greater than 2 bytes,  verification of lgfi %r4 <line number> would be essential

kpatch-build/create-diff-object.c
```diff
@@ -771,6 +775,86 @@ static bool kpatch_line_macro_change_only(struct section *sec)
...
+		 * warn_slowpath_* or another similar rela.
+		 */
+		found = 0;
+		list_for_each_entry(rela, &sec->rela->relas, list) {
+			if (rela->offset < offset + length)
+				continue;
+			if (rela->string)
+				continue;
+			if (!strncmp(rela->sym->name, "__warned.", 9) ||
+			    !strncmp(rela->sym->name, "__already_done.", 15))
```
sumanthkorikkar, line 838: Addition of .__func__ is necessary

jpoimboe: I have a `__func__` check coming soon in #1254.

kpatch-build/create-diff-object.c
```diff
@@ -771,6 +775,86 @@ static bool kpatch_line_macro_change_only(struct section *sec)
...
+				continue;
+			if (!strcmp(rela->sym->name, "__warn_printk")) {
```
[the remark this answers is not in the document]
jpoimboe: I guess this corresponds to the following line in find_build_dev_info():
```
pr_err("out of memory at line %d\n", __LINE__);
```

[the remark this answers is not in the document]
sumanthkorikkar: Yes.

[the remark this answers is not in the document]
jpoimboe: I've added a `_printk` check to #1254, since we're already checking for `printk`, which has been renamed to `_printk` for newer kernels.

kpatch-build/create-diff-object.c
C-Erastus: Hi @sumanthkorikkar : after investigating the above concern, @joe-lawrence and I found that we don't need to verify lgfi %4 when the line number is greater than 2 bytes as no c file in the kernel has lines exceeding 50k.

Verification method: find . -name '*.c' -exec wc -l {} \; | sort -n | tail -n5

sumanthkorikkar: ok, fine with it. We could add that in the future, if there is a need.

joe-lawrence: For posterity:  large numbers might result in other instruction combinations in addition to lgfi, for example from [Compiler Explorer's](https://godbolt.org/z/xv3vTnqx9) gcc 11.2.0:
```C
printf("%l\n", 0x7fff);      // lghi    %r3,32767
printf("%l\n", 0x8000);      // llill   %r3,32768
printf("%l\n", 0x10000);     // llilh   %r3,1
printf("%l\n", 0x10001);     // larl    %r13,.L3
                             // lg      %r3,.L4-.L3(%r13)
                             // .L3:
                             // .L4:
                             // .quad   65537

```

joe-lawrence:
Since @jpoimboe folded this into #1254, I'm going to consider this work as complete (at least on par with other arches).  @sumanthkorikkar raised a few review issues, which I think were all covered in #1254.  He also noted that large files with __LINE__ > 2 bytes may generate additional code patterns.  At the moment, no such files appear to exist in the upstream kernel tree, so we will defer implementing those for now.  Thanks everyone for code and review comments.

PR #1246, opened 2022-01-13
kpatch-build: Add missing malloc failure check

Decave:
In kpatch_create_mcount_sections(), there is a section of code wherein
on x86-64 builds, if the rela type for a symbol is R_X86_64_NONE, a
buffer into which a writeable copy of the symbol's .text is allocated.
This is done so that the .text section may be copied into the buffer,
and the NOP call that is generated for the function entry (by older
versions of the kernel/gcc) can be updated to be a call to fentry.

In all other calls in kpatch-build, any allocation is checked for
failures, and gracefully failed with a call to ERROR(). This diff
corrects the one aforementioned spot where this was missed.

Signed-off-by: David Vernet <void@manifault.com>

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -911,6 +911,8 @@ do {								\
 		log_debug("renaming %s %s to %s\n",		\
 			  kindstr, e2->name, e1->name);		\
 		e2->name = strdup(e1->name);			\
+		if (!e2->name)				        \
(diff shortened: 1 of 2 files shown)


sm00th (approved):

sm00th:
Hi, forcepush sounds better.

sm00th (approved):
Looks good, thanks for the patch.

joe-lawrence (approved):

PR #1248, opened 2022-01-21
Remove host / target architecture assumption from create-diff-object

joe-lawrence:
This is a remix of @gwelymernans 's #1179 with a few slight tweaks:

1) Do not add an `--arch` command line argument to various tools and kpatch-build.  Detect the ELF architecture of the input files and pivot based on that.  This would allow for a (theoretical at this point) cross-arch-build to build a ppc64le reference/patched kernel and when processing its object files, rely on the ELF header to indicate the target architecture.
2) Modifications limited to create-diff-object for now.  Without testing full cross-arch-build support, this is still useful to the project as we can test the ppc64le unit tests (with a very small patch: [unit-objs-cross.diff.txt](https://github.com/dynup/kpatch/files/7916207/unit-objs-cross.diff.txt)):
``` bash
$ arch
x86_64

$ make -s unit OBJDIR=objs/ppc64le
BUILD function-ptr-new
BUILD ASSERT_RTNL-detect
BUILD meminfo-init-FAIL
BUILD bug-table-section
BUILD smp-locks-section
BUILD fixup-section
BUILD meminfo-string
BUILD gcc-static-local-var-5
BUILD meminfo-init2-FAIL
BUILD gcc-static-local-var-3
BUILD mcount_loc-issue-1102
BUILD gcc-constprop-2-issue-935
BUILD ppc64le-bundled-localentry-issue-1007
BUILD static-local-size-mismatch-issue-1053
BUILD data-read-mostly
BUILD sec-addralign
BUILD tracepoints-section
BUILD data-rel
BUILD jump-label-issue-946
BUILD data-new
BUILD rela-common-symbols
BUILD gcc-constprop
BUILD special-static
BUILD warn-detect-FAIL
BUILD ppc64le-sibling-call-issue-1003-FAIL
BUILD new-function
TEST gcc-constprop
TEST gcc-constprop-2-issue-935
TEST gcc-static-local-var-5
TEST function-ptr-new
TEST bug-table-section
TEST meminfo-string
TEST gcc-static-local-var-3
TEST ppc64le-bundled-localentry-issue-1007
TEST ASSERT_RTNL-detect
TEST special-static
TEST static-local-size-mismatch-issue-1053
TEST data-read-mostly
TEST fixup-section
TEST mcount_loc-issue-1102
TEST sec-addralign
TEST tracepoints-section
TEST smp-locks-section
TEST data-rel
TEST data-new
TEST rela-common-symbols
TEST new-function
```

With the preparatory first commit and @gwelymernans 's previous heavy lifting, this removes all of the architecture-specific  preprocessor blocks from kpatch-build:
``` bash
$ grep -c -e '__powerpc64__' -e '__x86_64__' $(git ls-files kpatch-build/) 
kpatch-build/Makefile:0
kpatch-build/create-diff-object.c:0
kpatch-build/create-klp-module.c:0
kpatch-build/create-kpatch-module.c:0
kpatch-build/gcc-plugins/gcc-common.h:0
kpatch-build/gcc-plugins/gcc-generate-rtl-pass.h:0
kpatch-build/gcc-plugins/ppc64le-plugin.c:0
kpatch-build/insn/asm/inat.h:0
kpatch-build/insn/asm/inat_types.h:0
kpatch-build/insn/asm/insn.h:0
kpatch-build/insn/inat-tables.c:0
kpatch-build/insn/inat.c:0
kpatch-build/insn/insn.c:0
kpatch-build/kpatch-build:0
kpatch-build/kpatch-cc:0
kpatch-build/kpatch-elf.c:0
kpatch-build/kpatch-elf.h:0
kpatch-build/kpatch-intermediate.h:0
kpatch-build/kpatch.h:0
kpatch-build/list.h:0
kpatch-build/log.h:0
kpatch-build/lookup.c:0
kpatch-build/lookup.h:0
```
Leaving as draft for now as there would be a unit-test-objs update required and I also haven't tried it on a ppc64le host yet.

### DIFF
--- a/kpatch-build/Makefile
+++ b/kpatch-build/Makefile
@@ -10,11 +10,10 @@ SOURCES = create-diff-object.c kpatch-elf.c \
 		  create-kpatch-module.c \
 		  create-kpatch-module.c lookup.c
 
-ifeq ($(ARCH),x86_64)
 SOURCES += insn/insn.c insn/inat.c
 INSN     = insn/insn.o insn/inat.o
 insn/%.o: CFLAGS := $(filter-out -Wconversion, $(CFLAGS))
-else ifeq ($(ARCH),ppc64le)
+ifeq ($(ARCH),ppc64le)
 SOURCES += gcc-plugins/ppc64le-plugin.c
 PLUGIN   = gcc-plugins/ppc64le-plugin.so
 TARGETS += $(PLUGIN)
@@ -22,7 +21,7 @@ GCC_PLUGINS_DIR := $(shell gcc -print-file-name=plugin)
 PLUGIN_CFLAGS := $(filter-out -Wconversion, $(CFLAGS))
 PLUGIN_CFLAGS += -shared -I$(GCC_PLUGINS_DIR)/include \
 		   -Igcc-plugins -fPIC -fno-rtti -O2 -Wall
-else
+else ifneq ($(ARCH),x86_64)
 $(error Unsupported architecture ${ARCH}, check https://github.com/dynup/kpatch/#supported-architectures)
 endif
 
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -3665,6 +3665,35 @@ static void kpatch_no_sibling_calls_ppc64le(struct kpatch_elf *kelf)
 #endif
 }
 
+/* Check which functions have fentry/mcount calls; save this info for later use. */
+static void kpatch_find_func_profiling_calls(struct kpatch_elf *kelf)
+{
+	struct symbol *sym;
+	struct rela *rela;
+	list_for_each_entry(sym, &kelf->symbols, list) {
+		if (sym->type != STT_FUNC || !sym->sec || !sym->sec->rela)
+			continue;
+#ifdef __powerpc64__
+		list_for_each_entry(rela, &sym->sec->rela->relas, list) {
+			if (!strcmp(rela->sym->name, "_mcount")) {
+				sym->has_func_profiling = 1;
+				break;
+			}
+		}
+#else
+		rela = list_first_entry(&sym->sec->rela->relas, struct rela,
+					list);
+		if ((rela->type != R_X86_64_NONE &&
+		     rela->type != R_X86_64_PC32 &&
+		     rela->type != R_X86_64_PLT32) ||
+		    strcmp(rela->sym->name, "__fentry__"))
+			continue;
+
+		sym->has_func_profiling = 1;
+#endif
+	}
+}
+
 struct arguments {
 	char *args[7];
 	bool debug, klp_arch;
@@ -3741,7 +3770,9 @@ int main(int argc, char *argv[])
 	childobj = basename(orig_obj);
 
 	kelf_orig = kpatch_elf_open(orig_obj);
+	kpatch_find_func_profiling_calls(kelf_orig);
 	kelf_patched = kpatch_elf_open(patched_obj);
+	kpatch_find_func_profiling_calls(kelf_patched);
 
 	kpatch_compare_elf_headers(kelf_orig->elf, kelf_patched->elf);
 	kpatch_check_program_headers(kelf_orig->elf);
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -60,12 +60,6 @@
 	error(EXIT_STATUS_DIFF_FATAL, 0, "unreconcilable difference"); \
 })
 
-#ifdef __powerpc64__
-#define ABSOLUTE_RELA_TYPE R_PPC64_ADDR64
-#else
-#define ABSOLUTE_RELA_TYPE R_X86_64_64
-#endif
-
 char *childobj;
 
 enum subsection {
@@ -83,6 +77,7 @@ bool KLP_ARCH;
  * ****************/
 struct special_section {
 	char *name;
+	enum architecture arch;
 	int (*group_size)(struct kpatch_elf *kelf, int offset);
 };
 
@@ -137,7 +132,6 @@ static int is_bundleable(struct symbol *sym)
 	return 0;
 }
 
-#ifdef __powerpc64__
 /* Symbol st_others value for powerpc */
 #define STO_PPC64_LOCAL_BIT     5
 #define STO_PPC64_LOCAL_MASK    (7 << STO_PPC64_LOCAL_BIT)
@@ -161,17 +155,21 @@ static int is_bundleable(struct symbol *sym)
  * the object file.  The local entry point is 8 bytes after the global entry
  * point.
  */
-static int is_gcc6_localentry_bundled_sym(struct symbol *sym)
+static bool is_gcc6_localentry_bundled_sym(struct kpatch_elf *kelf,
+					  struct symbol *sym)
 {
-	return ((PPC64_LOCAL_ENTRY_OFFSET(sym->sym.st_other) != 0) &&
-		sym->sym.st_value == 8);
+	switch(kelf->arch) {
+	case PPC64:
+		return ((PPC64_LOCAL_ENTRY_OFFSET(sym->sym.st_other) != 0) &&
+			sym->sym.st_value == 8);
+	case X86_64:
+		return false;
+	default:
+		ERROR("unsupported arch");
+	}
+
+	return false;
 }
-#else
-static int is_gcc6_localentry_bundled_sym(struct symbol *sym)
-{
-	return 0;
-}
-#endif
 
 /*
  * On ppc64le, when a function references data, it does so indirectly, via the
@@ -226,7 +224,7 @@ static void kpatch_bundle_symbols(struct kpatch_elf *kelf)
 	list_for_each_entry(sym, &kelf->symbols, list) {
 		if (is_bundleable(sym)) {
 			if (sym->sym.st_value != 0 &&
-			    !is_gcc6_localentry_bundled_sym(sym)) {
+			    !is_gcc6_localentry_bundled_sym(kelf, sym)) {
 				ERROR("symbol %s at offset %lu within section %s, expected 0",
 				      sym->name, sym->sym.st_value,
 				      sym->sec->name);
@@ -579,7 +577,6 @@ out:
 		log_debug("section %s has changed\n", sec->name);
 }
 
-#ifdef __x86_64__
 /*
  * Determine if a section has changed only due to a WARN* or might_sleep
  * macro call's embedding of the line number into an instruction operand.
@@ -611,7 +608,7 @@ out:
  * 3) (optional) __warned.xxxxx or __already_done.xxxxx static local rela
  * 4) warn_slowpath_* or __might_sleep or some other similar rela
  */
-static bool kpatch_line_macro_change_only(struct section *sec)
+static bool kpatch_line_macro_change_only_x86_64(struct section *sec)
 {
 	struct insn insn1, insn2;
 	unsigned long start1, start2, size, offset, length;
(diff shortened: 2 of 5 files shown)


joe-lawrence:
bumped the unit-test-objs submodule reference and tweaked the travis file... @sm00th : not sure if you prefer making a change like this, or modify one of the Makefiles (top-level or unit test) to run for both arches.

sm00th:
> @sm00th : not sure if you prefer making a change like this, or modify one of the Makefiles (top-level or unit test) to run for both arches.

I am starting to think that we might want to remove the arch-specific logic from unit-tests completely now and just run everything every time. But this can come later.

Travis change is ok as a first step (although you also want to update .github/workflows/unit.yml), the problem with this approach is that people running unit tests locally won't run into possible ppc64le issues until they push an MR.

joe-lawrence:
> > @sm00th : not sure if you prefer making a change like this, or modify one of the Makefiles (top-level or unit test) to run for both arches.
> 
> I am starting to think that we might want to remove the arch-specific logic from unit-tests completely now and just run everything every time. But this can come later.
> 
> Travis change is ok as a first step (although you also want to update .github/workflows/unit.yml), the problem with this approach is that people running unit tests locally won't run into possible ppc64le issues until they push an MR.

Oh right, I was wondering why I didn't see my update run here in the github PR checks.  I can add the change over in unit.yml, but first:  if this were to be introduced gradually, does it make sense to implement it for local execution first, then add to gitlab/travis automation?  Alternatively, we could take the plunge now and just remove the arch checks from the unit tests.

sm00th:
> if this were to be introduced gradually, does it make sense to implement it for local execution first, then add to gitlab/travis automation?

I don't see why won't we want it running all the time, ci checks for this repo are not required to pass for us to merge a change so it shouldn't hurt.

>  Alternatively, we could take the plunge now and just remove the arch checks from the unit tests.

Might be the most straightforward course.

joe-lawrence:
> > Alternatively, we could take the plunge now and just remove the arch checks from the unit tests.
> 
> Might be the most straightforward course.

Ok, bombs away.  I removed the RFC, so kpatch-build source code style / nits and Makefile gotchas welcome. :)

sm00th (approved):
Looks good

joe-lawrence:
Fwiw, there is more to do in create-diff-object for making it endian safe.  I have some of it already done, but since s390x is not officially supported yet, I'll post as new PRs as I progress/have time.

jpoimboe (commented):
Very nice change, I added some style nits, as you requested ;-)

joe-lawrence:
Updated for @jpoimboe's review comments.  Incremental diff: [changes.diff.txt](https://github.com/dynup/kpatch/files/7988954/changes.diff.txt)

kpatch-build/create-diff-object.c
```diff
@@ -76,13 +72,21 @@ enum subsection {
 
 enum loglevel loglevel = NORMAL;
 
+enum architecture {
+	X86_64 = 0x1 << 0,
+	PPC64  = 0x1 << 1,
+};
+
+enum architecture target_arch;
```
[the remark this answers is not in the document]
joe-lawrence: Ok, this required some plumbing of struct kpatch_elf *kelf through the API, but not too unreasonable.

kpatch-build/create-diff-object.c
```diff
@@ -163,15 +166,13 @@ static int is_bundleable(struct symbol *sym)
  */
 static int is_gcc6_localentry_bundled_sym(struct symbol *sym)
```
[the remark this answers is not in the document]
joe-lawrence: Ok

kpatch-build/create-diff-object.c
```diff
@@ -1541,22 +1542,23 @@ static void kpatch_replace_sections_syms(struct kpatch_elf *kelf)
...
-			    rela->type == R_X86_64_PLT32) {
-				struct insn insn;
-				rela_insn(sec, rela, &insn);
-				add_off = (unsigned int)((long)insn.next_byte -
-					  (long)sec->base->data->d_buf -
-					  rela->offset);
-			} else if (rela->type == R_X86_64_64 ||
-				   rela->type == R_X86_64_32S)
+			if (target_arch == PPC64)
 				add_off = 0;
```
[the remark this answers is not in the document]
joe-lawrence: I ended up reverting the change to move the initializer to the top.   I believe some of the earlier code shuffling invoked a use-but-not-initialized warning from the -O2 build.  I no longer hit that warning, and think the `add_off = 0` here helps differentiate the two cases (ppc64le vs x86).

kpatch-build/create-diff-object.c
```diff
@@ -2169,68 +2168,80 @@ static int fixup_group_size(struct kpatch_elf *kelf, int offset)
...
 	},
 	{
 		.name		= ".static_call_sites",
+		.arch		= X86_64,
 		.group_size	= static_call_sites_group_size,
 	},
 	{
 		.name		= ".retpoline_sites",
+		.arch		= X86_64,
+		.arch		= X86_64,
```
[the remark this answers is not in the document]
joe-lawrence: Good eye.  That was me telling the compiler I really want it to assign that value.

kpatch-build/create-diff-object.c
```diff
@@ -163,15 +166,13 @@ static int is_bundleable(struct symbol *sym)
...
-		sym->sym.st_value == 8);
-}
-#else
-static int is_gcc6_localentry_bundled_sym(struct symbol *sym)
-{
+	if (target_arch == PPC64) {
+		return ((PPC64_LOCAL_ENTRY_OFFSET(sym->sym.st_other) != 0) &&
+			sym->sym.st_value == 8);
+	}
+
```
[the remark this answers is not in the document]
joe-lawrence: Good point.  I updated all of those sites and also used this opportunity to alphabetize them.  (OCD, maybe.)

kpatch-build/create-diff-object.c
```diff
@@ -3662,7 +3682,56 @@ static void kpatch_no_sibling_calls_ppc64le(struct kpatch_elf *kelf)
...
+	switch (ehdr.e_machine) {
+	case (EM_X86_64) :
+		target_arch = X86_64;
+		ABSOLUTE_RELA_TYPE = R_X86_64_64;
+		break;
+	case (EM_PPC64) :
+		target_arch = PPC64;
+		ABSOLUTE_RELA_TYPE = R_PPC64_ADDR64;
+		break;
+	default: ERROR("Unsupported target architecture");
```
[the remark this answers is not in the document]
joe-lawrence: Fixed

kpatch-build/Makefile
```diff
@@ -10,19 +10,18 @@ SOURCES = create-diff-object.c kpatch-elf.c \
 		  create-kpatch-module.c \
 		  create-kpatch-module.c lookup.c
 
-ifeq ($(ARCH),x86_64)
 SOURCES += insn/insn.c insn/inat.c
 INSN     = insn/insn.o insn/inat.o
 insn/%.o: CFLAGS := $(filter-out -Wconversion, $(CFLAGS))
-else ifeq ($(ARCH),ppc64le)
+ifeq ($(ARCH),ppc64le)
```
jpoimboe, line 16: It's probably not a big deal, but believe it or not, there are people out there who cross-compile for x86 on a ppc64le host.  Does the plugin not compile on x86?

kpatch-build/create-diff-object.c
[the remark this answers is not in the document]
jpoimboe: IMO a switch statement would still help readability here (and be consistent with the other arch checks)

kpatch-build/create-diff-object.c
```diff
@@ -161,17 +155,16 @@ static int is_bundleable(struct symbol *sym)
...
-static int is_gcc6_localentry_bundled_sym(struct symbol *sym)
+static bool is_gcc6_localentry_bundled_sym(struct kpatch_elf *kelf,
+					  struct symbol *sym)
 {
+	if (kelf->target_arch == PPC64) {
+		return ((PPC64_LOCAL_ENTRY_OFFSET(sym->sym.st_other) != 0) &&
+			sym->sym.st_value == 8);
+	}
+
 	return 0;
```
jpoimboe, line 166: This should return false instead of 0

kpatch-build/create-diff-object.c
```diff
@@ -1541,22 +1540,27 @@ static void kpatch_replace_sections_syms(struct kpatch_elf *kelf)
...
-					  (long)sec->base->data->d_buf -
-					  rela->offset);
-			} else if (rela->type == R_X86_64_64 ||
-				   rela->type == R_X86_64_32S)
-				add_off = 0;
-			else
-				continue;
-#endif
+			switch(kelf->target_arch) {
+				case PPC64:
```
jpoimboe, line 1544: The kernel style is to put 'case' on the same indentation level as 'switch', which helps reduce excessive indentation

kpatch-build/create-diff-object.c
```diff
@@ -161,17 +155,16 @@ static int is_bundleable(struct symbol *sym)
...
-{
-	return ((PPC64_LOCAL_ENTRY_OFFSET(sym->sym.st_other) != 0) &&
-		sym->sym.st_value == 8);
-}
-#else
-static int is_gcc6_localentry_bundled_sym(struct symbol *sym)
+static bool is_gcc6_localentry_bundled_sym(struct kpatch_elf *kelf,
+					  struct symbol *sym)
 {
+	if (kelf->target_arch == PPC64) {
```
jpoimboe, line 161: I think the 'target_' prefix in 'target_arch' is redundant since the 'kelf' object already represents the target

kpatch-build/Makefile
joe-lawrence: I haven't looked at the plugin just yet.  The scope of this PR was to only incrementally help create-diff-object so that it may read/write cross-arch ELF for the pre-built unit test objects.  AFAIK, the plugin only comes into play for the kernel builds, right?

kpatch-build/create-diff-object.c
joe-lawrence: Is it?  I've observed 50/50 across the kernel, and in fact create-diff-object.c as well :P  It looks weird when the last case  adds it's own bracketed block like:
```C
switch(foo) {
case 1:
    /* code */
case 99: {
    /* code */
}                          <- appears unbalanced
}
```
but none of my redundant default cases introduce a local scope, so that never happens here.

Anyway, it's not a big deal, I'll update to help the greater indentation cause.

joe-lawrence:
All review comments implemented, Incremental diff: [changes.diff.txt](https://github.com/dynup/kpatch/files/7989947/changes.diff.txt)

kpatch-build/create-diff-object.c
jpoimboe: > Is it? I've observed 50/50 across the kernel

Really?  `git grep -A1 switch` seems to show otherwise.

> and in fact create-diff-object.c as well :P

I'm not surprised by that, but it's never too late to change :-)

> It looks weird when the last case adds it's own bracketed block like:
> 
> ```c
> switch(foo) {
> case 1:
>     /* code */
> case 99: {
>     /* code */
> }                          <- appears unbalanced
> }
> ```

Ew, hopefully we can avoid the need for that.


> Anyway, it's not a big deal, I'll update to help the greater indentation cause.

Thanks :-)

kpatch-build/Makefile
jpoimboe: Ok, makes sense

kpatch-build/create-diff-object.c
jpoimboe: > Ew, hopefully we can avoid the need for that.

I see now that we did need it.  Looks reasonable though :-)

jpoimboe (approved):

joe-lawrence:
Thanks for the review.  One (hopefuly last) update.. it seems like I lost a `rela->type` assignment along the way in  kpatch_create_kpatch_arch_section(), so I added that back in.  With that, I'll step back and let our internal integration test bot churn on the PR.

PR #1249, opened 2022-02-03
Fix sym->type condition in locals_match()

sm00th:
Second loop in locals_match() checks table_sym->type instead of
sym->type.

Fixes: 7207687 Switch to per-file lookup table pointers.


joe-lawrence (approved):
Ah, copy/paste bug?  How did you spot this one?

sm00th:
> Ah, copy/paste bug?

Yes, looks like one.

> How did you spot this one?

Have been working on lto and it kept failing in locals_match() in a strange fashion. :)

 PS. I think we'll need to backport this to rhel9.

PR #1251, opened 2022-02-22
symtab_shndx support

sm00th:
Symbol indexes over 64k don't fit into st_shndx and are stored in
extended symbol table. Make sure we properly handle these cases. It is
not normally useful during kpatch builds but will come up if we start
diffing linked objects.

### DIFF
--- a/kpatch-build/kpatch-elf.c
+++ b/kpatch-build/kpatch-elf.c
@@ -263,6 +263,10 @@ void kpatch_create_section_list(struct kpatch_elf *kelf)
 
 		sec->index = (unsigned int)elf_ndxscn(scn);
 
+
+		if (sec->sh.sh_type == SHT_SYMTAB_SHNDX)
+			kelf->symtab_shndx = sec->data;
+
 		log_debug("ndx %02d, data %p, size %zu, name %s\n",
 			sec->index, sec->data->d_buf, sec->data->d_size,
 			sec->name);
@@ -278,6 +282,7 @@ void kpatch_create_symbol_list(struct kpatch_elf *kelf)
 	struct section *symtab;
 	struct symbol *sym;
 	unsigned int symbols_nr, index = 0;
+	Elf32_Word shndx;
 
 	symtab = find_section_by_name(&kelf->sections, ".symtab");
 	if (!symtab)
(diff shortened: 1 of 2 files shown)


joe-lawrence (approved):

jpoimboe:
This can be done more cleanly if the shndx conversion is done before the valid section range check:

https://github.com/jpoimboe/kpatch/commit/ba576df25b62c8b4e799a6d3a24ae1cf0d4ca8ba

sm00th:
> This can be done more cleanly if the shndx conversion is done before the valid section range check:
> 
> [jpoimboe@f159007](https://github.com/jpoimboe/kpatch/commit/f1590074515bb7a4a57c67ff38d991a7ba43c669)

Unfortunately I don't think that'll work. `gelf_getsymshndx()` can yield results well over `SHN_LORESERVE` so the `shndx < SHN_LORESERVE` on line 319 in your version will always fail.

jpoimboe:
> > This can be done more cleanly if the shndx conversion is done before the valid section range check:
> > [jpoimboe@f159007](https://github.com/jpoimboe/kpatch/commit/f1590074515bb7a4a57c67ff38d991a7ba43c669)
> 
> Unfortunately I don't think that'll work. `gelf_getsymshndx()` can yield results well over `SHN_LORESERVE` so the `shndx < SHN_LORESERVE` on line 319 in your version will always fail.

Ah, right.  What do you think about this version which adds a `sym->sym.st_shndx == SHN_XINDEX` check to that?

https://github.com/jpoimboe/kpatch/commit/0f7b69ae7b9dfaab358324b53100322d1faf34bc

sm00th:
> > > This can be done more cleanly if the shndx conversion is done before the valid section range check:
> > > [jpoimboe@f159007](https://github.com/jpoimboe/kpatch/commit/f1590074515bb7a4a57c67ff38d991a7ba43c669)
> > 
> > 
> > Unfortunately I don't think that'll work. `gelf_getsymshndx()` can yield results well over `SHN_LORESERVE` so the `shndx < SHN_LORESERVE` on line 319 in your version will always fail.
> 
> Ah, right. What do you think about this version which adds a `sym->sym.st_shndx == SHN_XINDEX` check to that?
> 
> [jpoimboe@0f7b69a](https://github.com/jpoimboe/kpatch/commit/0f7b69ae7b9dfaab358324b53100322d1faf34bc)

This should work. I've also changed the limits check to check against `sym->sym.st_shndx` instead of `shndx` because the latter might change by this point. Thank you for your suggestions, it does look cleaner now.

joe-lawrence (approved):
update looks good

Issue #1253, opened 2022-03-23
Handling changed line number __LINE__ and maybe some other issues


liu-song-6:
Hi, 

We hit the following error while building some patch
```
ld: final link failed: Bad value
```
After digging into it, we found the issue was caused by undefined local section (?)
```
readelf -s output/fs/btrfs/ctree.o   | grep "SECTION LOCAL" | grep UND
    60: 0000000000000000     0 SECTION LOCAL  DEFAULT  UND
```

The patch is actually very simple:
```
diff --git a/fs/btrfs/ctree.c b/fs/btrfs/ctree.c
index 5104ba0504a5..8dd02c755206 100644
--- a/fs/btrfs/ctree.c
+++ b/fs/btrfs/ctree.c
@@ -1458,6 +1459,7 @@ read_block_for_search(struct btrfs_root *root, struct btrfs_path *p,
                        *eb_ret = tmp;
                        return 0;
                }
+               btrfs_warn_rl(fs_info, "found not uptodate eb; btrfs_read_buffer failed; return EIO");
                free_extent_buffer(tmp);
                btrfs_release_path(p);
                return -EIO;
@@ -1485,8 +1487,10 @@ read_block_for_search(struct btrfs_root *root, struct btrfs_path *p,
                 * and give up so that our caller doesn't loop forever
                 * on our EAGAINs.
                 */
-               if (!extent_buffer_uptodate(tmp))
+               if (!extent_buffer_uptodate(tmp)) {
+                       btrfs_warn_rl(fs_info, "read_tree_block returned not-uptodate eb; set ret to EIO");
                        ret = -EIO;
+               }
                free_extent_buffer(tmp);
        } else {
                ret = PTR_ERR(tmp);
```

After digging into this, we found that `create-diff-object` detected changed functions that we didn't change in the patch, which is caused by `__LINE__` in `btrfs_abort_transaction`. After matching line numbers, by removing empty lines, the issue was gone. 

I have two questions from this case:
1. Is there a better way to handle `__LINE__` and changed line numbers?
2. What's the connection between changed `__LINE__` and the ld error above (`ld: final link failed: Bad value`)? I guess it is something with `create-diff-object`, but haven't figure out why.

jpoimboe:
> 1. Is there a better way to handle `__LINE__` and changed line numbers?

[...]

PR #1254, opened 2022-03-29
Multiple `__LINE__` macro detection fixes and improvements

jpoimboe:
Fix a couple of issues reported in #1253.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -577,14 +577,79 @@ out:
 		log_debug("section %s has changed\n", sec->name);
 }
 
+static unsigned int insn_length(struct kpatch_elf *kelf, void *addr)
+{
+	struct insn insn;
+
+	switch(kelf->arch) {
+
+	case X86_64:
+		insn_init(&insn, addr, 1);
+		insn_get_length(&insn);
+		return insn.length;
+
+	case PPC64:
+		return 4;
+
+	default:
+		ERROR("unsupported arch");
+	}
+
+	return 0;
+}
+
 /*
- * Determine if a section has changed only due to a WARN* or might_sleep
- * macro call's embedding of the line number into an instruction operand.
+ * This function is not comprehensive, i.e. it doesn't detect immediate loads
+ * to *all* registers.  It only detects those which have been found in the wild
+ * to be involved in the load of a __LINE__ immediate.  If we miss some, that's
+ * ok, we'll find them later when someone notices a function falsely being
+ * reported as changed ;-)
+ *
+ * (Alternatively we may want to consider just checking all registers which can
+ * be used as function arguments.)
+ */
+static bool insn_is_load_immediate(struct kpatch_elf *kelf, void *addr)
+{
+	struct insn insn;
+	unsigned int instr;
+
+	switch(kelf->arch) {
+
+	case X86_64:
+		insn_init(&insn, addr, 1);
+		insn_get_opcode(&insn);
+
+		if (insn.opcode.value == 0xba)	/* mov $imm, %edx */
+			return true;
+		if (insn.opcode.value == 0xbe)	/* mov $imm, %esi */
+			return true;
+
+		break;
+
+	case PPC64:
+		instr = *(unsigned int *)addr >> 16;
+
+		if (instr == 0x38a0)		/* li r5, imm */
+			return true;
+
+		break;
+
+	default:
+		ERROR("unsupported arch");
+	}
+
+	return false;
+}
+
+/*
+ * Determine if a section has changed only due to a __LINE__ number change,
+ * e.g. a WARN() or might_sleep() macro's embedding of the line number into an
+ * instruction operand.
  *
  * Warning: Hackery lies herein.  It's hopefully justified by the fact that
  * this issue is very common.
  *
- * Example WARN*:
+ * Example WARN():
  *
  *  938:   be 70 00 00 00          mov    $0x70,%esi
  *  93d:   48 c7 c7 00 00 00 00    mov    $0x0,%rdi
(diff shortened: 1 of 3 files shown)


jpoimboe:
Going to rework the "line macro change only" patch a bit.  In the process I found some bugs in the ppc-specific version of the function.  Also there's a lot of copy/paste between the x86 and ppc versions which I'll just converge into a single function.

jpoimboe:
> 1. I'm pretty sure there is an endian bug in the ppc64le version (it compares whole word instruction encoding values, which wouldn't work in cross-arch cases)

Is this only seen when running ppc64le unit tests on an s390 host?  I hadn't really considered that scenario.

> 2. Take a look at Erastus's WIP [s390x version](https://github.com/dynup/kpatch/pull/1243/commits/da775bfe1906916e5ecfab30e3a37244f8079118), it should be very easy to fold into a common function along with the other arches.

What exactly should I look at?  It looks really similar to the ppc and x86 versions..  What I was thinking was the main differences between the arch implementations could be abstracted with a few functions like 'insn_length()' and `is_potential_line_immedate_load_insn()` or such.

jpoimboe:
v2:
- unify kpatch_line_macro_change_only() for all arches
- check for both `__func__` and `__btrfs_abort_transaction`

joe-lawrence:
> > 1. I'm pretty sure there is an endian bug in the ppc64le version (it compares whole word instruction encoding values, which wouldn't work in cross-arch cases)
> 
> Is this only seen when running ppc64le unit tests on an s390 host? I hadn't really considered that scenario.
> 

Yes, I think that would do it.  An odd scenario for sure, but I think checking per-byte would be safer and avoid anyone else from copy/pasting that pattern in the future.  (I think Erastus and I hit this on x86 when we were testing s390 object files and modeled after the ppc64le implementation of checking long int values.)

> > 2. Take a look at Erastus's WIP [s390x version](https://github.com/dynup/kpatch/pull/1243/commits/da775bfe1906916e5ecfab30e3a37244f8079118), it should be very easy to fold into a common function along with the other arches.
> 
> What exactly should I look at? It looks really similar to the ppc and x86 versions.. What I was thinking was the main differences between the arch implementations could be abstracted with a few functions like `insn_length()` and `is_potential_line_immediate_load_insn()` or such.

Sorry to be vague, but yeah that's pretty much it.  If you were creating a common function, I just wanted to give you a heads up that the pattern pretty much holds for s390, too.

kpatch-build/create-diff-object.c
```diff
@@ -623,153 +680,86 @@ static bool kpatch_line_macro_change_only_x86_64(struct section *sec)
...
-		/* verify it's a load immediate to r5 */
-		if (!(instr1 == 0x38a0 && instr2 == 0x38a0))
-			return false;
+			if (r->offset < offset + insn1_len)
+				continue;
 
-		found = 0;
-		list_for_each_entry(rela, &sec->rela->relas, list) {
-			if (rela->offset < offset + PPC_INSTR_LEN)
+			rela = toc_rela(r);
```
joe-lawrence, line 725: Ok, **toc_rela()** should just return the **rela** back for non-ppc64le.

kpatch-build/create-diff-object.c
```diff
@@ -577,14 +577,76 @@ static void kpatch_compare_correlated_section(struct section *sec)
...
+	struct insn insn;
+	unsigned int instr;
+
+	switch(kelf->arch) {
+
+	case X86_64:
+		insn_init(&insn, addr, 1);
+		insn_get_opcode(&insn);
+
+		if (insn.opcode.value == 0xba)	/* mov $imm, %edx */
```
joe-lawrence, line 619: Looks like there is a union aliasing **value** and a character array of **bytes**:
``` C
struct insn_field {
        union {
                insn_value_t value;
                insn_byte_t bytes[4];
        };
```
I think checking specific **bytes[x]** would be endian safer.

kpatch-build/create-diff-object.c
jpoimboe: Right, and also for the ppc64le case where the rela isn't a toc-style nested rela.

kpatch-build/create-diff-object.c
```diff
@@ -577,14 +577,76 @@ static void kpatch_compare_correlated_section(struct section *sec)
...
+			return true;
+		if (insn.opcode.value == 0xbe)	/* mov $imm, %esi */
+			return true;
+
+		return false;
+
+	case PPC64:
+		instr = *(unsigned int *)addr >> 16;
+
+		if (instr == 0x38a0)		/* li r5, imm */
```
joe-lawrence, line 629: Likewise here, **instr** is a native numerical (endian) type.  Something like this should work regardless of host/file endianness:
``` C
char *instr = addr + PPC_RA_OFFSET / 8;
if (instr[0] == 0xa0 || instr[1] == 0x38)
```

joe-lawrence (commented):
This is a nice cleanup (and fixes an outstanding issue).

I left two comments about endianness.  If you want to defer that work until we do a fuller audit of the code when s390x support goes live, we can punt it down the road.  I happened to have a half-baked local branch that started addressing a few spots like this.

jpoimboe:
> This is a nice cleanup (and fixes an outstanding issue).

Oh it does fix an issue, doesn't it ;-)  I assume you mean the toc_rela() change, I should move that to its own patch.  I'll also try to fix the endian issues in another patch.

jpoimboe:
v3:
- Split toc_rela() change into a separate patch

(will also do a v4 with endian fixes)

jpoimboe:
v4:
- Fix endian issues
- Add theoretical s390 support for `__LINE__` macro detection.  This is a modified version of the patch in #1243.  I went ahead and included this patch here because it helped me test the endian fixes, and it works fine as a standalone patch which can be easily combined with #1203 for further s390 testing going forward.

kpatch-build/create-diff-object.c
```diff
@@ -742,7 +743,8 @@ static bool kpatch_line_macro_change_only(struct kpatch_elf *kelf,
 			    !strcmp(rela->sym->name, "___might_sleep") ||
 			    !strcmp(rela->sym->name, "__might_fault") ||
 			    !strcmp(rela->sym->name, "printk") ||
-			    !strcmp(rela->sym->name, "lockdep_rcu_suspicious")) {
+			    !strcmp(rela->sym->name, "lockdep_rcu_suspicious") ||
+			    !strcmp(rela->sym->name, "__btrfs_abort_transaction")) {
```
joe-lawrence, line 769: **fs/btrfs/ctree.h** also contains **btrfs_handle_fs_error** and **btrfs_panic** macros that similarly pass `__LINE__` to a function as the third parameter.  The latter two count up to the same number of affected .c files as the first, would they be worth handling, too?
```
$ grep -R -o -e btrfs_abort_transaction fs/btrfs | sort -u | wc -l
21
$ grep -R -o -e btrfs_handle_fs_error fs/btrfs | sort -u | wc -l
15
$ grep -R -o -e btrfs_panic fs/btrfs | sort -u | wc -l
5
```

jpoimboe: Good call, I'll add `__btrfs_handle_fs_error` and `__btrfs_panic`.

jpoimboe:
v5:
- added `__btrfs_handle_fs_error` and `__btrfs_panic`

jpoimboe:
v6:
- check arg2/arg3 on s390/ppc64le for parity with x86
- add "_printk" check (which was renamed from "printk" on newer kernels)

jpoimboe:
Bumped unit test submodule reference

PR #1260, opened 2022-04-12
Bump to version v0.9.6

joe-lawrence:
Updates of interest:

v0.9.6:
- Allow OOT modules to be built with non-distro kernels
- Add cross-arch unit testing support
- Support ELF extended symbol section indexes
- Allow setting kernel version if --sourcedir and --vmlinux are used
- Cleanup and enhance __LINE__ macro detection for all arches
- Fix segfault on .LCx string literal symbols
- Include __dyndbg section when referenced by jump table
- Honor user provided KBUILD_EXTRA_SYMBOLS
- Support .retpoline_sites section
- Add native compiler selection via CROSS_COMPILE

### DIFF
--- a/contrib/kpatch.spec
+++ b/contrib/kpatch.spec
@@ -6,7 +6,7 @@
 
 Name: kpatch
 Summary: Dynamic kernel patching
-Version: 0.9.5
+Version: 0.9.6
 License: GPLv2
 Group: System Environment/Kernel
 URL: http://github.com/dynup/kpatch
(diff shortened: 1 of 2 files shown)


jpoimboe:
Also:
- Support .retpoline_sites section
- Add cross-compilation support? (not sure if this fully works yet with the kpatch-build script since ARCH is based on host arch?)

joe-lawrence:
> Support .retpoline_sites section

Hmm, I thought I copy/pasted that one, but somehow it disappeared from my editor.  Added.

> Add cross-compilation support? (not sure if this fully works yet with the kpatch-build script since ARCH is based on host arch?)

As soon as I see "cross" I think "cross-architecture", and on that front, it's only implemented as far as create-diff-object.  I think that would be most visible via the unit-tests.

We did merge #1248 and those commits call it "cross-compilation" and we now allow the user to pick their compiler via `CROSS_COMPILE`.  The (implied) limitation is that it must be for the local arch.  Maybe something like "Add native compiler selection via CROSS_COMPILE" ?

jpoimboe (approved):

yhcote (approved):
Looks good to me.

PR #1261, opened 2022-04-22
readme: bring README.md into the 21st century

jpoimboe:
Update a lot of the old cruft and move the boring installation stuff to
a separate file.


joe-lawrence (approved):
Changes look good, thanks for updating... so many random kpatch.ko comments in there :)

PR #1174, opened 2021-04-20
Additional rhel 8.4 support

joe-lawrence:
- `pahole` and large ELF sections already merged (4cc2f062794c184c31bc90f5688a8b7479f3f90e and e25450f5a80349b88607634a80a3fa4e02823caa) 
- Drop `klp.arch relocations` support for rhel-8.4
- Add rebased and now enabled integration tests for rhel-8.4 `kernel-4.18.0-304.el8`

PR #1163, opened 2021-03-17
kpatch-build: drop klp.arch support in RHEL-8.4

joe-lawrence:
RHEL-8.4 backported the "livepatch,module: Remove .klp.arch and
module_disable_ro()" patchset [1], so update kpatch-build accordingly.

[1] https://lore.kernel.org/live-patching/nycvar.YFH.7.76.2005080016330.25812@cbobk.fhfr.pm/T/#t

Signed-off-by: Joe Lawrence <joe.lawrence@redhat.com>

PR #1264, opened 2022-04-28
kpatch-build: remove duplicated use_klp_arch

liu-song-6:
The two versions are not the same (4.18.0-240.el8 vs. 4.18.0-284.el8).
But I am not quite sure which one is accurate. Remove the first one as the
second one is being used before this change.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -206,21 +206,6 @@ rhel_kernel_version_gte() {
 }
 
 # klp.arch relocations were supported prior to v5.8
-# and prior to 4.18.0-240.el8
-use_klp_arch()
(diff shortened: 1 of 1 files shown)


jpoimboe (approved):
Wow, good catch.  It looks like #1163 and #1174 had two different versions of 
the same patch.  When they were merged, git silently did the wrong thing.  Then 
I guess shellcheck doesn't detect duplicate functions.

Luckily the 2nd function was correct, which I assume takes precedence since it overwrites the first.

joe-lawrence:
Yeah, oops.  kernel-4.18.0-284.el8 was when we backported the klp.arch removal code, so that is correct.

PR #1270, opened 2022-05-18
kpatch: also unload a loaded, but no longer enabled kernel live patch

rikvanriel:
Currently "kpatch force unload <klp_some_version>" will fail to unload
an old kernel live patch module that is no longer seeing any use, when
the "enabled" file for that module is no longer present.

Check for the presence of such modules by checking /sys/module, and
return success from disable_patch() for a module that is already
disabled.

This allows "kpatch force unload <klp_some_version>" to succeed for
already disabled modules.

Signed-off-by: Rik van Riel <riel@surriel.com>

### DIFF
--- a/kpatch/kpatch
+++ b/kpatch/kpatch
@@ -378,8 +378,12 @@ disable_patch () {
 
 	local enabled="$SYSFS/$modname/enabled"
 	if ! [[ -e "$enabled" ]]; then
-	    warn "patch module $modname is not loaded"
-	    return 1
+		if [[ -d "/sys/module/$modname" ]] ; then
+			# Module is loaded, but already disabled
+			return 0
+		fi
+		warn "patch module $modname is not loaded"
+		return 1
 	fi
 
 	if [[ "$(cat "$enabled")" -eq 1 ]]; then


joe-lawrence:
Hi @rikvanriel , the change looks good, thanks!

joe-lawrence (approved):

kpatch/kpatch
```diff
@@ -378,6 +378,10 @@ disable_patch () {
 
 	local enabled="$SYSFS/$modname/enabled"
 	if ! [[ -e "$enabled" ]]; then
+            if [[ -d "/sys/module/$modname" ]] ; then
```
jpoimboe, line 381: Tabs, not spaces please.  Otherwise looks good...

rikvanriel: OK, let me push a new version with tabs. Sorry about that.

jpoimboe: No worries... you can force push to the same branch.

rikvanriel: Done.

jpoimboe (approved):

kpatch/kpatch
joe-lawrence: Oh shellcheck... it can figure out the strangest single/double quotation compatibility bug, but not whitespace :)

joe-lawrence:
Btw, "force unload" is a RHEL-specific behavior.  Would you expect this fix to land on any particular elX release?  Otherwise, we generally rebase the utility on the next minor rhel.

rikvanriel:
Next minor RHEL should be fine. I already have another workaround for this issue too, just thought it would be worth fixing here, too :)

PR #1271, opened 2022-05-23
kpatch build dependency cleanup

joe-lawrence:
`yum-builddep kernel` has been bugging me for a _long_ time.  Sometimes it installs all kernel build dependencies, other times it leaves out obvious packages like bison, flex, bc, etc.  The RHEL kernel.spec file is not the most straightforward specfile and I think effort has gone into paring down a minimal dependency set (think about skipping debug builds, kABI, all the other things that the specfile can do).  Whether specfile complexity is the root cause or not, the result is that I have to often install additional packages after running test/integration/lib.sh's **kpatch_dependencies()** function.

This patchset modifies **kpatch_dependencies()**'s `yum-builddep` invocation to add `--skip-unavailable`, which in my limited tests, seems to render a more complete dependency install.  My theory is that without the flag, yum bails out early once it encounters a package it can't install (perhaps a dep for another architecture).  With the flag, many more packages do get installed, at least the ones that kpatch-build's kernel build actually requires.

A few cleanups were added along the way, including a few top-level Makefile flourishes that I thought would be helpful.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ UNITTEST_DIR = test/unit
 INTEGRATION_DIR = test/integration
 CLEAN_DIRS  += clean-$(UNITTEST_DIR)
 
-.PHONY: all install uninstall clean check unit
+.PHONY: all dependencies install uninstall clean check unit
 .PHONY: $(SUBDIRS) $(BUILD_DIRS) $(INSTALL_DIRS) $(CLEAN_DIRS)
 .PHONY: integration integration-slow integration-quick
 .PHONY: vagrant-integration-slow vagrant-integration-quick vagrant-integration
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -71,3 +72,24 @@ check:
...
+	@echo "Test targets:"
+	@echo "    make check                        run static code analyzers"
+	@echo "    make integration                  build and run integration tests [2]"
+	@echo "    make integration-slow             build and run integration tests [2]"
+	@echo "    make integration-quick            build and run integration tests [2]"
+	@echo "    make unit                         run unit tests"
+	@echo
+	@echo "[1] requires admin privileges"
+	@echo "[2] installs test kpatch kernel modules, run at your own risk"
+	@echo
```
joe-lawrence, line 95: (I skipped the vagrant targets since we don't document them on any markdown page.)

sm00th (approved):
Neat trick, indeed builddep has been a pain-point for a while. One might have thought `--skip-unavailable` will yield you less packages, not more.

Didn't try it but the changes look good to me.

jpoimboe (approved):

PR #1274, opened 2022-06-16
rewrite the is_bundleable function

anatasluo:
The implementation of "is_bundleable" function has too many cases.
I summarize these cases and provide a new implementation.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -80,52 +80,66 @@ struct special_section {
 	int (*group_size)(struct kpatch_elf *kelf, int offset);
 };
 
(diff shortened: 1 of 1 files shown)


jpoimboe:
Thank you for the patches.

I believe this introduces a bug: For `.text.unlikely.foo`, it will only strip the `.text.` prefix, leaving `unlikely.foo`, which will no longer match `foo`, causing is_bundleable() to return false.

Also, while it does reduce duplication, I still find the original code simpler and easier to understand.

I would rather leave it the way it is.

anatasluo:
@jpoimboe thanks for your review.

Issue #1278, opened 2022-07-12
rhel-7.9/syscall.patch fails to build on 3.10.0-1160.71.1.el7.ppc64le


joe-lawrence:
[build.log.tar.gz](https://github.com/dynup/kpatch/files/9093028/build.log.tar.gz)

```In file included from kernel/sys.c:1395:0:
/root/kpatch/kmod/patch/kpatch-syscall.h:99:2: error: data definition has no type or storage class [-Werror]
  __diag_push();       \
  ^
/root/kpatch/kmod/patch/kpatch-syscall.h:37:2: note: in expansion of macro ‘__KPATCH_SYSCALL_DEFINEx’
  __KPATCH_SYSCALL_DEFINEx(x, sname, __VA_ARGS__)
  ^
/root/kpatch/kmod/patch/kpatch-syscall.h:28:43: note: in expansion of macro ‘KPATCH_SYSCALL_DEFINEx’
 #define KPATCH_SYSCALL_DEFINE1(name, ...) KPATCH_SYSCALL_DEFINEx(1, _##name, __VA_ARGS__)
                                           ^
kernel/sys.c:1396:1: note: in expansion of macro ‘KPATCH_SYSCALL_DEFINE1’
 KPATCH_SYSCALL_DEFINE1(newuname, struct new_utsname __user *, name)
 ^
/root/kpatch/kmod/patch/kpatch-syscall.h:99:2: error: type defaults to ‘int’ in declaration of ‘__diag_push’ [-Werror=implicit-int]
  __diag_push();       \
  ^
/root/kpatch/kmod/patch/kpatch-syscall.h:37:2: note: in expansion of macro ‘__KPATCH_SYSCALL_DEFINEx’
  __KPATCH_SYSCALL_DEFINEx(x, sname, __VA_ARGS__)
  ^
/root/kpatch/kmod/patch/kpatch-syscall.h:28:43: note: in expansion of macro ‘KPATCH_SYSCALL_DEFINEx’
 #define KPATCH_SYSCALL_DEFINE1(name, ...) KPATCH_SYSCALL_DEFINEx(1, _##name, __VA_ARGS__)
                                           ^
kernel/sys.c:1396:1: note: in expansion of macro ‘KPATCH_SYSCALL_DEFINE1’
 KPATCH_SYSCALL_DEFINE1(newuname, struct new_utsname __user *, name)
 ^
[...]

PR #1279, opened 2022-07-13
macros: tweak syscall patching macros

joe-lawrence:
Kernel version specific __KPATCH_SYSCALL_DEFINEx macros were added to
kpatch-syscall.h for x86, but only single versions for other arches.
This works out for s390x, but not ppc64le for which kpatch-build support
goes back to kernel versions that require slightly different macros.

Reorder the __KPATCH_SYSCALL_DEFINEx macros to define the arch-specific
ones first (arch/.../include/asm/syscall_wrapper.h) and then fall back
to using generic ones (include/linux/syscalls.h versions).

Fixes: #1278
Fixes: 9c0b678621b9 ("macros: add syscall patching macros")

### DIFF
--- a/kmod/patch/kpatch-syscall.h
+++ b/kmod/patch/kpatch-syscall.h
@@ -38,9 +38,11 @@
 
 #ifdef CONFIG_X86_64
 
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 7, 0)
+/* x86/include/asm/syscall_wrapper.h versions */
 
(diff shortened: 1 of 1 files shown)


joe-lawrence:
Internal integration tests passed for most combinations (a few timed out due to unrelated infra issues), so marking this one ready.

yhcote (approved):

jpoimboe (approved):

Issue #1280, opened 2022-07-15
kpatch-build: .rela.return_sites not selected for inclusion on 5.18.12-rc1+ kernels


dobsonj:
Using fedora 36 / kernel-5.18.11-200.fc36.x86_64 with kpatch built from the following commit:

```
commit dc7e5cb03de7ba1d7781ab826bdb7107e3557f56 (HEAD -> master, origin/master, origin/HEAD)
Merge: d1423a3 48d997f
Author: Joe Lawrence <joe.lawrence@redhat.com>
Date:   Wed Jun 29 09:55:53 2022 -0400

    Merge pull request #1203 from sumanthkorikkar/s390x-kpatch-support
    
    S390x kpatch support
```

I tried a simple example similar to the [quick start guide](https://github.com/dynup/kpatch#quick-start) but it fails with `.rela.return_sites not selected for inclusion`.

```
[vagrant@fedora36 linux-5.18.11-200.fc36.x86_64]$ cat meminfo-string.diff
diff --git a/fs/proc/meminfo.c b/fs/proc/meminfo.c
index 6fa761c9..c1a3ab6c 100644
--- a/fs/proc/meminfo.c
+++ b/fs/proc/meminfo.c
@@ -55,6 +55,7 @@ static int meminfo_proc_show(struct seq_file *m, void *v)
        sreclaimable = global_node_page_state_pages(NR_SLAB_RECLAIMABLE_B);
        sunreclaim = global_node_page_state_pages(NR_SLAB_UNRECLAIMABLE_B);

+       available = 1337;
        show_val_kb(m, "MemTotal:       ", i.totalram);
        show_val_kb(m, "MemFree:        ", i.freeram);
        show_val_kb(m, "MemAvailable:   ", available);
[vagrant@fedora36 linux-5.18.11-200.fc36.x86_64]$ SRCRPM=~/kernel-5.18.11-200.fc36.src.rpm kpatch-build meminfo-string.diff
Using cache at /home/vagrant/.kpatch/src
Testing patch file(s)
Reading special section data
Building original source
Building patched source
Extracting new and modified ELF sections
ERROR: meminfo.o: 1 unsupported section change(s)
meminfo.o: changed function: meminfo_proc_show
meminfo.o: changed section .rela.return_sites not selected for inclusion
/usr/local/libexec/kpatch/create-diff-object: unreconcilable difference
ERROR: 1 error(s) encountered. Check /home/vagrant/.kpatch/build.log for more details.
```

[...]

PR #1281, opened 2022-07-16
create-diff-object: add support for .return_sites section (x86)

dobsonj:
Fixes https://github.com/dynup/kpatch/issues/1280

Simple test with this change applied:
```
[vagrant@fedora36 linux-5.18.11-200.local.fc36.x86_64]$ cat meminfo-string.diff 
index 6fa761c9..c1a3ab6c 100644
--- a/fs/proc/meminfo.c
+++ b/fs/proc/meminfo.c
@@ -55,6 +55,7 @@ static int meminfo_proc_show(struct seq_file *m, void *v)
        sreclaimable = global_node_page_state_pages(NR_SLAB_RECLAIMABLE_B);
        sunreclaim = global_node_page_state_pages(NR_SLAB_UNRECLAIMABLE_B);
 
+       seq_printf(m, "I can haz kpatch?\n");
        show_val_kb(m, "MemTotal:       ", i.totalram);
        show_val_kb(m, "MemFree:        ", i.freeram);
        show_val_kb(m, "MemAvailable:   ", available);

[vagrant@fedora36 linux-5.18.11-200.local.fc36.x86_64]$ kpatch-build -s . meminfo-string.diff
Using source directory at /home/vagrant/kernel/kernel-5.18.11/linux-5.18.11-200.local.fc36.x86_64
Testing patch file(s)
Reading special section data
Building original source
Building patched source
Patched objects: vmlinux
Building patch module: livepatch-meminfo-string.ko
SUCCESS

[vagrant@fedora36 linux-5.18.11-200.local.fc36.x86_64]$ sudo kpatch load livepatch-meminfo-string.ko
loading patch module: livepatch-meminfo-string.ko

[vagrant@fedora36 linux-5.18.11-200.local.fc36.x86_64]$ head -1 /proc/meminfo 
I can haz kpatch?
```

/cc @joe-lawrence @jpoimboe

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2090,6 +2090,11 @@ static int retpoline_sites_group_size(struct kpatch_elf *kelf, int offset)
 	return 4;
 }
 
+static int return_sites_group_size(struct kpatch_elf *kelf, int offset)
+{
+	return 4;
+}
+
 static int fixup_entry_group_size(struct kpatch_elf *kelf, int offset)
 {
 	static int size = 0;
@@ -2226,6 +2231,11 @@ static struct special_section special_sections[] = {
 		.name		= ".retpoline_sites",
 		.arch		= X86_64,
 		.group_size	= retpoline_sites_group_size,
+	},
+	{
+		.name		= ".return_sites",
+		.arch		= X86_64,
+		.group_size	= return_sites_group_size,
 	},
 	{
 		.name		= "__ftr_fixup",


jpoimboe (approved):

dobsonj:
> I don't see how a return site entry could change without a corresponding code change that **create-diff-object** is already picking up anyway. Maybe we should assert this somehow?

Sorry, I'm not sure if I follow: are you just suggesting an assertion that .return_sites did not change?
Or a more nuanced assertion that .return_sites did not change without a corresponding code change? (I don't see an obvious way to codify this, but open to suggestion)

joe-lawrence:
> > I don't see how a return site entry could change without a corresponding code change that **create-diff-object** is already picking up anyway. Maybe we should assert this somehow?
> 
> Sorry, I'm not sure if I follow: are you just suggesting an assertion that .return_sites did not change? Or a more nuanced assertion that .return_sites did not change without a corresponding code change? (I don't see an obvious way to codify this, but open to suggestion)

It was more just thinking out loud, but the nuanced assertion was what I had been pondering.  AFAIK, we don't do the same for any of the sections that we ignore, so I don't think it's necessary here.  Thanks for finding this and posting a MR to fix!

joe-lawrence (approved):

Issue #1282, opened 2022-07-27
kpatch-cc build error when source tree prefix matches substring in changed object file


jerdfelt:
Example setup:
- Patch makes changes to `arch/x86/kernel/cpu/bugs.c`
- Source tree is at `/kernel`

When using `kpatch-build` in this or similar setup, it can result in a build error like this:

```
Using source directory at /kernel
Testing patch file(s)
Reading special section data
Building original source
Building patched source
cp: cannot stat '/kernel/arch/x86cpu/bugs.o': No such file or directory
ERROR: kpatch build failed. Check /root/.kpatch/build.log for more details.
```

This is because a change made in 51a8fad34ff3 to support out-of-tree modules. The change made `kpatch-cc` attempt to strip the source tree prefix from the object file path. When not using out-of-tree modules, this path is already relative and can strip a substring instead, leading to errors.

PR #1283, opened 2022-07-27
kpatch-cc: fix stripping of source tree prefix

jerdfelt:
To support building out-of-tree kernel modules, the source tree prefix
is attempted to be stripped from change object file paths to make them
relative. However, if the path is already relative, the change can
strip a substring instead, resulting in build errors.

Ensure just the prefix is stripped instead of any substring.

Fixes: #1282
Fixes: 51a8fad34ff3 ("Add support for building out-of-tree modules")


joe-lawrence (approved):

yhcote (approved):

joe-lawrence:
Thanks @jerdfelt

PR #1287, opened 2022-08-05
kpatch-build: use err.h instead of error.h for musl support

androw:
Musl libc doesn't support error.h
Moving to err.h would allow using it on Musl system


jpoimboe (approved):

joe-lawrence (approved):
Thanks, @androw

PR #1289, opened 2022-08-05
kpatch-build: support Linux 5.19 .cmd file format

### LINKED ISSUE #1277: kpatch-build error with 5.19 kernel
I hit kpatch-build error like
```
ERROR: invalid ancestor fs/proc/array.o for fs/proc/array.o.
```
with 5.19 kernel. Some debugging shows that this is caused by changes in .cmd files, i.e., array.o (in 5.19) instead of fs/proc/array.o (in 5.18). However, I haven't figured out what change in the kernel caused this change in .cmd files. 

Something like [this commit](https://github.com/liu-song-6/kpatch/commit/b52499b7b0932971d212e9d40a696128108acdf6), fixes the issue. But I am not quite sure that's the right fix. Actually, I wonder whether we should fix it in the kernel.

jpoimboe:
Rewrite kobj_find() to deal with Linux 5.19, where the .cmd files use
object file paths relative to the .cmd file rather than relative to the
root of the kernel tree.

Fixes #1277.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -434,82 +434,215 @@ find_special_section_data() {
 	return
 }
 
-filter_parent_obj()
-{
-  local dir="${1}"
-  local file="${2}"
+# path of file, relative to dir
+# adapted from https://stackoverflow.com/a/24848739
+relpath() {
+	local file="$1"
+	local dir="$2"
 
-  grep -v "\.mod\.cmd$" | grep -Fv "${dir}/.${file}.cmd"
+	local filedir
+	local common
+	local result
+
+	filedir="$(dirname "$(readlink -f "$file")")"
+	common="$(readlink -f "$dir")"
+
+	if [[ "$filedir" = "$common" ]]; then
+		basename "$file"
+		return
+	fi
+
+	while [[ "${filedir#$common/}" = "$filedir" ]]; do
+		common="$(dirname "$common")"
+		result="../$result"
+	done
+
+	result="${result}${filedir#$common/}"
+	echo "${result}/$(basename "$file")"
+}
+
+cmd_file_to_o_file() {
+	local parent="$1"
+
+	# convert cmd file name to corresponding .o
+	parent_dir="$(dirname "$parent")"
+	parent_dir="${parent_dir#./}"
+	parent="$(basename "$parent")"
+	parent="${parent#.}"
+	parent="${parent%.cmd}"
+	parent="$parent_dir/$parent"
+
+	[[ -f $parent ]] || die "can't find $parent associated with $1"
+
+	echo "$parent"
+}
+
+get_parent_from_parents() {
+	local parents=("$@")
+
+	[[ ${#parents[@]} -eq 0 ]] && PARENT="" && return
+	[[ ${#parents[@]} -eq 1 ]] && PARENT="${parents[0]}" && return
+
+	# multiple parents:
+	local parent
+	local mod_name="${parents[0]%.*}"
+	local mod_file
+	for parent in "${parents[@]}"; do
+		# for modules, there can be multiple matches.  Some
+		# combination of foo.o, foo.mod, and foo.ko, depending
+		# on kernel version and whether the module is single or
+		# multi-object.  Make sure a .mod and/or .ko exists, and no
+		# more than one .mod/.ko exists.
+
+		[[ $parent = *.o ]] && continue
+
+		if [[ ${parent%.*} != "$mod_name" ]]; then
+			mod_file=""
+			break
+		fi
+
+		if [[ $parent = *.mod || $parent = *.ko ]]; then
+			mod_file=$parent
+			continue
+		fi
+
+		mod_file=""
+		break
+	done
+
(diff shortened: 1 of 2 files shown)


kpatch-build/kpatch-build
```diff
@@ -434,76 +434,121 @@ find_special_section_data() {
...
-	else
-		parent="$(grep -lw "$grepname" "$dir"/.*.cmd | filter_parent_obj "${dir}" "${file}" | head -n1)"
-		num="$(grep -lw "$grepname" "$dir"/.*.cmd | filter_parent_obj "${dir}" "${file}" | wc -l)"
+	filedir="$(dirname "$file")"
+	filename="$(basename "$file")"
+
+	# Linux < 5.19 file path is relative to kernel source tree root
+	while IFS='' read -r line; do
+		parents+=("$line")
+	done < <(grep -Flw "$file" "$dir"/.*.cmd | grep -Fv "$filedir/.$filename.cmd")
```
t-msn, line 458: Hello,
Compared to previous filter_parent_obj, should this line be as follows?
```
         done < <(grep -Flw "$file" "$dir"/.*.cmd | grep -v "\.mod\.cmd$" | grep -Fv "$filedir/.$filename.cmd")
```
At least this fixes the "two parent matches" build error of macro-callback integration test on kernel 5.18 with using this new version. (I have not tested fully nor 5.19 kernel yet)

joe-lawrence: Also, "ERROR: two parent matches for fs/nfsd/export.o" from module.patch on rhel-9.

jpoimboe:
> Is it just me or do the new functions take _substantially_ longer than the old ones? For reference, it previously took ~220 minutes to find parents for all objects in a rhel-9 kernel build. I kicked off the same search using the new code last night and it's still running ~12 hrs later (I estimate about 20% complete). For real kpatches, this should be a big problem, but I was hoping to compare before/after for various kernel versions.

Weird, was that with your test script?  When I tested it on an upstream kernel it took like 5 minutes to find all the objects.  I'll try with RHEL9.

joe-lawrence:
> > Is it just me or do the new functions take _substantially_ longer than the old ones? For reference, it previously took ~220 minutes to find parents for all objects in a rhel-9 kernel build. I kicked off the same search using the new code last night and it's still running ~12 hrs later (I estimate about 20% complete). For real kpatches, this should be a big problem, but I was hoping to compare before/after for various kernel versions.
> 
> Weird, was that with your test script? When I tested it on an upstream kernel it took like 5 minutes to find all the objects. I'll try with RHEL9.

Hmm, maybe I had a really slow beaker box??  I did the same `find . -name '*.o' -exec ~/kpatch/find-kobj-test-new {} \;` from the top of the kernel build tree as before.

joe-lawrence:
See results of test: `find . -name '*.o' -exec find-kobj-test-jp {} \;` from the top-level kernel build directory for:

3.10.0-1160.76.1.el7.x86_64
4.18.0-416.el8.x86_64
5.14.0-70.13.1.el9.x86_64

[results.tar.gz](https://github.com/dynup/kpatch/files/9355916/results.tar.gz)

In each case, the new algorithm found a few more ancestors than previously.  :+1:

jpoimboe:
Nice.  I've been testing and tweaking it more.  v2 may look like another rewrite by the time I'm done with it...

jpoimboe:
The only thing I didn't get to test yet was with an OOT module, if somebody gets a chance to test that I didn't break anything there that would be helpful.

/cc @liu-song-6

liu-song-6:
> The only thing I didn't get to test yet was with an OOT module, if somebody gets a chance to test that I didn't break anything there that would be helpful.
> 
> /cc @liu-song-6

Our build system doesn't have OOT support for 5.19 kernel yet. I tried to hack it, but didn't get much luck. 

I will give this a try when the build system is ready.

jpoimboe:
> Our build system doesn't have OOT support for 5.19 kernel yet. I tried to hack it, but didn't get much luck.
> 
> I will give this a try when the build system is ready.

Thanks!  Testing an OOT module with an older kernel would also be helpful.

jpoimboe:
v3:
- add bash relpath() instead of relying on `realpath --relative-to=`
- fix an imprecision in the grep regex in __find_parent_obj_in_dir() which can cause an infinite loop with `drivers/accessibility/speakup/makemapdata`

liu-song-6:
> 
> Thanks! Testing an OOT module with an older kernel would also be helpful.

Yeah, I can build KLP for OOT module with this version (on an older kernel).

jpoimboe:
v4:
- rebased to trigger integration testing (not sure if previous failure was a false positive)

jpoimboe:
No apparent failures this time, @joe-lawrence did you want to give this one any further testing or review before merging?

joe-lawrence:
> No apparent failures this time, @joe-lawrence did you want to give this one any further testing or review before merging?

The code looks reasonable and definitely more readable than before.  Thanks for tackling this annoying problem.

joe-lawrence (approved):

Issue #1296, opened 2022-08-31
What do I need to do to support riscv64?


yinxx:
[  313s] + /usr/bin/make -O -j2 V=1 VERBOSE=1
[  313s] /usr/bin/make -C kpatch-build
[  313s] /usr/bin/make -C kpatch
[  313s] make[1]: Nothing to be done for 'all'.
[  313s] /usr/bin/make -C kmod
[  313s] make[1]: Entering directory '/home/abuild/rpmbuild/BUILD/kpatch-0.9.5/kpatch-build'
[  313s] Makefile:30: *** Unsupported architecture riscv64, check https://github.com/dynup/kpatch/#supported-architectures.  Stop.
[  313s] make[1]: Leaving directory '/home/abuild/rpmbuild/BUILD/kpatch-0.9.5/kpatch-build'
[  313s] make: *** [Makefile:22: build-kpatch-build] Error 2
[  313s] make: *** Waiting for unfinished jobs....
[  313s] make[1]: Nothing to be done for 'all'.
[  313s] error: Bad exit status from /var/tmp/rpm-tmp.OmUGeL (%build)

euspectre:
To support RISC-V in live patching, you need much more that updating kpatch tools.

Many things need to be done at the kernel side, e.g. to make stack traces reliable, etc.

You can take a look at the activities to add livepatch support for ARM:
https://lore.kernel.org/linux-arm-kernel/20220117145608.6781-1-madvenka@linux.microsoft.com/T/
https://lwn.net/Articles/890741/

euspectre:
I guess, reliable kernel stack traces and support for RISC-V in objtool are the areas to start to eventually enable livepatch on this architecture. Likely, doable but not that easy.

yinxx:
could you give me some clue or advice, or web, etc

for me to study 
to add the riscv64 support!
thanks a lot







At 2022-09-01 17:39:16, "Eugene A. Shatokhin" ***@***.***> wrote:

I guess, reliable kernel stack traces and support for RISC-V in objtool are the areas to start to eventually enable livepatch on this architecture. Likely, doable but not that easy.

euspectre:
I'd suggest to learn how they did it for ARM first, see my earlier comment (https://github.com/dynup/kpatch/issues/1296#issuecomment-1234021957).

euspectre:
By the way, I am very interested in having livepatch on RISC-V too. 

[...]

PR #1297, opened 2022-09-01
README: Add arch porting question

joe-lawrence:
This gets asked from time to time and Josh provided a good answer to #1296, so how about we add it to the FAQ.


jpoimboe (commented):
Good idea

euspectre:
Good idea, thanks!

PR #1301, opened 2022-09-15
kpatch-build: rela section could disappear after patched

anatasluo:
Hi, guys. I meet a segment fault when I use the Kpatch.

My patch is

```
From d4238410bab5b907ee514c15ab4d9109d6da2796 Mon Sep 17 00:00:00 2001
From: Luo Longjun <luolongjuna@gmail.com>
Date: Wed, 14 Sep 2022 16:38:28 +0800
Subject: [PATCH] kpatch test

---
 fs/proc/version.c | 6 +-----
 1 file changed, 1 insertion(+), 5 deletions(-)

diff --git a/fs/proc/version.c b/fs/proc/version.c
index b449f186577f..734ddfc7d316 100644
--- a/fs/proc/version.c
+++ b/fs/proc/version.c
@@ -6,12 +6,8 @@
 #include <linux/seq_file.h>
 #include <linux/utsname.h>
 
-static int version_proc_show(struct seq_file *m, void *v)
+notrace noinline static int version_proc_show(struct seq_file *m, void *v)
 {
-	seq_printf(m, linux_proc_banner,
-		utsname()->sysname,
-		utsname()->release,
-		utsname()->version);
 	return 0;
 }
 
-- 
2.37.3
```

My test environment:

```
NAME="Fedora Linux"
VERSION="36 (Server Edition)"
ID=fedora
VERSION_ID=36
VERSION_CODENAME=""
PLATFORM_ID="platform:f36"
PRETTY_NAME="Fedora Linux 36 (Server Edition)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:36"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f36/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=36
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=36
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
VARIANT="Server Edition"
VARIANT_ID=server
```

The output of the Kpatch:

```
[...]

PR #1303, opened 2022-09-23
kpatch-build: strengthen conditions for changed sections

anatasluo:
Based on this [kpatch-build: rela section could disappear after patched](https://github.com/dynup/kpatch/pull/1301).

I found another problem, considering the following example in X86.

original code:
```
__noreturn noinline int kpatch_func(void)
{
	while(1) { };
}
EXPORT_SYMBOL(kpatch_func);
```

patched code:
```
__noreturn notrace noinline int kpatch_func(void)
{
	asm(".byte 0xe8, 0x00, 0x00, 0x00, 0x00");
	while(1) { };
}
EXPORT_SYMBOL(kpatch_func);
```

These two functions are different, but the output of Kpatch is
`
ERROR: no functional changes found.
`

The problem is explained in the log:

------------------------------

If two sections want to be the same, they need to satisfy
two conditions:

1) the result of memcpy is zero, which means they have the same content.

2) they have the same relocation entries.

In one specific situation, two sections have the same content.
But one section has relocation entries while the other one has
no relocation entries. For example, in X86, consider the
following code:

original code
```
__noreturn noinline int kpatch_func(void)
{
	while(1) {};
}
```

patched code
```
__noreturn notrace noinline int kpatch_func(void)
{
	asm(".byte 0xe8, 0x00, 0x00, 0x00, 0x00");
	while(1){};
}
```

Since the original code has a fentry call, these two functions have
the same compile result. But obviously, they are different functions.
Currently, Kpatch would not find their differences since the patched
code has no relocation entries.

For the situation that one section has relocation entries while the
other one doesn't have, it should be set to be changed directly.

------------------------------

I am not sure if it is useful for Kpatch, so I open another PR.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -579,7 +579,9 @@ static void kpatch_compare_correlated_section(struct section *sec)
 	}
 
 	if (sec1->sh.sh_size != sec2->sh.sh_size ||
-	    sec1->data->d_size != sec2->data->d_size) {
+	    sec1->data->d_size != sec2->data->d_size ||
+		(sec1->rela && !sec2->rela) ||
+		(sec2->rela && !sec1->rela)) {
 		sec->status = CHANGED;
 		goto out;
 	}


joe-lawrence:
Hi @anatasluo : curious cases!  Have you seen these occur in a real-life patch?  FWIW, I don't think we've ever toggled trace/notrace in a patch, so maybe these are just a proof of concept.

So the summary is:
- #1301 handles the case where function's .rela section disappears,
- #1303 handles the fallout: if the .rela section (dis)appears, note the difference so .text is considered changed and included in the kpatch

Small nit in the commit msg for this one: s/result of memcpy is zero/result of **memcmp** is zero/

anatasluo:
@joe-lawrence You are right, it is rare to make a patch like that in the kernel. Actually, we are trying to implement a livepatch mechanism in userspace. And we borrow many ideas from Kpatch. In userspace, there is no trace or return thunk mechanism. After theoretical analysis， we found these corner cases. And we use these cases to test our program so that any modification for the program works as expected. I hope these cases will not bother the community.

anatasluo:
@joe-lawrence About this case [kpatch-build: strengthen conditions for changed sections](https://github.com/dynup/kpatch/pull/1303#top), my beginning idea is: relocation entry will fill the content with zeros temporarily. For code like:
```
mov $0, $rax
```
It could be two situations:
1) "$0" is the data "0".
2) "$0" is filled by the relocation entry.

These two situations are different, but they have the same memory content. 
So, to make the comparison more accurate, we need to ensure they both have the same relocation entry.

jpoimboe:
Looks good to me, can you rebase on master?

joe-lawrence:
> Actually, we are trying to implement a livepatch mechanism in userspace. And we borrow many ideas from Kpatch. In userspace, there is no trace or return thunk mechanism. After theoretical analysis， we found these corner cases. And we use these cases to test our program so that any modification for the program works as expected. I hope these cases will not bother the community.

Sounds like a cool project.  Thanks for upstreaming issues you encounter, reports and patches are welcome.

anatasluo:
@jpoimboe already rebase on master

jpoimboe (approved):

yhcote (approved):

PR #1307, opened 2022-10-19
kpatch-build: relax vmlinux "Linux version" grep

joe-lawrence:
When determining the linux version from a vmlinux file, we assume that `strings` output will show "Linux version <version>" at the beginning of the line.

This assumption was observed to not hold true for clang on s390x:

  $ strings ./build/linux-5.18/vmlinux | grep -m1 -e "Linux version"
  mY*Linux version 5.18.0 (root@xyz.com) (clang version 14.0.5 (Fedora 14.0.5-1.fc36), GNU ld version 2.37-36.fc36) #7 SMP Wed Oct 19 12:47:00 EDT 2022

Relax the regular expression to look for "Linux version <word>", but not anchored at the beginning of the line.

Reported-by: Yulia Kopkova <ykopkova@redhat.com>

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -806,7 +806,7 @@ if [[ -n "$USERSRCDIR" ]]; then
 	[[ ! -e "$VMLINUX" ]] && die "can't find vmlinux"
 
 	# Extract the target kernel version from vmlinux in this case.
-	VMLINUX_VER="$(strings "$VMLINUX" | grep -m 1 -e "^Linux version" | awk '{ print($3); }')"
+	VMLINUX_VER="$(strings "$VMLINUX" | grep -m1 -o -e 'Linux version \S*' | awk '{print $3}')"
 	if [[ -n "$ARCHVERSION" ]]; then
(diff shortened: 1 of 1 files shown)


joe-lawrence:
Hi @sumanthkorikkar : I don't think we ever tested s390x support with the llvm toolset.  Our internal QE tried that case and discovered this bug.  At the same time, even with this change in place, clang-14 doesn't support kpatch-build's s390x ARCH_KCFLAGS (-mno-pic-data-is-text-relative -fno-section-anchors).  Do you know if there are comparable clang flags, or should we have kpatch-build declare no support for this combination (s390x and clang).  Thanks.

sumanthkorikkar:
Hi Joe, I will check with our team. Thanks.

sumanthkorikkar:
Hi Joe, our compiler team will look into enabling -mno-pic-data-is-text-relative for s390 clang. Thanks.

sumanthkorikkar:
Hi Joe, After discussion with our compiler team, mcmodel flag can be used for clang + s390. I compile tested it and kpatch module loads fine with mcmodel=large. Our compiler team will also likely provide clang -mno-pic-data-is-text-relative as an alias to -mcmodel=medium (which is also mcmodel=large on SystemZ)

PR #1311, opened 2022-11-07
examples: add /proc/version kpatch sample

joe-lawrence:
Create a simple kpatch test that should apply across a wide range of kernels.  The version_proc_show() is a good candidate as it's easy to verify and hasn't been touched upstream since 2008.


joe-lawrence:
This patch is pretty similar to the cmdline-string.patch, but touches a function that hasn't changed in forever.  Having a patch like this _really_ useful for quick tests, so I stashed it in the examples/ directory.

jpoimboe (approved):

PR #1312, opened 2022-11-08
three fix of elf file creation

ZhangHongchen1:
Hi Josh,
 We are preparing to support a new architecture:LoongArch . But we find some common problems of this tool,So I think we can submit the common patches firstly,Then the fllowing LoongArch support.
These three patches solve the following problem:
1. e_flags is not written in the end file,we should save it,beause it holds processor-specific flags.
2. clean up the symbol replacement algorithm. When we want add LoongArch support,we find out we should add
  !strcmp(rela->sym->name, ".orc_unwind_ip"). IMO, too many compares in the code is not a good idea, so I clean the code and
  make it easy to read.
3. when one symbol is not included in the output elf file,the related rela of debug info section should also not be included. If we
  don't follow it,then the rela would refer to an not included symbol which contains random data.

Thanks
Hongchen Zhang

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1558,9 +1558,9 @@ static void kpatch_replace_sections_syms(struct kpatch_elf *kelf)
 				start = sym->sym.st_value;
 				end = sym->sym.st_value + sym->sym.st_size;
 
-				if (is_text_section(relasec->base) &&
+				if (rela->type == R_X86_64_32S &&
+				    is_text_section(relasec->base) &&
 				    !is_text_section(sym->sec) &&
-				    rela->type == R_X86_64_32S &&
 				    rela->addend == (long)sym->sec->sh.sh_size &&
 				    end == (long)sym->sec->sh.sh_size) {
 
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2705,8 +2705,9 @@ static void kpatch_include_debug_sections(struct kpatch_elf *kelf)
 		if (!is_rela_section(sec) || !is_debug_section(sec))
 			continue;
 		list_for_each_entry_safe(rela, saferela, &sec->relas, list)
-			if (!rela->sym->sec->include)
+			if (!rela->sym->include || !rela->sym->sec->include) {
 				list_del(&rela->list);
+			}
 	}
 }
 
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1544,6 +1544,10 @@ static void kpatch_replace_sections_syms(struct kpatch_elf *kelf)
 
 			target_off = rela_target_offset(kelf, relasec, rela);
 
+			if (target_off >= (long)rela->sym->sec->sh.sh_size)
+				continue;
+
+			found = false;
 			/*
 			 * Attempt to replace references to unbundled sections
 			 * with their symbols.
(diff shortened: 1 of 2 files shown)


kpatch-build/create-diff-object.c
```diff
@@ -1558,9 +1546,9 @@ static void kpatch_replace_sections_syms(struct kpatch_elf *kelf)
 				start = sym->sym.st_value;
 				end = sym->sym.st_value + sym->sym.st_size;
 
-				if (is_text_section(relasec->base) &&
+				if (rela->type == R_X86_64_32S &&
+				    is_text_section(relasec->base) &&
 				    !is_text_section(sym->sec) &&
-				    rela->type == R_X86_64_32S &&
```
jpoimboe, line 1563: Please make this a separate commit.

kpatch-build/create-diff-object.c
```diff
@@ -1609,18 +1597,11 @@ static void kpatch_replace_sections_syms(struct kpatch_elf *kelf)
...
 				rela->sym = sym;
 				rela->addend -= start;
 				break;
 			}
-
-			if (!found && !is_string_literal_section(rela->sym->sec) &&
-			    strncmp(rela->sym->name, ".rodata", 7)) {
-				ERROR("%s+0x%x: can't find replacement symbol for %s+%ld reference",
-				      relasec->base->name, rela->offset, rela->sym->name, rela->addend);
-			}
```
jpoimboe, line 1623: If this error is getting removed, there needs to be more justification than "we think it is just ok".

jpoimboe: Also, why does .rela.orc_unwind_ip reference .orc_unwind_ip itself?  That sounds like a different format compared to the x86 implementation, which only references text addresses.

kpatch-build/create-diff-object.c
```diff
@@ -2705,8 +2686,9 @@ static void kpatch_include_debug_sections(struct kpatch_elf *kelf)
 		if (!is_rela_section(sec) || !is_debug_section(sec))
 			continue;
 		list_for_each_entry_safe(rela, saferela, &sec->relas, list)
-			if (!rela->sym->sec->include)
+			if (!rela->sym->include || !rela->sym->sec->include) {
```
jpoimboe, line 2689: Hm, why haven't we seen this problem on x86?

kpatch-build/create-diff-object.c
ZhangHongchen1: OK

kpatch-build/create-diff-object.c
ZhangHongchen1: > If this error is getting removed, there needs to be more justification than "we think it is just ok".
OK,Let's leave it not changed.

ZhangHongchen1: > Also, why does .rela.orc_unwind_ip reference .orc_unwind_ip itself? That sounds like a different format compared to the x86 implementation, which only references text addresses.
Yes,the rela format is a little different.

kpatch-build/create-diff-object.c
ZhangHongchen1: Yes, I checked the compiled c file fs/aio.c of x86_64.It reference to __dyndbg  using the __dyndbg + offset,it did not use the 
symbols in the __dyndbg section ,so it has no problem.
But LoongArch generate relas using the symbols in __dyndbg(or __verbose in old version kernel) and triggers the problem.
I think we should avoid that condition even if there is no problem now for x86_64.

ZhangHongchen1:
Hi @jpoimboe 
As I make the found flag check every symbol replace loop，a x86 problem triggerd:

create-diff-object: ERROR: static-local-moved-subfunction-issue-1054.ORIG.o: kpatch_replace_sections_syms: 1622: .orc_unwind_ip+0xc8: can't find replacement symbol for .init.text+711 reference: Success

The reason seens to bee the offset 711 is outside of .init.text section.
And an exception for this condition?

PR #1313, opened 2022-11-10
kpatch:use vmlinux from first build

robinYin996:
We can use the first compiled vmlinux file from origin source code, and  get the gcc version from the config file ，so we don't have to download kernel debuginfo(it's too big),and use kpatch-build command without '-v'

Signed-off-by: yinbinbin <yinbinbin001@linux.alibaba.com>

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -271,14 +271,29 @@ find_core_symvers() {
 }
 
 gcc_version_from_file() {
-	"$READELF" -p .comment "$1" | grep -m 1 -o 'GCC:.*'| awk '{$1=$2=""; print $0}'|awk '{gsub(/^\s+|\s+$/, "");print}'
+	"$READELF" -p .comment "$1" | grep -m 1 -o 'GCC:.*'| cut -d ' ' -f 3-
 }
 
 gcc_version_from_config() {
-    if [[ -n "$CONFIGFILE" ]]; then
-        cat "$CONFIGFILE" | grep -w "Compiler: gcc" | awk '{$1=$2=$3=$4=""; print $0}' | awk '{gsub(/^\s+|\s+$/, "");print}'
+    local gccver
+    if [[ -f "$CONFIGFILE" ]]; then
+	    # Compiler: gcc (GCC) 8.5.0 20210514 (Red Hat 8.5.0-4)
+	    gccver="$(sed -n '/Compiler:/p' "$CONFIGFILE" | cut -d ' ' -f 5-)"
+	    if [[ -z "$gccver" ]];then
+		    #CONFIG_CC_VERSION_TEXT="gcc (GCC) 6.5.1 20190307 (Red Hat 6.5.1-1 2.17)"
+		    gccver="$(sed -n '/CONFIG_CC_VERSION_TEXT/p' "$CONFIGFILE" | cut -d ' ' -f 3-)"
+		    gccver=${gccver%\"*}
+	    fi
+    fi
+    echo "$gccver"
+}
+
+kernel_version_from_config() {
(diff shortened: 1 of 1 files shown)


kpatch-build/kpatch-build
```diff
@@ -271,9 +271,14 @@ find_core_symvers() {
 }
 
 gcc_version_from_file() {
-	"$READELF" -p .comment "$1" | grep -m 1 -o 'GCC:.*'
+	"$READELF" -p .comment "$1" | grep -m 1 -o 'GCC:.*'| awk '{$1=$2=""; print $0}'|awk '{gsub(/^\s+|\s+$/, "");print}'
```
jpoimboe, line 274: This would be easier to read with the use of cut and sed.

kpatch-build/kpatch-build
```diff
@@ -271,9 +271,14 @@ find_core_symvers() {
...
 }
 
+gcc_version_from_config() {
+    if [[ -n "$CONFIGFILE" ]]; then
+        cat "$CONFIGFILE" | grep -w "Compiler: gcc" | awk '{$1=$2=$3=$4=""; print $0}' | awk '{gsub(/^\s+|\s+$/, "");print}'
```
jpoimboe, line 279: I think this doesn't work with newer .config files which have CONFIG_CC_VERSION_TEXT instead.

jpoimboe (commented):
This is an interesting approach.  If the original vmlinux isn't needed for the `--sourcedir` option, then it shouldn't be needed for any of the distro-specific options, and maybe we can just get rid of the `--vmlinux` option altogether, as well as the need for the debuginfo RPM.

Also, I think this highlights an existing bug in kpatch-build for the `--sourcedir` case where the vmlinux is in `$USERSRCDIR`.  It reads the symbol table from the patched version of vmlinux, which could be dangerous if the patch altered the symbol table.

kpatch-build/kpatch-build
robinYin996: Sorry，I submit it again

robinYin996:
> This is an interesting approach. If the original vmlinux isn't needed for the `--sourcedir` option, then it shouldn't be needed for any of the distro-specific options, and maybe we can just get rid of the `--vmlinux` option altogether, as well as the need for the debuginfo RPM.
> 
> Also, I think this highlights an existing bug in kpatch-build for the `--sourcedir` case where the vmlinux is in `$USERSRCDIR`. It reads the symbol table from the patched version of vmlinux, which could be dangerous if the patch altered the symbol table.

if "$VMLINUX" equal  "$KERNEL_SRCDIR"/vmlinux file ， kpatch-build  will save original vmlinux before it gets overwritten

jpoimboe:
> if "$VMLINUX" equal "$KERNEL_SRCDIR"/vmlinux file ， kpatch-build will save original vmlinux before it gets overwritten

But it's not restored before `readelf -s` is run on it.

jpoimboe:
> > if "$VMLINUX" equal "$KERNEL_SRCDIR"/vmlinux file ， kpatch-build will save original vmlinux before it gets overwritten
> 
> But it's not restored before `readelf -s` is run on it.

Never mind, I now see that VMLINUX gets changed to the original vmlinux in `"$TEMPDIR/vmlinux"` in that case.

jpoimboe:
> In fact, I'd like to improve that detection by doing an explicit comparison of their symbol tables.

So it turns out that on powerpc and s390 the symbol tables can get shuffled around when you add `-ffunction-sections -fdata-sections`.  And sometimes symbols can even be removed.  So it doesn't seem to be safe to use the first built vmlinux for symbol sympos lookups.  And we don't want to have x86-specific functionality here, so as far as I can tell, debuginfo will be required.

PR #1314, opened 2022-11-22
create-diff-object: fix __UNIQUE_ID() variable correlation

jpoimboe:
kpatch_mangled_strcmp() only ignores the digits after the period, but in the case of __UNIQUE_ID(), the symbol names have random digits before the period due to the use of `__COUNTER__`.  Make sure such symbols are properly correlated.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -397,6 +397,35 @@ static bool has_digit_tail(char *tail)
 }
 
 /*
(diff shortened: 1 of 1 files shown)


joe-lawrence (approved):
Code looks good, lmk if you want any help generating unit test object files.

jpoimboe:
Thanks for the review.  I already did a unit test, see https://github.com/dynup/kpatch-unit-test-objs/pull/45

PR #1315, opened 2022-11-22
Static call fixes

jpoimboe:
Improve handling of unsupported static calls
- for tracepoints, silently disable them
- otherwise give a better error message
- add KPATCH_STATIC_CALL macro


jpoimboe:
I will make unit tests for this and #1314 tomorrow-ish

joe-lawrence:
Thanks for figuring out that errx() thing, I thought I was crazy when I kept seeing "Error... :Success" messages :laughing:   The code looks fine, but I'd like to try going thru a patch iteratively to see the various warnings and recommendations in action.  If all goes well, I can try to put together some words for the patch author guide as well (some time next week).

jpoimboe:
Oh right, I forgot about the patch author guide.  I can do that as well (post Turkey-day most likely).

joe-lawrence (approved):

jpoimboe:
Oops, forgot to update the patch author guide.  Working on that now.

PR #1316, opened 2022-11-23
Draft: Add proc version example v2

joe-lawrence:
Red Hat QE reports that the recently added proc-version.patch prevents the crash utility from running as it parses /proc/version looking for the kernel version string.  This patch tweaks the example to move the "kpatch" string later in the banner.  Unfortunately the existing banner string already contains a trailing newline, so this seemed like the simplest solution while keeping the patch small and portable across kernel versions.


joe-lawrence:
Unfortunately this isn't good enough either as crash utility compares the entire vmlinux version string with /proc/version.

PR #1318, opened 2022-12-01
patch-author-guide: update jump label / static call descriptions

jpoimboe:
Now that we have KPATCH_STATIC_CALL(), document its usage.  While at it, give a more thorough description for why jump labels and static calls aren't supported in some scenarios.

### DIFF
--- a/doc/patch-author-guide.md
+++ b/doc/patch-author-guide.md
@@ -24,7 +24,7 @@ Table of contents
 - [Code removal](#code-removal)
 - [Once macros](#once-macros)
 - [inline implies notrace](#inline-implies-notrace)
-- [Jump labels](#jump-labels)
+- [Jump labels and static calls](#jump-labels-and-static-calls)
 - [Sibling calls](#sibling-calls)
 - [Exported symbol versioning](#exported-symbol-versioning)
 - [System calls](#system-calls)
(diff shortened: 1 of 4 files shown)


joe-lawrence:
Terminology question: "klp relas" or "klp-relocations" or "dynrelas" (as found in the source code) and is it time for a jargon file?

Also, feel free to steal any ascii art from https://lpc.events/event/4/contributions/507/attachments/316/533/LPC2019.pdf (slides 16, 17 for example).

jpoimboe:
> Terminology question: "klp relas" or "klp-relocations" or "dynrelas" (as found in the source code) and is it time for a jargon file?

I think the only valid names are "klp relocations" (because that's what they're called upstream), "klp relocs" (abbreviated, easier to type/read), and "klp relas" (after the section name).  Though "reloc" is more general of a name than "rela".  The "dynrela" naming is obsolete.  Any objections if I do a mass rename in the code (basically s/dynrela/klp_reloc/)?

jpoimboe:
v2:
- fix typo
- RIP "dynrela"

jpoimboe:
v2.1:
- rebase to fix unit test failure
- oops, forgot to git add my latest changes

joe-lawrence (approved):

PR #1319, opened 2022-12-01
github actions: fix warning with node 16

jpoimboe:
Attempt to fix the following warning:

  Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout@v2

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -11,7 +11,7 @@ jobs:
     env:
       CFLAGS: ${{ matrix.cflags }}
     steps:
-    - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


joe-lawrence (approved):
Sure, let's fix the warning with node 16!  Where did this even emanate from?  Did github send email this to you about a specific PR?

jpoimboe:
Yeah, I was getting email about github action warnings on my private branches like:

https://github.com/jpoimboe/kpatch/actions/runs/3596879345

PR #1322, opened 2023-01-13
Upstream kernel v6.2 support

joe-lawrence:
Collection of changes to support up to v6.2.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2835,6 +2835,12 @@ static void kpatch_mark_ignored_sections(struct kpatch_elf *kelf)
 		    !strncmp(sec->name, ".llvm_addrsig", 13) ||
 		    !strncmp(sec->name, ".llvm.", 6))
 			sec->ignore = 1;
+
+		if (kelf->arch == X86_64) {
+			if (!strcmp(sec->name, ".rela__patchable_function_entries") ||
+			    !strcmp(sec->name, "__patchable_function_entries"))
+				sec->ignore = 1;
+		}
 	}
 
 	sec = find_section_by_name(&kelf->sections, ".kpatch.ignore.sections");
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -236,14 +236,21 @@ static struct rela *toc_rela(const struct rela *rela)
 static void kpatch_bundle_symbols(struct kpatch_elf *kelf)
 {
 	struct symbol *sym;
+	unsigned int expected_offset;
 
 	list_for_each_entry(sym, &kelf->symbols, list) {
 		if (is_bundleable(sym)) {
-			if (sym->sym.st_value != 0 &&
-			    !is_gcc6_localentry_bundled_sym(kelf, sym)) {
-				ERROR("symbol %s at offset %lu within section %s, expected 0",
+			if (sym->pfx)
+				expected_offset = 16;
+			else if (is_gcc6_localentry_bundled_sym(kelf, sym))
+				expected_offset = 8;
+			else
+				expected_offset = 0;
+
+			if (sym->sym.st_value != expected_offset) {
+				ERROR("symbol %s at offset %lu within section %s, expected %u",
 				      sym->name, sym->sym.st_value,
-				      sym->sec->name);
+				      sym->sec->name, expected_offset);
 			}
 
 			sym->sec->sym = sym;
@@ -1913,6 +1920,8 @@ static int kpatch_include_changed_functions(struct kpatch_elf *kelf)
 		    sym->type == STT_FUNC) {
 			changed_nr++;
 			kpatch_include_symbol(sym);
+			if (sym->pfx)
+				kpatch_include_symbol(sym->pfx);
 		}
 
 		if (sym->type == STT_FILE)
@@ -1927,7 +1936,8 @@ static void kpatch_print_changes(struct kpatch_elf *kelf)
 	struct symbol *sym;
 
 	list_for_each_entry(sym, &kelf->symbols, list) {
-		if (!sym->include || !sym->sec || sym->type != STT_FUNC || sym->parent)
+		if (!sym->include || !sym->sec || sym->type != STT_FUNC ||
+		    sym->parent || sym->is_pfx)
 			continue;
 		if (sym->status == NEW)
 			log_normal("new function: %s\n", sym->name);
@@ -3922,7 +3932,8 @@ static void kpatch_find_func_profiling_calls(struct kpatch_elf *kelf)
 	struct rela *rela;
 	unsigned char *insn;
 	list_for_each_entry(sym, &kelf->symbols, list) {
-		if (sym->type != STT_FUNC || !sym->sec || !sym->sec->rela)
+		if (sym->type != STT_FUNC || sym->is_pfx ||
+		    !sym->sec || !sym->sec->rela)
 			continue;
 
 		switch(kelf->arch) {
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2124,6 +2124,11 @@ static int static_call_sites_group_size(struct kpatch_elf *kelf, int offset)
 	return size;
 }
 
+static int call_sites_group_size(struct kpatch_elf *kelf, int offset)
+{
+	return 4;
+}
+
 static int retpoline_sites_group_size(struct kpatch_elf *kelf, int offset)
 {
 	return 4;
@@ -2430,6 +2435,11 @@ static struct special_section special_sections[] = {
 		.arch		= X86_64,
 		.group_size	= static_call_sites_group_size,
 		.group_filter	= static_call_sites_group_filter,
+	},
+	{
+		.name		= ".call_sites",
+		.arch		= X86_64,
+		.group_size	= call_sites_group_size,
 	},
 	{
 		.name		= ".retpoline_sites",
--- a/kpatch-build/kpatch-elf.c
+++ b/kpatch-build/kpatch-elf.c
@@ -397,6 +397,97 @@ static void kpatch_create_section_list(struct kpatch_elf *kelf)
 		ERROR("expected NULL");
 }
 
+/*
+ * Some x86 kernels have NOP function padding [1] for which objtool [2]
+ * adds ELF function symbols with prefix "__pfx_" to indicate the start
+ * of a function, inclusive of NOP-padding.  Find the prefix symbols and
+ * link them to their corresponding function symbols at an expected
+ * offset.
+ *
+ * A few examples:
+ *
+ *    Value          Size Type    Bind   Vis      Ndx Name
+ * (fork.o, simple case)
+ * 0000000000000000     0 SECTION LOCAL  DEFAULT   31 .text.get_task_mm
+ * 0000000000000000    16 FUNC    GLOBAL DEFAULT   31 __pfx_get_task_mm
+ * 0000000000000010    91 FUNC    GLOBAL DEFAULT   31 get_task_mm
+ *
+ * (fork.o, multiple function aliases)
+ * 0000000000000000     0 SECTION LOCAL  DEFAULT  190 .text.__do_sys_fork
+ * 0000000000000000    16 FUNC    GLOBAL DEFAULT  190 __pfx___x64_sys_fork
+ * 0000000000000010    49 FUNC    LOCAL  DEFAULT  190 __do_sys_fork
+ * 0000000000000010    49 FUNC    GLOBAL DEFAULT  190 __ia32_sys_fork
+ * 0000000000000010    49 FUNC    GLOBAL DEFAULT  190 __x64_sys_fork
+ *
+ * (fork.o multiple functions in one section)
+ * 0000000000000000     0 SECTION LOCAL  DEFAULT   59 .init.text
+ * 0000000000000000    16 FUNC    LOCAL  DEFAULT   59 __pfx_coredump_filter_setup
+ * 0000000000000010    40 FUNC    LOCAL  DEFAULT   59 coredump_filter_setup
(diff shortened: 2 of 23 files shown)


joe-lawrence:
@liu-song-6 : Here are some more commits broken out to support some of the things I encountered with v6.2-rc3.  Still draft state and needs review.

@swine : Interestingly enough, x86 will start using patchable_function_entries, too.  I integrated the initial patches from your arm64 branch to support what I'm seeing with x86 and my toolchain (no multiple pfe sections).  The major change here is that it I think we'll need to detect and support both, as it's no longer a single switch on the arch (for x86 anyway).

kpatch-build/create-diff-object.c
```diff
@@ -228,6 +228,54 @@ static struct rela *toc_rela(const struct rela *rela)
...
+	if (sym->type != STT_FUNC)
+		return false;
+
+	for (offset = 0; offset < sym->sym.st_value; offset += insn_len) {
+		if (!insn_is_nop_pad(kelf, sym->sec->data->d_buf + offset, end, &insn_len))
+			return false;
+	}
+
+	return true;
+}
```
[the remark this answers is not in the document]
joe-lawrence: Good observation.  IIRC I developed nop-detection for an earlier commit before "__pfx" symbols were implemented.  Maybe this occurred from rc2 to rc3?  Regardless, searching for "__pfx" symbols is a lot easier (and potentially easier on other arches if ever needed).

kpatch-build/create-diff-object.c
```diff
@@ -1669,11 +1720,11 @@ static void kpatch_check_func_profiling_calls(struct kpatch_elf *kelf)
 	int errs = 0;
 
 	list_for_each_entry(sym, &kelf->symbols, list) {
-		if (sym->type != STT_FUNC || sym->status != CHANGED ||
+		if (sym->type != STT_FUNC || sym->func_prefix || sym->status != CHANGED ||
```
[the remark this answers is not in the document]
joe-lawrence: I'll take a second look, but I probably searched for all the symbol list traversals and just added the condition as another "don't bother processing it" without thinking about the greater context.  I'll clean up in the next version.

kpatch-build/create-diff-object.c
```diff
@@ -3637,13 +3708,51 @@ static void kpatch_create_callbacks_objname_rela(struct kpatch_elf *kelf, char *
 }
 
 /*
+ * Allocate the mcount/patchable_function_entry sections which must be done
+ * before the patched object is torn down so that the section flags can be
+ * copied.
+ */
+static void kpatch_alloc_mcount_sections(struct kpatch_elf *kelf, struct kpatch_elf *kelfout)
```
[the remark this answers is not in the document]
joe-lawrence: Ooo, I'll definitely take a second look at this one.  I was only looking at the object files and assumed the section made its way out to vmlinux / module.ko.

test/integration/linux-6.2.0/module.patch
```diff
@@ -0,0 +1,95 @@
...
++		printk("kpatch: memcg_kmem_enabled_key\n");
++
++	BUG_ON(!static_branch_likely(&kpatch_key));
++	static_branch_disable(&kpatch_key);
++	BUG_ON(static_branch_likely(&kpatch_key));
++	static_branch_enable(&kpatch_key);
++#endif
++       return 0;
++}
++KPATCH_PRE_PATCH_CALLBACK(pre_patch_callback);
```
joe-lawrence, line 48: @jpoimboe : what do you think of this change?  I don't know why we've never seen dmesg complaints before, but static_branch_disable() may block and provoke RCU read-side critical section warnings.  Moving this to a call back puts it in a safe context, but I didn't know if there were other reasons for these changes to exist in a replaced function (and not a callback).

jpoimboe: Hm, I think moving it to a callback defeats the point of the test.  It's intended to test jump labels in patched code, which rely on create-diff-object and klp relas working correctly.

joe-lawrence:
@jpoimboe : sorry long the long delay on this PR, I've been swamped other w/day-to-day work.  Here is my current progress on implementing your suggestions for 6.1, 6.2 support.  IIRC there are still some integration test warnings to work through (special-static.patch and lockdep warnings for example).

kpatch-build/kpatch-elf.c
```diff
@@ -397,6 +411,89 @@ static void kpatch_create_section_list(struct kpatch_elf *kelf)
...
+
+	log_debug("\n=== prefixed function links ===\n");
+	list_for_each_entry(sym, &kelf->symbols, list) {
+
+		if (sym->type != STT_FUNC || sym->func_prefix ||
+		    sym->sym.st_value < pfx_nop_offset)
+			continue;
+
+		pfx_sym = find_symbol_by_offset(&kelf->symbols, sym->sec,
+				(unsigned) sym->sym.st_value - pfx_nop_offset);
```
jpoimboe, line 482: The symbol being 16 bytes away doesn't necessarily mean it's a prefix symbol.  It could just be a small function.  It should have the same name as sym with the prefix added.

It might be cleaner to do the lookup in the other direction.  For each `__pfx_*` or `__cfi_*` symbol, do find_symbol_by_name_offset() 16 bytes later (and error out if it can't be found).

kpatch-build/kpatch-elf.c
```diff
@@ -397,6 +411,89 @@ static void kpatch_create_section_list(struct kpatch_elf *kelf)
...
+		if (!pfx_sym) {
+			printf("%s NOTE: could not find corresponding prefix function symbol for %s in section %s\n",
+				childobj, sym->name, sym->sec->name);
+			continue;
+		}
+
+		pfx_sym->func_prefix = true;
+		sym->prefix_link = pfx_sym;
```
jpoimboe, line 490: With `prefix_link` not being bidirectional, it's not clear which direction it means or how it differs from `func_prefix`.  Maybe `prefix_link` could just be called `prefix` (consistent in naming with `parent` and `children`) and `func_prefix` could be called `is_prefix`.

kpatch-build/create-diff-object.c
```diff
@@ -239,7 +239,7 @@ static void kpatch_bundle_symbols(struct kpatch_elf *kelf)
 
 	list_for_each_entry(sym, &kelf->symbols, list) {
 		if (is_bundleable(sym)) {
-			if (sym->sym.st_value != 0 &&
+			if (sym->sym.st_value != 0 && !sym->prefix_link &&
 			    !is_gcc6_localentry_bundled_sym(kelf, sym)) {
 				ERROR("symbol %s at offset %lu within section %s, expected 0",
 				      sym->name, sym->sym.st_value,
```
jpoimboe, line 245: This warning probably needs to be updated, i.e. shouldn't we warn about a bundled prefixed symbol which is not at 16 bytes, or a bundled localentry symbol which is not at 8 bytes?

kpatch-build/kpatch-elf.c
```diff
@@ -121,6 +121,20 @@ struct symbol *find_symbol_by_name(struct list_head *list, const char *name)
 	return NULL;
 }
 
+struct symbol *find_symbol_by_offset(struct list_head *list, struct section *sec,
+				     unsigned int offset)
+{
+	struct symbol *sym;
+
+	list_for_each_entry(sym, list, list)
+		if (sym->type == STT_FUNC &&
```
jpoimboe, line 130: Based on this check this would be `find_func_by_offset()`

jpoimboe (commented):
@joe-lawrence if you're busy, let me know if you want me to take over the patches.  I could also help look at the integration test failures.

kpatch-build/create-diff-object.c
joe-lawrence: Yea, there's a bunch of different conditions that are now crammed in here that should be called out.

kpatch-build/kpatch-elf.c
joe-lawrence: Yup

kpatch-build/kpatch-elf.c
joe-lawrence: Right, that seems like a more accurate way to about it.

joe-lawrence:
> @joe-lawrence if you're busy, let me know if you want me to take over the patches. I could also help look at the integration test failures.

If you currently have some bandwidth, that would be great.

jpoimboe:
I've got a branch [here](https://github.com/jpoimboe/kpatch/tree/6.2-fixes) which is the same as yours except I reworked the nops commit.  I still need to look at the integration tests.

joe-lawrence:
> I've got a branch [here](https://github.com/jpoimboe/kpatch/tree/6.2-fixes) which is the same as yours except I reworked the nops commit. I still need to look at the integration tests.

Ok, I'll take a look.  Do you want to open a new MR, or force push on top of my branch?  BTW, you may need the additional gcc-plugin update I pushed yesterday if you're building with Fedora 38 / ppc64le.

jpoimboe:
> Ok, I'll take a look. Do you want to open a new MR, or **force push on top of my branch**? ~~

hm, yeah, I didn't know that was possible.  Lemme try to figure out how to do that :-)

joe-lawrence:
> > Ok, I'll take a look. Do you want to open a new MR, or **force push on top of my branch**? ~~
> 
> hm, yeah, I didn't know that was possible. Lemme try to figure out how to do that :-)

I've never tried it, but the webUI (in the righthand column) has this checked "Allow edits and access to secrets by maintainers" and the tooltip says, "If checked, users with write access to dynup/kpatch can add new commits to your v6.2-support branch. You can always change this setting later."

jpoimboe:
> I've never tried it, but the webUI (in the righthand column) has this checked "Allow edits and access to secrets by maintainers" and the tooltip says, "If checked, users with write access to dynup/kpatch can add new commits to your v6.2-support branch. You can always change this setting later."

Nice, looks like I managed to do it.

This latest version has a fix to kpatch_find_func_profiling_calls().  I was seeing integration test failures in ftrace due to some pfx functions getting added to `__mcount_loc`.

jpoimboe:
I replaced the gcc plugin commit from this PR with the one from #1337.  Integration tests for x86 and ppc are working for me.

jpoimboe (approved):

jpoimboe:
Added a commit to fix an integration test issue found by Yulia.

joe-lawrence:
Yulia's latest tests look good for gcc x86_64, s390x and ppc64le.  Clang tests on x86_64 show two failures.

The first is related to the `__fentry__` relocation type and the new `__pfx_<func>` nops:
```
create-diff-object: ERROR: cmdline.o: kpatch_create_mcount_sections: 3754: kpatch_print_message: unexpected instruction at the start of the function
cmdline.o: new function: kpatch_print_message
```
In kpatch_create_mcount_sections() , if the rela is R_X86_64_NONE, it assumes the beginning of the section will start the `0f 1f 44 00 00          nopl   0x0(%rax,%rax,1)` sequence.   For clang version 15.0.7 (Fedora 15.0.7-2.fc37), we see something like:
```
Disassembly of section .text.kpatch_print_message:

0000000000000000 <__pfx_kpatch_print_message>:
   0:   90                      nop
   1:   90                      nop
   2:   90                      nop
   3:   90                      nop
   4:   90                      nop
   5:   90                      nop
   6:   90                      nop
   7:   90                      nop
   8:   90                      nop
   9:   90                      nop
   a:   90                      nop
   b:   90                      nop
   c:   90                      nop
   d:   90                      nop
   e:   90                      nop
   f:   90                      nop

0000000000000010 <kpatch_print_message>:
  10:   0f 1f 44 00 00          nopl   0x0(%rax,%rax,1)
                        11: R_X86_64_NONE       __fentry__-0x4
```
So I think it will need to handle the `__pfx_<func>` nops case well and make the comparison at the real function (not section) start.  I assume an older gcc compiler would generate the same code and relocation type as well?

[repro-instr-start.tar.gz](https://github.com/dynup/kpatch/files/11127212/repro-instr-start.tar.gz)

joe-lawrence:
The second clang failure occurs for `.L.str[.X]` symbols (added debugging):
```
create-diff-object: ERROR: af_netlink.o: find_local_syms: 191: couldn't find matching af_netlink.c local symbols in vmlinux symbol table
  locals_match failed for table_sym: .L.str
create-diff-object: ERROR: fork.o: find_local_syms: 191: couldn't find matching fork.c local symbols in vmlinux symbol table
  locals_match failed for table_sym: .L.str
create-diff-object: ERROR: aio.o: find_local_syms: 191: couldn't find matching aio.c local symbols in vmlinux symbol table
  locals_match failed for table_sym: .L.str
create-diff-object: ERROR: exit.o: find_local_syms: 191: couldn't find matching exit.c local symbols in vmlinux symbol table
  locals_match failed for table_sym: .L.str.33
create-diff-object: ERROR: sys.o: find_local_syms: 191: couldn't find matching sys.c local symbols in vmlinux symbol table
  locals_match failed for table_sym: .L.str.105
create-diff-object: ERROR: n_tty.o: find_local_syms: 191: couldn't find matching n_tty.c local symbols in vmlinux symbol table
  locals_match failed for table_sym: .L.str.6
```
These symbols appear in the vmlinux.symtab for their given FILE, but not in the individual original or patched object files:
```
$ grep 'fork\.c' vmlinux.symtab 
 14271: 0000000000000000     0 FILE    LOCAL  DEFAULT  ABS fork.c
 ...
 14464: ffffffff82804060    14 OBJECT  LOCAL  DEFAULT    3 .L.str
 14465: ffffffff82804058     0 NOTYPE  LOCAL  DEFAULT    3 .LCPI20_0

$ readelf --wide --symbols ~/repro-local_syms/fork.PATCHED.o | grep '.L.str'
$ readelf --wide --symbols ~/repro-local_syms/fork.ORIG.o | grep '.L.str'
```
Skipping these symbols in the table_sym check in locals_match() does appease the build.

joe-lawrence:
@jpoimboe : should we merge this one and handle additional clang issues in a follow up PR?

jpoimboe:
@joe-lawrence sounds good to me

PR #1323, opened 2023-01-25
examples: add cmdline-string.patch as a compatible example

lulinqing:
This is a test example currently preferable to [proc-version.patch](https://github.com/dynup/kpatch/blob/master/examples/proc-version.patch) as crash utility has difficulty parsing /proc/version content after being altered.

Signed-off-by: Linqing Lu <lilu@redhat.com>


jpoimboe (approved):

joe-lawrence:
Thanks @lulinqing

Issue #1320, opened 2022-12-09
Need suggestions for CONFIG_LTO_CLANG support


liu-song-6:
I am try to build livepatch for kernel built with CONFIG_LTO_CLANG. 
The first issue was these .o files are not ELF files, but LLVM IR bitcode. To resolve this, I made [some hack](https://github.com/liu-song-6/kpatch/commit/e7be2deee4519a0b697680c2673dc3704c3383fa) so that we run create-diff-object on vmlinux.o, which is ELF. However, create-diff-object spent hours in the following stack:

     find_symbol_by_index()                                                                                                                                                                                           
     kpatch_create_rela_list()                                                                                                                                                                                        
     kpatch_elf_open()

which makes it unusable. 

So my question is: Shall we optimize create-diff-object to be faster? Or shall we try some other solutions?

Thanks in advance!

joe-lawrence:
@sm00th : This sounds familiar from your previous study of clang / LTO.  Did you have any notes or branches left over from that investigation that may help @liu-song-6 ?

sm00th:
Hi there. Yeah, I ran into this issue as well and used this [hacky commit](https://github.com/sm00th/kpatch/commit/ac336fe35415ae334180ae61c431641ebb0470ec) to alleviate the problem during development. However I didn't think that is a proper solution to find_symbol_by_index() slowness.

The might be some other parts you might find usable on that branch: https://github.com/sm00th/kpatch/commits/lto

It is all very raw though.

liu-song-6:
@joe-lawrence @sm00th Thanks for sharing this work. 

I tried the lto branch (on top of some of our patches for pgo), and it seems there is still some issues. 

@sm00th , do you have plan to continue working on this? We are hoping to use kpatch on our LTO kernel in 2023 (hopefully H1). Thanks!

sm00th:
> @joe-lawrence @sm00th Thanks for sharing this work.
> 
> I tried the lto branch (on top of some of our patches for pgo), and it seems there is still some issues.

Yes, there are. I worked on this a while ago and there definitely were issues left, but afaik it produced loadable and working (but unwieldy, they were roughly the same size as vmlinux because of all the debuginfo and rodata sections being copied over) kpatches for simple patches for vmlinux (maybe not for modules). The better parts of the branch were merged to mainline and what is left over are hacky commits that probably shouldn't be used as is but might give you some insights and early warnings about things I encountered. 

[...]

PR #1325, opened 2023-02-08
kpatch-build: support CONFIG_LTO_CLANG_THIN

liu-song-6:
Support CONFIG_LTO_CLANG_THIN with ld.lld --lto-obj-path option.

With CONFIG_LTO_CLANG_THIN, .o files are LLVM IR binary, so CDO doesn't work on .o file. To solve this issue, we CDO the thinlto files generated by the --lto-obj-path option. Clang LTO generates the thinlto files after cross file inline, so they are good candidates for CDO. See [1] for more discussions about this.

To achieve this, we need:

  1. kpatch-build to update kernel Makefile(s) so it generates thinlto files;
  2. kpatch-build and kpatch-cc to save the thinlto file properly;
  3. kpatch-build to feed these thinlto files to CDO;
  4. The user need to supply vmlinux.o, from which we generate the symtab file. We need this because GLOBAL symbols may be marked as LOCAL in LTO vmlinux;
  4. A small workaround in CDO that ignores changes in init.text for vmlinux.o.thinlto.*

[1] https://github.com/dynup/kpatch/issues/1320

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -430,6 +430,23 @@ static int __kpatch_unique_id_strcmp(char *s1, char *s2)
 	return 1;
 }
 
+static int __kpatch_skip_digits_strcmp(char *s1, char *s2)
+{
+	while (*s1 == *s2) {
+		if (!*s1)
+			return 0;
+		s1++;
+		s2++;
+
+		while (isdigit(*s1))
+			s1++;
+		while (isdigit(*s2))
+			s2++;
+	}
+
+	return 1;
+}
+
 /*
  * This is like strcmp, but for gcc-mangled symbols.  It skips the comparison
  * of any substring which consists of '.' followed by any number of digits.
@@ -445,6 +462,19 @@ static int kpatch_mangled_strcmp(char *s1, char *s2)
 
 	if (!strncmp(s1, "__UNIQUE_ID_", 12))
 		return __kpatch_unique_id_strcmp(s1, s2);
+
+	/*
+	 * Hack for __initcall_stub() with CONFIG_LTO_CLANG.
+	 * The following should match:
+	 *
+	 *   __initstub__kmod_syscall__728_5326_bpf_syscall_sysctl_init7
+	 *   __initstub__kmod_syscall__728_5324_bpf_syscall_sysctl_init7
+	 *
+	 * Please refer to __initcall_stub() in include/linux/init.h for
+	 * more details.
+	 */
+	if (!strncmp(s1, "__initstub__kmod_syscall__", 26))
+		return __kpatch_skip_digits_strcmp(s1, s2);
 
 	while (*s1 == *s2) {
 		if (!*s1)
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -1118,6 +1118,11 @@ else
 	MAKEVARS+=("LD=${KPATCH_CC_PREFIX}${LD}")
 fi
 
+# It is common for the user to enable all LLVM utilities with LLVM=1.
+# Detech such case with CONFIG_AS_IS_LLVM flag.
+if [[ -n "$CONFIG_AS_IS_LLVM" ]]; then
+	MAKEVARS+=("LLVM=1")
+fi
 
 # $TARGETS used as list, no quotes.
 # shellcheck disable=SC2086
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -173,11 +173,11 @@ cleanup() {
 	# restore original vmlinux if it was overwritten by sourcedir build
 	[[ -e "$TEMPDIR/vmlinux" ]] && mv -f "$TEMPDIR/vmlinux" "$KERNEL_SRCDIR/"
 
-	# restore original link-vmlinux.sh if we updated it for the build
+	# restore any files that were modified for the build
 	[[ -e "$TEMPDIR/link-vmlinux.sh" ]] && mv -f "$TEMPDIR/link-vmlinux.sh" "$KERNEL_SRCDIR/scripts"
-
-	# restore original Makefile.modfinal if we updated it for the build
 	[[ -e "$TEMPDIR/Makefile.modfinal" ]] && mv -f "$TEMPDIR/Makefile.modfinal" "$KERNEL_SRCDIR/scripts"
+	[[ -e "$TEMPDIR/Makefile.build" ]] && mv -f "$TEMPDIR/Makefile.build" "$KERNEL_SRCDIR/scripts"
+	[[ -e "$TEMPDIR/Makefile" ]] && mv -f "$TEMPDIR/Makefile" "$KERNEL_SRCDIR"
 
 	[[ "$DEBUG" -eq 0 ]] && rm -rf "$TEMPDIR"
 	rm -rf "$RPMTOPDIR"
@@ -1054,6 +1054,23 @@ if [[ -n "$CONFIG_DEBUG_INFO_BTF" ]]; then
 	fi
 fi
 
+if [[ -n "$CONFIG_LTO_CLANG" ]]; then
+	[[ -n "$CONFIG_LTO_CLANG_THIN" ]] || die "Non-thin LTO is not supported. Please enable CONFIG_LTO_CLANG_THIN"
+
+	# This is a heuristic: use -x to check vmlinux vs. vmlinux.o
+	[[ -x "$VMLINUX" ]] && die "For kernel with CONFIG_LTO_CLANG, please supply vmlinux.o instead of vmlinux via -v|--vmlinux option."
+
+	# update Makefile so that ld.lld generate vmlinux.o.thinlto.o* files for vmlinux.o
+	cp -f "$KERNEL_SRCDIR/Makefile" "$TEMPDIR/Makefile" || die
+	sed -i "s/--thinlto-cache-dir=\$(extmod_prefix).thinlto-cache/--lto-obj-path=vmlinux.o.thinlto.o/g" "$KERNEL_SRCDIR"/Makefile
+
+	# update scripts/Makefile.build so that ld.lld generate XX.o.thinlto.o* files for modules
+	cp -f "$KERNEL_SRCDIR/scripts/Makefile.build" "$TEMPDIR/Makefile.build" || die
+	sed -i "s/\$(ld_flags)/\$(ld_flags) --lto-obj-path=\$@.thinlto.o/g" "$KERNEL_SRCDIR"/scripts/Makefile.build
+
+	export KPATCH_CC_LTO=1
+fi
+
 if [[ -n "$CONFIG_CC_IS_CLANG" ]]; then
 	echo "WARNING: Clang support is experimental"
 fi
@@ -1179,9 +1196,14 @@ if [[ -n "$CONFIG_MODVERSIONS" ]]; then
 	trace_on
 fi
 
+if [[ -n "$CONFIG_LTO_CLANG" ]]; then
+	DIFF_OBJS="$TEMPDIR/thinlto_objs"
+else
+	DIFF_OBJS="$TEMPDIR/changed_objs"
+fi
 # Read as words, no quotes.
 # shellcheck disable=SC2013
-for i in $(cat "$TEMPDIR/changed_objs")
+for i in $(cat "$DIFF_OBJS")
 do
 	mkdir -p "$TEMPDIR/patched/$(dirname "$i")" || die
 	cp -f "$BUILDDIR/$i" "$TEMPDIR/patched/$i" || die
@@ -1210,7 +1232,8 @@ if [[ -z "$MODNAME" ]] ; then
 
 	MODNAME="$(module_name_string "$MODNAME")"
 fi
-FILES="$(cat "$TEMPDIR/changed_objs")"
+FILES="$(cat "$DIFF_OBJS")"
+
 cd "$TEMPDIR" || die
 mkdir output
 declare -a objnames
@@ -1234,7 +1257,11 @@ for i in $FILES; do
 
 	mkdir -p "output/$(dirname "$i")"
 	cd "$BUILDDIR" || die
-	find_kobj "$i"
+	if [[ -n "$CONFIG_LTO_CLANG" ]] ; then
+		KOBJFILE=${i/.o.thinlto.o*/}
+	else
+		find_kobj "$i"
+	fi
 	cd "$TEMPDIR" || die
 	if [[ -e "orig/$i" ]]; then
 		if [[ -n $OOT_MODULE ]]; then
@@ -1251,14 +1278,28 @@ for i in $FILES; do
 		else
 			KOBJFILE_NAME=$(basename "${KOBJFILE%.ko}")
 			KOBJFILE_NAME="${KOBJFILE_NAME//-/_}"
-			KOBJFILE_PATH="${TEMPDIR}/module/$KOBJFILE"
+			if [[ -n "$CONFIG_LTO_CLANG" ]] ; then
+				KOBJFILE_PATH="${TEMPDIR}/module/$KOBJFILE.ko"
+			else
+				KOBJFILE_PATH="${TEMPDIR}/module/$KOBJFILE"
+			fi
 			SYMTAB="${KOBJFILE_PATH}.symtab"
 			SYMVERS_FILE="$BUILDDIR/Module.symvers"
 		fi
 
-		"$READELF" -s --wide "$KOBJFILE_PATH" > "$SYMTAB"
+		# With CONFIG_LTO_CLANG, multiple .thinlto files share a
+		# symtab file. Only generate the symtab file once.
+		[[ -e "$SYMTAB" ]] || "$READELF" --symbols --wide "$KOBJFILE_PATH" > "$SYMTAB"
 		if [[ "$ARCH" = "ppc64le" ]]; then
 			sed -ri 's/\s+\[<localentry>: 8\]//' "$SYMTAB"
+		fi
+
+		if [[ -n "$CONFIG_LTO_CLANG" ]] ; then
+			# skip .thinlto file that didn't change at all
+			diff "orig/$i" "patched/$i" 2> /dev/null && continue
+			# skip .thinlto file without any functions
+			num_func=$("$READELF" --symbols "orig/$i" | grep -c FUNC)
+			[[ $num_func -eq 0 ]] && continue
 		fi
 
 		# create-diff-object orig.o patched.o parent-name parent-symtab
@@ -1313,7 +1354,7 @@ fi
 cd "$TEMPDIR/output" || die
 # $KPATCH_LDFLAGS and result of find used as list, no quotes.
 # shellcheck disable=SC2086,SC2046
-"$LD" -r $KPATCH_LDFLAGS -o ../patch/tmp_output.o $(find . -name "*.o") 2>&1 | logger || die
+"$LD" -r $KPATCH_LDFLAGS -o ../patch/tmp_output.o $(find . -name "*.o*") 2>&1 | logger || die
 
 if [[ "$USE_KLP" -eq 1 ]]; then
 	cp -f "$TEMPDIR"/patch/tmp_output.o "$TEMPDIR"/patch/output.o || die
--- a/kpatch-build/kpatch-cc
+++ b/kpatch-build/kpatch-cc
@@ -79,6 +79,24 @@ elif [[ "$TOOLCHAINCMD" =~ ^(.*-)?ld || "$TOOLCHAINCMD" =~ ^(.*-)?ld.lld ]] ; th
 					args+=(--warn-unresolved-symbols)
 					break
 					;;
+				*/.tmp_*.o)
+					# .tmp_*.o is used for single file modules.
+					# See "cmd_ld_single_m" in scripts/Makefile.build.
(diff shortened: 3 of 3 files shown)


kpatch-build/kpatch-build
```diff
@@ -1054,6 +1060,35 @@ if [[ -n "$CONFIG_DEBUG_INFO_BTF" ]]; then
...
+	[[ -e "$VMLINUX" ]] || die "For kernel with CONFIG_LTO_CLANG, please supply vmlinux.o via -v|--vmlinux option."
+
+	# This is a heuristic: use -x to check vmlinux vs. vmlinux.o
+	[[ -x "$VMLINUX" ]] && die "For kernel with CONFIG_LTO_CLANG, please supply vmlinux.o instead of vmlinux via -v|--vmlinux option."
+
+	cp -f "$KERNEL_SRCDIR/Makefile" "$TEMPDIR/Makefile" || die
+	sed -i "s/--thinlto-cache-dir=\$(extmod_prefix).thinlto-cache/--lto-obj-path=vmlinux.o.thinlto.o/g" "$KERNEL_SRCDIR"/Makefile
+
+	cp -f "$KERNEL_SRCDIR/scripts/Makefile.build" "$TEMPDIR/Makefile.build" || die
+	sed -i "s/\$(ld_flags)/\$(ld_flags) --lto-obj-path=\$@.thinlto.o/g" "$KERNEL_SRCDIR"/scripts/Makefile.build
```
jpoimboe, line 1087: These changes to Makefile and Makefile.build need short comments describing what they're doing and why

kpatch-build/kpatch-cc
```diff
@@ -79,6 +79,24 @@ elif [[ "$TOOLCHAINCMD" =~ ^(.*-)?ld || "$TOOLCHAINCMD" =~ ^(.*-)?ld.lld ]] ; th
 					args+=(--warn-unresolved-symbols)
 					break
 					;;
+				*/.tmp_*.o)
+					# .tmp_*.o is used for single file modules.
+					# See "cmd_ld_single_m" in scripts/Makefile.build.
+					if [[ $KPATCH_CC_HANDLE_LTO -ne 0 ]] ; then
```
jpoimboe, line 85: I think "[[ -n $KPATCH_CC_LTO -eq 1 ]]" would be clearer since we never actually initialize the variable to 0 (and also it avoids the double negative check)

kpatch-build/kpatch-cc
```diff
@@ -79,6 +79,24 @@ elif [[ "$TOOLCHAINCMD" =~ ^(.*-)?ld || "$TOOLCHAINCMD" =~ ^(.*-)?ld.lld ]] ; th
...
+						mkdir -p "$KPATCH_GCC_TEMPDIR/orig/$(dirname "$relobj")"
+						cp "${obj/.tmp_/}".thinlto.o* "$KPATCH_GCC_TEMPDIR/orig/$(dirname "$relobj")"
+						echo "${obj/.tmp_/}".thinlto.o* >> "$KPATCH_GCC_TEMPDIR/thinlto_objs"
+					fi
+					break
+					;;
+				*.o)
+					if [[ $KPATCH_CC_HANDLE_LTO -ne 0 ]] ; then
```
jpoimboe, line 93: same

kpatch-build/kpatch-build
```diff
@@ -1238,6 +1284,19 @@ for i in $FILES; do
...
+				KOBJFILE_PATH="$VMLINUX"
+				SYMTAB="${TEMPDIR}/${KOBJFILE_NAME}.symtab"
+				SYMVERS_FILE="$BUILDDIR/Module.symvers"
+			else
+				KOBJFILE_NAME=$(basename "${KOBJFILE%.ko}")
+				KOBJFILE_NAME="${KOBJFILE_NAME//-/_}"
+				KOBJFILE_PATH="${TEMPDIR}/module/$KOBJFILE.ko"
+				SYMTAB="${KOBJFILE_PATH}.symtab"
+				SYMVERS_FILE="$BUILDDIR/Module.symvers"
+			fi
```
jpoimboe, line 1299: This change is almost completely copy-pasted from existing code and could surely be simplified?

kpatch-build/kpatch-build
```diff
@@ -1229,7 +1271,11 @@ for i in $FILES; do
 
 	mkdir -p "output/$(dirname "$i")"
 	cd "$BUILDDIR" || die
-	find_kobj "$i"
+	if [[ -z "$CONFIG_LTO_CLANG" ]] ; then
+		find_kobj "$i"
+	else
+		KOBJFILE=${i/.o.thinlto.o*/}
+	fi
```
jpoimboe, line 1278: Please do the `[[ -n "$CONFIG_LTO_CLANG ]]` case first

kpatch-build/kpatch-build
```diff
@@ -1107,6 +1142,7 @@ declare -a MAKEVARS
 if [[ -n "$CONFIG_CC_IS_CLANG" ]]; then
 	MAKEVARS+=("CC=${KPATCH_CC_PREFIX}${CLANG}")
 	MAKEVARS+=("HOSTCC=clang")
+	MAKEVARS+=("LLVM=1")
```
jpoimboe, line 1145: This is only a one-line change but it has a big impact and belongs in a separate commit with proper justification.  Also I wonder how it interacts with the CONFIG_LD_IS_LLD check.

kpatch-build/kpatch-build
```diff
@@ -1054,6 +1060,35 @@ if [[ -n "$CONFIG_DEBUG_INFO_BTF" ]]; then
...
+	# This is a heuristic: use -x to check vmlinux vs. vmlinux.o
+	[[ -x "$VMLINUX" ]] && die "For kernel with CONFIG_LTO_CLANG, please supply vmlinux.o instead of vmlinux via -v|--vmlinux option."
+
+	cp -f "$KERNEL_SRCDIR/Makefile" "$TEMPDIR/Makefile" || die
+	sed -i "s/--thinlto-cache-dir=\$(extmod_prefix).thinlto-cache/--lto-obj-path=vmlinux.o.thinlto.o/g" "$KERNEL_SRCDIR"/Makefile
+
+	cp -f "$KERNEL_SRCDIR/scripts/Makefile.build" "$TEMPDIR/Makefile.build" || die
+	sed -i "s/\$(ld_flags)/\$(ld_flags) --lto-obj-path=\$@.thinlto.o/g" "$KERNEL_SRCDIR"/scripts/Makefile.build
+
+	export KPATCH_CC_HANDLE_LTO=1
```
jpoimboe, line 1089: I don't think "HANDLE" adds anything, can we just call it KPATCH_CC_LTO?

kpatch-build/kpatch-build
```diff
@@ -1174,9 +1210,14 @@ if [[ -n "$CONFIG_MODVERSIONS" ]]; then
 	trace_on
 fi
 
+if [[ -n "$CONFIG_LTO_CLANG" ]]; then
+	DIFF_OBJS="$TEMPDIR/thinlto_objs"
+else
+	DIFF_OBJS="$TEMPDIR/changed_objs"
+fi
```
jpoimboe, line 1217: Why does lto need its own special file name?

kpatch-build/kpatch-build
```diff
@@ -1251,11 +1310,21 @@ for i in $FILES; do
 			SYMVERS_FILE="$BUILDDIR/Module.symvers"
 		fi
 
-		"$READELF" -s --wide "$KOBJFILE_PATH" > "$SYMTAB"
+		# With CONFIG_LTO_CLANG, multiple .thinlto files share a
+		# symtab file. Only generate the symtab file once.
+		[[ -e "$SYMTAB" ]] || "$READELF" -s --wide "$KOBJFILE_PATH" > "$SYMTAB"
```
jpoimboe, line 1315: For more readability use the long arg: "--symbols"

kpatch-build/create-diff-object.c
```diff
@@ -2825,6 +2825,21 @@ static void kpatch_mark_ignored_sections(struct kpatch_elf *kelf)
...
+		 * while the original function is very similar, like:
+		 * __initstub__kmod_syscall__728_5324_bpf_syscall_sysctl_init7
+		 *
+		 * We don't have very good solution for it yet. To workaround
+		 * the issue, it seems safe to ignore .init.text changes in
+		 * vmlinux.o. This will skip such new functions.
+		 */
+		if (!strncmp(childobj, "vmlinux.o", 9) &&
+		    !strncmp(sec->name, ".init.text", 10))
+			sec->ignore = 1;
```
jpoimboe, line 2842: Also this (or whatever the change eventually is) should probably be in its own commit.

kpatch-build/create-diff-object.c
```diff
@@ -2825,6 +2825,21 @@ static void kpatch_mark_ignored_sections(struct kpatch_elf *kelf)
 		    !strncmp(sec->name, ".llvm_addrsig", 13) ||
 		    !strncmp(sec->name, ".llvm.", 6))
 			sec->ignore = 1;
+
+		/*
+		 * With CONFIG_LTO_CLANG, we see some weird new functions,
+		 * such as:
+		 * __initstub__kmod_syscall__728_5326_bpf_syscall_sysctl_init7
+		 * while the original function is very similar, like:
+		 * __initstub__kmod_syscall__728_5324_bpf_syscall_sysctl_init7
```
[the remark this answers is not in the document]
liu-song-6: I played with kpatch_mangled_strcmp() a little bit, but didn't solve all the problem. Let me dig more into it.

kpatch-build/kpatch-build
```diff
@@ -179,6 +179,12 @@ cleanup() {
 	# restore original Makefile.modfinal if we updated it for the build
 	[[ -e "$TEMPDIR/Makefile.modfinal" ]] && mv -f "$TEMPDIR/Makefile.modfinal" "$KERNEL_SRCDIR/scripts"
 
+	# restore original Makefile.build if we updated it for the build
+	[[ -e "$TEMPDIR/Makefile.build" ]] && mv -f "$TEMPDIR/Makefile.build" "$KERNEL_SRCDIR/scripts"
+
+	# restore Makefile if we updated it for the build
+	[[ -e "$TEMPDIR/Makefile" ]] && mv -f "$TEMPDIR/Makefile" "$KERNEL_SRCDIR"
+
```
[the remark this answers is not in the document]
liu-song-6: Fixed.

kpatch-build/kpatch-build
```diff
@@ -1054,6 +1060,35 @@ if [[ -n "$CONFIG_DEBUG_INFO_BTF" ]]; then
...
+	#
+	# readelf -s vmlinux  | grep perf_event_bpf_event
+	#
+	# non-LTO:
+	#  124946: ffffffff81287070   858 FUNC    GLOBAL DEFAULT    1 perf_event_bpf_event
+	#
+	# LTO:
+	#  122028: ffffffff81298cd0   861 FUNC    LOCAL  HIDDEN     1 perf_event_bpf_event
+	#
+	# To work around this, use vmlinux.o instead.
```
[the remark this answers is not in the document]
liu-song-6: Sounds great! Fixed it.

kpatch-build/kpatch-build
liu-song-6: Sure. Fixed.

kpatch-build/kpatch-build
liu-song-6: It appears to me that I need to pass in LLVM=1 to the make command. Otherwise, the make will fail. I have the same change for PGO support.

kpatch-build/kpatch-build
liu-song-6: Initially, I thought we may still use changed_objs for something else. But I think we didn't use it at the moment. 

OTOH, "changed_objs" is not accurate for thin LTO, as not all these files are changed.

kpatch-build/kpatch-build
liu-song-6: Fixed.

kpatch-build/kpatch-build
liu-song-6: Yeah, updating kpatch_mangled_strcmp() solves the problem.

kpatch-build/kpatch-build
liu-song-6: It is possible for the user to specify CC=clang and LD=ld.lld separately. And LLVM=1 will enable a set of utilities. From llvm.rst:
```
LLVM has substitutes for GNU binutils utilities. They can be enabled individually.
The full list of supported make variables::
        make CC=clang LD=ld.lld AR=llvm-ar NM=llvm-nm STRIP=llvm-strip \
          OBJCOPY=llvm-objcopy OBJDUMP=llvm-objdump READELF=llvm-readelf \
          HOSTCC=clang HOSTCXX=clang++ HOSTAR=llvm-ar HOSTLD=ld.lld
To simplify the above command, Kbuild supports the ``LLVM`` variable::
        make LLVM=1
```
I changed the code to use CONFIG_AS_IS_LLVM to detect the use of LLVM=1. I think this should cover most use cases.

liu-song-6:
@jpoimboe Does this version look reasonable?

kpatch-build/kpatch-build
```diff
@@ -1118,6 +1137,11 @@ else
 	MAKEVARS+=("LD=${KPATCH_CC_PREFIX}${LD}")
 fi
 
+# It is common for the user to enable all LLVM utilities with LLVM=1.
+# Detech such case with CONFIG_AS_IS_LLVM flag.
+if [[ -n "$CONFIG_AS_IS_LLVM" ]]; then
+	MAKEVARS+=("LLVM=1")
+fi
```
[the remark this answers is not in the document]
liu-song-6: Yeah, I couldn't think of a better test for LLVM=1.

kpatch-build/kpatch-build
```diff
@@ -1054,6 +1054,25 @@ if [[ -n "$CONFIG_DEBUG_INFO_BTF" ]]; then
 	fi
 fi
 
+if [[ -n "$CONFIG_LTO_CLANG" ]]; then
+	[[ -n "$CONFIG_LTO_CLANG_THIN" ]] || die "Non-thin LTO is not supported. Please enable CONFIG_LTO_CLANG_THIN"
+
+	# With LTO, many GLOBAL symbols are converted to LOCAL in vmlinux,
+	# which confuses the klp reloc logic.  Require vmlinux.o for LTO.
+	[[ -e "$VMLINUX" ]] || die "For kernel with CONFIG_LTO_CLANG, please supply vmlinux.o via -v|--vmlinux option."
```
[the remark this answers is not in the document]
liu-song-6: Will remove.

kpatch-build/create-diff-object.c
```diff
@@ -430,6 +430,30 @@ static int __kpatch_unique_id_strcmp(char *s1, char *s2)
...
+		if (*s1 == '\0' || *s2 == '\0')
+			break;
+
+		s1++;
+		s2++;
+	}
+
+	return *s1 == *s2;
+}
+
```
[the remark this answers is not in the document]
liu-song-6: Hmm.. I am also confused with 1/0 return value. Let me double check..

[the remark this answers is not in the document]
liu-song-6: > Here's a simplified version (untested)
> 
> ```
> static int __kpatch_skip_digits_strcmp(char *s1, char *s2)
> {
> 	while (*s1 == *s2) {
> 		if (!*s1)
> 			return 0;
> 		s1++;
> 		s2++;
> 
> 		while (isdigit(*s1))
> 			s1++;
> 		while (isdigit(*s2))
> 			s2++;
> 	}
> 
> 	return 1;
> }
> ```

This is really clean. Thanks!

liu-song-6:
@jpoimboe Could you please share your comments on the latest version? Thanks in advance!

jpoimboe:
I tried out the patches against a 6.2 kernel and got the following errors:
```
Binary files orig/vmlinux.o.thinlto.o2 and patched/vmlinux.o.thinlto.o2 differ
Binary files orig/vmlinux.o.thinlto.o223 and patched/vmlinux.o.thinlto.o223 differ
create-diff-object: ERROR: vmlinux.o.thinlto.o223: kpatch_bundle_symbols: 244: symbol trace_event_raw_event_task_newtask at offset 16 within section .text.trace_event_raw_event_task_newtask, expected 0
Binary files orig/vmlinux.o.thinlto.o227 and patched/vmlinux.o.thinlto.o227 differ
create-diff-object: ERROR: vmlinux.o.thinlto.o227: kpatch_bundle_symbols: 244: symbol delayed_put_task_struct at offset 16 within section .text.delayed_put_task_struct, expected 0
Binary files orig/vmlinux.o.thinlto.o232 and patched/vmlinux.o.thinlto.o232 differ
create-diff-object: ERROR: vmlinux.o.thinlto.o232: kpatch_bundle_symbols: 244: symbol __ptrace_may_access at offset 16 within section .text.__ptrace_may_access, expected 0
Binary files orig/vmlinux.o.thinlto.o234 and patched/vmlinux.o.thinlto.o234 differ
create-diff-object: ERROR: vmlinux.o.thinlto.o234: kpatch_bundle_symbols: 244: symbol trace_event_raw_event_signal_generate at offset 16 within section .text.trace_event_raw_event_signal_generate, expected 0
Binary files orig/vmlinux.o.thinlto.o307 and patched/vmlinux.o.thinlto.o307 differ
create-diff-object: ERROR: vmlinux.o.thinlto.o307: kpatch_bundle_symbols: 244: symbol __klp_shadow_get_or_alloc at offset 16 within section .text.__klp_shadow_get_or_alloc, expected 0
Binary files orig/vmlinux.o.thinlto.o469 and patched/vmlinux.o.thinlto.o469 differ
create-diff-object: ERROR: vmlinux.o.thinlto.o469: kpatch_bundle_symbols: 244: symbol trace_event_raw_event_vm_unmapped_area at offset 16 within section .text.trace_event_raw_event_vm_unmapped_area, expected 0
Binary files orig/vmlinux.o.thinlto.o510 and patched/vmlinux.o.thinlto.o510 differ
create-diff-object: ERROR: vmlinux.o.thinlto.o510: kpatch_bundle_symbols: 244: symbol get_arg_page at offset 16 within section .text.get_arg_page, expected 0
Binary files orig/vmlinux.o.thinlto.o593 and patched/vmlinux.o.thinlto.o593 differ
create-diff-object: ERROR: vmlinux.o.thinlto.o593: kpatch_bundle_symbols: 244: symbol do_task_stat at offset 16 within section .text.do_task_stat, expected 0
Binary files orig/vmlinux.o.thinlto.o809 and patched/vmlinux.o.thinlto.o809 differ
create-diff-object: ERROR: vmlinux.o.thinlto.o809: kpatch_bundle_symbols: 244: symbol selinux_set_mnt_opts at offset 16 within section .text.selinux_set_mnt_opts, expected 0
ERROR: 9 error(s) encountered. Check /home/jpoimboe/.kpatch/build.log for more details.
```
The cause for the errors is the new CONFIG_CALL_DEPTH_TRACKING feature which enables CONFIG_PREFIX_SYMBOLS, which relies on objtool to create a `__pfx_<func>` symbol for the 16 bytes of NOPS preceding each function.

The problem is that objtool didn't run on the vmlinux.o.thinlto.o* objects.

Objtool has become an integral part of the x86 toolchain and removing it from patch module objects could create all kinds of surprises.  So I'm thinking this approach may not work and we may have to add vmlinux.o support to CDO after all.

jpoimboe:
> How should we enable CDOing vmlinux.o? @jpoimboe IIRC, you plan to work on this soon?

I could certainly do it, though I'm probably not free for at least a few weeks.

liu-song-6:
> > How should we enable CDOing vmlinux.o? @jpoimboe IIRC, you plan to work on this soon?
> 
> I could certainly do it, though I'm probably not free for at least a few weeks.

At the moment, I am hoping this version will work for our 5.19 based kernel. If it doesn't work, we would probably need CDOing vmlinux.o soon..

jpoimboe:
> At the moment, I am hoping this version will work for our 5.19 based kernel.

If you rely on the ORC unwinder, static calls or return thunks then it may have problems...

liu-song-6:
> > At the moment, I am hoping this version will work for our 5.19 based kernel.
> 
> If you rely on the ORC unwinder, static calls or return thunks then it may have problems...

Hmm... we do use most of these. Let me see how bad it would be.. Maybe we can run objtool on these thinlto files before CDOing?

jpoimboe:
> > > At the moment, I am hoping this version will work for our 5.19 based kernel.
> > 
> > 
> > If you rely on the ORC unwinder, static calls or return thunks then it may have problems...
> 
> Hmm... we do use most of these. Let me see how bad it would be.. Maybe we can run objtool on these thinlto files before CDOing?

Yes, maybe (but to be clear I'd rather avoid this type of Makefile hacking for kpatch upstream)

liu-song-6:
> Yes, maybe (but to be clear I'd rather avoid this type of Makefile hacking for kpatch upstream)

Agreed. CDOing vmlinux.o is a much cleaner approach. 

We already use some non-upstream kpatch commits for PGO. If this hack works for short term, we will do the same.

liu-song-6:
I am testing this change that run objtool on thinlto files: https://github.com/liu-song-6/kpatch/commit/1203c9fc4901c8a53074a36d55166d4cdd037109 . 

I understand we don't want this in upstream, but just in case someone would like to test/use it before we can CDO vmlinux.o.

PR #1327, opened 2023-02-20
kpatch-build: add KERNELRELEASE to MAKEVARS

zhaozhengyu96:
If the ${KERNEL_SRCDIR} is a git repository, `scripts/setlocalversion` will append a plus sign to the kpatch module version. This will cause a version mismatch when installing this patch module:

    invalid module version ${kernel-release} for kernel ${kernel-release}+

Use $ARCHVERSION instead.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -1117,7 +1117,8 @@ if [[ -n "$CONFIG_LD_IS_LLD" ]]; then
 else
 	MAKEVARS+=("LD=${KPATCH_CC_PREFIX}${LD}")
 fi
-
+#Force the kpatch module version to be consistent with ${ARCHVERSION}
+MAKEVARS+=("KERNELRELEASE=${ARCHVERSION}")
 
 # $TARGETS used as list, no quotes.
 # shellcheck disable=SC2086


jpoimboe:
Yeah, I guess this happens when CONFIG_DEBUG_INFO_BTF is set, there are hacks in kpatch-build which modify files in the git tree.

jpoimboe:
Can you confirm you have CONFIG_DEBUG_INFO_BTF set?  If so I'm wondering if we can solve the pahole problem a different way, by for example refusing to build with CONFIG_DEBUG_INFO_BTF if pahole is below a certain version.

zhaozhengyu96:
Thanks for the reminder.
I checked my config file and find that CONFIG_DEBUG_INFO_BTF is not set. I can not find the kpatch-build BPF hacks so I do not know how `kpatch-build` modify files in the git tree. If it does not commit the changes, it would result in `--dirty` according to scm_version(). 

Actually I suspect that the `+` is from the building of patched source, which will fail for `git describe --exact-match` in scm_version(). I haven't understood where the version of $MODNAME.ko comes from. If my guess is correct, any  kernel source directory using git will have this `+` and  I have three ways to to solve this problem:
1. `rm -rf .git/`
2. Using this patch. Version from original vmlinux seems like a better choice?
3. Introduce an additional parameter for $MAKEVARS. I can use kpatch-build like:
   kpatch-build -m "LOCALVERSION=XXX KERNELRELEASE=XXX"

Hope there is a better way！

jpoimboe:
> Thanks for the reminder. I checked my config file and find that CONFIG_DEBUG_INFO_BTF is not set. I can not find the kpatch-build BPF hacks so I do not know how `kpatch-build` modify files in the git tree. If it does not commit the changes, it would result in `--dirty` according to scm_version().
> 
> Actually I suspect that the `+` is from the building of patched source, which will fail for `git describe --exact-match` in scm_version().

kpatch-build has `./scripts/setlocalversion --save-scmversion`, which locks in the version so that applying the patch shouldn't add a `+`.  And that still seems to be working as I can't recreate your issue.

BTW, here's how to check the module version:
```
modinfo livepatch-shadow-newpid.ko |grep vermagic
vermagic:       6.2.0 SMP preempt mod_unload 
```
It comes from `include/config/kernel.release` which is auto-generated on every build of the kernel.

zhaozhengyu96:
Finally, I found out that my git repository was using a bad tag. I would fail on "git describe --exact-match" even if I use "git reset --hard target_tag".  I think I should fix the repository instead of modifying kpatch-build.

THANKS!!!

PR #1329, opened 2023-02-23
kpatch-syscall: update syscall macros for s390

sumanthkorikkar:
Since linux 'commit 2213d44e140f ("s390/syscalls: get rid of system call alias functions")', s390 syscall wrappers are modified. Adjust it accordingly for kpatch

### DIFF
--- a/kmod/patch/kpatch-syscall.h
+++ b/kmod/patch/kpatch-syscall.h
@@ -82,6 +82,39 @@
 #elif defined(CONFIG_S390)
 
 /* arch/s390/include/asm/syscall_wrapper.h versions */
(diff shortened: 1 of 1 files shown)


jpoimboe (approved):

joe-lawrence:
Since 2213d44e140f is destined for v6.3 ... could we merge this after I finish up v6.2 support in #1322 .  IIRC there were some updates to the syscall macros in that set as well.

jpoimboe:
@sumanthkorikkar #1322 has been merged so you can rebase now.  Thanks!

sumanthkorikkar:
rebased.

joe-lawrence (approved):

Issue #1328, opened 2023-02-21
ubsan for kpatch


sumanthkorikkar:
Hi all,

This is regarding ubsan support for kpatch. ubuntu recently queried regarding this option on s390.
I quickly tested it on x86 and it seems to fail on x86 in a similar way.

I did perform initial preliminary analysis. These are my current findings.
Debug data:
2a. buggy program:
```
#include <stdio.h>
int main(int argc, char **argv) {
        int arr[100];
        arr[101] = 1;
        printf("arr[101] = %d", arr[101]);
        return 0;
}
```

2b. Disassembly and relocs:
```
  1a:   50 10 b0 ac             st      %r1,172(%r11)
        int arr[100];
        arr[101] = 1;
  1e:   a7 39 00 65             lghi    %r3,101
  22:   c0 20 00 00 00 00       larl    %r2,22 <main+0x22>
                        24: R_390_PC32DBL       .data..Lubsan_data1+0x2
  28:   c0 e5 00 00 00 00       brasl   %r14,28 <main+0x28>
                        2a: R_390_PLT32DBL      __ubsan_handle_out_of_bounds+0x2
```

2c. .data.*Lubsan represents struct out_of_bounds_data, which is declared in linux/lib/ubsan.h
```
struct out_of_bounds_data {
      struct source_location location;
      struct type_descriptor *array_type;
      struct type_descriptor *index_type;
};

 struct source_location {
      const char *file_name;
      union {
            unsigned long reported;
            struct {
                  u32 line;
                  u32 column;
            };
      };
};
```

2d. Disassembly of section .data..Lubsan_data1:
```
0000000000000000 <.data..Lubsan_data1>:
       
                        0: R_390_64     .rodata      <=== source_location.location->file_name
   8:   00 00 00 04             .long   0x00000004   <=== source_location.location->line
   c:   00 00 00 05             .long   0x00000005   <=== source_location.location->column  
       
                        10: R_390_64    .data..Lubsan_type0   <== source_location->array_type
                        18: R_390_64    .data..Lubsan_type1   <=== source_location->index_type
```

2e. readelf -p .rodata sample.o
```
String dump of section '.rodata':
  [     0]  sample.c
  [     a]  arr[101] = %d
```
 
2f. readelf -p .data..Lubsan_type0 sample.o
```
String dump of section '.data..Lubsan_type0':
[     4]  'int [100]'
```
 
2g. readelf -p .data..Lubsan_type1 sample.o :
```
[...]

PR #1332, opened 2023-03-01
support ubsan for kpatch

sumanthkorikkar:
ubsan generates .data..Lubsan_data* sections as follows:

```
1. int main(int argc, char **argv) {
    int arr[100]; 
    arr[101] = 1; 
    printf("arr[101] = %d", arr[101]); 
    return 0; 
}
```
reloc:
```
2. 1a:   50 10 b0 ac          st      %r1,172(%r11)
      int arr[100];
      arr[101] = 1;
1e:   a7 39 00 65             lghi    %r3,101
22:   c0 20 00 00 00 00       larl    %r2,22 <main+0x22>
            24: R_390_PC32DBL       .data..Lubsan_data1+0x2
28:   c0 e5 00 00 00 00       brasl   %r14,28 <main+0x28>
            2a: R_390_PLT32DBL      __ubsan_handle_out_of_bounds+0x2
```

3. ubsan data section info:
``` 
0000000000000000 <.data..Lubsan_data1>:
0: R_390_64     .rodata              <=== source_location.location->file_name
8: 00 00 00 04  .long   0x00000004   <=== source_location.location->line
c: 00 00 00 05  .long   0x00000005   <=== source_location.location->column

10: R_390_64    .data..Lubsan_type0   <== source_location->array_type
18: R_390_64    .data..Lubsan_type1   <=== source_location->index_type```
```

4. Avoid correlating the *.data.Lubsan* sections. This means included function points to new *.data.Lubsan* sections.
 
5. Misc:
 * .data..Lubsan_type section does not have symbol, hence skip this case in kpatch_replace_sections_syms()
 * skip data.ubsan* from lookup table, as this in not present in vmlinux.symtab. 

6. Fixes the issue reported in https://github.com/dynup/kpatch/issues/1328

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1036,6 +1036,9 @@ static void kpatch_correlate_sections(struct list_head *seclist_orig,
 			    sec_patched->twin)
 				continue;
 
+			if (is_ubsan_sec(sec_orig->name))
+				continue;
+
 			if (is_special_static(is_rela_section(sec_orig) ?
 					      sec_orig->base->secsym :
 					      sec_orig->secsym))
@@ -1070,6 +1073,9 @@ static void kpatch_correlate_symbols(struct list_head *symlist_orig,
 		list_for_each_entry(sym_patched, symlist_patched, list) {
 			if (kpatch_mangled_strcmp(sym_orig->name, sym_patched->name) ||
 			    sym_orig->type != sym_patched->type || sym_patched->twin)
+				continue;
+
+			if (is_ubsan_sec(sym_orig->name))
 				continue;
 
 			if (is_special_static(sym_orig))
@@ -1545,6 +1551,13 @@ static void kpatch_replace_sections_syms(struct kpatch_elf *kelf)
 		list_for_each_entry(rela, &relasec->relas, list) {
 
 			if (rela->sym->type != STT_SECTION || !rela->sym->sec)
+				continue;
+
+			/*
+			 * UBSAN data will be taken wholesale, no need to
+			 * replace section symbols.
+			 */
+			if (is_ubsan_sec(rela->sym->name))
 				continue;
 
 			/*
--- a/kpatch-build/kpatch-elf.c
+++ b/kpatch-build/kpatch-elf.c
@@ -587,6 +587,18 @@ bool is_local_sym(struct symbol *sym)
 	return sym->bind == STB_LOCAL;
 }
 
+bool is_ubsan_sec(const char *name) {
+	if (!strncmp(name, ".data.rel.local..Lubsan_data", 28) ||
+		!strncmp(name, ".data..Lubsan_type", 18) ||
+		!strncmp(name, ".Lubsan_data", 12) ||
+		!strncmp(name, ".data..Lubsan_data", 18) ||
+		!strncmp(name, ".rela.data..Lubsan_data", 23) ||
+		!strncmp(name, ".rela.data.rel.local..Lubsan_data", 33))
+		return true;
+	else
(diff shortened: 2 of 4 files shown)


kpatch-build/create-diff-object.c
```diff
@@ -1555,7 +1561,8 @@ static void kpatch_replace_sections_syms(struct kpatch_elf *kelf)
 			    !strcmp(rela->sym->name, ".fixup") ||
 			    !strcmp(rela->sym->name, ".altinstr_replacement") ||
 			    !strcmp(rela->sym->name, ".altinstr_aux") ||
-			    !strcmp(rela->sym->name, ".text..refcount"))
+			    !strcmp(rela->sym->name, ".text..refcount") ||
+			    !strncmp(rela->sym->name, ".data..Lubsan_type", 18))
 				continue;
```
jpoimboe, line 1566: When building a sample patch I got the following error
```
create-diff-object: ERROR: array.o: kpatch_replace_sections_syms: 1716: .text.do_task_stat+0xe0a: can't find replacement symbol for .data..Lubsan_data132+0 reference
```
so here I think we need something like this instead
```
			/*
			 * UBSAN data will be taken wholesale, no need to
			 * replace section symbols.
			 */
			if (is_ubsan_data(rela->sym->name))
				continue;

```

sumanthkorikkar: When I looked at the kelf->symbols list,
sym->sec->sym->name: .Lubsan_data151 sym->sec->name: .data.rel.local..Lubsan_data151 sym->name: .data.rel.local..Lubsan_data151

Relocation section '.rela.text.meminfo_proc_show' at offset 0x1e18 contains 154 entries:
0000000000000ed6  0000004f0000001a R_390_GOTENT           0000000000000000 .Lubsan_data150 + 2
0000000000000edc  0000009600000014 R_390_PLT32DBL         0000000000000000 __ubsan_handle_out_of_bounds + 2
**0000000000000ee6  0000004e0000001a R_390_GOTENT           0000000000000000 .Lubsan_data151 + 2  <----**
0000000000000eec  0000009600000014 R_390_PLT32DBL         0000000000000000 __ubsan_handle_out_of_bounds + 2

The above rela entry **.Lubsan_data151** already had section symbol, so it wasnt replaced in kpatch_replace_sections_syms.

But I assume if the compiler generated rela entry such that symbol name is .data.rel.local..Lubsan_data151 
in .rela.text.meminfo_proc_show, then it should be replaced to .Lubsan_data151+offset in .rela.text.meminfo_proc_show. 

So, to be sure, I just added only .data..Lubsan_type case in kpatch_replace_sections_syms().
I will also add .data..Lubsan_data in addition to that. As this was missed  as you reported above.

In any case, if you think, we should cover all the case, I can add is_ubsan_sec() in kpatch_replace_sections_syms().

kpatch-build/kpatch-elf.c
```diff
@@ -587,6 +587,18 @@ bool is_local_sym(struct symbol *sym)
 	return sym->bind == STB_LOCAL;
 }
 
+bool is_ubsan_data(const char *name) {
```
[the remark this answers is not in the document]
sumanthkorikkar: Added

kpatch-build/create-diff-object.c
sumanthkorikkar: Added the changes you mentioned, Agree to what you mentioned above. Thanks

jpoimboe (approved):

joe-lawrence (approved):

joe-lawrence:
Thanks, @sumanthkorikkar

PR #1336, opened 2023-03-13
kpatch-build: Fix setlocalversion issue with 6.3 kernel

### LINKED ISSUE #1335: 6.3 failure with setlocalversion
6.3-based kernel is failing with
```
$ kpatch-build/kpatch-build -s ~/git/linux test/integration/linux-5.18.0/shadow-newpid.patch
Using source directory at /home/jpoimboe/git/linux
Testing patch file(s)
Reading special section data
Building original source
Usage: ./scripts/setlocalversion [srctree]
ERROR: kpatch build failed. Check /home/jpoimboe/.kpatch/build.log for more details.
```
The setlocalversion script no longer supports the `--save-scmversion` hack so we'll have to come up with another hack.  Replace setlocalversion with a version which does what we want?

jpoimboe:
kpatch-build: Fix setlocalversion issue with 6.3 kernel
    
The kernel has a VERMAGIC_STRING, e.g. "6.2.0".  The module loader uses that string to ensure that all loaded modules' version strings match the kernel's.

If the kernel source is in a git tree, and if there are uncommitted changes, the version string will have a '+' or "-dirty" appended to it, like "6.1.0+" or "6.2.0-dirty".  This dirty tree detection is done by the setlocalversion script.

This affects kpatch-build in a few ways.  When it builds the original kernel, in some cases there are uncommitted changes to the makefiles.  When it builds the patched kernel, there are additional uncommitted changes due to the .patch file being applied.

We want to avoid the VERMAGIC_STRING changing between builds.  Otherwise it would cause problems:

- object code which uses that string would change unnecessarily, causing a false positive change detected by create-diff-object

- the linked patch module would report the wrong version, resulting in the module failing to load due to version mismatch.
    
Up until now, the version was prevented from changing by running `setlocalversion --save-scmversion` before the build.  That command hard-codes the version by saving it to a file which is then automatically read later during future invocations of the kernel build.

Unfortunately that feature was removed in the 6.3 merge window with commit f6e09b07cc12 ("kbuild: do not put .scmversion into the source tarball").  So we need to come up with a new approach.

Fix it by temporarily replacing the setlocalversion script with a one-liner which just echo's the original version.  I think this is unfortunately the best we can do, as we really can't handle VERMAGIC_STRING changing, either during/between kernel builds or during the module link.

Fixes #1335.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -166,18 +166,13 @@ remove_patches() {
 }
 
 cleanup() {
-	rm -f "$BUILDDIR/.scmversion"
-
 	remove_patches
 
-	# restore original vmlinux if it was overwritten by sourcedir build
+	# restore any files that were modified for the build
 	[[ -e "$TEMPDIR/vmlinux" ]] && mv -f "$TEMPDIR/vmlinux" "$KERNEL_SRCDIR/"
-
-	# restore original link-vmlinux.sh if we updated it for the build
 	[[ -e "$TEMPDIR/link-vmlinux.sh" ]] && mv -f "$TEMPDIR/link-vmlinux.sh" "$KERNEL_SRCDIR/scripts"
-
-	# restore original Makefile.modfinal if we updated it for the build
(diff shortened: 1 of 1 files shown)


joe-lawrence (approved):
(Assuming internal integration likes this on older kernels.)

PR #1338, opened 2023-03-30
Fix shellcheck error

iwamatsu:
Fix the following ShellCheck error.
  
      - SC2317 (info): Command appears to be unreachable. Check usage (or ignore if
        invoked indirectly).
      - SC2086 (info): Double quote to prevent globbing and word splitting.
      - SC2004 (style): $/${} is unnecessary on arithmetic variables.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -909,7 +909,7 @@ else
 		echo "Unpacking kernel source"
 
 		rpm -D "_topdir $RPMTOPDIR" -ivh "$SRCRPM" 2>&1 | logger || die
-		rpmbuild -D "_topdir $RPMTOPDIR" -bp --nodeps "--target=$(uname -m)" "$RPMTOPDIR"/SPECS/kernel$ALT.spec 2>&1 | logger ||
+		rpmbuild -D "_topdir $RPMTOPDIR" -bp --nodeps "--target=$(uname -m)" "$RPMTOPDIR/SPECS/kernel$ALT.spec" 2>&1 | logger ||
 			die "rpmbuild -bp failed.  you may need to run 'yum-builddep kernel' first."
 
 		if [[ "$DISTRO" = openEuler ]]; then
@@ -1271,7 +1271,7 @@ for i in $FILES; do
 
 		# create-diff-object orig.o patched.o parent-name parent-symtab
 		#		     Module.symvers patch-mod-name output.o
-		"$TOOLSDIR"/create-diff-object $CDO_FLAGS "orig/$i" "patched/$i" "$KOBJFILE_NAME" \
+		"$TOOLSDIR"/create-diff-object "$CDO_FLAGS" "orig/$i" "patched/$i" "$KOBJFILE_NAME" \
 			"$SYMTAB" "$SYMVERS_FILE" "${MODNAME//-/_}" \
 			"output/$i" 2>&1 | logger 1
 		check_pipe_status create-diff-object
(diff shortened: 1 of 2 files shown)


kpatch-build/kpatch-build
```diff
@@ -1271,7 +1271,7 @@ for i in $FILES; do
 
 		# create-diff-object orig.o patched.o parent-name parent-symtab
 		#		     Module.symvers patch-mod-name output.o
-		"$TOOLSDIR"/create-diff-object $CDO_FLAGS "orig/$i" "patched/$i" "$KOBJFILE_NAME" \
+		"$TOOLSDIR"/create-diff-object "$CDO_FLAGS" "orig/$i" "patched/$i" "$KOBJFILE_NAME" \
```
joe-lawrence, line 1274: Something to watch out for,  quoting a string that contains multiple arguments will consolidate them into a single argument.  For example:
```bash
#!/bin/bash

foo()
{
    [[ -n "$1" ]] && echo "arg1: $1"
    [[ -n "$2" ]] && echo "arg2: $2"
    [[ -n "$3" ]] && echo "arg3: $3"
}

args="--one --two --three"

echo "foo \$args"
foo $args
echo "foo \"\$args\""
foo "$args"
```
Results in two different behaviors:
``` bash
$ ./test.sh 
foo $args
arg1: --one
arg2: --two
arg3: --three
foo "$args"
arg1: --one --two --three
```
So it might be more future proof if **$CDO_FLAGS** was an array.

Also, there is a real bug when **$CDO_FLAGS** is empty.  A slightly modified repro:
``` bash
#!/bin/bash

foo()
{
    echo "arg1: $1"
    echo "arg2: $2"
}

args=""

echo "foo \"\$args\" \"next-option\""
foo "$args" "next-option"

echo "foo \$args \"next-option\""
foo $args "next-option"
```
Results again in two different behaviors:
``` bash
$ ./test.sh 
foo "$args" "next-option"
arg1: 
arg2: next-option
foo $args "next-option"
arg1: next-option
arg2:
```
The second problem breaks builds.

Modifying to store and pass though an array should solve both problems.

kpatch/kpatch
```diff
@@ -355,7 +355,6 @@ load_module () {
 			i=$((i+1))
 			if [[ $i -eq $MAX_LOAD_ATTEMPTS ]]; then
 				die "failed to load module $module"
-				break
```
joe-lawrence, line 358: Heh, when you're dead you can't break :skull:

kpatch-build/kpatch-build
```diff
@@ -1357,7 +1357,7 @@ if [[ "$USE_KLP" -eq 1 ]]; then
 		extra_flags="--no-klp-arch-sections"
 	fi
 	cp -f "$TEMPDIR/patch/$MODNAME.ko" "$TEMPDIR/patch/tmp.ko" || die
-	"$TOOLSDIR"/create-klp-module $extra_flags "$TEMPDIR/patch/tmp.ko" "$TEMPDIR/patch/$MODNAME.ko" 2>&1 | logger 1
+	"$TOOLSDIR"/create-klp-module "$extra_flags" "$TEMPDIR/patch/tmp.ko" "$TEMPDIR/patch/$MODNAME.ko" 2>&1 | logger 1
```
joe-lawrence, line 1360: Same comments apply for the **$extra_flags**

joe-lawrence (changes requested):
Hi @iwamatsu, thanks for posting.  These look like shellcheck complaints new to v0.9.0, so always interesting to see what new advice it's come up with.  See my comments below regarding the two variables that hold command line flags.  To appease shellcheck, a little more elaborate fixup will be needed to handle them safely.

PR #1339, opened 2023-04-14
kpatch-build: support Photon OS distribution

bhllamoreaux:
VMware's Photon OS uses kpatch to build and load kernel livepatches, starting with 3.0.
Ref: https://github.com/vmware/photon/tree/master/SPECS/kpatch

It would be nice to have support for Photon OS in kpatch-build, so that kpatch-build works for
Photon out of the box. Ideally, this will work the same as for other supported distros,  where
kpatch-build can be invoked just as:
     `kpatch-build <name>.patch.`

This commit handles setup of all kernel sources, vmlinux files, etc for Photon OS.

cc @sshedi

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -831,15 +831,6 @@ fi
 
 [[ "$SKIPCLEANUP" -eq 0 ]] && trap cleanup EXIT INT TERM HUP
 
-KVER="${ARCHVERSION%%-*}"
-if [[ "$ARCHVERSION" =~ - ]]; then
-	KREL="${ARCHVERSION##*-}"
-	KREL="${KREL%.*}"
-fi
-[[ "$ARCHVERSION" =~ .el7a. ]] && ALT="-alt"
-
-[[ -z "$TARGETS" ]] && TARGETS="vmlinux modules"
-
 # Don't check external file.
 # shellcheck disable=SC1090
 if [[ -z "$USERSRCDIR" ]] && [[ -f "$RELEASE_FILE" ]]; then
@@ -847,7 +838,37 @@ if [[ -z "$USERSRCDIR" ]] && [[ -f "$RELEASE_FILE" ]]; then
 	DISTRO="$ID"
 fi
 
-if [[ "$DISTRO" = fedora ]] || [[ "$DISTRO" = rhel ]] || [[ "$DISTRO" = ol ]] || [[ "$DISTRO" = centos ]] || [[ "$DISTRO" = openEuler ]]; then
+KVER="${ARCHVERSION%%-*}"
+if [[ "$ARCHVERSION" =~ - ]]; then
+	# handle flavor extension on Photon ex) -rt, -esx
+	if [[ "$DISTRO" = photon ]]; then
+		KREL="${ARCHVERSION#*-}"
+
+		# strip rt patchset version if present
+		# remove trailing -flavor and starting -rt### patchset
+		KREL="${KREL%-*}"
+		KREL="${KREL#*-}"
+
+		PH_TAG="${ARCHVERSION##*.}"
+		PH_FLAVOR="${PH_TAG##*-}"
+		PH_TAG="${PH_TAG%%-*}"
+
+		# if no flavor, these will be the same
+		[[ "$PH_FLAVOR" = "$PH_TAG" ]] && PH_FLAVOR=""
+	else
+		KREL="${ARCHVERSION##*-}"
+	fi
+	KREL="${KREL%.*}"
+fi
+
+[[ "$ARCHVERSION" =~ .el7a. ]] && ALT="-alt"
+
+[[ -z "$TARGETS" ]] && TARGETS="vmlinux modules"
+
+if [[ "$DISTRO" = fedora ]] || [[ "$DISTRO" = rhel ]] || [[ "$DISTRO" = ol ]] ||
+   [[ "$DISTRO" = centos ]] || [[ "$DISTRO" = openEuler ]] ||
+   [[ "$DISTRO" = photon ]]; then
+
 	[[ -z "$VMLINUX" ]] && VMLINUX="/usr/lib/debug/lib/modules/$ARCHVERSION/vmlinux"
 	[[ -e "$VMLINUX" ]] || die "kernel-debuginfo-$ARCHVERSION not installed"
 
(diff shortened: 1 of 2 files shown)


jpoimboe:
@bhllamoreaux thanks for the contribution!

The code is doing several of the same things the Fedora/RHEL/OL/CentOS/openEuler code is doing, yet in an a slightly different way, making it harder to review.  Since Photon OS is RPM-based, I would much rather try to converge it with the `if [[ "$DISTRO" = fedora ]]...` code.

I know there are some inherent differences between the distros, but even if there have to be several Photon OS checks within, that would still make it much easier to review for correctness, i.e. that all the right variables are being set and that the printouts are consistent between distributions.  And it would also make it easier to spot the commonalities and differences between the distros if we want to refactor the code later.

bhllamoreaux:
Thanks for the quick response! That sounds reasonable, I'll give it a shot.

bhllamoreaux:
Updated the changes to be integrated into the existing code, instead of in a Photon specific section, as suggested.

kpatch-build/kpatch-build
```diff
@@ -885,40 +903,74 @@ elif [[ -e "$KERNEL_SRCDIR"/.config ]] && [[ -e "$VERSIONFILE" ]] && [[ "$(cat "
...
+
+				PHOTON_VERSION="${PH_TAG//[^0-9]/}".0
+				wget -P "$TEMPDIR" "https://packages.vmware.com/photon/$PHOTON_VERSION/photon_srpms_${PHOTON_VERSION}_x86_64/$SRC_RPM_NAME" 2>&1 | logger || die
+				SRCRPM="$TEMPDIR/$SRC_RPM_NAME"
 			else
 				command -v yumdownloader &>/dev/null || die "yumdownloader (yum-utils or dnf-utils) not installed"
 				yumdownloader --source --destdir "$TEMPDIR" "kernel$ALT-$KVER-$KREL" 2>&1 | logger || die
+				SRCRPM="$TEMPDIR/kernel$ALT-$KVER-$KREL.src.rpm"
 			fi
-			SRCRPM="$TEMPDIR/kernel$ALT-$KVER-$KREL.src.rpm"
```
sshedi, line 906:
```
if [ -z "$SRCRPM" ]; then
  SRCRPM="$TEMPDIR/kernel$ALT-$KVER-$KREL.src.rpm"
fi
```
And remove it from line 921 & 935

kpatch-build/kpatch-build
```diff
@@ -885,40 +903,74 @@ elif [[ -e "$KERNEL_SRCDIR"/.config ]] && [[ -e "$VERSIONFILE" ]] && [[ "$(cat "
...
+				SRCRPM="$TEMPDIR/kernel$ALT-$KVER-$KREL.src.rpm"
+			elif [[ "$DISTRO" = photon ]]; then
+				if [[ -n "$PH_FLAVOR" ]]; then
+					SRC_RPM_NAME="linux-$PH_FLAVOR-$KVER-$KREL.$PH_TAG.src.rpm"
+				else
+					SRC_RPM_NAME="linux-${ARCHVERSION}.src.rpm"
+				fi
+
+				PHOTON_VERSION="${PH_TAG//[^0-9]/}".0
+				wget -P "$TEMPDIR" "https://packages.vmware.com/photon/$PHOTON_VERSION/photon_srpms_${PHOTON_VERSION}_x86_64/$SRC_RPM_NAME" 2>&1 | logger || die
```
sshedi, line 930: _x86_64 --> $(uname -m)
we might support aarch64 someday

test/integration/lib.sh
```diff
@@ -102,6 +102,24 @@ kpatch_openEuler_dependencies()
...
+kpatch_photon_dependencies()
+{
+	local flavor
+	flavor=$(uname -r | cut -d "-" -f 3)
+	if [[ $(uname -r | cut -d "-" -f 2) == "rt" ]]; then
+		flavor="rt"
+	fi
+
+	tdnf install -y kmod bash rpm-build coreutils gawk util-linux binutils sed findutils \
+		make gcc elfutils elfutils-devel systemd-rpm-macros build-essential
```
sshedi, line 114: nit:
`build-essentail` implicitly brings in `gcc, make, gawk, binutils`
`elfutils-devel` brings in `elfutils` so we can remove the redundant entries from here

sshedi (changes requested):
Good work @bhllamoreaux. Some minor tweaks needed.

bhllamoreaux:
Updated with suggested tweaks

sshedi:
Hi @jpoimboe can you please take a look again. Thanks.

kpatch-build/kpatch-build
```diff
@@ -831,23 +831,41 @@ fi
...
 
-# Don't check external file.
-# shellcheck disable=SC1090
-if [[ -z "$USERSRCDIR" ]] && [[ -f "$RELEASE_FILE" ]]; then
-	source "$RELEASE_FILE"
-	DISTRO="$ID"
-fi
-
-if [[ "$DISTRO" = fedora ]] || [[ "$DISTRO" = rhel ]] || [[ "$DISTRO" = ol ]] || [[ "$DISTRO" = centos ]] || [[ "$DISTRO" = openEuler ]]; then
+if [[ "$DISTRO" = fedora ]] || [[ "$DISTRO" = rhel ]] || [[ "$DISTRO" = ol ]] || [[ "$DISTRO" = centos ]] || [[ "$DISTRO" = openEuler ]] || [[ "$DISTRO" = photon ]]; then
```
jpoimboe, line 868: Can we split this in two lines?
```
if [[ "$DISTRO" = fedora ]] || [[ "$DISTRO" = rhel ]] || [[ "$DISTRO" = ol ]] ||
   [[ "$DISTRO" = centos ]] || [[ "$DISTRO" = openEuler ]] ||
   [[ "$DISTRO" = photon ]]; then

	[[ -z "$VMLINUX" ]] && VMLINUX="/usr/lib/debug/lib/modules/$ARCHVERSION/vmlinux"
```

kpatch-build/kpatch-build
```diff
@@ -885,51 +903,97 @@ elif [[ -e "$KERNEL_SRCDIR"/.config ]] && [[ -e "$VERSIONFILE" ]] && [[ "$(cat "
...
 		if [[ "$DISTRO" = openEuler ]]; then
 			[[ -z "$CONFIGFILE" ]] && CONFIGFILE="/boot/config-${ARCHVERSION}"
+		elif [[ "$DISTRO" = photon ]]; then
+			[[ -z "$CONFIGFILE" ]] && CONFIGFILE="$KERNEL_SRCDIR/$SRC_CFG"
+
+			# modify config file here to get the right vermagic, as Photon does not always listen to localversion file
+			if [[ -z "$PH_FLAVOR" ]]; then
+				sed -i s/^CONFIG_LOCALVERSION=\".*\"/CONFIG_LOCALVERSION=\"-"$KREL"."$PH_TAG"\"/g "$CONFIGFILE"
+			else
+				sed -i s/^CONFIG_LOCALVERSION=\".*\"/CONFIG_LOCALVERSION=\"-"$KREL"."$PH_TAG"-"$PH_FLAVOR"\"/g "$CONFIGFILE"
```
jpoimboe, line 994: These should have `|| die`

jpoimboe (commented):
Looking good!  I added a few minor comments

bhllamoreaux:
Apologies for the delay, I was distracted with some other work for a bit. Hopefully the latest revision looks good

jpoimboe (approved):

sshedi:
@jpoimboe can we merge this PR? Been pending from a while now. Thanks.

jpoimboe:
@joe-lawrence does this look ok to you?

joe-lawrence (approved):

joe-lawrence:
> @joe-lawrence does this look ok to you?

:+1:

sshedi:
Thanks a lot for your time and inputs @joe-lawrence and @jpoimboe.

Have a nice time ahead.

bhllamoreaux:
^^ Thanks everyone @jpoimboe @sshedi @joe-lawrence

PR #1342, opened 2023-04-26
INSTALL.md: update OpenEuler prereqs formatting

ryanbsull:
shrinks the size of the sub-section header below  
the OpenEuler section, reducing confusion as to  
what section that is a part of, as it is only  
necessary for OpenEuler machines

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -203,7 +203,7 @@ Install the dependencies for compiling kpatch and running kpatch-build:
 make dependencies
 ```
 
-Before running kpatch-build, two more things need to be checked:
+#### Before running kpatch-build, two more things need to be checked:
 -------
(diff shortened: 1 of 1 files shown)


joe-lawrence:
Thanks for fixing!  The new markdown renders the instructions in a better fashion now that the OpenEuler "Before running kpatch-build..." is a sub-heading.

Onto the nits, about the commit msg:
1. Since this we're a kernel-related open source project, we add Signed-off-by lines to commits
2. While you're at it, s/shrinks/Shrink and end with period.  Also, word wrap at 72 is about what most kernel commits do (except in cases of command or log quoting).

ryanbsull:
@joe-lawrence Fixed the commit formatting, let me know if there are any other issues, thanks.

joe-lawrence:
Thanks looks good, one last nit: "shrinks" should be "Shrink".

joe-lawrence (approved):

PR #1345, opened 2023-05-16
lookup: local_match can be empty

lzw3232:
symtab may have two STT_FILE symbols with the same name, but neither of them has a local symbol.
we can choose either one,
because the lookup_table_file_sym is used to match the local symbol.

for example:
in x86:
arch/x86/mm/kaslr.c have two local symbols (can modify the code to have no local symbols)
arch/x86/lib/kaslr.c have no local symbols

if we make the patch for arch/x86/lib/kaslr.c  or arch/x86/mm/kaslr.c, it will be error.

PR #1346, opened 2023-05-31
lookup: skip duplicate local symbol table check for files without locals

jpoimboe:
If the file doesn't have local object/func symbols, any empty match will do, and duplicate matching local symbol lists aren't a problem.

Fixes #1345.

Reported-by: lzwycc <lzw32321226@163.com>


jpoimboe:
v2:
- add maybe_discarded_sym() check

Issue #1347, opened 2023-06-22
x86 NOP padded functions without __pfx_ symbol


joe-lawrence:
Reported by @ryanbsull in v6.3 integration tests:
```
create-diff-object: ERROR: meminfo.o: kpatch_bundle_symbols: 251: symbol meminfo_proc_show at offset 16 within section .text.meminfo_proc_show, expected 0
```
I thought I had added support for the x86 NOP padding in 3e54c63b ("create-diff-object: support x86 NOP-padded functions") however we missed a configuration combination that does not create prefix symbols:
``` bash
$ git describe HEAD
v6.4-rc7-72-gdad9774deaf1

# test 1) Initial default config turns on Indirect Branch Tracking and
#         16-NOP padding bytes
$ make defconfig
$ grep -e CONFIG_X86_KERNEL_IBT -e CONFIG_FUNCTION_PADDING_BYTES .config
CONFIG_X86_KERNEL_IBT=y
CONFIG_FUNCTION_PADDING_BYTES=16

# test 1) Generates no "__pfx_" symbols
$ make -j$(nproc) fs/proc/cmdline.o
$ objdump -Dr -j .text fs/proc/cmdline.o  | grep -e '<.*>:' -e 'Disassembly'
Disassembly of section .text:
0000000000000000 <cmdline_proc_show-0x10>:
0000000000000010 <cmdline_proc_show>:

# test 2) Turn off Indirect Branch Tracking
$ make clean
$ ./scripts/config --set-val CONFIG_X86_KERNEL_IBT n

# test 2) Results in "__pfx_" symbols
$ make -j$(nproc) fs/proc/cmdline.o
$ objdump -Dr -j .text fs/proc/cmdline.o  | grep -e '<.*>:' -e 'Disassembly'
Disassembly of section .text:
0000000000000000 <__pfx_cmdline_proc_show>:
0000000000000010 <cmdline_proc_show>:
```

The test which Ryan was running was using a .config with CONFIG_X86_KERNEL_IBT turned off,  create-diff-object didn't find **__pfx_cmdline_proc_show** so it expected that **cmdline_proc_show** would sit at the beginning of its section.

Without tracing all the details of kbuild, I _think_ options like CONFIG_LTO_CLANG and CONFIG_X86_KERNEL_IBT execute objtool at a later (linking?) stage, so even in the presence of NOP function padding bytes, create-diff-object won't necessarily see objtool-generated `__pfx_` symbols in the individual object files.

[...]

PR #1350, opened 2023-07-13
kpatch-build: account for __pfx_-less NOP padding

joe-lawrence:
Some kernel configurations generate function NOP padding, but without associated __pfx_<function> symbols.  For example:
```
  $ git describe HEAD
  v6.4-rc7-72-gdad9774deaf1

  # Initial default config turns on Indirect Branch Tracking and 16-NOP padding bytes
  $ make defconfig
  $ grep -e CONFIG_X86_KERNEL_IBT -e CONFIG_FUNCTION_PADDING_BYTES .config
  CONFIG_X86_KERNEL_IBT=y
  CONFIG_FUNCTION_PADDING_BYTES=16

  # Test .o build generates 16 bytes of NOPs but no "__pfx_" symbol
  $ make -j$(nproc) fs/proc/cmdline.o
  $ objdump -Dr -j .text fs/proc/cmdline.o  | grep -e '<.*>:' -e 'Disassembly'
  Disassembly of section .text:
  0000000000000000 <cmdline_proc_show-0x10>:
  0000000000000010 <cmdline_proc_show>:
```
This is because objtool operation on the object code may be delayed under certain configurations.  As such, create-diff-object should be prepared to encounter NOP padded functions in object files before any __pfx_ symbols are created.

Fixes: 3e54c63b175b ("create-diff-object: support x86 NOP-padded functions")
Closes: #1347 ("x86 NOP padded functions without __pfx_ symbol")

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -229,6 +229,54 @@ static struct rela *toc_rela(const struct rela *rela)
 }
 
 /*
+ * Some x86 kernels have NOP function padding, see upstream commit
+ * bea75b33895f ("x86/Kconfig: Introduce function padding").
+ * Check for any amount of NOPs between the beginning of a function's
+ * section and its offset value within that section.
+ */
+static bool insn_is_nop_pad(struct kpatch_elf *kelf, void *addr, void *end,
+			    unsigned long *insn_len)
+{
+	unsigned char *insn = addr;
+
+	switch(kelf->arch) {
+
+	case X86_64:
+		if (insn[0] == 0x90) {
+			*insn_len = 1;
+			return true;
+		}
+		break;
+
+	case PPC64:
+	case S390:
+		/* kernel feature not present on these arches  */
+		return false;
+
+	default:
+		ERROR("unsupported arch");
+	}
+
+	return false;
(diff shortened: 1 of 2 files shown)


joe-lawrence:
Internal testing notes: Ryan reports that this fixes x86-64 integration tests for the v6.3 upstream kernel when CONFIG_X86_KERNEL_IBT=y

jpoimboe:
Unfortunately I'm not sure this approach is safe (see the discussion in #1320).  We may have to add support for running create-diff-object on vmlinux.o.

joe-lawrence:
> Unfortunately I'm not sure this approach is safe (see the discussion in #1320). We may have to add support for running create-diff-object on vmlinux.o.

Ah right thanks for referencing that, I had filed that in my brain as an LTO-specific issue.

I'll close this PR, but should we consider throwing an error on `CONFIG_X86_KERNEL_IBT=y` until we figure out the eventual game plan?

jpoimboe:
> I'll close this PR, but should we consider throwing an error on `CONFIG_X86_KERNEL_IBT=y` until we figure out the eventual game plan?

Yeah, that would probably be a good idea.

PR #1354, opened 2023-08-10
Add loongarch64 support

zhangshaoning1:


jpoimboe:
Adding support for a new arch would probably need a lot more changes than this.  What testing have you done?

PR #1355, opened 2023-08-10
add kpatch core module option

wardenjohn:
For some cases, kpatch-build will force to use livepatch even though user wants kpatch core module, which is discommodious to use under such scenario.

Therefore, we should provide an option to achieve kpatch core module. Let users decide whether to use this feature.

### DIFF
--- a/kmod/patch/patch-hook.c
+++ b/kmod/patch/patch-hook.c
@@ -17,7 +17,7 @@
  * 02110-1301, USA.
  */
 
-#if IS_ENABLED(CONFIG_LIVEPATCH)
+#if IS_ENABLED(CONFIG_LIVEPATCH) && (USE_KLP == 1)
 #include "livepatch-patch-hook.c"
 #else
 #include "kpatch-patch-hook.c"
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -57,6 +57,7 @@ declare -a PATCH_LIST
 APPLIED_PATCHES=0
 OOT_MODULE=
 KLP_REPLACE=1
+KPATCH_MODULE_ENABLE=0
 
 GCC="${CROSS_COMPILE:-}gcc"
 CLANG="${CROSS_COMPILE:-}clang"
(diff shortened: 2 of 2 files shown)


jpoimboe:
Unfortunately the kpatch core module is deprecated because of known bugs and limitations.  The current plan is to remove it completely for the upcoming 1.0 release.

joe-lawrence:
> Unfortunately the kpatch core module is deprecated because of known bugs and limitations. The current plan is to remove it completely for the upcoming 1.0 release.

I've got a local branch that pulls out the legacy module and its appendages.  I still need to go through some of the kpatch/livepatch indirection to see if there is more to remove.  I was hoping to post soon, maybe next week.

wardenjohn:
I am sorry for my limitations, can you describe the known bugs and limitations of kpatch core module?

wardenjohn:
Because there still some kpatch.ko cases in my running system. I should know the issuse of using kpatch.ko to see if it is a big problem or not.

jpoimboe:
For starters, it doesn't compile with Linux 5.7+.  There are also some initialization ordering issues which may cause subtle bugs if the patched code uses jump labels or other special sections.

euspectre:
Moreover, you can patch more kernel functions with livepatching than with the kpatch.ko-based patches.
I cannot provide a concrete example right now, but I definitely encountered such situations, where we were able to provide a livepatch-based kernel update to fix certain bugs but it was impossible with kpatch.ko.

And on top of that, if I remember it correctly, redirecting kernel functions with kpatch.ko is slower than with livepatch. It does not matter much if the functions you change do some heavy work, but it might be significant for smaller functions.

I'd go for kpatch.ko only if one has to maintain a really old kernel series like 3.x or 4.x, where the in-kernel livepatch support was not fully functional yet (I mean atomic replacement of patches, etc.).

For kernels 5.x livepatch is the way to go, IMO.

wardenjohn:
Thanks, for all my system which greater than 5.7 is using livepatch now. However, I have to maintain the rest running kernel which is 4.19\4.9 or even 3.10, which is very old but still have to support them...lol..

wardenjohn:
@joe-lawrence I agree. Because not all SRE can persuade their client to upgrade their running system( even they think their system is stable enough). These old system will always have some bugs to fix. If kpatch try to remove the kpatch core mode completely in the future. Maybe we can just use the previous version of kpatch which contains kpatch.ko.

Kpatch core mode should not be a stumbling block to move on. But we should also care then needs of this mode.

PR #1358, opened 2023-08-28
Fix undefined behavior problem when using list_foreach_entry

heuza:
This upstream list.h offsetof implementation rely on undefined behavior implementation. I got SIGILL when building with flag -fsanitize=undefined by clang. Using __builtin_offsetof will fix this problem.

### DIFF
--- a/kpatch-build/list.h
+++ b/kpatch-build/list.h
@@ -30,7 +30,7 @@
 /**
  * Get offset of a member
  */
-#define offsetof(TYPE, MEMBER) ((size_t) &((TYPE *)0)->MEMBER)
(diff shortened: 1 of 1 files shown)


swine (approved):
that cleans it up nicely, supported by both gcc+clang :)

joe-lawrence:
LOL, from https://lore.kernel.org/all/20220202102147.326672-1-linux@rasmusvillemoes.dk/T/#u
> "RIP `((size_t)&((TYPE *)0)->MEMBER)`; you were beautiful in a way only a mother could love," Nick Desaulniers

joe-lawrence (approved):

Issue #1360, opened 2023-09-11
ERROR in find_local_syms, couldn't find matching XXX local symbols in vmlinux symbol table


ziyangc97:
hi, I met the following error when making kpatch.

My environment infor
```
kpatch version: 0.9.7
kernel version: 6.4.0
architecture: x86_64
linux: openEuler
gcc: gcc-12
```

/root/.kpatch/build.log:
```
  NM      System.map
  SORTTAB vmlinux
create-diff-object: ERROR: time.o: kpatch_bundle_symbols: 242: symbol __do_sys_adjtimex at offset 16 within section .text.__do_sys_adjtimex, expected 0: Success
ERROR: version-timestamp.o: 1 unsupported section change(s)
create-diff-object: unreconcilable difference: Success
```
The following is my kernel patch:
```
--- kernel/time/time.c  2023-09-06 11:19:31.000000000 +0800
+++ kernel/time/time.c.new      2023-09-11 19:06:21.674000162 +0800
@@ -170,6 +170,9 @@ int do_sys_settimeofday64(const struct t
 {
        static int firsttime = 1;
        int error = 0;
+
+       static int testdata=1;
+       printk("Test testdata=%d\n", testdata);

        if (tv && !timespec64_valid_settod(tv))
                return -EINVAL;
```

Addition info produced by debug_info flag:
```
+ [[ x86_64 = \p\p\c\6\4\l\e ]]
+ /usr/libexec/kpatch/create-diff-object orig/init/version-timestamp.o patched/init/version-timestamp.o vmlinux /root/.kpatch/tmp/vmlinux.symtab /usr/src/linux-6.4.0-6.0.0.14.oe2309.x86_64/Module.symvers klp_testklp output/init/version-timestamp.o
+ logger 1
+ local to_stdout=1
+ [[ 1 -ge 2 ]]
+ [[ 1 -eq 1 ]]
+ tee -a /root/.kpatch/build.log
ERROR: version-timestamp.o: 1 unsupported section change(s)
create-diff-object: unreconcilable difference: Success
version-timestamp.o: changed section .data not selected for inclusion
+ check_pipe_status create-diff-object
+ rc=2
+ [[ 2 = 139 ]]
+ [[ 2 -eq 0 ]]
+ [[ 2 -eq 3 ]]
+ ERROR=2
+ [[ 2 -eq 0 ]]
+ [[ 2 -ne 0 ]]
+ die '2 error(s) encountered'
[...]

PR #1361, opened 2023-09-16
lookup.c: maybe_discarded_sym() add __pfx_ pattern

he7850:
Linux-6.4 may generate `__pfx_{func}` symbols for function symbol `{func}`, however, vmlinux symtab will not contain these symbols, which will cause error when matching local symbols for changed objects.

fix part of https://github.com/dynup/kpatch/issues/1360


kpatch-build/lookup.c
```diff
@@ -83,6 +83,7 @@ static bool maybe_discarded_sym(const char *name)
 	    !strncmp(name, "__brk_reservation_fn_", 21) ||
 	    !strncmp(name, "__func_stack_frame_non_standard_", 32) ||
 	    strstr(name, "__addressable_") ||
+	    strstr(name, "__pfx_") ||
```
jpoimboe, line 86: It would be more precise to use `!strncmp(name, "__pfx_", 6)`

PR #1366, opened 2023-12-04
test/integration/rhel-8.9: integration testing for rhel-8.9

ryanbsull:


joe-lawrence:
A few small things:

- This PR is missing the *.test files in `test/integration/rhel-8.9/`.  You can copy them from the `test/integration/rhel-8.8/` directory  (note that module-LOADED.test may need the updated linux-6.2 or 9.x version)
- symvers-disagreement-FAIL.patch - somehow Julien's commit message at the top of the .path file got dropped.  Can you copy/paste from and existing version.
- Commit message
  - add the GA kernel version "kernel-4.18.0-513.5.1.el8_9" in the commit msg for reference
  - module.patch looks very different from 8.8, I *think* we had to use the linux-6.2 or 9.x version.  Can you just add a quick note in the commit msg that we updated this patch/test.  Otherwise, the rest appear to be a straight rebase.

PR #1367, opened 2023-12-06
kpatch-build: check if gawk exist

ZhangHongchen1:
kpatch-build uses gawk to find special section's struct size, but gawk is not always installed. So check if gawk is installed.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -674,6 +674,10 @@ usage() {
 	echo "		                        (not recommended)" >&2
 }
 
+if ! command -v gawk &> /dev/null; then
+	die "gawk not installed"
+fi
+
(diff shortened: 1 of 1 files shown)


kpatch-build/kpatch-build
```diff
@@ -674,6 +674,10 @@ usage() {
 	echo "		                        (not recommended)" >&2
 }
 
+if ! command -v gawk &> /dev/null; then
+	die "gawk not exist, please install gawk first"
```
jpoimboe, line 678: Thanks.  Please change the message to "gawk not installed" for consistency with some of the other error messages.

ZhangHongchen1: @jpoimboe  OK,I have modified as you said,please help review again.

kpatch-build/kpatch-build
```diff
@@ -674,6 +674,10 @@ usage() {
 	echo "		                        (not recommended)" >&2
 }
 
+if ! command -v gawk &> /dev/null; then
+	die "gawk not installed, please install gawk first"
```
jpoimboe, line 678: Sorry, to be clear, I meant to also remove the "please install gawk first", as that bit is obvious and more importantly it makes the error message consistent with other similar errors.

ZhangHongchen1: Hi @jpoimboe,
 OK,removed the "please install gawk first",please help check again.
Thanks.

jpoimboe (approved):
Looks good to me, thanks!

PR #1372, opened 2024-01-16
kpatch-build: add support for OpenCloudOS

anatasluo:
OpenCloudOS is a centos-like Linux distribution.
I test kpatch in OpenCloudOS V8 and V9.
It works well in V9. But v8 itself has two problems:
1. no available epol repo, so kpatch can't install ccache.
2. executing 'uname -r' can't get an accurate kernel version.

Both problems have been notified to the OpenCloudOS community. After they fix these problems, kpatch will work well in all versions.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -71,6 +71,7 @@ declare -rA SUPPORTED_DEB_DISTROS=(
 	["ubuntu"]="Ubuntu OS")
 
 declare -rA SUPPORTED_RPM_DISTROS=(
+	["opencloudos"]="OpenCloudOS"
 	["anolis"]="Anolis OS"
 	["centos"]="CentOS"
 	["fedora"]="Fedora"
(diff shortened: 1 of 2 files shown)


joe-lawrence:
Thanks for contributing @anatasluo.  FYI, we have an inflight MR #1368 that is touching this very code, let's get that one finished and merged, than we can rebase this one on top.

anatasluo:
@joe-lawrence Sure, it makes code seem better.

anatasluo:
@joe-lawrence I have adapted this patch based on #1368 . Can you please help review again ?

joe-lawrence:
@anatasluo : kpatch-build changes look good, but can you provide a function for test/integration/lib.sh to install the OpenCloudOS build dependencies?  If these are exactly the same as RHEL, then feel free to copy the CentOS strategy of just wrapping the kpatch_rhel_dependencies() funciton.

anatasluo:
@joe-lawrence Thanks for your review, I have fixed this problem. Is there anything else I should do?

joe-lawrence (approved):

Issue #1375, opened 2024-02-29
kpatch-build error when modifying an object file's only syscall


joe-lawrence:
### Repro:

Try building a kpatch like the following, which modifies the _only_ syscall in the resulting object file:
```patch
diff --git a/kernel/printk/printk.c b/kernel/printk/printk.c
index f1f9ce9b23f6..b8aaea260673 100644
--- a/kernel/printk/printk.c
+++ b/kernel/printk/printk.c
@@ -1903,8 +1903,10 @@ int do_syslog(int type, char __user *buf, int len, int source)
 	return error;
 }
 
-SYSCALL_DEFINE3(syslog, int, type, char __user *, buf, int, len)
+#include "kpatch-syscall.h"
+KPATCH_SYSCALL_DEFINE3(syslog, int, type, char __user *, buf, int, len)
 {
+	asm("nop");
 	return do_syslog(type, buf, len, SYSLOG_FROM_READER);
 }
 ```
kpatch-build will error out:
```
create-diff-object: ERROR: printk.o: kpatch_mark_ignored_sections: 2895: KPATCH_IGNORE_SECTION: can't find __syscalls_metadata
printk.o: ignoring section: _ftrace_events
```
### Cause:
The kpatch-syscall.h macros specifically call out `__syscalls_metadata` and `_ftrace_events` as sections for kpatch-build to ignore:
```C
#define KPATCH_IGNORE_SYSCALL_SECTIONS					\
	KPATCH_IGNORE_SECTION("__syscalls_metadata");			\
	KPATCH_IGNORE_SECTION("_ftrace_events")

#define KPATCH_SYSCALL_DEFINE1(name, ...) KPATCH_SYSCALL_DEFINEx(1, _##name, __VA_ARGS__)

...

#define KPATCH_SYSCALL_DEFINEx(x, sname, ...)				\
	KPATCH_IGNORE_SYSCALL_SECTIONS;					\
	__KPATCH_SYSCALL_DEFINEx(x, sname, __VA_ARGS__)
```
However, create-diff-object's `kpatch_mark_ignored_sections()` code to ignore such sections also requires that the sections are found in the object file, else it will error out with the above warning.

Object file sections from the repro patch build:
```bash
[root@hpe-bl460cgen10-01 kpatch]# readelf --wide --sections ~/.kpatch/tmp/orig/kernel/printk/printk.o | grep -w -e '_ftrace_events' -e '__syscalls_metadata'
[...]

PR #1376, opened 2024-02-29
kmod: maintain syscall metadata sections in kpatch syscall macros

joe-lawrence:
The KPATCH_SYSCALL_DEFINEn macros in kpatch-syscall.h do not provide the same syscall metadata (saved in the __syscalls_metadata and _ftrace_events ELF sections) as the kernel.  These same macros also instruct kpatch-build to ignore changes to these sections.  This works fine as long as there are other unmodified syscalls present in the object file.  However, if not, the kpatch syscall macros may result in either metadata ELF sections not appearing in the patched object file. The create-diff-object program expects to encounter any ELF section that has been marked by KPATCH_IGNORE_SECTION in the patched object file.

To avoid this limitation, create dummy __syscalls_metadata and _ftrace_events entries for the kpatch-modified syscall.  The specific values shouldn't matter since their sections will still be marked with KPATCH_IGNORE_SECTION and now their presence will be guarenteed for create-diff-object.

Closes: #1375 ("kpatch-build error when modifying an object file's only syscall")

### DIFF
--- a/kmod/patch/kpatch-syscall.h
+++ b/kmod/patch/kpatch-syscall.h
@@ -15,14 +15,24 @@
  * being a 'kpatch' prefix added to the __do_sys##name() function name.  This
  * causes kpatch-build to treat it as a new function (due to
  * its new name), and its caller __se_sys##name() function is inlined by its own
- * caller __x64_sys##name() function, which has an fentry hook.
-
+ * caller __x64_sys##name() function, which has an fentry hook.  Since the
+ * kpatch versions do not provide SYSCALL_METADATA, specifically entries in the
+ * __syscalls_metadata and _ftrace_events sections, provide dummy values in
+ * these sections and instruct kpatch-build to ignore changes to them.
+ *
  * To patch a syscall, just replace the use of the SYSCALL_DEFINE1 (or similar)
  * macro with the "KPATCH_" prefixed version.
  */
 
-#define KPATCH_IGNORE_SYSCALL_SECTIONS					\
+#define KPATCH_SYSCALL_METADATA(sname)					\
+	static struct syscall_metadata __used				\
+	  __section("__syscalls_metadata")			 	\
+	  *__p_syscall_meta_##sname = NULL;				\
 	KPATCH_IGNORE_SECTION("__syscalls_metadata");			\
+									\
+	static struct trace_event_call __used				\
+	  __section("_ftrace_events")					\
+	  *__event_enter_##sname = NULL;				\
 	KPATCH_IGNORE_SECTION("_ftrace_events")
 
 #define KPATCH_SYSCALL_DEFINE1(name, ...) KPATCH_SYSCALL_DEFINEx(1, _##name, __VA_ARGS__)
@@ -33,7 +43,7 @@
 #define KPATCH_SYSCALL_DEFINE6(name, ...) KPATCH_SYSCALL_DEFINEx(6, _##name, __VA_ARGS__)
 
 #define KPATCH_SYSCALL_DEFINEx(x, sname, ...)				\
-	KPATCH_IGNORE_SYSCALL_SECTIONS;					\
+	KPATCH_SYSCALL_METADATA(sname);					\
 	__KPATCH_SYSCALL_DEFINEx(x, sname, __VA_ARGS__)
 
 #ifdef CONFIG_X86_64


kmod/patch/kpatch-syscall.h
```diff
@@ -15,15 +15,25 @@
...
+	static struct syscall_metadata __used				\
+	  __section("__syscalls_metadata")			 	\
+	  *__p_syscall_meta_##sname = NULL;				\
 	KPATCH_IGNORE_SECTION("__syscalls_metadata");			\
-	KPATCH_IGNORE_SECTION("_ftrace_events")
+									\
+	static struct trace_event_call __used				\
+	  __section("_ftrace_events")					\
+	  *__event_enter_##sname = NULL;				\
+	KPATCH_IGNORE_SECTION("_ftrace_events");
```
jpoimboe, line 36: remove trailing semicolon

jpoimboe (commented):
Looks good other than the trailing semicolon

joe-lawrence:
A few follow ups:

* Since the kernel's include/linux/syscalls.h only defines SYSCALL_METADATA inside an #ifdef CONFIG_FTRACE_SYSCALLS block, should we do the same for KPATCH_SYSCALL_METADATA, too?
* In my limited testing, at least on ppc64le, I have observed that the dummy entries do ensure a __syscalls_metadata section, but not an associated rela section.  This can SEGFAULT create-diff-object's __kpatch_correlate_section() code as it's passed a NULL sec_patched pointer.  Should we assign the dummy metadata entries a pointer to something real (like &jiffies) to force the rela section, too?

jpoimboe:
>     * Since the kernel's include/linux/syscalls.h only defines SYSCALL_METADATA inside an #ifdef CONFIG_FTRACE_SYSCALLS block, should we do the same for KPATCH_SYSCALL_METADATA, too?

I think either way is fine.
 
>     * In my limited testing, at least on ppc64le, I have observed that the dummy entries do ensure a __syscalls_metadata section, but not an associated rela section.  This can SEGFAULT create-diff-object's __kpatch_correlate_section() code as it's passed a NULL sec_patched pointer.  Should we assign the dummy metadata entries a pointer to something real (like &jiffies) to force the rela section, too?

Hm, how does it get passed a NULL sec_patched pointer?

joe-lawrence:
> > In my limited testing, at least on ppc64le, I have observed that the dummy entries do ensure a __syscalls_metadata section, but not an associated rela section.  This can SEGFAULT create-diff-object's __kpatch_correlate_section() code as it's passed a NULL sec_patched pointer.  Should we assign the dummy metadata entries a pointer to something real (like &jiffies) to force the rela section, too?
> 
> Hm, how does it get passed a NULL sec_patched pointer?

```
Program received signal SIGSEGV, Segmentation fault.
0x0000000010004cf8 in __kpatch_correlate_section (sec_orig=0x10057060, sec_patched=0x0) at create-diff-object.c:1005
1005            CORRELATE_ELEMENT(sec_orig, sec_patched, "section");

(gdb) bt
#0  0x0000000010004cf8 in __kpatch_correlate_section (sec_orig=0x10057060, sec_patched=0x0) at create-diff-object.c:1005
#1  0x000000001000523c in kpatch_correlate_section (sec_orig=0x10056fc0, sec_patched=0x10111cf0) at create-diff-object.c:1032
#2  0x000000001000541c in kpatch_correlate_sections (seclist_orig=0x10054100, seclist_patched=0x1010f3f0) at create-diff-object.c:1071
#3  0x0000000010006d20 in kpatch_correlate_elfs (kelf_orig=0x100540f0, kelf_patched=0x1010f3e0) at create-diff-object.c:1505
#4  0x000000001000f6d4 in main (argc=8, argv=0x7fffffffec48) at create-diff-object.c:3992

(gdb) up
#1  0x000000001000523c in kpatch_correlate_section (sec_orig=0x10056fc0, sec_patched=0x10111cf0) at create-diff-object.c:1032
1032                    __kpatch_correlate_section(sec_orig->rela, sec_patched->rela);

(gdb) list
1027            if (is_rela_section(sec_orig)) {
1028                    __kpatch_correlate_section(sec_orig->base, sec_patched->base);
1029                    sec_orig = sec_orig->base;
1030                    sec_patched = sec_patched->base;
1031            } else if (sec_orig->rela) {
1032                    __kpatch_correlate_section(sec_orig->rela, sec_patched->rela);
1033            }
1034
1035            if (sec_orig->secsym)
1036                    kpatch_correlate_symbol(sec_orig->secsym, sec_patched->secsym);

(gdb) p sec_orig->name
$1 = 0x7ffff7c5ff75 "__syscalls_metadata"
(gdb) p sec_orig->rela
$2 = (struct section *) 0x10057060
(gdb) p sec_patched->name
$3 = 0x7ffff7bef861 "__syscalls_metadata"
(gdb) p sec_patched->rela
$4 = (struct section *) 0x0
```

jpoimboe:
Your version of the code seems to be missing the fix for that: 9fac261e ?

joe-lawrence:
> Your version of the code seems to be missing the fix for that: [9fac261](https://github.com/dynup/kpatch/commit/9fac261ed0c9586b7de78f95663fb6feebb063b5) ?

Ah yes, good find, my older kpatch-build was indeed missing that commit (which fixes the problem).

Anything more to tweak for this PR?

jpoimboe (approved):

PR #1374, opened 2024-02-20
create-diff-object: check NOP-padded instruction when symbol missing

anatasluo:
commit 3e54c63b175b68cf4 has added support for x86 NOP-padded functions.

However, it works only when pfx symbol exists.

In fedora 39 with kernel 6.7.4-200.fc39.x86_64.
I met the following error:
kpatch_bundle_symbols: 251: symbol uptime_proc_show at offset 16 within section .text.uptime_proc_show, expected 0.

Problem here is, the symbol table of uptime.o has no __pfx symbol.

To solve this problem, we can add a check if the padding area is filled with nop instructions.

PR #1379, opened 2024-03-18
create-diff-object: Add support for CONFIG_X86_KERNEL_IBT

jpoimboe:
With IBT enabled, objtool runs on the final linked vmlinux.o object instead of the individual translation units, creating the __pfx symbols at the end.  But create-diff-object still runs on the individual .o objects, in which case the __pfx symbols may be missing.  Manually detect function padding for that case.

With this change, it should be fine [*] to patch a kernel with CONFIG_X86_KERNEL_IBT enabled.

[*] Unless your patch adds an indirect call to an existing function
    which doesn't have any other indirect callers, in which case the
    callee might have been sealed, which will trigger a "Missing ENDBR"
    warning/panic.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -228,6 +228,28 @@ static struct rela *toc_rela(const struct rela *rela)
 				   (unsigned int)rela->addend);
 }
 
+static unsigned int function_padding_size(struct kpatch_elf *kelf, struct symbol *sym)
+{
+	unsigned int size = 0;
+
+	switch (kelf->arch) {
+	case X86_64:
+	{
+		unsigned char *insn = sym->sec->data->d_buf;
+		unsigned int i;
+
+		for (i = 0; i < sym->sym.st_value && *insn == 0x90; i++, insn++)
+			size++;
+
+		break;
+	}
+	default:
+		break;
+	}
+
+	return size;
+}
+
 /*
  * When compiling with -ffunction-sections and -fdata-sections, almost every
  * symbol gets its own dedicated section.  We call such symbols "bundled"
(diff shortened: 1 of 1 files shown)


jpoimboe:
Closes: #1374 
See also: #1320

jpoimboe:
TODO before merging

- [ ] Create unit test PR
- [ ] Do integration testing

Then after merging I'll port integration tests to the latest upstream kernel (or Fedora) version.

jpoimboe:
This isn't quite ready, will reopen later

liu-song-6:
I pushed my hack + some debug change here: https://github.com/liu-song-6/kpatch/tree/for-6.4-fb-ibt-test

jpoimboe:
@liu-song-6: My current thinking is that any solution that runs create-diff-object *before* objtool is flawed, because create-diff-object converts some relas to klp relas, and objtool doesn't know how to read klp relas so it doesn't have the full picture it needs.

That could maybe be fixed by teaching objtool to be able to handle klp relas.  I get the feeling that would be complex and fragile but I haven't looked at it much.

Some months back I was working on speeding up create-diff-object on vmlinux.o, but unfortunately other work popped up and that stalled.  I don't know if it would ever be fast enough, it probably needs more research.

I've also been thinking about a different approach, which is to teach objtool how to diff binaries.  I actually have a POC of that already and it's pretty straightforward.  My hope is that if objtool does the diff, then kpatch-build could do the function extraction.  Or, maybe even objtool could do the extraction and we could retire kpatch ;-)

I'm sorry this has stalled for so long, but getting IBT working is a big priority for me so I hope to get it done soon.

liu-song-6:
> any solution that runs create-diff-object before objtool is flawed

My hack for LTO is to run objtool first on special .o files then run create-diff-object on them. AFAICT, it works when there is no IBT. 

Do we need to CDO vmlinux for IBT w/o LTO? If so, I guess we also need to improve the speed of CDO?

PR #1384, opened 2024-04-08
Declaration and definition are inconsistent

xky1306102chenhong:
【In English】
Declaration and definition are inconsistent
declaration:
image

definition:
image

【中文】
声明和定义不一致

### DIFF
--- a/kpatch-build/insn/asm/inat.h
+++ b/kpatch-build/insn/asm/inat.h
@@ -106,7 +106,7 @@ extern insn_attr_t inat_get_group_attribute(insn_byte_t modrm,
 					    insn_attr_t esc_attr);
 extern insn_attr_t inat_get_avx_attribute(insn_byte_t opcode,
 					  insn_byte_t vex_m,
-					  insn_byte_t vex_pp);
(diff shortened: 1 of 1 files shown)


joe-lawrence:
Hi @xky1306102chenhong: good catch.  Are you using an automated tool to find such mismatches?  Finally, since the insn/ files are copied out of the kernel source (for example see tools/arch/x86/include/asm/inat.h and arch/x86/include/asm/inat.h), this update should be targeted there first.

xky1306102chenhong:
> Hi @xky1306102chenhong: good catch. Are you using an automated tool to find such mismatches? Finally, since the insn/ files are copied out of the kernel source (for example see tools/arch/x86/include/asm/inat.h and arch/x86/include/asm/inat.h), this update should be targeted there first.

Thank you very much for your review comments. I noticed it while reading through the code, not through an automated tool. I'm attempting to submit a patch to the Linux Kernel, but it's taking some time as I'm still relatively new to this. If the Linux Kernel accepts my patch, I'll be sure to inform you promptly.

PR #1390, opened 2024-04-26
add loongarch64 support

loong-hy:


joe-lawrence:
Adding support for a new arch would probably need a lot more changes than this. What testing have you done?

Issue #1380, opened 2024-03-21
x86 paravirt code uses alternatives v6.8+


joe-lawrence:
Upstream kernel commit torvalds/linux@f7af6977621a41661696d94c0c0a20c761404476 removes `struct paravirt_patch_site` as it's no longer used to implement the feature.  

Kpatch-build and friends will need to handle accordingly for upstream v6.8+ and future downstream backports.

PR #1392, opened 2024-05-02
kpatch-build: handle paravirt absence in Linux v6.8+

ryanbsull:
Adds linux kernel version checks to all functions pertaining to struct paravirt_patch_site as it is removed in Linux v6.8+

Fixes: https://github.com/dynup/kpatch/issues/1380

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -358,7 +358,7 @@ find_special_section_data() {
 		"x86_64")
 			check[a]=true					# alt_instr
 			kernel_version_gte 5.10.0 && check[s]=true	# static_call_site
-			[[ -n "$CONFIG_PARAVIRT" ]] && check[p]=true	# paravirt_patch_site
+			[[ -n "$CONFIG_PARAVIRT" ]] && ! kernel_version_gte 6.8.0 && check[p]=true	# paravirt_patch_site
 			;;
 		"ppc64le")
 			check[f]=true					# fixup_entry


kpatch-build/kpatch-build
```diff
@@ -438,7 +438,7 @@ find_special_section_data() {
 	[[ ${check[i]} && -z "$PRINTK_INDEX_STRUCT_SIZE" ]] && die "can't find special struct pi_entry size"
 	[[ ${check[j]} && -z "$JUMP_STRUCT_SIZE" ]] && die "can't find special struct jump_entry size"
 	[[ ${check[o]} && -z "$ORC_STRUCT_SIZE" ]] && die "can't find special struct orc_entry size"
-	[[ ${check[p]} && -z "$PARA_STRUCT_SIZE" ]] && die "can't find special struct paravirt_patch_site size"
+	[[ ${check[p]} && -z "$PARA_STRUCT_SIZE" ]] && ! kernel_version_gte 6.8.0 && die "can't find special struct paravirt_patch_site size"
```
joe-lawrence, line 441: I think this second kernel_version_gte check is redundant as ${check[p]} should only be set if it's >= 6.8.0, right?

kpatch-build/create-diff-object.c
```diff
@@ -2100,10 +2101,17 @@ static int printk_index_group_size(struct kpatch_elf *kelf, int offset)
 
 static int parainstructions_group_size(struct kpatch_elf *kelf, int offset)
 {
+	struct utsname name;
 	static int size = 0;
 	char *str;
+	int major, minor;
 
-	if (!size) {
+	if (uname(&name) < 0) {
```
joe-lawrence, line 2109: Unfortunately create-diff-object can't assume that it's building for the currently running kernel.  (Think about how it operates when kpatch-build is passed the -r, -s, -v options.)

joe-lawrence (changes requested):

kpatch-build/create-diff-object.c
```diff
@@ -2100,10 +2101,14 @@ static int printk_index_group_size(struct kpatch_elf *kelf, int offset)
...
-	char *str;
+	static int size = 0, omit = 0;
+	char *str, *omit_str;
 
-	if (!size) {
+	omit_str = getenv("OMIT_PARA");
+	if (omit_str)
+		omit = atoi(omit_str);
+
+	if (!size && !omit) {
```
joe-lawrence, line 2111: Does this work out if you omitted this "omit" code and left create-diff-object untouched?  If I read the code correctly, the group_size callbacks are only executed when their sections are encountered:

```
kpatch_process_special_sections()
  	for (special = special_sections; special->name; special++) {
		...
		sec = find_section_by_name(&kelf->sections, special->name);
		if (!sec || !sec->rela)
			continue;

		kpatch_regenerate_special_section(kelf, lookup, special, sec->rela);

			group_size = special->group_size(kelf, src_offset);
```

So I'd expect create-diff-object would only care about PARA_STRUCT_SIZE when the corresponding .parainstructions section is present.  (Which I believe torvalds/linux@f7af6977 removes.)

ryanbsull: You are correct, my bad do not need to even pass omit

joe-lawrence (approved):

PR #1396, opened 2024-06-05
Recognize Amazon Linux ID in kpatch-build

mihails-strasuns:
Make kpatch-build aware about the ID of Amazon Linux distributions. No other special changes are needed.


joe-lawrence (approved):

joe-lawrence:
Thanks @mihails-strasuns and @puranjaymohan

PR #1397, opened 2024-06-06
Small upstream kernel fixes

joe-lawrence:
Three minor fix-ups for kpatch-build and upstream kernel trees:

1. The supported_{deb,rpm}_distro() functions are now called when running with `USERSRCDIR`.  Protect against indexing their associated arrays with empty strings
2. Set the correct `VMLINUX` path when `USERSRCDIR` is specified ($TEMPDIR vs. $KERNEL_BACKUPDIR)
3. Fix setlocalversion for pre-v6.3 kernels - we fixed this for newer kernels, but sometimes it's handy to run tests against older kernels... it's not too difficult to support both pre and post v6.3 kernels here.


joe-lawrence:
@jpoimboe , @yhcote , @ryanbsull : can I get a quick review on this MR?  Thanks.

jpoimboe (approved):

yhcote (approved):

PR #1398, opened 2024-06-06
kpatch-build: Ignore missing PARA_STRUCT_SIZE for 6.8.0+ kernels

liu-song-6:
struct paravirt_patch_site was removed in 6.8.0+ kernels. We should not fail kpatch-build for missing PARA_STRUCT_SIZE.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -272,6 +272,11 @@ support_klp_replace()
 	fi
 }
 
+check_paravirt_patch_site()
+{
+    ! kernel_version_gte 6.8.0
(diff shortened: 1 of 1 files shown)


joe-lawrence:
Hi @liu-song-6 , does #1392 not solve this already?

liu-song-6:
> Hi @liu-song-6 , does #1392 not solve this already?

Ah, I guess it does fix it. I was running tests with my fork, and missed this one. I think we can close this one.

liu-song-6:
#1392 should already fix this. Closing.

PR #1401, opened 2024-07-12
test/integration/rhel-9.5: update tests to kernel-5.14.0-479.el9

ryanbsull:


joe-lawrence:
(Converting this one to a Draft as it is an intermediate kernel release for 9.5 development, and the PR is useful to trigger internal CI jobs.)

Issue #1404, opened 2024-07-24
Kernel Warning: Unpatched return thunk in use. This should not happen!


blitz:
When livepatching Linux 6.10, I get the warning below from the kernel. This seems to be harmless.

My kernel config: [kernel-config.txt](https://github.com/user-attachments/files/16360923/kernel-config.txt)
The patch: [0001-XXX-kvm-log-PIT-creation.patch.txt](https://github.com/user-attachments/files/16360945/0001-XXX-kvm-log-PIT-creation.patch.txt)
GCC: 13.2.0

```
[ 5229.900430] livepatch_7z4lq0wn99kgss3d32qjv1hn15xzl934_0001_XXX_kvm: tainting kernel with TAINT_LIVEPATCH
[ 5229.900929] ------------[ cut here ]------------
[ 5229.900930] Unpatched return thunk in use. This should not happen!
[ 5229.900933] WARNING: CPU: 2 PID: 35461 at arch/x86/kernel/cpu/bugs.c:3023 __warn_thunk+0x2c/0x40
[...]

PR #1405, opened 2024-07-24
Workaround create-diff-object issue

blitz:
We've seen this issue on Linux 6.9 and 6.10:

```
create-diff-object: ERROR: x86.o: kpatch_bundle_symbols: 251: symbol get_cpu_tsc_khz at offset 16 within section .text.get_cpu_tsc_khz, expected 0
```

We found this workaround at:
https://gitee.com/src-openeuler/kpatch/issues/I9I3EZ

We are not sure whether this fix is sane, but the result seems to work. :)

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -237,6 +237,7 @@ static void kpatch_bundle_symbols(struct kpatch_elf *kelf)
 {
 	struct symbol *sym;
 	unsigned int expected_offset;
+	unsigned int directcall_offset = 16;
 
 	list_for_each_entry(sym, &kelf->symbols, list) {
 		if (is_bundleable(sym)) {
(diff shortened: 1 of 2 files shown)


joe-lawrence:
Hi @blitz, thanks for the report.

Is there any chance you could provide a kernel config (maybe it's the same as #1403 ?) and an input kpatch that will reproduce this error.  This looks very familiar in my KERNEL_IBT travels (the aborted https://github.com/dynup/kpatch/pull/1350 perhaps?)

blitz:
> Hi @blitz, thanks for the report.
> 
> Is there any chance you could provide a kernel config (maybe it's the same as #1403 ?) and an input kpatch that will reproduce this error. This looks very familiar in my KERNEL_IBT travels (the aborted #1350 perhaps?)

You should be able to reproduce it with the config and patch from here: https://github.com/dynup/kpatch/issues/1404

I'll try without KERNEL_IBT.

PR #1406, opened 2024-08-01
patch-author-guide: correct the example CVE ID information

georgejguo:
The official website says CVE users should reference CVE-2016-5696 instead of CVE-2016-5389.


joe-lawrence:
Good catch @georgejguo !  I think what happened was that Red Hat originally tagged this internally as CVE-2016-5389, but as you found out, it is (now) CVE-2016-5696.

PR #1407, opened 2024-08-02
kpatch-build: introduce version print option

wardenjohn:
Introduce an option to print the version of kpatch-build.
When kpatch-build is installed in to a system, we and run for some time. It will have some difficulty to find out the version of the kpatch-build in the system.

So, I introduce '--version' option for user to see it. The tag released now seems to be v0.9.9, so I set version to be 0.9.9 of master branch.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -33,6 +33,7 @@
 # - Builds the patched kernel/module and monitors changed objects
 # - Builds the patched objects with gcc flags -f[function|data]-sections
 # - Runs kpatch tools to create and link the patch kernel module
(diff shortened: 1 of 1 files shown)


jpoimboe:
@joe-lawrence any objection to this?  It would have to be added to the list of files to be updated on release.

joe-lawrence:
> @joe-lawrence any objection to this? It would have to be added to the list of files to be updated on release.

Sure if it's handy, adding another file is not a huge problem.  I would have liked something that incorporated `git describe` output in case of bug reports from a repo version, but that could always be done via later enhancement.

jpoimboe (approved):

PR #1415, opened 2024-10-09
create-diff-object: __patchable_function_entries section support

joe-lawrence:
The __mcount_loc section contains the addresses of patchable ftrace sites which is used by the ftrace infrastructure in the kernel to create a list of tracable functions and to know where to patch to enable tracing of them.  On some kernel configurations, section is called __patchable_function_entries and is generated by the compiler.  Either of __mcount_loc or __patchable_function_entries is recognised by the kernel but for these configurations, use __patchable_function_entries as it is what is expected.

The x86_64 arch is special (of course).  Unlike other arches (ppc64le and aarch64) a x86_64 kernel built with -fpatchable-function-entry will generate nops AND create rela__patchable_function_entries for functions even marked as notrace.  For this arch, always create __mount_loc sections and rely on __fentry__ relocations to indicate ftrace call sites.

Note: this patch is a non-arch translation of original code by Pete Swain <swine@google.com> for aarch64.  At the same time, this version squashes several follow up commits from him and zimao <zimao@microsoft.com>.  The intent is minimize the eventual changeset for aarch64 support now that other arches are making use of __patchable_function_entries sections.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -615,9 +615,14 @@ static void kpatch_compare_correlated_section(struct section *sec)
 	     !is_text_section(sec1)))
 		DIFF_FATAL("%s section header details differ from %s", sec1->name, sec2->name);
 
-	/* Short circuit for mcount sections, we rebuild regardless */
+	/*
+	 * Short circuit for mcount and patchable_function_entries
+	 * sections, we rebuild regardless
+	 */
 	if (!strcmp(sec->name, ".rela__mcount_loc") ||
-	    !strcmp(sec->name, "__mcount_loc")) {
+	    !strcmp(sec->name, "__mcount_loc") ||
+	    !strcmp(sec->name, ".rela__patchable_function_entries") ||
+	    !strcmp(sec->name, "__patchable_function_entries")) {
 		sec->status = SAME;
 		goto out;
 	}
@@ -3677,20 +3682,46 @@ static void kpatch_create_callbacks_objname_rela(struct kpatch_elf *kelf, char *
 }
 
 /*
+ * Create links between text sections and their corresponding
+ * __patchable_function_entries sections (as there may be multiple pfe
+ * sections).
+ */
+static void kpatch_set_pfe_link(struct kpatch_elf *kelf)
+{
+	struct section* sec;
+	struct rela *rela;
+
+	if (!kelf->has_pfe)
+		return;
+
+	list_for_each_entry(sec, &kelf->sections, list) {
+		if (strcmp(sec->name, "__patchable_function_entries"))
+			continue;
+
+		if (!sec->rela)
+			continue;
+
+		list_for_each_entry(rela, &sec->rela->relas, list)
+			rela->sym->pfe = sec;
+	}
+}
+
+/*
  * This function basically reimplements the functionality of the Linux
  * recordmcount script, so that patched functions can be recognized by ftrace.
  *
  * TODO: Eventually we can modify recordmount so that it recognizes our bundled
  * sections as valid and does this work for us.
  */
-static void kpatch_create_mcount_sections(struct kpatch_elf *kelf)
+static void kpatch_create_ftrace_callsite_sections(struct kpatch_elf *kelf, bool has_pfe)
 {
 	int nr, index;
-	struct section *sec, *relasec;
-	struct symbol *sym;
-	struct rela *rela, *mcount_rela;
+	struct section *sec = NULL;
+	struct symbol *sym, *rela_sym;
+	struct rela *rela;
 	void **funcs;
 	unsigned long insn_offset = 0;
+	unsigned int rela_offset;
 
 	nr = 0;
 	list_for_each_entry(sym, &kelf->symbols, list)
@@ -3698,9 +3729,18 @@ static void kpatch_create_mcount_sections(struct kpatch_elf *kelf)
 		    sym->has_func_profiling)
 			nr++;
 
-	/* create text/rela section pair */
-	sec = create_section_pair(kelf, "__mcount_loc", sizeof(void*), nr);
-	relasec = sec->rela;
+	if (has_pfe)
+		/*
+		 * Create separate __patchable_function_entries sections
+		 * for each function in the following loop.
+		 */
+		kelf->has_pfe = true;
+	else
+		/*
+		 * Create a single __mcount_loc section pair for all
+		 * functions.
+		 */
+		sec = create_section_pair(kelf, "__mcount_loc", sizeof(void*), nr);
 
 	/* populate sections */
 	index = 0;
@@ -3709,25 +3749,37 @@ static void kpatch_create_mcount_sections(struct kpatch_elf *kelf)
 			continue;
 
 		if (!sym->has_func_profiling) {
-			log_debug("function %s has no fentry/mcount call, no mcount record is needed\n",
+			log_debug("function %s has no ftrace callsite, no __patchable_function_entries/mcount record is needed\n",
 				  sym->name);
 			continue;
 		}
 
 		switch(kelf->arch) {
 		case PPC64: {
-			bool found = false;
+			unsigned char *insn;
 
-			list_for_each_entry(rela, &sym->sec->rela->relas, list)
-				if (!strcmp(rela->sym->name, "_mcount")) {
-					found = true;
-					break;
-				}
+			if (kelf->has_pfe) {
+				insn_offset = sym->sym.st_value + PPC64_LOCAL_ENTRY_OFFSET(sym->sym.st_other);
+				insn = sym->sec->data->d_buf + insn_offset;
 
-			if (!found)
-				ERROR("%s: unexpected missing call to _mcount()", __func__);
+				/* verify nops */
+				if (insn[0] != 0x00 || insn[1] != 0x00 || insn[2] != 0x00 || insn[3] != 0x60 ||
+				    insn[4] != 0x00 || insn[5] != 0x00 || insn[6] != 0x00 || insn[7] != 0x60)
+					ERROR("%s: unexpected instruction in patch section of function\n", sym->name);
+			} else {
+				bool found = false;
 
-			insn_offset = rela->offset;
+				list_for_each_entry(rela, &sym->sec->rela->relas, list)
+					if (!strcmp(rela->sym->name, "_mcount")) {
+						found = true;
+						break;
+					}
+
+				if (!found)
+					ERROR("%s: unexpected missing call to _mcount()", __func__);
+
+				insn_offset = rela->offset;
+			}
 			break;
 		}
 		case X86_64: {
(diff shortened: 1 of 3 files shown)


joe-lawrence:
Tested internally (beaker job #10108322) on s390x and ppc64le running early rhel-10 kernel.  Also tested by Amazon on top of the rest of the aarch64 arch support code.

kpatch-build/create-diff-object.c
```diff
@@ -3676,6 +3681,206 @@ static void kpatch_create_callbacks_objname_rela(struct kpatch_elf *kelf, char *
...
+			 * This arch still relies on __mcount_loc section, so we will check then.
+			 */
+			break;
+		default:
+			ERROR("unsupported arch");
+		}
+
+		/* Allocate __patchable_function_entries for symbol */
+		sec = create_section_pair(kelf, "__patchable_function_entries", sizeof(void *), 1);
+		sec->sh.sh_flags |= SHF_WRITE | SHF_LINK_ORDER;
```
joe-lawrence, line 3854: George Guo noted in #1424's 4ce275484aa76 that there were linker complaints on loong64 due to incorrect flags, it was fixed by adding `SHF_ALLOC` and setting sh_link to 1.  I can confirm the same WAL / link flags for these sections on ppc64le and aarch64, so I will fix that up here, too.

kpatch-build/create-diff-object.c
```diff
@@ -3676,6 +3681,206 @@ static void kpatch_create_callbacks_objname_rela(struct kpatch_elf *kelf, char *
...
+		}
+
+		if (!sec->rela) {
+			continue;
+		}
+		list_for_each_entry(rela, &sec->rela->relas, list) {
+			rela->sym->sec->pfe = sec;
+		}
+	}
+}
```
jpoimboe, line 3704: What's the purpose of this function?  Is it really possible to have multiple `__patchable_function_entries` sections?  If so, it would be a good idea to have a comment at the top.

kpatch-build/create-diff-object.c
```diff
@@ -3676,6 +3681,206 @@ static void kpatch_create_callbacks_objname_rela(struct kpatch_elf *kelf, char *
...
+		if (sym->type == STT_FUNC && sym->status != SAME &&
+		    sym->has_func_profiling)
+			nr++;
+
+	/*
+	 * We will create separate __patchable_function_entries
+	 * sections for each symbols.
+	 */
+	kelf->has_pfe = true;
+	sec = find_section_by_name(&kelf->sections, "__patchable_function_entries");
```
jpoimboe, line 3724: This has no purpose apparently?

Also since this function is very similar to kpatch_create_mcount_sections() it would be a good idea to mirror that logic as much as possible so it's easier to review.

kpatch-build/create-diff-object.c
```diff
@@ -3676,6 +3681,206 @@ static void kpatch_create_callbacks_objname_rela(struct kpatch_elf *kelf, char *
...
+			/* optional endbr64 instruction */
+			if (insn[0] == 0xf3 && insn[1] == 0x0f && insn[2] == 0x1e && insn[3] == 0xfa)
+				insn+=4;
+
+			/*
+			 * We can't verify the ftrace call site as x86_64 adds nops (and
+			 * rela__patchable_function_entries) regardless of whether it is ftrace-able or not.
+			 * This arch still relies on __mcount_loc section, so we will check then.
+			 */
+			break;
```
jpoimboe, line 3847: I believe all this x86-specific code can go away, as it doesn't actually use the section.  x86's use of `-fpatchable-function-entry=16,16` is a hack for prepending every function with nops.  Maybe the elf code can just not set kelf->has_pfe for x86.

kpatch-build/create-diff-object.c
```diff
@@ -3676,6 +3681,206 @@ static void kpatch_create_callbacks_objname_rela(struct kpatch_elf *kelf, char *
...
+		default:
+			ERROR("unsupported arch");
+		}
+
+		/* Allocate __patchable_function_entries for symbol */
+		sec = create_section_pair(kelf, "__patchable_function_entries", sizeof(void *), 1);
```
jpoimboe, line 3853: Not sure why "1" instead of "nr"?  Also this should be up higher (so it mirrors the mcount version of the function).

kpatch-build/create-diff-object.c
```diff
@@ -4067,6 +4323,12 @@ int main(int argc, char *argv[])
 
 	kelf_orig = kpatch_elf_open(orig_obj);
 	kelf_patched = kpatch_elf_open(patched_obj);
+
+	kpatch_set_pfe_link(kelf_orig);
```
jpoimboe, line 4327: I don't think it matters whether the original object had pfe.

kpatch-build/kpatch-elf.c
```diff
@@ -88,17 +88,29 @@ struct section *find_section_by_index(struct list_head *list, unsigned int index
...
+	if (!list || !list->next || !name)
+		return NULL;
+
+	list_for_each_entry(sec, list, list) {
+		if (strcmp(sec->name, name))
+			continue;
+		if (--nth >= 0)
+			continue;
+		return sec;
+	}
```
jpoimboe, line 104: Why is this function needed?

kpatch-build/create-diff-object.c
joe-lawrence: Yes, on more recent kernel/toolchain combinations you'll see object files with sections like:
```
readelf --wide --sections fs/proc/meminfo.o | grep ' __patchable'
  [10] __patchable_function_entries PROGBITS        0000000000000000 000eb0 000008 00 WAL  8   0  8
  [12] __patchable_function_entries PROGBITS        0000000000000000 000eb8 000008 00 WAL  1   0  8
  [14] __patchable_function_entries PROGBITS        0000000000000000 000ec0 000008 00 WAL  1   0  8
```
(Older combinations collated the entries into a section pfe / rela section pair.)

kpatch-build/create-diff-object.c
joe-lawrence: Oops, looks like `sec` and `relasec` assignments are just overwritten here.

With respect to kpatch_create_mcount_sections() - yes, ideally there will be a single function, but I found it much easier to debug and review a standalone copy, at least to start with.  Once the pfe functionality is set (and the commonality determined), I can merge the two functions back together.

kpatch-build/create-diff-object.c
joe-lawrence: Fair enough.  x86_64's (ab)use of these sections is rather strange, so if their omission given a .config / compiler that would otherwise create them in a kernel object file is safe, then I can cut it out.

kpatch-build/kpatch-elf.c
joe-lawrence: Ah, I think this is another piece of the arm64 port that turned out to not be needed.  (It was introduced in 17a90ecf9bb7629363e03100b5545bb254e03557 but then mostly reverted later in c48342cbf87fb8867114d98fef5b2487ed67bceb  Since this MR tries to squash that branch's journey of discovery, I think the original find_section_by_name should suffice w/o needing a nth search.)

kpatch-build/create-diff-object.c
joe-lawrence: Unlike the combined mcount section, we're creating a pfe section per function, so there's a new 1 entry pair created every time through the loop.

kpatch-build/create-diff-object.c
joe-lawrence: I'll look into this tomorrow...

kpatch-build/create-diff-object.c
jpoimboe: They don't necessarily need to be combined into a single function, if it's easier to keep them apart.  I was just looking for more symmetry between the two functions so it's easier to see they're correct.

kpatch-build/create-diff-object.c
jpoimboe: Ah... is that really necessary though?  can they just be placed in a single section?

joe-lawrence:
v2:
- add a comment above `kpatch_set_pfe_link()` explaining why the function iterates over multiple pft sections
- remove unneeded x86_64 code from `kpatch_create_pfe_sections()`
- `kpatch_create_ftrace_callsite_sections()` and `main()` rename `create_pfe` to `has_pfe` -- just a slight change in meaning since `kpatch_create_ftrace_callsite_sections()` decides what to do about x86_64 objects that "have pfe" sections... the decision to "create pfe" sections lives within the function.
  - assuming v2 looks to be mostly solid, I can still double back and see about folding `kpatch_create_pfe_sections()` back into `kpatch_create_mcount_sections()`.  The code is about 90% the same, I'd just like to avoid having half a dozen "if pfe" blocks for the two sections.
- remove unneeded `find_nth_section_by_name()` and revert to original `find_section_by_name()` implementation

kpatch-build/create-diff-object.c
```diff
@@ -3945,44 +4109,84 @@ static void kpatch_no_sibling_calls_ppc64le(struct kpatch_elf *kelf)
...
 		case S390:
-			/* Check for compiler generated fentry nop - jgnop 0 */
-			insn = sym->sec->data->d_buf;
-			if (insn[0] == 0xc0 && insn[1] == 0x04 &&
-				insn[2] == 0x00 && insn[3] == 0x00 &&
-				insn[4] == 0x00 && insn[5] == 0x00)
-				sym->has_func_profiling = 1;
+			if (kpatch_symbol_has_pfe_entry(kelf, sym)) {
+				ERROR("unsupported arch");
+			} else if (sym->sec->rela) {
```
jpoimboe, line 4181: This check isn't needed, the s390 nop doesn't have a relocation.

kpatch-build/kpatch-elf.c
```diff
@@ -977,11 +984,17 @@ void kpatch_reindex_elements(struct kpatch_elf *kelf)
...
 			sym->sym.st_shndx = (unsigned short)sym->sec->index;
-		else if (sym->sym.st_shndx != SHN_ABS &&
-			 sym->sym.st_shndx != SHN_LIVEPATCH)
+			if (sym->sec->pfe) {
+				sym->sec->pfe->sh.sh_link = sym->sec->index;
+				if (sym->sec->pfe->rela)
+					sym->sec->pfe->rela->sh.sh_info = sym->sec->index;
+			}
+		} else if (sym->sym.st_shndx != SHN_ABS &&
+			 sym->sym.st_shndx != SHN_LIVEPATCH) {
```
jpoimboe, line 995: wrong indentation

kpatch-build/create-diff-object.c
```diff
@@ -3945,44 +4109,84 @@ static void kpatch_no_sibling_calls_ppc64le(struct kpatch_elf *kelf)
...
-				if (!strcmp(rela->sym->name, "_mcount")) {
-					sym->has_func_profiling = 1;
-					break;
+			if (kpatch_symbol_has_pfe_entry(kelf, sym)) {
+				sym->has_func_profiling = 1;
+			} else if (sym->sec->rela) {
+				list_for_each_entry(rela, &sym->sec->rela->relas, list) {
+					if (!strcmp(rela->sym->name, "_mcount")) {
+						sym->has_func_profiling = 1;
+						break;
```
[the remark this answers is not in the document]
joe-lawrence: Looking at a ppc64le object file, I don't think it's the first rela:
```
Relocation section '.rela.text.meminfo_proc_open' at offset 0x1698 contains 6 entries:
0000000000000000  000000220000002c R_PPC64_REL64          0000000000000000 .TOC. - 8
0000000000000008  0000000000000076 R_PPC64_ENTRY                             0
0000000000000014  000000330000000a R_PPC64_REL24          0000000000000000 _mcount + 0
0000000000000028  0000000800000032 R_PPC64_TOC16_HA       0000000000000000 .toc + 10
0000000000000030  0000000800000040 R_PPC64_TOC16_LO_DS    0000000000000000 .toc + 10
0000000000000034  000000320000000a R_PPC64_REL24          0000000000000000 single_open + 0
```
Though maybe some other assumption could be made (considering global/local entry and TOC relas)?

joe-lawrence:
v3:
- Implement all of the code cleanup suggestions
- Re: "I don't think it matters whether the original object had pfe"- it's used when determining if the original function had an ftrace site (ie, what to look for, an mcount relocation or pfe nop site)
- Re: "Ah... is that [a new pfe section per function] really necessary though? can they just be placed in a single section?" - technically I think they could (it seems to be separated for garbage collection purposes), it is just easier to debug when the create-diff-object output.o has similar layout to its kernel build created counterpart
- Re: "I'm pretty sure there's no need to use the section symbol [in pfe sections], either way it points to the same thing in the code" - true, but again, it is easier to debug when the comparing readelf from the original kernel object to the create-diff-object generated output
- Re: "Shouldn't it [ppc64le mcount relocation check in kpatch_find_func_profiling_calls()] be the first rela in the section?" - no, sample ppc64le object files have a TOC and entry relocation before their mcount relocations, unless I am misreading the question

Sorry again for making the review a hash.  GitHub hides/buries earlier comments when a subsequent push removes the associated code.

jpoimboe:
>     * Re: "Ah... is that [a new pfe section per function] really necessary though? can they just be placed in a single section?" - technically I think they could (it seems to be separated for garbage collection purposes), it is just easier to debug when the create-diff-object output.o has similar layout to its kernel build created counterpart
> 
>     * Re: "I'm pretty sure there's no need to use the section symbol [in pfe sections], either way it points to the same thing in the code" - true, but again, it is easier to debug when the comparing readelf from the original kernel object to the create-diff-object generated output

I'd disagree with both points, pfe/mcount_loc are quite simple and should be bug-free at this point, especially with mostly shared code.   More robust/readable code is more important than hypothetical (and rare for this code IMO) debugging.  I don't see how that would help much with debugging anyway, as the layouts of these sections are dirt simple.  But I defer to you.

> Sorry again for making the review a hash. GitHub hides/buries earlier comments when a subsequent push removes the associated code.

No worries, it's one of github's quirks and the comments can be shown by unhiding the "resolved" comments.  The above summary is actually pretty helpful, but not necessary.

BTW I think you forgot to move 'pfe' to struct symbol?

joe-lawrence:
> BTW I think you forgot to move 'pfe' to struct symbol?
Oops, I meant to double back and do that last.  I assume you mean to associate it with the text section symbol?

joe-lawrence:
v4:
- Move the struct section *pfe down into function rela symbol, update assignments and reads accordingly

Issue #1416, opened 2024-10-10
C99 code vs. gcc defaults?


joe-lawrence:
Internal CI testing on RHEL-7.9 popped up this build error for #1415:
``` bash
gcc -MMD -MP -I../kmod/patch -Iinsn -Wall -Wsign-compare -Wconversion -Wno-sign-conversion -g -Werror   -c -o create-diff-object.o create-diff-object.c
create-diff-object.c: In function ‘kpatch_create_pfe_sections’:
create-diff-object.c:3831:4: error: ‘for’ loop initial declarations are only allowed in C99 mode
    for (int i=0; i<16; i++) {
    ^
create-diff-object.c:3831:4: note: use option -std=c99 or -std=gnu99 to compile your code
```

Up to now, we haven't merged any C99-isms like this into the code base, but it seems that default gcc options in distros like Fedora and the Ubuntu? github Travis CI must be assuming C99, so I never saw this error until the internal CI ran on RHEL-7.

A few things we could do:
- Add `-std=gnu99` to kpatch-build Makefile
- Add `-std=gnu89` to kpatch-build/Makefile
- Add `-std=gnu89` test to .travis.yml to catch them earlier in the PR

joe-lawrence:
@jpoimboe : any preferences given the maturity of the code base?

jpoimboe:
I think we can add `-std=gnu11` to kpatch-build/Makefile as that's what the kernel uses and we tend to follow kernel style.

PR #1417, opened 2024-10-10
kpatch-build: set C language standard to gnu11

joe-lawrence:
Add -std=gnu11 to CFLAGS for kpatch-build tooling.  This aligns with the kernel build and avoids confusion when older tooling may default to earlier versions.

Closes: #1416 ("C99 code vs. gcc defaults?")


joe-lawrence:
v2: filter out the -std option for C++ `PLUGIN_CFLAGS` to avoid:
```
cc1plus: error: command line option ‘-std=gnu11’ is valid for C/ObjC but not for C++ [-Werror]
```

jpoimboe (approved):

PR #1424, opened 2024-11-08
Add LoongArch support for kpatch

georgejguo:


joe-lawrence:
Thanks for opening the PR, @georgejguo .  The plan is to first merge PR #1415 as the patchable_function_entries section support is not limited to a single arch.  Once that code is in place, the arm64 support PR shrinks drastically.  I would think a similar simplification for loongarch would be true as well.

georgejguo:
I think you can merge it if it's not harmful.
I don't know when you will be done with PR#1415?

georgejguo:
Could you merge it if it's not harmful?
I'm not sure when you will be done with PR#1415

PR #1426, opened 2024-12-04
Update s390x syscall wrappers for RHEL-9.6

joe-lawrence:
RHEL-9.6 backported upstream s390x syscall wrappers [1] that now need a distro-specific kernel version check.

[1] https://gitlab.com/redhat/centos-stream/src/kernel/centos-stream-9/-/merge_requests/5019/diffs?file=4c7281905bacd504a8205467bfd8e28ac241dbed#diff-content-4c7281905bacd504a8205467bfd8e28ac241dbed

### DIFF
--- a/kmod/patch/kpatch-syscall.h
+++ b/kmod/patch/kpatch-syscall.h
@@ -91,40 +91,23 @@
 
 #elif defined(CONFIG_S390)
 
-/* arch/s390/include/asm/syscall_wrapper.h versions */
 # if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 3, 0)
-
-#define __KPATCH_S390_SYS_STUBx(x, name, ...)                                          \
-	long __s390_sys##name(struct pt_regs *regs);                            \
-	ALLOW_ERROR_INJECTION(__s390_sys##name, ERRNO);                         \
-	static inline long ___se_sys##name(__MAP(x, __SC_LONG, __VA_ARGS__));   \
-	long __s390_sys##name(struct pt_regs *regs)                             \
-	{                                                                       \
-		return ___se_sys##name(SC_S390_REGS_TO_ARGS(x, __VA_ARGS__));   \
-	}                                                                       \
-	static inline long ___se_sys##name(__MAP(x, __SC_LONG, __VA_ARGS__))    \
-	{                                                                       \
-		__MAP(x, __SC_TEST, __VA_ARGS__);                               \
-		return __kpatch_do_sys##name(__MAP(x, __SC_COMPAT_CAST, __VA_ARGS__)); \
-	}
-
-#define __KPATCH_SYSCALL_DEFINEx(x, name, ...)					\
-       long __s390x_sys##name(struct pt_regs *regs);                           \
-       ALLOW_ERROR_INJECTION(__s390x_sys##name, ERRNO);                        \
-       static inline long __se_sys##name(__MAP(x, __SC_LONG, __VA_ARGS__));    \
-       static inline long __kpatch_do_sys##name(__MAP(x, __SC_DECL, __VA_ARGS__));    \
-       __KPATCH_S390_SYS_STUBx(x, name, __VA_ARGS__);                                 \
-       long __s390x_sys##name(struct pt_regs *regs)                            \
-       {                                                                       \
-               return __se_sys##name(SC_S390_REGS_TO_ARGS(x, __VA_ARGS__));    \
-       }                                                                       \
-       static inline long __se_sys##name(__MAP(x, __SC_LONG, __VA_ARGS__))     \
-       {                                                                       \
-               __MAP(x, __SC_TEST, __VA_ARGS__);                               \
-               return __kpatch_do_sys##name(__MAP(x, __SC_CAST, __VA_ARGS__));        \
-       }                                                                       \
-       static inline long __kpatch_do_sys##name(__MAP(x, __SC_DECL, __VA_ARGS__))
+#  define KPATCH_SYSCALL_WRAPPERS_V2
 # else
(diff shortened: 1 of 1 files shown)


kmod/patch/kpatch-syscall.h
```diff
@@ -92,7 +92,8 @@
 #elif defined(CONFIG_S390)
 
 /* arch/s390/include/asm/syscall_wrapper.h versions */
-# if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 3, 0)
+# if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 3, 0) || \
+     RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(9, 6)
```
jpoimboe, line 96: Does it need to "#ifdef RHEL_RELEASE_CODE" before using it?

joe-lawrence: Ah yes, that's right.

BTW, do you know if the preprocessor can handle this on a single line like:
```C
#if defined(RHEL_RELEASE_CODE) && RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(9, 6)
```
 w/o resorting to defining the macros in the non-RH case:
```C
#ifndef RHEL_RELEASE_CODE
# define RHEL_RELEASE_CODE 1
# define RHEL_RELEASE_VERSION(x,y) 0
#endif
```
In other places in this repo, we defined `HAVE_FEATURE` macros based on up/down-stream versions and then switched around that instead.  Do you have a preference?

jpoimboe: I *think* that should work.  Either way works for me.

jpoimboe: Actually I'm not sure if that one-liner will work, i.e., I don't know whether the preprocessor stops evaluating the line when defined() returns false.

joe-lawrence: > Actually I'm not sure if that one-liner will work, i.e., I don't know whether the preprocessor stops evaluating the line when defined() returns false.

Seems like it doesn't:
```C
$ cat test.c
#if defined(RHEL_RELEASE_CODE) && defined(RHEL_RELEASE_VERSION) && RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(9, 6)
#endif

$ gcc -c test.c -o test
test.c:1:109: error: missing binary operator before token "("
    1 | #if defined(RHEL_RELEASE_CODE) && defined(RHEL_RELEASE_VERSION) && RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(9, 6)
      |                                                                                                             ^
```
And my thought of trying to define RHEL_RELEASE_CODE / RHEL_RELEASE_VERSION wouldn't work too well either as it would only be accurate if the code checks RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION.  Flip it around to <= and it will enable unintended blocks :(

So I think it will have to be the longer HAVE_xyz form.

joe-lawrence: I don't see a shorter way to do it, how about (on top of this MR):
````diff
diff --git a/kmod/patch/kpatch-syscall.h b/kmod/patch/kpatch-syscall.h
index 8bd94f3a897f..058380bd88b3 100644
--- a/kmod/patch/kpatch-syscall.h
+++ b/kmod/patch/kpatch-syscall.h
@@ -91,9 +91,22 @@
 
 #elif defined(CONFIG_S390)
 
+# if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 3, 0)
+#  define KPATCH_SYSCALL_WRAPPERS_V2
+# else
+#  define KPATCH_SYSCALL_WRAPPERS_V1
+# endif
+
+# if defined(RHEL_RELEASE_CODE)
+#  if RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(9, 6)
+#   define KPATCH_SYSCALL_WRAPPERS_V2
+#  else
+#   define KPATCH_SYSCALL_WRAPPERS_V1
+# endif
+
+
 /* arch/s390/include/asm/syscall_wrapper.h versions */
-# if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 3, 0) || \
-     RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(9, 6)
+#if defined(KPATCH_SYSCALL_WRAPPERS_V2)
 
 #define __KPATCH_S390_SYS_STUBx(x, name, ...)                                          \
        long __s390_sys##name(struct pt_regs *regs);                            \
@@ -158,7 +171,7 @@
        __diag_pop();                                                                   \
        static inline long __kpatch_do_sys##name(__MAP(x,__SC_DECL,__VA_ARGS__))
 
-#endif /* LINUX_VERSION_CODE */
+#endif /* KPATCH_SYSCALL_WRAPPERS_V2 */
 
 #elif defined(CONFIG_PPC64)
```

joe-lawrence:
v2
- Created KPATCH_SYSCALL_WRAPPERS_Vx macros for the s390x code to switch on

jpoimboe:
Looks good, for readability can we do something like below (basically V1 first and add a comment to the '#else')?
```
#ifdef KPATCH_SYSCALL_WRAPPERS_V1

....

#else /* KPATCH_SYSCALL_WRAPPERS_V2 */

....

#endif /* KPATCH_SYSCALL_WRAPPERS_V2 */

joe-lawrence:
v3:
- Tweak #else line as suggested by Josh

jpoimboe:
I was also suggesting the V1 should come first, just so they're in logical order.

joe-lawrence:
v4:
- Move the s390x syscall wrappers into ascending version order now that they don't (directly) rely on LINUX_VERSION_CODE inequalities.

jpoimboe (approved):

PR #1431, opened 2025-01-03
kpatch-syscall: Fix kpatch s390 error when CONFIG_COMPAT is disabled

sumanthkorikkar:
When CONFIG_COMPAT is disabled (No 31 bit emulation support), macro __S390_SYS_STUBx() is nop.
(See arch/s390/include/asm/syscall_wrapper.h).
Also, __SC_COMPAT_CAST definition is not available, which leads to build error.

Hence, __KPATCH_S390_SYS_STUBx() inherited from __S390_SYS_STUBx() macro, should point also point to nop when CONFIG_COMPAT is disabled.


joe-lawrence (approved):
Thanks, @sumanthkorikkar !

Issue #1430, opened 2024-12-23
kpatch-build fails: unable to locate patch-hook.o since linux commit 13b25489b6f8


sumanthkorikkar:
Hi All,

Starting from linux commit 13b25489b6f8 ("kbuild: change working directory to external module directory with M="), the kpatch-build fails

```
Error log:
meminfo.o: changed function: meminfo_proc_show                                  
make V=1 -C /mnt/linux M=/root/.kpatch/tmp/patch CFLAGS_MODULE=''               
make[1]: Entering directory '/mnt/linux'                                        
make  -C /root/.kpatch/tmp/patch \                                              
-f /mnt/linux/Makefile                                                          
make[2]: Entering directory '/root/.kpatch/tmp/patch'                           
make --no-print-directory -C /root/.kpatch/tmp/patch \                          
-f /mnt/linux/Makefile                                                          
make -f /mnt/linux/scripts/Makefile.build obj=. need-builtin=1 need-modorder=1  
# cmd_mod test-data-new.mod                                                     
  printf '%s\n'   patch-hook.o output.o | awk '!x[$0]++ { print("./"$0) }' > test-data-new.mod
# LDS     kpatch.lds                                                            
  gcc -E -Wp,-MMD,./.kpatch.lds.d -nostdinc -I/mnt/linux/arch/s390/include -I/mnt/linux/arch/s390/include/generated -I/mnt/linux/include -I/mnt/linux/include -I/mnt/linux/arch/s390/include/uapi -I/mnt/linux/arch/s390/include/generated/uapi -I/mnt/linux/include/uapi -I/mnt/linux/include/generated/uapi -include /mnt/linux/include/linux/compiler-version.h -include /mnt/linux/include/linux/kconfig.h -D__KERNEL__ -fmacro-prefix-map=/mnt/linux/=   -P -Us390 -D__ASSEMBLY__ -DLINKER_SCRIPT -o kpatch.lds kpatch.lds.S
make V=1 -C /mnt/linux M=/root/.kpatch/tmp/patch CFLAGS_MODULE='' patch-hook.o  
make -f /mnt/linux/scripts/Makefile.build obj=. need-builtin=1 need-modorder=1 ./patch-hook.o
make[6]: *** No rule to make target 'patch-hook.o'.  Stop. `
```

Error notes:

```
- make -C changes the directory to /mnt/linux.
[...]

PR #1433, opened 2025-01-17
kmod: fix OOT kpatch kmod build

joe-lawrence:
Starting from linux commit 13b25489b6f8 ("kbuild: change working directory to external module directory with M="), the kpatch-build fails:

  make -C /root/linux M=/root/.kpatch/tmp/patch CFLAGS_MODULE=''
  make[1]: Entering directory '/root/linux'
  make[2]: Entering directory '/root/.kpatch/tmp/patch'
    LDS     kpatch.lds
  make -C /root/linux M=/root/.kpatch/tmp/patch CFLAGS_MODULE='' patch-hook.o
  ***
  *** The external module source tree is not clean.
  *** Please run 'make -C /root/linux M=/root/linux clean'
  ***

The easiest and quickest way to fix the build is to remove the $(KPATCH_MAKE) patch-hook.o in Makefile.  Unfortunately this would mean make `patch-hook.o` cannot be performed directly.  However the Makefile still lists patch-hook.o's source dependencies, so if any are updated, the kpatch module would be rebuilt anyway.

Thanks for Sumanth Korikkar for reporting, debugging, and suggesting this workaround.

Closes: #1430 ("kpatch-build fails: unable to locate patch-hook.o since linux commit 13b25489b6f8")
Reported-by: Sumanth Korikkar <sumanthk@linux.ibm.com>

### DIFF
--- a/kmod/patch/Makefile
+++ b/kmod/patch/Makefile
@@ -19,7 +19,6 @@ $(KPATCH_NAME).ko:
 $(obj)/$(KPATCH_NAME).o: $(src)/kpatch.lds
 
 patch-hook.o: patch-hook.c kpatch-patch-hook.c livepatch-patch-hook.c
-	$(KPATCH_MAKE) patch-hook.o
 
 clean:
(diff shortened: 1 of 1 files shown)


yhcote (approved):
Looks good to me.

PR #1442, opened 2025-02-27
kpatch-build: Ignore missing PARA_STRUCT_SIZE for 6.8.0+ kernels

liu-song-6:
struct paravirt_patch_site was removed in 6.8.0+ kernels. We should not fail kpatch-build for missing PARA_STRUCT_SIZE.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -275,6 +275,11 @@ support_klp_replace()
 	fi
 }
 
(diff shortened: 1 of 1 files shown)


joe-lawrence:
Hi @liu-song-6 , didn't PR #1392 solve this problem?  Is setting `check[p]=true` when `! kernel_version_gte 6.8.0` not sufficient?

PR #1444, opened 2025-03-06
create-diff-object: Include full paths in build.log output for changed object files

georgejguo:
Hi Joe,

It‘s better  to show full paths in ~/.kpatch/build.log for changed object files.

For examble:
Now output in build.log:
tcp.o: changed function: tcp_sendmsg

Improved output in build.log:
net/ipv4/tcp.o: changed function: tcp_sendmsg

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -4161,7 +4161,7 @@ int main(int argc, char *argv[])
 	patch_name    = arguments.args[5];
 	output_obj    = arguments.args[6];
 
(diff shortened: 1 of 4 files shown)


joe-lawrence:
Hi @georgejguo, I agree, this change helps clarify the log by using the full kernel paths than just the basename (especially so when building large patches).  With that, create-klp-module.c and create-kpatch-module.c contain the same pattern, those should be updated as well.

joe-lawrence (approved):
Thanks, @georgejguo

Issue #1437, opened 2025-01-28
Question about pointer equality


iii-i:
I've been looking into kpatch on s390x in context of a [compiler issue](https://sourceware.org/bugzilla/show_bug.cgi?id=29655) and came up with the following experiment: https://github.com/iii-i/linux/commit/0f400db024fb3faaeb2bc39f114ea5805ad8c963.

One part of it is a kernel function that returns its own address:

```
void *patchme(void) { return patchme; }
```

and we replace it using kpatch:

```
void *patchme(void) { printk(KERN_NOTICE "patched\n"); return patchme; }
```

Another part is a sysfs knob to compare `patchme` with `patchme()`:

```
	printk(KERN_NOTICE "patchme=0x%lx patchme()=0x%lx\n",
	       (unsigned long)patchme, (unsigned long)patchme());
```

The output before patching is as expected, the values match:

```
patchme=0x3ffe0962f78 patchme()=0x3ffe0962f78
```

After patching the values don't match anymore:

```
patched
patchme=0x3ffe0962f78 patchme()=0x3ff600a16c8
```

One may argue, and most likely be correct, that this is an artificially created scenario, which will probably not affect a real workload.

But still I would like to ask: is this an acceptable behavior?
If yes, would it be reasonable to add this to the list of limitations / things to watch out for?
If no, what would be a good way to deal with this? The only thing I can come up with now is to perform all address loads via GOT instead of using relative addressing, but this will probably slow the kernel down quite a bit.

Edit: adding @Andreas-Krebbel and @sumanthkorikkar to the discussion.

rui314:
One thing to note is that the pointer equality is guaranteed by the C language standard, so it is technically clearly a violation of the language spec (i.e. this may break valid code.) In addition to that, we found actual use cases of this guarantee in several user-land programs, as they break if pointer equality is not upheld. We learned this the hard way while developing linkers (lld and mold), where debugging issues caused by breaking pointer equality was very difficult.

joe-lawrence:
Hmm, at a high level, this reminds me of the discussion on https://github.com/dynup/kpatch/pull/755 and https://github.com/dynup/kpatch/pull/789 where kpatch-build was enhanced to use dynrelas (or klp-relocations in upstream parlance) on function pointer assignments -- this way the original function would be resolved and used on kpatch load.

I don't remember if we implemented the same for s390x, if not, would doing so help answer or simplify this problem?

iii-i:
[...]

PR #1445, opened 2025-03-07
create-diff-object: Fix function pointer rela on s390

sumanthkorikkar:
Typically, symbols within the same object file can be referenced directly for better performance. However, when the patched code stores a function pointer (R_390_GOTENT and rela symbol type is STT_FUNC), using a dynrela is a safer approach. This ensures that if the function is used as a asynchronous callback, the kernel does not attempt to execute an invalid pointer after the module is unloaded. Instead, with a dynrela, the kernel will invoke the original function, preventing potential crashes.

Test program: Test if the original function ptr address is printed during patch load / unload.
https://github.com/iii-i/linux/commit/0f400db024fb3faaeb2bc39f114ea5805ad8c963

`Patched code (function ptr rela):`
`void *patchme(void) { printk(KERN_NOTICE "patched\n"); return patchme; }`
 `  0:   c0 04 00 00 00 00       jgnop   0 <patchme>`
`   6:   eb ef f0 88 00 24       stmg    %r14,%r15,136(%r15)`
 `  c:   c4 28 00 00 00 00       lgrl    %r2,c <patchme+0xc>`
`                        e: R_390_GOTENT .LC0+0x2`
`  12:   b9 04 00 ef             lgr     %r14,%r15`
`  16:   e3 f0 ff e8 ff 71       lay     %r15,-24(%r15)`
`  1c:   e3 e0 f0 98 00 24       stg     %r14,152(%r15)`
`  22:   c0 e5 00 00 00 00       brasl   %r14,22 <patchme+0x22>`
`                        24: R_390_PLT32DBL      _printk+0x2`
`  28:   c4 28 00 00 00 00       lgrl    %r2,28 <patchme+0x28>`
`                        2a: R_390_GOTENT        patchme+0x2  <- func ptr`
`  2e:   eb ef f0 a0 00 04       lmg     %r14,%r15,160(%r15)`
`  34:   c0 f4 00 00 00 00       jg      34 <patchme+0x34>`
`                        36: R_390_PC32DBL       __s390_indirect_jump_r14+0x2`
`  3a:   07 07                   nopr    %r7`
`  3c:   07 07                   nopr    %r7`
`  3e:   07 07                   nopr    %r7`

Reference:
1. https://github.com/dynup/kpatch/issues/1437
2. https://github.com/dynup/kpatch/pull/755

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -3237,16 +3237,32 @@ static bool is_expoline(struct kpatch_elf *kelf, char *name)
 	return kelf->arch == S390 && !strncmp(name, "__s390_indirect_jump_r", 22);
 }
 
-static int function_ptr_rela(const struct rela *rela)
+static int function_ptr_rela(const struct rela *rela, struct kpatch_elf *kelf)
 {
 	const struct rela *rela_toc = toc_rela(rela);
+	bool funcptr = false;
 
-	return (rela_toc && rela_toc->sym->type == STT_FUNC &&
-		!rela_toc->sym->parent &&
-		rela_toc->addend == (int)rela_toc->sym->sym.st_value &&
-		(rela->type == R_X86_64_32S ||
-		rela->type == R_PPC64_TOC16_HA ||
-		rela->type == R_PPC64_TOC16_LO_DS));
+	if (rela_toc && rela_toc->sym->type == STT_FUNC && !rela_toc->sym->parent) {
+		switch (kelf->arch) {
+		case X86_64:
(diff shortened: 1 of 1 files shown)


joe-lawrence (approved):
Seems reasonable to me, and thanks for untangling that massive return line.  Arch-specific switch may be longer to read, but IMHO way more readable for non-PowerPC cases.

Before merging, I'm trying a few more tests patching an OOT kernel module as trying to find perfect kernel code to test combinations (function pointers in .text, .data, etc.) for an integration test is difficult (as you found noticed).   The best compromise we may currently have is to use an OOT module to generate a unit test (cached .o files), we'll see.

PR #1452, opened 2025-04-21
kpatch-build: restore source code from patched

wardenjohn:
The purpose of using `--skip-cleanup` is to save
the intermediate build products. Leaving the patch 
in the source code does not make much sense and
may even increase the workload for debugging.

Let's restore the source code from the patch.

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -900,7 +900,11 @@ else
 	BUILDDIR="$KERNEL_SRCDIR"
 fi
 
-[[ "$SKIPCLEANUP" -eq 0 ]] && trap cleanup EXIT INT TERM HUP
+if [[ "$SKIPCLEANUP" -eq 0 ]]; then
+	trap cleanup EXIT INT TERM HUP
+else
+	trap remove_patches EXIT INT TERM HUP
+fi
 
 # Don't check external file.
 # shellcheck disable=SC1090


wardenjohn:
Well, if we use '--skip-cleanup', the intermediate build products will store in ~/.kpatch/tmp, I think this would be enough for 
 us to analysis (we can store more other products here if we need more in the future).

Do we need to store the source code to ~/.kpatch/src? Like we are building kernel source, it may take too much space unnecessary (I guess...).

And according to my own experience, we usually need to adjust our patch file to make it right to build to one same kernel source. Restoring the target kernel source will make us convenient to rebuild the hotfix with kpatch again instead of reverse the patch operation manually. So, leave it as ~/.kpatch/src may be not so good?

joe-lawrence:
Would the existing `--debug` flag suite your workflow better?  With that option, you'll get:
* Kpatches reverted
* Kernel files (like Makefile, scripts/link-vmlinux.sh, etc.) restored
* $RPMTOPDIR removed (but that shouldn't matter for subsequent kpatch-builds as it will detect the previous build, aka, "Using cache at $KERNEL_SRCDIR")

If the debug message flow is too verbose, we could perhaps add a dial to tone it down.  With that, we wouldn't have to change the behavior of `--skip-cleanup` that has been in place since 2016.

wardenjohn:
ok, I will try to use this flag and test if it can cover what I need.

I will report my feedback here.

Thanks, Joe @joe-lawrence

wardenjohn:
@joe-lawrence 
Thanks for your suggestion. Also the output seem a lot, but the `debug` option do can provide my  intermediate build products for analysis. I think this is good for me to use.

You can close this PR without merging. Thanks~

PR #1455, opened 2025-04-25
CDO: Ignore section "alloc_tags"

liu-song-6:
Upstream kernel commit [1] introduces a new "alloc_tags" section. Ignore it while building patches.

[1] commit 0db6f8d7820a ("alloc_tag: load module tags into separate contiguous memory")

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2862,6 +2862,7 @@ static void kpatch_mark_ignored_sections(struct kpatch_elf *kelf)
 		if (!strncmp(sec->name, ".discard", 8) ||
 		    !strncmp(sec->name, ".rela.discard", 13) ||
 		    !strncmp(sec->name, ".llvm_addrsig", 13) ||
+		    !strcmp(sec->name, "alloc_tags") ||
 		    !strncmp(sec->name, ".llvm.", 6))
 			sec->ignore = 1;
 


joe-lawrence:
Hi Song, I have a dumb question : do these sections need to be regenerated for the livepatch module, or?

liu-song-6:
> Hi Song, I have a dumb question : do these sections need to be regenerated for the livepatch module, or?

Well, this will probably break some allocation profile. I was thinking discrepancy in the profile is acceptable for KLPs.

joe-lawrence:
I don't know much about the alloc tags, but from a quick glance it sounds like a way to track allocations to a specific place in the code.  I wonder if 1) livepatch in general may confuse the tracking as we rebuild sources with the same source paths and load them via kpatch or 2) the kpatch module build already contains sane tags and gets tracked separately, in which case we would need a (subset of) the ones generated for by the patched build?  (Or perhaps 3) none at all and we just don't track any allocations by kpatched code.)

If you know more (I hope so :) perhaps we can set up a few experiments to play with the feature.  LMK or point me to  references to learn more.  (https://origin.kernel.org/doc/html/latest/mm/allocation-profiling.html details a little bit, but it may just be faster to try it with a simple kpatch.)

liu-song-6:
I am testing kpatch for 6.13 kernel, but I haven't fixed all issues, so we cannot try this out yet. The next issue, after this fix and 1454, is that commit 11b3d5175e6b changed how we build out of tree kernel modules.

joe-lawrence:
Tried this on an x86_64 kernel build ([config.txt](https://github.com/user-attachments/files/19968359/config.txt)).  Didn't get far before EOD, but tried with the example cmdline.patch w/o kpatch-build changes:
``` bash
$ grep -e 'Kernel Config' -e CONFIG_MEM_ALLOC_PROFILING ~/linux/.config
# Linux/x86 6.14.0 Kernel Configuration
CONFIG_MEM_ALLOC_PROFILING=y
CONFIG_MEM_ALLOC_PROFILING_ENABLED_BY_DEFAULT=y
CONFIG_MEM_ALLOC_PROFILING_DEBUG=y

$ readelf --wide --sections livepatch-cmdline-string.ko | grep tags
  [36] .codetag.alloc_tags PROGBITS        0000000000000540 003200 0000c8 00  WA  0   0  8
  [37] .rela.codetag.alloc_tags RELA            0000000000000000 0032c8 0001e0 18   I 70  36  8
```
Next experiment is to change a function that actually uses the tags to invoke a section change and pickup this fix.

liu-song-6:
I looked more into this. The "alloc_tags" section only exists in vmlinux.o, but not in vmlinx. Since my version of kpatch-build use vmlinux.o (to support LTO), I need to ignore "alloc_tags" section. For upstream, this is probably not needed.

joe-lawrence:
With the builds from the config I attached in an earlier comment, I do see "alloc_tags" and ".relaalloc_tags" in the individual object files.  So a patch to kernel/pid.c :: alloc_pid() will result in a change in the section:
```
ERROR: kernel/pid.o: 1 unsupported section change(s)
create-diff-object: unreconcilable difference
kernel/pid.o: changed function: alloc_pid
kernel/pid.o: changed section alloc_tags not selected for inclusion
```
The sections I reported seeing in the cmdline.patch generated example were the "alloc_tags" corresponding to the support code provided in kmod/patch and not from the changed objects themselves.

Oddly enough, even with my livepatch to alloc_pid() loaded and enabled, the results of `grep alloc_pid /proc/allocinfo` reported the allocation size and calls (at it's original code line) moving up and down with system activity... so there may be more than simply including the patched object files "alloc_tags" and related sections.  I didn't notice any adverse effect with the livepatch loaded, but can't explain why the reported size and calls were incrementing.

PR #1456, opened 2025-05-05
CDO: Introduce log_error()

liu-song-6:
log_normal() is used in some critical errors, such as:

"changed section XXXX not selected for inclusion".

It is not very obvious this is a fatal error.

Introduce log_error() and use it for critical messages.

Note that we are not using ERROR() here, because ERROR() will terminate CDO immediately, which is not the desired behavior.


joe-lawrence (approved):
Seems reasonable.  On a related note, too bad there are so many existing ERROR() calls, otherwise we should have called it FATAL() or something to indicate that it means business and will halt the program :)

joe-lawrence:
(The latest two CDO MRs both touch kpatch_verify_patchability(), so I quickly rebased this one to smooth over the merge conflict.)

PR #1457, opened 2025-05-05
CDO: Handle .data..unlikely sections.

liu-song-6:
Upstream kernel commit [1] renamed .data.unlikely section as .data..unlikely. Handle it the same as .data.unlikely sections.

[1] bb43a59944f4 ("Rename .data.unlikely to .data..unlikely")


joe-lawrence (approved):
Thanks, Song.

PR #1460, opened 2025-05-21
kpatch-build: handle paravirt absence for RHEL in kernel-5.14.0-582.el9+

ryanbsull:


kpatch-build/kpatch-build
```diff
@@ -373,6 +373,14 @@ clang_version_check() {
 	return
 }
 
+check_paravirt() {
```
joe-lawrence, line 376: Small nit: can you move this up to the other version checks (use_klp_arch) so they're all together.

joe-lawrence (changes requested):

kpatch-build/kpatch-build
ryanbsull: Can do, just moved

joe-lawrence (approved):

Issue #1461, opened 2025-05-23
misuse for sym->type in function file_has_locals()


wbing-1927:
in kpatch-build/lookup.c function file_has_locals
check for sym is not STB_LOCAL should aim for sym->bind instead sym->type:

is this a misspell or something else

```
static bool file_has_locals(struct symbol *file_sym, struct list_head *sym_list)
{
	struct symbol *sym = file_sym;

	list_for_each_entry_continue(sym, sym_list, list) {
		if (sym->type == STT_FILE)
			break;
		if (sym->type != STB_LOCAL)
			continue;
		if (maybe_discarded_sym(sym->name))
			continue;

		if (sym->type == STT_FUNC || sym->type == STT_OBJECT)
			return true;
	}

	return false;
}
```

PR #1462, opened 2025-05-23
lookup: fix local bind check

xieyuanbin1-huawei:
STB_LOCAL is a bind attribute, not a type attribute. Just fix it.


xieyuanbin1-huawei:
This is for https://github.com/dynup/kpatch/issues/1461

PR #1463, opened 2025-05-27
Added support for AlmaLinux

KosukeH-he:
Hello,

We would like to use `kpatch-build` on AlmaLinux, so we have added AlmaLinux to the list of supported distributions.

## Details
- Tested with AlmaLinux 9.5 and 9.6.
- Kernel versions used:
  - 5.14.0-503.40.1.el9_5.x86_64 (AlmaLinux 9.5)
  - 5.14.0-570.12.1.el9_6.x86_64 (AlmaLinux 9.6)
- Ran the included test suites:
  - `make unit`: passed
  - `make integration`: passed
- Integration tests were performed using the same approach as for CentOS, utilizing the RHEL test suite.

Thank you for your time and consideration. I look forward to your feedback.

Please excuse any language mistakes, and feel free to suggest improvements.


test/integration/lib.sh
```diff
@@ -79,6 +79,50 @@ kpatch_rhel_dependencies()
...
+	[[ "$arch" == "x86_64" ]] && sudo yum install -y pesign
+
+	# ccache
+	if ! command -v ccache &> /dev/null; then
+		if ! sudo yum install -y ccache; then
+#			sudo yum install -y "https://dl.fedoraproject.org/pub/epel/epel-release-latest-${rhel_major}.noarch.rpm" && \
+#			sudo yum install -y ccache && \
+			sudo yum install -y ccache &&
+			sudo yum remove -y epel-release
+		fi
```
joe-lawrence, line 122: This looks to be a copy of kpatch_rhel_dependencies() with the -rt kernel, a few build dependency lines commented out, and the epel / ccache install lines modified... was this for debugging?

Would it be possible to kpatch_alma_dependencies() to just call kpatch_rhel_dependencies() or are there parts of the rhel setup that don't apply to alma?

joe-lawrence (commented):
Hi KosukeH-he, thanks for the MR and apologies for the tardy review.  See the comments about test/integration/lib.sh.

KosukeH-he:
Thank you very much for your feedback.
When I initially began working on this, I was using a customized environment, and it appears that some remnants of that configuration may have remained.
I am currently in the process of verifying the details of the environment I used at the time, so I would appreciate your patience a little while longer.

test/integration/lib.sh
KosukeH-he: Thank you very much for your patience.

After investigating the issue, I found that residual configuration related to the previously customized environment was still present.

I have updated the code so that kpatch_almalinux_dependencies() now calls kpatch_rhel_dependencies().
No issues were found when setting up AlmaLinux using the revised kpatch_almalinux_dependencies().

Kindly review when convenient.

joe-lawrence:
@KosukeH-he : these changes look good to me (as far as I understand AlmaLinux's relationship to RHEL).  Your last comment noted that you were still verifying some parts of your environment?  Do you still need to check, or is this MR ready for review?

joe-lawrence:
Looks good, thanks @KosukeH-he

PR #1466, opened 2025-06-26
kpatch-macros: Add KPATCH_PRINTK_DEFERRED macros

wardenjohn:
When we call printk in the context, it may cause
dead lock in the context of scheduler:

sched_func -> rq_lock -> warning -> console_print
-> soft_cursor -> drm_fb_helper_sys_imageblit ->
drm_fb_helper_dirty -> schedule_work -> queue_work (select current cpu) -> try_to_wake_up -> rq_lock (hard lockup)

Every console printing level greater than 4 may
cause this problem in the path of the scheduler.

Using printk_deferred can avoid this situation,
because printk_deferred will use irq_work_queue
instead of schedule_work.

So, we support printk_deferred for kpatch print
macros, because printk_deferred is annotated with
the __cold attribute as printk.

### DIFF
--- a/kmod/patch/kpatch-macros.h
+++ b/kmod/patch/kpatch-macros.h
@@ -141,6 +141,34 @@ struct kpatch_post_unpatch_callback {
 		printk(_fmt, ## __VA_ARGS__); \
 })
 
+
+/*
+ * KPATCH_PRINTK_DEFERRED macro
+ *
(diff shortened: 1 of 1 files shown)


wardenjohn:
@joe-lawrence Sure, I will push it again.

If you dont mind, I may copy your fixed message... -_<

PR #1467, opened 2025-06-30
kpatch-build: Add support for module symbol namespaces

joe-lawrence:
Roberto Bergantinos Corpas <rbergant@redhat.com> reported a problem when building a kpatch for:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/patch/?id=021ba7f1babd029e714d13a6bf2571b08af96d0f

```
ERROR: modpost: module livepatch-021ba7f1babd029e714d13a6bf2571b08af96d0f uses symbol dma_buf_export from namespace DMA_BUF, but does not import it.
ERROR: modpost: module livepatch-021ba7f1babd029e714d13a6bf2571b08af96d0f uses symbol dma_buf_put from namespace DMA_BUF, but does not import it.
ERROR: modpost: module livepatch-021ba7f1babd029e714d13a6bf2571b08af96d0f uses symbol dma_buf_fd from namespace DMA_BUF, but does not import it.
```

The problem is that the livepatch needs to access symbols exported via the "DMA_BUF" symbol namespace.  A workaround is to add `MODULE_IMPORT_NS("DMA_BUF")` to the kpatch, but then kpatch-build complains about unhandled .modinfo changes.

Update kpatch-build to allow .modinfo changes and extract the `import_ns=<value>` strings from .modinfo for inclusion in the output ELF object file.  Coupled with `MODULE_IMPORT_NS("DMA_BUF")`, this allows kpatch to build a livepatch that accesses a module symbol namespace.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1804,6 +1804,74 @@ static void kpatch_include_symbol(struct symbol *sym)
 		kpatch_include_section(sym->sec);
 }
 
+/*
+ * Module namespaces are saved as key value pairs in the .modinfo
+ * section.  For modules, they are in the form "import_ns=<value" and
+ * for built-ins, "<object>.import_ns=<value>".  Extract just these
+ * namespace strings from the patched .modinfo to put back into the
+ * output ELF.
+ */
+static void kpatch_get_modinfo_namespaces(struct kpatch_elf *kelf_patched,
+					  struct section *sec)
+{
+	char *data = NULL;
+	int data_len = 0;
+	const char *p, *end;
+
+	if (!sec->data || !sec->data->d_buf)
+		return;
+
+       /*
+        * Copy all the "import_ns=<value>" strings into
+        * a temporary data array (ignore other key-values)
+        */
+	p = sec->data->d_buf;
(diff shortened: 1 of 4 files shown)


kpatch-build/create-diff-object.c
```diff
@@ -1804,6 +1804,71 @@ static void kpatch_include_symbol(struct symbol *sym)
...
+	kelf_patched->modinfo_data = data;
+	kelf_patched->modinfo_data_len = data_len;
+}
+
+static void kpatch_put_modinfo_namespaces(struct kpatch_elf *kelf_patched,
+					  struct kpatch_elf *kelf_out)
+{
+	struct section *sec;
+
+	sec = create_section(kelf_out, ".modinfo",
```
jpoimboe, line 1865: Should this be skipped if .modinfo doesn't exist in the patched object?

joe-lawrence: Yup, no need to create an empty section.

joe-lawrence: True, that's a better test, as it would skip instances with .modinfo but no imports.

PR #1469, opened 2025-07-14
patch-author-guide: add a blurb about module namespaces

joe-lawrence:
Now that kpatch-build passes along .modinfo "import_ns" key/value pairs, add a hint in the patch author guide about using MODULE_IMPORT_NS.  The most likely use case will be formerly built-in code that now needs the import when extracted out into a kpatch.ko.

### DIFF
--- a/doc/patch-author-guide.md
+++ b/doc/patch-author-guide.md
@@ -28,6 +28,7 @@ Table of contents
 - [Sibling calls](#sibling-calls)
 - [Exported symbol versioning](#exported-symbol-versioning)
 - [System calls](#system-calls)
+- [Symbol Namespaces](#symbol-namespaces)
 
 Patch analysis
 --------------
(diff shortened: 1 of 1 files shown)


doc/patch-author-guide.md
```diff
@@ -945,3 +946,18 @@ missing because of the 'inline' annotation, which invokes 'notrace'.
...
+-----------------
+
+While kpatch modules automatically inherit namespace imports from
+already-patched object files, a manual import may be required when
+patching kernel code that is normally built-in.
+
+The original built-in code doesn't need to explicitly import namespaces.
+However, when converted into a kpatch module, it must declare any namespace
+dependencies. Without this explicit import, the kpatch-build command will fail
+with modpost errors for using symbols from a namespace without importing it.
```
jpoimboe, line 960: Can an example of the error be shown here?

joe-lawrence:
v2:
- Added kpatch-build error from modpost [Josh]

jpoimboe (approved):

PR #1471, opened 2025-08-19
kpatch-build: Enable cross compiling with env TARGET_ARCH

liu-song-6:
llvm/clang supports cross compiling of the kernel. To build livepatch in the same cross compile environment, enable specifying TARGET_ARCH for kpatch-build. For example, in a x86_64 host, we can build livepatch for aarch64 kernel with:

   TARGET=aarch64 kpatch-build ...

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -40,6 +40,7 @@ set -o pipefail
 BASE="$PWD"
 SCRIPTDIR="$(readlink -f "$(dirname "$(type -p "$0")")")"
 ARCH="$(uname -m)"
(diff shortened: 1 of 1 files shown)


liu-song-6:
I am creating this PR for comments. We probably want to wait until #1439 is landed.

joe-lawrence:
This is interesting and something I hoped we'd eventually implement :)  Have you run cross-compiled kpatch modules against native-built kernels?

joe-lawrence:
Thanks @liu-song-6 , I dropped out some of the extraneous whitespace edits (probably an editor setting).

I'll work out a release next week (need to run all the internal tests and then push out a tag).

Issue #1470, opened 2025-07-31
cs-10: ppc64le builds fail


joe-lawrence:
The latest CentOS-Stream 10 kernels fail kpatch-build builds:

```
$ ./kpatch-build/kpatch-build -d examples/cmdline-string.patch

[ ... snip ... ]

Extracting new and modified ELF sections
arch/powerpc/tools/vmlinux.arch.o'
create-diff-object: ERROR: fs/proc/cmdline.o: kpatch_create_ftrace_callsite_sections: 3860: cmdline_proc_show: unexpected instruction in patch sect

fs/proc/cmdline.o: changed function: cmdline_proc_show
doing "deep find" for parent object
ERROR: invalid ancestor arch/powerpc/tools/vmlinux.arch.o for arch/powerpc/tools/vmlinux.arch.o. Check /root/.kpatch/build.log for more details.

$ tree ~/.kpatch/tmp/{orig,patched}
/root/.kpatch/tmp/orig
├── arch
│   └── powerpc
│       └── tools
│           └── vmlinux.arch.o
└── fs
    └── proc
        └── cmdline.o
/root/.kpatch/tmp/patched
├── arch
│   └── powerpc
│       └── tools
│           └── vmlinux.arch.o
└── fs
    └── proc
        └── cmdline.o

12 directories, 4 files
```

[config-6.12.0-113.el10.ppc64le.txt](https://github.com/user-attachments/files/21535009/config-6.12.0-113.el10.ppc64le.txt)

[repro-unexpected-instruction.tar.gz](https://github.com/user-attachments/files/21579203/repro-unexpected-instruction.tar.gz)

PR #1477, opened 2025-09-05
kpatch-build: fix ppc64le trampoline handling after BPF changes

emmil:
kpatch-build: fix ppc64le trampoline handling after BPF changes

Starting from linux commit eec37961a56a (powerpc64/ftrace: Move
ftrace sequence out of line) the kpatch-build fails

        create-diff-object: ERROR: fs/proc/cmdline.o: kpatch_create_ftrace_callsite_sections: 3860: cmdline_proc_show: unexpected instruction in patch section of function

        fs/proc/cmdline.o: changed function: cmdline_proc_show
        doing "deep find" for parent object
        ERROR: invalid ancestor arch/powerpc/tools/vmlinux.arch.o for arch/powerpc/tools/vmlinux.arch.o. Check /root/.kpatch/build.log for more details.

Moving the ftrace stubs out-of-line modified the prologue on ppc64le
and it now uses a single nop instead of two.
Update the nop check in create-diff-object.c accordingly.

Also add vmlinux.arch.o to the list of objects excluded from
instrumentation, since it only contains the out-of-line ftrace stubs
for vmlinux.
The module loader will create any necessary (livepatch) module
ftrace stubs, so we don't care about modifications to this file.

Closes: #1470 ("cs-10: ppc64le builds fail")
Reported-by: Joe Lawrence <joe.lawrence@redhat.com>
Signed-off-by: Radomir Vrbovsky <rvrbovsk@redhat.com>

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -3854,9 +3854,8 @@ static void kpatch_create_ftrace_callsite_sections(struct kpatch_elf *kelf)
 				insn_offset = sym->sym.st_value + PPC64_LOCAL_ENTRY_OFFSET(sym->sym.st_other);
 				insn = sym->sec->data->d_buf + insn_offset;
 
-				/* verify nops */
-				if (insn[0] != 0x00 || insn[1] != 0x00 || insn[2] != 0x00 || insn[3] != 0x60 ||
-				    insn[4] != 0x00 || insn[5] != 0x00 || insn[6] != 0x00 || insn[7] != 0x60)
+				/* verify nop */
+				if (insn[0] != 0x00 || insn[1] != 0x00 || insn[2] != 0x00 || insn[3] != 0x60)
 					ERROR("%s: unexpected instruction in patch section of function\n", sym->name);
 			} else {
 				bool found = false;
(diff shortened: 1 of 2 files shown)


joe-lawrence (changes requested):
Code changes look fine.  As we discussed off list, we could be more deliberate about detecting one or two NOPs, but since this is only a minor assertion it's not worth the effort.

A few nits with the commit msg:
- I usually use my email alias in this repo:  Joe Lawrence <joe.lawrence@redhat.com>
- As I understand the patchset: [[PATCH v7 00/17] powerpc: Core ftrace rework, support for ftrace direct and bpf trampolines](https://lore.kernel.org/linuxppc-dev/20241030070850.1361304-1-hbathini@linux.ibm.com/):
  - We don't need to specify that these were "BPF changes", I think it could just say, "Moving the ftrace stubs out-of-line modified the prologue ..."
  - And then regarding vmlinux.arch.o, "... excluded from
instrumentation, since it only contains the out-of-line ftrace stubs for vmlinux.  The module loader will create any necessary (livepatch) module ftrace stubs, so we don't care about modifications to this file."

joe-lawrence (approved):
Thanks, @emmil

PR #1483, opened 2025-10-28
Remove trailing slash from $SRCDIR and check for required env vars

Tomcat-42:
If $SRCDIR has a trailing slash (e.g., `kernel/`), the command `cp -r "$SRCDIR" "${SRCDIR}.orig"` incorrectly copies the backup *inside* the source directory instead of alongside it. This causes the subsequent `diff -Nupr` to fail or compute an incorrect diff.

Also checks for the definitions of required env vars.

### DIFF
--- a/test/integration/rebase-patches
+++ b/test/integration/rebase-patches
@@ -18,9 +18,15 @@
 #    ID=rhel VERSION_ID=7.8 ./rebase-patches rhel-7.7/*{.patch,.disabled}
 # % cp rhel-7.7/*.test rhel-7.8/
 
-OUTDIR=$(pwd)/${ID}-${VERSION_ID}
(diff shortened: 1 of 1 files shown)


joe-lawrence (changes requested):
LOL, I've stubbed my toe on this many times when using tab completion to fill out the SRCDIR path, but never bothered to fix it.

Nit 1 - same as the other PR, if you can format the commit msg like a kernel commit.  That is, a subject format "subsys/section: summary ...", the description, finished by your signed-off-by line.

Nit 2 - completely up to you, but would `realpath` be any more useful?  I think it may solve this problem and also do some additional cleanup.  I'd be happy either way and just wanted to mention it.

PR #1485, opened 2025-11-17
create-diff-object: fix Clang UBSAN diff failure by ignoring unnamed data

FlorentRevest:
This fixes a diff failure when generating livepatches for objects compiled with Clang and UBSAN (Undefined Behavior Sanitizer).

When UBSAN is enabled, Clang heavily utilizes `.data..L__unnamed_XX` sections to store type descriptors and source location metadata. kpatch-build currently fails to handle these sections, resulting in errors like:

  fsnotify.o: changed section .data..L__unnamed_1 not selected for inclusion
  ERROR: fsnotify.o: 1 unsupported section change(s)
  create-diff-object: unreconcilable difference

The numeric suffix (e.g., `_1`) in these section names is an unstable internal counter. If the patch modifies the code, these indices often shift, causing `create-diff-object` to incorrectly correlate unrelated UBSAN metadata between the original and patched objects. Since kpatch cannot support changes to existing data sections, the build aborts.

Fix this by treating `.data..L__unnamed_` sections as unstable and uncorrelatable.

This patch introduces `is_clang_unnamed_data()` to:
1. Explicitly skip these sections in `kpatch_correlate_sections()`.
2. Prevent `kpatch_correlate_static_local_variables()` from forcing correlation based on symbol usage.

This ensures that UBSAN metadata sections are always marked as 'NEW' and allocated safely in the patch module, enabling support for livepatching UBSAN-instrumented kernels.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -400,6 +400,18 @@ static bool is_string_literal_section(struct section *sec)
 }
 
 /*
+ * Clang generates .data..L__unnamed_XX sections for anonymous constants.
+ * The numeric suffix is unstable (it can change if code is added/removed).
+ * Therefore, we must never correlate these by name; the patched object
+ * must always allocate a fresh copy (Status: NEW).
+ */
+static bool is_clang_unnamed_data(const char *name)
+{
+	return !strncmp(name, ".data..L__unnamed_", 18) ||
+		!strncmp(name, ".data.__unnamed_", 16);
+}
+
+/*
  * This function detects whether the given symbol is a "special" static local
  * variable (for lack of a better term).
  *
@@ -426,6 +438,10 @@ static bool is_special_static(struct symbol *sym)
 
 	/* pr_debug() uses static local variables in the __verbose or __dyndbg section */
 	if (is_dynamic_debug_symbol(sym))
+		return true;
+
+	/* Do not try to correlate statics inside unstable Clang sections */
+	if (sym->sec && is_clang_unnamed_data(sym->sec->name))
 		return true;
 
 	if (sym->type == STT_SECTION) {
@@ -1150,6 +1166,11 @@ static void kpatch_correlate_sections(struct list_head *seclist_orig,
 			if (is_ubsan_sec(sec_orig->name))
 				continue;
 
+			/* Skip correlation for unstable Clang anonymous sections */
+			if (is_clang_unnamed_data(sec_orig->name) ||
+			    is_clang_unnamed_data(sec_patched->name))
+				continue;
+
 			if (is_special_static(is_rela_section(sec_orig) ?
 					      sec_orig->base->secsym :
 					      sec_orig->secsym))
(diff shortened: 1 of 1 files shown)


FlorentRevest:
Thanks for triggering the tests @joe-lawrence! :) I can't seem to be able to open any logs for those "5 failing checks" - do you know if they are flakes or indicative of a real issue with my PR ? Also, if you have access to logs and they point to an issue in my PR, could you share them here so I get a chance to fix the issue ?

joe-lawrence:
> Thanks for triggering the tests @joe-lawrence! :) I can't seem to be able to open any logs for those "5 failing checks" - do you know if they are flakes or indicative of a real issue with my PR ? Also, if you have access to logs and they point to an issue in my PR, could you share them here so I get a chance to fix the issue ?

Hi @FlorentRevest , the packit tests are relatively new so bear with us during the learning curve :)  Can you access any of these links:

https://github.com/dynup/kpatch/pull/1485/checks?check_run_id=56410963239
https://dashboard.packit.dev/jobs/testing-farm/1262780
https://artifacts.osci.redhat.com/testing-farm/ea560bbd-872d-4561-98e5-a7bada8e6e2e

And then @Tomcat-42, can you take a look at the test results to see if there are any real failures, or if these are false positives.  Thanks!

FlorentRevest:
> Hi @FlorentRevest , the packit tests are relatively new so bear with us during the learning curve :)

Of course! :) 

> Can you access any of these links:
> 
> https://github.com/dynup/kpatch/pull/1485/checks?check_run_id=56410963239

This one opens but doesn't contain much apart from a link to the next two

> https://dashboard.packit.dev/jobs/testing-farm/1262780

This one also opens but also doesn't contain much apart from a link to the next one

> https://artifacts.osci.redhat.com/testing-farm/ea560bbd-872d-4561-98e5-a7bada8e6e2e

This one fails to load on any device I tried with:

"This site can’t be reached
Check if there is a typo in artifacts.osci.redhat.com.
DNS_PROBE_FINISHED_NXDOMAIN"

FlorentRevest:
> > However, even if we wait for this migration to happen before merging, I'd really appreciate a code review in the meantime - when someone gets a chance of course - because the risk for us to have to livepatch a UBSAN-sanitized compilation unit grows every day and if we have to use this patch as a downstream addition, some sort of review from someone who understands kpatch better than me would give us a lot more confidence that my patch is not entirely unreasonable! :)
> 
> The patch looks reasonable to me. I'm not heavily experienced with either clang or UBSAN, but I can generate such sections with a simple userspace prog:

Thanks a lot for looking into it! I really appreciate! :) 

> now I know the PR is targeted to the UBSAN options, but should we care about the non-local naming for `-fsanitize=kernel-address`? (I don't know if you're using KASAN, but maybe their existence means clarifying any naming/descriptions in this PR?)

That's a very good point - I have not tried with KASAN so I didn't know about those sections but I agree, it looks like they should be taken care of in the same way as the local ones as part of this PR.

> If you like, we could probably add this to our unit test collection. The easiest way would be to build with `kpatch-build --debug` and then attach the `~/.kpatch/tmp/{orig/patched}` object files as well as the `~/.kpatch/tmp/kpatch.env` file here. Pushing those are a bit annoying due to the git submodule arrangement, but I can walk you through it if you are interested.

I agree some tests here would be great and I'd be happy to look into it. I'll let you know if I run into issues.

I am just gonna leave for some holidays around LPC now so I won't be able to take care of this short term but let me circle back to this after LPC.

joe-lawrence:
Hi @FlorentRevest , thanks for updating and including object file for the unit tests.  The changes look good, can you add a bump to the test/unit/obj submodule reference here to move it to the latest & greatest?  Thanks.

PR #1487, opened 2025-12-04
create-diff-object: Add support for x86_64 CONFIG_CALL_PADDING

lu21253:
For x86_64 this option uses -fpatchable-function-entry=$(CONFIG_FUNCTION_PADDING_BYTES),$(CONFIG_FUNCTION_PADDING_BYTES), so CONFIG_FUNCTION_PADDING_BYTES NOPs are placed before the function entry point (in order to store a pointer to ftrace_ops).When calculating function padding, check for the presence of the CONFIG_FUNCTION_PADDING_BYTES NOPs, and adjust the padding size by CONFIG_FUNCTION_PADDING_BYTES if they are found.

Signed-off-by: l21253 <lu.haoA@h3c.com>
Reviewed-by: w28894 <wang.shijie@h3c.com>
Reviewed-by: z17497 <zhang.chunA@h3c.com>

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -284,6 +284,43 @@ static unsigned int function_padding_size(struct kpatch_elf *kelf, struct symbol
 
 		break;
 	}
+	case X86_64: 
+	{
+		uint8_t *insn = sym->sec->data->d_buf;
+		uint8_t *insn_end = sym->sec->data->d_buf + sym->sym.st_value;
+	
+	       /*
+  	        * If the x86_64 kernel is compiled with CONFIG_CALL_PADDING
+  	        * then there are CONFIG_FUNCTION_PADDING_BYTES NOPs before the function.Verify the presence of the CONFIG_FUNCTION_PADDING_BYTES 
+  	        * NOPs before the function entry.
+  	        * Possible padding patterns before X86 functions:
+  	        * 1. NOP instructions (0x90) for alignment.
+  	        * 2. Multi-byte NOPs that compilers may insert (e.g., 0x66 0x90).
+  	        * 3. Jump tables or special instructions may exist before functions (adjust according to actual scenarios).
+                */
+		while (insn < insn_end) {
+
+			if (*insn == 0x90) {
+				size++;
+				insn++;
+			}
+			else if (insn + 1 < insn_end && 
+					  insn[0] == 0x66 && insn[1] == 0x90) {
+				size += 2;
+				insn += 2;
+			}
+			else {
+				break;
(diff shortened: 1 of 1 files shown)


lu21253:
Hi, @joe-lawrence :
My kernel configuration consists of some standard settings. I am debugging livepatch on Linux kernel 6.6.0 and want to use kpatch-master to generate the livepatch patch ko object file. When executing create-diff-object, the following error occurred:

make: Leaving directory '/home/l21253/V9/B93D001/PUBLIC/kernel'

/home/l21253/V9/B93D001/kpatch-master/kpatch-build/create-diff-object /home/l21253/V9/B93D001/objs64/orig/n_tty.o /home/l21253/V9/B93D001/objs64/patched/n_tty.o vmlinux /home/l21253/V9/B93D001/objs64/vmlinux.symtab /home/l21253/V9/B93D001/TARGET/arch/x86_64/release/Module.symvers livepatch-livepatch-vmlinux-tty.ko /home/l21253/V9/B93D001/objs64/output/n_tty.o

create-diff-object: ERROR: home/l21253/V9/B93D001/objs64/orig/n_tty.o: kpatch_bundle_symbols: 355: symbol echo_set_canon_col at offset 16 within section .text.echo_set_canon_col, expected 0

joe-lawrence:
> CONFIG_X86_KERNEL_IBT=y

In https://github.com/dynup/kpatch/pull/1411 we added a kpatch-build assertion against this option (IIRC there was some explanation in https://github.com/dynup/kpatch/issues/1320), so you should had gotten an "ERROR: kernel option 'CONFIG_X86_KERNEL_IBT' not supported" error message right off the bat.

The short of it is that with this kernel option, and a few others now, the kernel build will defer objtool object file manipulation to vmlinux.o instead of on individual object files.  This breaks kpatch-build assumptions and we never found a great workaround, so to address that Josh created an upstream, in-kernel-tree approach just recently merged: https://www.phoronix.com/news/Linux-6.19-objtool-klp-build 

That said, maybe this kpatch-build patch is reasonable for completeness on !CONFIG_X86_KERNEL_IBT configs.  How did you hit the create-diff-object error (ie, which version of kpatch-build are you using) and do you happen to have an input kpatch for repro?  Thanks!

lu21253:
@joe-lawrence 
The kpatch-build version we are using is 0.9.11, but we did not directly use the kpatch-build script; we streamlined the script. After retesting, removing the CONFIG_X86_KERNEL_IBT configuration item also resolves the issue we previously encountered, but we still require the CONFIG_X86_KERNEL_IBT configuration item.

PR #1494, opened 2026-02-10
create-diff-object: Remove undefined function symbols

sumanthkorikkar:
When building shadow-pid.patch on a debug kernel, it generates __bug_table, which contains an array of struct bug_entries.

.rela__bug_table contains references to bug address, line number and column.

create-diff-object identifies that .text.kernel_clone has changed and it includes .rela.text.kernel_clone rela section. Then later, it includes all symbols (in kpatch_include_symbols()) associated with it, which ends up including __bug_table and its rela section .rela__bug_table. Then, all the function symbols associated with .rela__bug_table is included irrespective of whether it's section is included or not.

This leads to the following modpost errors:
kernel/fork.o: changed function: kernel_clone
kernel/exit.o: changed function: do_exit
fs/proc/array.o: changed function: proc_pid_status make -C /root/linux M=/root/.kpatch/tmp/patch CFLAGS_MODULE='' make[1]: Entering directory '/root/linux'
make[2]: Entering directory '/root/.kpatch/tmp/patch'
  LDS     kpatch.lds
  CC [M]  patch-hook.o
  LD [M]  test-shadow-newpid.o
  MODPOST Module.symvers
WARNING: modpost: missing MODULE_DESCRIPTION() in test-shadow-newpid.o
ERROR: modpost: "replace_mm_exe_file" [test-shadow-newpid.ko] undefined!
ERROR: modpost: "put_task_stack" [test-shadow-newpid.ko] undefined!
ERROR: modpost: "release_task" [test-shadow-newpid.ko] undefined!
ERROR: modpost: "set_mm_exe_file" [test-shadow-newpid.ko] undefined!

Examining the /root/.kpatch/patch/ directory reveals, these symbols are never referenced in any relas.

readelf -Ws output.o |
grep -E 'put_task_stack|replace_mm_exe_file|release_task|set_mm_exe_file'
    27: 0000000000000000     0 SECTION LOCAL  DEFAULT   36 .rodata.release_task.str1.2
    45: 0000000000000000     0 SECTION LOCAL  DEFAULT   55 .rodata.set_mm_exe_file.str1.2
    47: 0000000000000000     0 SECTION LOCAL  DEFAULT   57 .rodata.replace_mm_exe_file.str1.2
   234: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND replace_mm_exe_file
   254: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND put_task_stack
   263: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND release_task
   269: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND set_mm_exe_file

readelf -Wr output.o |
grep -E 'put_task_stack|replace_mm_exe_file|release_task|set_mm_exe_file' <EMPTY>

Hence, exclude these unreferenced symbols to avoid modpost errors.

Fix:
* Identify all function symbols present in __bug_table and track their symbol indices.
* Exclude .rela__bug_table and .rela__mcount_loc, and for all other relocation sections, check whether any of these symbol indices are actually referenced.
* If a symbol index is never referenced in any relevant relocation section and the symbol’s section is not included in the patch, exclude the symbol from being added.

**PATCH RFC v2:**
 Note: Skipped need_klp_reloc()/kpatch_create_intermediate_sections()
    check for .rela__bug_table section.
    Reason: The function symbols that were not referenced by any sections
    other than .rela__bug_table were being initialized with include = 0 (via
    rela->sym->include = 0). As a result, kpatch_migrate_included_elements()
    did not migrate these function symbols into kelf_out.  However, later in
    kpatch_create_intermediate_sections(), when parsing the .rela__bug_table
    relasec and evaluating each symbol in need_klp_reloc(), the
    code ended up using the previous rela->sym reference (which had already
    been torn down). Since that symbol had its include field set to 0, the
    dereference led to a segmentation fault.  To prevent this, the
    .rela__bug_table section is excluded from consideration in
    kpatch_migrate_included_elements().  Additionally, if a function is
    modified, the assumption is that, it will be referenced by other
    relasec.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1886,6 +1886,23 @@ static void kpatch_verify_patchability(struct kpatch_elf *kelf)
 		DIFF_FATAL("%d unsupported section change(s)", errs);
 }
 
+/*
+ * Do not perform symbol inclusion initially from .rela__bug_table.
+ *
+ * WARN() is converted to a static call which emits relocations that point
+ * directly into __bug_table+off. If the usual "include every symbol referenced
+ * by this rela section" rule is applied here, it could end up dragging in the
+ * entire __bug_table via propagation.
+ *
+ * .rela__bug_table relocations are processed later in
+ * kpatch_regenerate_special_section(), which also adjusts relocations
+ * targetting __bug_table.
+ */
+static bool kpatch_skip_symbol_inclusion_from_relasec(struct section *relasec)
+{
+	return !strcmp(relasec->name, ".rela__bug_table");
+}
+
 static void kpatch_include_symbol(struct symbol *sym);
 
 static void kpatch_include_section(struct section *sec)
@@ -1906,6 +1923,8 @@ static void kpatch_include_section(struct section *sec)
 	if (!sec->rela)
 		return;
 	sec->rela->include = 1;
+	if (kpatch_skip_symbol_inclusion_from_relasec(sec->rela))
+		return;
 	list_for_each_entry(rela, &sec->rela->relas, list)
 		kpatch_include_symbol(rela->sym);
 }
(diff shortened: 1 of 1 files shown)


joe-lawrence:
@sumanthkorikkar : I haven't been following this PR closely, is this only relevant for debug kernels?  Just wondering if this would be needed as a bug-fix type change in light of Josh's recent maintenance-only mode announcement (#1499).  I was thinking of tagging a v0.9.12 to capture the last round of fixes and enhancements since the last version, and didn't know if this might qualify as well.

joe-lawrence:
@Tomcat-42 : any idea why [testing-farm:RHEL-8.10.0-Nightly-ppc64le:integration-rhel](https://github.com/dynup/kpatch/pull/1494/checks?check_run_id=67944162181) failed?  The other tests (save for the known Fedora infra issues) all passed, so I'm assuming something intermittent?

Tomcat-42:
@joe-lawrence Yep, I guess it was some transient failure on infra, I've re-runned the job and it passed.

The Fedora integration-upstream jobs are fixed in master (except ppc64le), so maybe @sumanthkorikkar wants to do a rebase? (I've noticed that push by maintainers is disabled in his fork)

sumanthkorikkar:
> @sumanthkorikkar : I haven't been following this PR closely, is this only relevant for debug kernels? Just wondering if this would be needed as a bug-fix type change in light of Josh's recent maintenance-only mode announcement (#1499). I was thinking of tagging a v0.9.12 to capture the last round of fixes and enhancements since the last version, and didn't know if this might qualify as well.

Thanks for the information. It would be nice to have this PR as bug fix in the kpatch. This is relevant for only debug kernel and problem occurs in x86 and s390

Relevant commit:
x86 - kernel version v6.19 - 5b472b6e5bd9 ("x86_64/bug: Implement __WARN_printf()") - v6.19
s390  - 04dabb4261c3 ("s390/bug: Implement __WARN_printf()")

sumanthkorikkar:
> @joe-lawrence Yep, I guess it was some transient failure on infra, I've re-runned the job and it passed.
> 
> The Fedora integration-upstream jobs are fixed in master (except ppc64le), so maybe @sumanthkorikkar wants to do a rebase? (I've noticed that push by maintainers is disabled in his fork)

@Tomcat-42 , no - i have already done the rebasing for v3 here.  Let me know, if I have to do anything else with my branch. Thank you.

joe-lawrence:
Hi @sumanthkorikkar , I'd like to merge this PR despite it aging out.  The fix and solution looks fine to me.  I created a unit test based on your bug description here: https://github.com/dynup/kpatch-unit-test-objs/pull/56  do you mind taking a look and then we can merge that PR, and then bump unit test pointer in this PR to pickup the test?

PR #1502, opened 2026-05-15
lookup: x86_64 __pfx_

lu21253:
Skip symbols starting with __pfx_

Signed-off-by: lu21253 <lu.haoA@h3c.com>


lu21253:
> Skip symbols starting with __pfx_
> 6.6 kernel version, each function symbol in vmlinux/ko has a corresponding __pfx_-prefixed symbol, but it is not present in .o files. When creating a patch file, the check for searching and matching local symbols fails.
> Signed-off-by: lu21253 [lu.haoA@h3c.com](mailto:lu.haoA@h3c.com)

PR #1509, opened 2026-06-29
test/integration/rhel-9.8: Rebase patches to kernel-5.14.0-687.15.1.el9_8

Tomcat-42:


joe-lawrence:
Hi @Tomcat-42 : 
- Can we use the updated module.patch/module-LOADED.test that uses xfs and not nfsd (see the rhel-8.10 / rhel-10.2 versions).  For whatever reason we didn't make that change in prior rhel-9 rebases
- Likewise, for some reason the rhel-9 tests dropped new-function.patch and symvers-disagreement-FAIL.patch somewhere along the way.  Can you see if those can be added back to the 9.8 suite?
- And do you have links to test runs on ~~arm64~~ / ppc64le / s390x?

edit: rhel-9 probably doesn't have arm64 livepatching support, nevermind

PR #1512, opened 2026-08-04
create-diff-object: don't mark prefix symbols as changed

Zinkelburger:
When a symbol moves between sections -- out of an ignored section, or between text subsections -- kpatch_compare_correlated_symbol() marks it CHANGED.  A function's __pfx_/__cfi_ prefix symbol moves along with its parent, so it gets marked CHANGED as well.

Prefix symbols are padding, not independently patchable functions. kpatch_find_func_profiling_calls() already skips them, so they never have has_func_profiling set, and kpatch_check_func_profiling_calls() then rejects the build:

  function __pfx_foo has no fentry/mcount call, unable to patch

Marking them CHANGED also emits a klp_func entry pointing at the padding instead of the function.

This shows up on x86_64 with CONFIG_CALL_PADDING when patching a function in a section that doesn't honor -ffunction-sections, such as .sched.text: dropping __sched from the function and adding
KPATCH_IGNORE_SECTION(".sched.text") moves both the function and its prefix symbol out of the shared section.

Leave the prefix symbol's status alone.  It is still carried into the output through its parent function's sym->pfx link.

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -1048,14 +1048,18 @@ static void kpatch_compare_correlated_symbol(struct symbol *sym)
 	 * If two symbols are correlated but their sections are not, then the
 	 * symbol has changed sections.  This is only allowed if the symbol is
 	 * moving out of an ignored section, or moving between normal/hot/unlikely
-	 * subsections.
+	 * subsections.  A prefix symbol moves along with its parent function,
+	 * but it's padding rather than a patchable function, so leave its status
+	 * alone; it's included via the parent's sym->pfx link.
 	 */
 	if (sym1->sec && sym2->sec && sym1->sec->twin != sym2->sec) {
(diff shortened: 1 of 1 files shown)


joe-lawrence:
Hi @Zinkelburger : thanks for the PR, though since we are mostly in maintenance-mode only for kpatch-build, what are the intended use-cases for this fix?  And then, which approach is better #1512 or #1503 ?

Zinkelburger:
Internal to Red Hat, Sustaining has to support older streams with KPatch, and it uses this tool. I ran into this issue getting my patch to build because I had to remove a `__sched` annotation. Doesn't seem like many people have had to do this!

#1503 seems to be a different issue. It should also cover __cfi_ not just pfx (mine does both). There's also #1361.
#1503 has an issue with matching between `.o` and the `vmlinux symtab`, which I do not have. Mine is the status assigned afterwards.

PR #1520, opened 2026-08-19
kpatch-build: fix find_parent_obj KVM shortcut for all arches

### LINKED ISSUE #1519: kpatch-build: add arm64 KVM to known deep finds
On arm64, the KVM generic objects are built from arch/arm64/kvm/Makefile
with KVM=../../../virt/kvm, so kpatch-build has to resolve the object
name arch/arm64/kvm/../../../virt/kvm/kvm_main.o, whose parent is
arch/arm64/kvm/kvm.o.

find_parent_obj() already has a "prevent known deep finds" shortcut for
arch/x86/kvm, which covers the same situation on x86 (see #735), but
there is no arm64 counterpart. Patching a function in
virt/kvm/kvm_main.c therefore falls through to the full-tree deep find,
which fails to resolve the parent:

  doing "deep find" for parent object
  ERROR: invalid ancestor arch/arm64/kvm/../../../virt/kvm/kvm_main.o for arch/arm64/kvm/../../../virt/kvm/kvm_main.o

Add the corresponding shortcut for arch/arm64/kvm.

Tested on aarch64 with a 4.19 kernel and gcc 8.2.

jexbow:
## Summary

The `find_parent_obj()` shortcut for `virt/kvm/*` files hardcodes `arch/x86/kvm`, which only works when the target arch is x86. On every other architecture the parent-object lookup for KVM files misses the shortcut and falls through to a deep (and slow) full-tree search — and as reported in #1519, on arm64 it can even miss the parent entirely.

Instead of adding per-arch `if` blocks, derive the arch-specific KVM directory from `$TARGET_ARCH`. The kpatch `TARGET_ARCH` values don't always match the kernel arch directory names, so a small mapping is needed:

| kpatch TARGET_ARCH | kernel arch dir |
|---|---|
| `x86_64` | `arch/x86/kvm` |
| `aarch64` | `arch/arm64/kvm` |
| `ppc64le` | `arch/powerpc/kvm` |
| `s390x` | `arch/s390/kvm` |
| `loongarch64` | `arch/loongarch/kvm` |

## Verification

All kpatch-supported arches build the common KVM code from `virt/kvm/` (v4.19: arm64/x86/s390/powerpc use `KVM=../../../virt/kvm` in their KVM Makefiles; modern kernels set `KVM ?= ../../../virt/kvm` in `virt/kvm/Makefile.kvm`; loongarch follows the same pattern). So the `virt/kvm/*` shortcut now correctly maps to the right arch directory on every supported arch.

Tested only on arm64 (4.19 kernel, gcc 8.2) — I don't have hardware for the other arches, but the change is mechanical and the mapping table above is complete.

Fixes: #1519

### DIFF
--- a/kpatch-build/kpatch-build
+++ b/kpatch-build/kpatch-build
@@ -662,7 +662,7 @@ find_parent_obj() {
 		[[ -n "$PARENT" ]] && return
 	fi
 	if [[ $file = virt/kvm/* ]]; then
-		find_parent_obj_in_dir "$file" "arch/x86/kvm"
+		find_parent_obj_in_dir "$file" "arch/$(kernel_src_arch "$TARGET_ARCH")/kvm"
 		[[ -n "$PARENT" ]] && return
 	fi
 	if [[ $file = drivers/oprofile/* ]]; then
@@ -757,14 +757,20 @@ print_supported_distro(){
 	fi
 }
 
-# Used in "make ARCH=xxx" when making the kernel.
-# Match "aarch64" to "arm64", while keep everything else the same.
-kernel_make_arch() {
-	if [[ "$1" == "aarch64" ]]; then
-		echo "arm64"
-	else
-		echo "$1"
-	fi
+# Translate a kpatch TARGET_ARCH into the corresponding kernel source arch
+# directory name (and the value expected by "make ARCH=").  kpatch's
+# TARGET_ARCH values don't always match the kernel arch names, e.g.
+# x86_64 -> x86, aarch64 -> arm64, ppc64le -> powerpc, s390x -> s390,
+# loongarch64 -> loongarch.
+kernel_src_arch() {
+	case "$1" in
+		x86_64)      echo "x86" ;;
+		aarch64)     echo "arm64" ;;
+		ppc64le)     echo "powerpc" ;;
+		s390x)       echo "s390" ;;
+		loongarch64) echo "loongarch" ;;
+		*)           echo "$1" ;;
+	esac
 }
 
 usage() {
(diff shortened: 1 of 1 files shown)


joe-lawrence:
Thanks, @jexbow two really small nits:
- Can you add your "Signed-off-by: <name> <email>" line to the commit?
- The code looks like it would work, but how about:
  - Consolidating into the existing `kernel_make_arch()` -> rename to `kernel_src_arch()`
  - Add the arches
  - Then the fix becomes: `find_parent_obj_in_dir "$file" "arch/$(kernel_make_arch "$TARGET_ARCH")/kvm"`

jexbow:
Thanks @joe-lawrence, both make sense.

1. **Signed-off-by** — added.
2. **Consolidate into kernel_make_arch()** — good call. The current `kernel_make_arch()` only maps `aarch64`→`arm64` and passes the rest through unchanged, which is also a latent bug for x86_64/ppc64le/s390x/loongarch64 (kernel `make ARCH=` expects the source arch name). I renamed it to `kernel_src_arch()`, extended it to handle all kpatch-supported arches, and updated both call sites (the `KARCH=` one and the new `find_parent_obj` shortcut), which simplifies the fix to:

   ```bash
   find_parent_obj_in_dir "$file" "arch/$(kernel_src_arch "$TARGET_ARCH")/kvm"
   ```

Just force-pushed the update (amended commit) with the Signed-off-by line.

jpoimboe (approved):

jexbow:
Thanks @jpoimboe for the review! And thanks again @joe-lawrence for the suggestion to consolidate the arch mapping — that made the fix much cleaner.

joe-lawrence (approved):
Looks good, thanks @jexbow !

PR #1523, opened 2026-09-08
create-diff-object: include __ftr_alt_* when regenerating __ftr_fixup

joe-lawrence:
When a changed ppc64le function is referenced by a kept __ftr_fixup group, create-diff-object must also include the corresponding __ftr_alt_* section data.  kpatch_regenerate_special_section() currently only sets rela->sym->include = 1, which does not migrate STT_SECTION companions and leaves UND relocs in .rela__ftr_fixup:

  ld: final link failed: bad value

Call kpatch_include_symbol() instead, matching the similar .altinstr_replacement fix in 13b00146715c ("kpatch-build: ensure that .altinstr_replacement section is included").

Also exempt __ftr_fixup -> __ftr_alt_* relocations from the out-of-range relocation check.  Those addends are linker inputs for fixup-entry offsets, not pointers into the alt section.

Closes: 1522 ("ppc64le __ftr_fixup / __ftr_alt_97 sections causing bad final link")

### DIFF
--- a/kpatch-build/create-diff-object.c
+++ b/kpatch-build/create-diff-object.c
@@ -2937,7 +2937,7 @@ static void kpatch_regenerate_special_section(struct kpatch_elf *kelf,
 				rela->offset -= src_offset - dest_offset;
 				rela->rela.r_offset = rela->offset;
 
-				rela->sym->include = 1;
+				kpatch_include_symbol(rela->sym);
 
 				if (!strcmp(special->name, ".fixup"))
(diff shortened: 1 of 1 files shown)


joe-lawrence:
@jpoimboe : could you give a sanity check on this ppc64le fix?  This currently affects rhel-10 ppc64le kpatch.

jpoimboe:
> @jpoimboe : could you give a sanity check on this ppc64le fix? This currently affects rhel-10 ppc64le kpatch.

LGTM

